PR #45, opened 2026-05-14
Optimize list_reverse benchmark and enhance method tracing

lifthrasiir:
## Summary

This pull request adds JIT (Just-In-Time) specialization for the `list.reverse()` method in the Pyre Python interpreter, mirroring existing optimizations for `list.append()` and `list.pop()`. The changes enable the interpreter to emit a direct, efficient helper call when `list.reverse()` is invoked on a proven built-in list, improving performance for this operation. Additionally, the `list_reverse.py` benchmark is updated to wrap its logic in a `main()` function for consistency with other benchmarks.

**JIT Specialization for `list.reverse()`**

- Added a new external function `jit_list_reverse` in `listobject.rs` to perform the list reversal operation, marked with `dont_look_inside` for JIT optimization.
- Registered the new `jit_list_reverse` function and its alias in the JIT function address table in `jit_fnaddr.rs`.

**JIT Trace Integration**

- Implemented a new `list_reverse_value` trace path in `trace_opcode.rs` to directly emit the specialized helper call when a `list.reverse()` is detected on a built-in list.
- Integrated the new specialization into the method dispatch logic for both bound and unbound `list.reverse()` calls within the JIT trace opcode handler.

**Benchmark Update**

- Refactored the `list_reverse.py` benchmark to wrap its logic in a `main()` function, aligning it with other benchmarks and preparing for future oopspec lowering.

### Prompt & Model

Model: gpt-5.5

Prompt:
```
are available locally.

4. Structural adaptations

“Structural adaptations.”
```

### Answer

```
  1. Cases Where Patch Regressed PyPy Parity Compared To Main

  None found.

  None found that should be classified as an incorrect port.

  The new list.reverse() trace path preserves PyPy-visible behavior for the list strategies Pyre currently implements: it guards an
  exact builtin list receiver, calls the existing w_list_reverse, and returns None, matching PyPy’s descr_reverse ->
  W_ListObject.reverse -> strategy.reverse.

  - pyre/pyre-object/src/listobject.rs:822 implements only Pyre’s current list strategies: Empty, Integer, Float, and Object.
    PyPy’s pypy/objspace/std/listobject.py:1362 also has range-list behavior where reverse() switches to integer strategy and
    recurses, plus other strategies like bytes/ascii/int-or-float. This was already true on upstream/main; the patch only exposes the
    existing Rust helper to the JIT trace path.

  4. Structural Adaptations

  - pyre/pyre-jit-trace/src/trace_opcode.rs:4762 adds a direct trace-time specialization for list.reverse(). PyPy/RPython does not
    have this exact shape: PyPy goes through pypy/objspace/std/listobject.py:741, then RPython lowers reverse via rpython/rtyper/
    rlist.py:140 to rpython/rtyper/rlist.py:676.
  - pyre/pyre-object/src/listobject.rs:1133 adds jit_list_reverse as a dont_look_inside helper. RPython’s ll_reverse is
    look_inside_iff for virtual constant-length lists, so this is not structurally identical. Given Pyre’s current partial rtyper/
    codewriter lowering, this is a reasonable adaptation rather than a parity regression.
  - pyre/pyre-interpreter/src/jit_fnaddr.rs:405 registers the new helper address for Pyre’s JIT plumbing. There is no direct PyPy
    source equivalent.
  - pyre/bench/list_reverse.py:8 is a benchmark harness change, not a PyPy/RPython port surface.
```

## Summary by CodeRabbit

* **Performance**
  * Improved execution speed for `list.reverse()` operations through runtime optimization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/youknowone/pyre/pull/45)

### DIFF
--- a/pyre/bench/list_reverse.py
+++ b/pyre/bench/list_reverse.py
@@ -5,24 +5,22 @@
 # On this branch, reverse() stays in Integer strategy (no boxing overhead).
 # REPS=9 (odd): final list is reversed, so lst[0]=N-1, lst[-1]=0.
 #
-# NOTE: kept at module level intentionally — wrapping in def main() lets the
-# JIT fire and improves cranelift from ~18x to ~13x cpython, but the result
-# straddles the 10x cpython threshold and fails check.py flakily depending
-# on cpython measurement noise. Re-wrap when list oopspec lowering lands and
-# brings cranelift comfortably under threshold.
+def main():
+    N = 200000
+    REPS = 9
 
-N = 200000
-REPS = 9
+    lst = []
+    i = 0
+    while i < N:
+        lst.append(i)
+        i = i + 1
 
-lst = []
-i = 0
-while i < N:
-    lst.append(i)
-    i = i + 1
+    r = 0
+    while r < REPS:
+        lst.reverse()
+        r = r + 1
 
-r = 0
-while r < REPS:
-    lst.reverse()
-    r = r + 1
+    print(lst[0], lst[-1])
 
-print(lst[0], lst[-1])
+
+main()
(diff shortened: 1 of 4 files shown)


youknowone (approved):
👍

PR #48, opened 2026-05-14
Fix blackhole exception handling and align LOAD_GLOBAL behavior

lifthrasiir:
## Summary

This pull request refactors the JIT's handling of Python's LOAD_GLOBAL opcode to more closely mirror PyPy's semantics by explicitly passing the live frame object to the residual helper. This allows the global lookup to fall back to the frame's builtins module, not just the globals dict, matching Python's runtime behavior. The change also updates the relevant IR building, code generation, and tests to handle the new function signature and argument shape.

Key changes:

### LOAD_GLOBAL helper refactor

* The `bh_load_global_fn` helper now takes an explicit `frame` argument in addition to `namespace` and `code`, allowing it to access builtins via the live frame, not just the globals dict. The lookup logic now matches PyPy's `_load_global`, first checking globals, then builtins, and raising `NameError` if not found.

* The CPU struct's `load_global_fn` signature is updated to accept four arguments (`namespace`, `code`, `frame`, `namei`).

### IR and codewriter updates

* The IR builder and codewriter are updated so that all residual calls to `load_global_fn` now pass the frame as an additional argument, and the IR shape reflects this (`[ns, code, frame, namei]`).

### Test and documentation updates

* All related tests are updated to expect the new argument shape and verify the correct IR and call signatures.

### Exception handling improvement

* Exception stack handling in the codewriter is improved to use a scratch register for the exception object, preventing accidental overwrites when reusing registers.

These changes ensure that LOAD_GLOBAL in the JIT now matches Python's semantics more closely, especially in cases involving custom builtins, and that the IR and tests are consistent with the new calling convention.

### Prompt & Model

Model: gpt-5.5

Prompt:
```
are available locally.

4. Structural adaptations

“Structural adaptations.”
```

### Answer

```
  1. Regressed PyPy Parity Compared To Main

  None found.

  The patch improves parity in the two changed behavioral areas:

  - LOAD_GLOBAL now checks frame-picked builtins after globals, matching PyPy _load_global() in pypy/interpreter/pyopcode.py:958.
  - PUSH_EXC_INFO now preserves the popped exception in a scratch register before pushing prev, matching PyPy’s local w_exc lifetime
    in pypy/interpreter/pyopcode.py:819.

  None found.

  The changed helper signature is internally consistent across:

  - pyre/pyre-jit/src/call_jit.rs:3324: bh_load_global_fn(ns, code, frame, namei)
  - pyre/pyre-jit/src/jit/cpu.rs:41: extern fn type updated to 4 args
  - pyre/pyre-jit/src/jit/codewriter.rs:5557: SSA residual call passes ns, code, frame
  - pyre/pyre-jit/src/jit/flatten.rs:2850: call descr uses [Ref, Ref, Ref, Int]

  PUSH_EXC_INFO still models current exception state as a raw exception object in TLS, while PyPy stores/restores an OperationError
  and distinguishes hidden applevel frames. PyPy branches on self.hide() and uses hidden_operationerr in pypy/interpreter/
  pyopcode.py:823, then restores via _restore_exc_info() in pypy/interpreter/pyframe.py:185. The Rust JIT helper path uses
  get_current_exception / set_current_exception object helpers in pyre/pyre-jit/src/call_jit.rs:3652. This was already the shape on
  upstream/main; the patch only fixes register lifetime.

  LOAD_GLOBAL still uses Pyre’s split DictStorage globals representation instead of PyPy’s direct object-space
  finditem_str(self.get_w_globals(), varname) object path. The direct globals lookup existed before this patch in the helper; the new
  builtins fallback brings it closer to PyPy rather than introducing the mismatch.

  4. Structural Adaptations

  LOAD_GLOBAL passes the live frame explicitly as a residual helper argument. PyPy has self._load_global(varname) and can call
  self.get_builtin() directly; the Rust C-style helper has no implicit receiver, so pyre/pyre-jit/src/jit/codewriter.rs:5592 adds
  frame_var to the ref args.

  namei >> 1 in pyre/pyre-jit/src/call_jit.rs:3335 is a CPython-compatible bytecode adaptation for LOAD_GLOBAL’s low-bit NULL marker.
  PyPy’s local source uses nameindex directly, but this falls under the opcode/compiler mismatch exception.

  The PUSH_EXC_INFO scratch copy in pyre/pyre-jit/src/jit/codewriter.rs:6125 is a Rust SSA/register adaptation. PyPy keeps w_exc as a
  Python local while mutating the value stack; Rust needs ref_copy so the later push(exc) does not reread the stack slot overwritten
  by push(prev).
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed exception value handling in JIT compilation to prevent stack slot corruption.

* **Refactor**
  * Improved global variable lookup to support frame-aware builtin resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/youknowone/pyre/pull/48)

### DIFF
--- a/pyre/pyre-jit/src/jit/flatten.rs
+++ b/pyre/pyre-jit/src/jit/flatten.rs
@@ -5761,7 +5761,7 @@ mod tests {
         // `_load_global(self, ...)` receiver.
         let insn = build_load_global_fn_residual_call_ir_r_insn(
             /* load_global_fn_idx */ 12, /* namei */ 5, /* ns_reg */ 3,
-            /* code_reg */ 4, /* frame_reg */ 6, /* dst_reg */ 3,
+            /* code_reg */ 4, /* frame_reg */ 6, /* dst_reg */ 7,
         );
         match insn {
             Insn::Op {
@@ -5770,7 +5770,7 @@ mod tests {
                 result: Some(reg),
             } => {
                 assert_eq!(opname, "residual_call_ir_r");
-                assert_eq!(reg, Register::new(Kind::Ref, 3));
+                assert_eq!(reg, Register::new(Kind::Ref, 7));
                 assert_eq!(args.len(), 4);
                 match &args[0] {
                     Operand::ConstInt(v) => assert_eq!(*v, 12),
@@ -5851,7 +5851,7 @@ mod tests {
         let ns_var = Variable::new(VariableId(3), Kind::Ref);
         let code_var = Variable::new(VariableId(4), Kind::Ref);
         let frame_var = Variable::new(VariableId(6), Kind::Ref);
-        let dst_var = Variable::new(VariableId(3), Kind::Ref);
+        let dst_var = Variable::new(VariableId(7), Kind::Ref);
         let descr = intern_call_descr_stub(
             effect_info_for_call_flavor(CallFlavor::Plain),
             vec![Kind::Ref, Kind::Ref, Kind::Ref, Kind::Int],
@@ -5876,7 +5876,7 @@ mod tests {
         let mut lower_constant = probe_constant_lowering();
         let dual = flatten_op_to_insn(&dual_op, &mut get_register)
             .expect("residual_call SpaceOperation must lower");
-        let prod = build_load_global_fn_residual_call_ir_r_insn(12, 5, 3, 4, 6, 3);
+        let prod = build_load_global_fn_residual_call_ir_r_insn(12, 5, 3, 4, 6, 7);
         assert_eq!(format!("{prod:?}"), format!("{dual:?}"));
     }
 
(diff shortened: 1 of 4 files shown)


pyre/pyre-jit/src/jit/flatten.rs
```diff
@@ -5740,15 +5755,13 @@ mod tests {
...
-        // codewriter.rs:2176-2185 — `bh_load_global_fn` cannot force
-        // virtuals, matches `EF_CAN_RAISE`).
+        // code), Reg(frame)]), Descr(CallDescrStub{Plain,
+        // [Ref, Ref, Ref, Int]})] → Reg(Ref, dst)`.  The explicit
+        // frame Ref is the helper-level stand-in for PyPy's
+        // `_load_global(self, ...)` receiver.
         let insn = build_load_global_fn_residual_call_ir_r_insn(
             /* load_global_fn_idx */ 12, /* namei */ 5, /* ns_reg */ 3,
-            /* code_reg */ 4, /* dst_reg */ 3,
+            /* code_reg */ 4, /* frame_reg */ 6, /* dst_reg */ 7,
```
youknowone, line 5764: we'd better to make constants  for these literals later

youknowone (approved):

PR #59, opened 2026-05-16
Optimize tuple unpacking and specialization in JIT trace

lifthrasiir:
## Summary

This pull request enhances support for Python object subclass identity in the JIT trace, particularly for tuples and integer subclasses. The main changes introduce descriptors and helpers to accurately track and guard the dynamic Python-level class (`w_class`) of objects, ensuring correct behavior when handling subclassed types and specialized tuple variants.

Key changes include:

### Tuple and Specialized Tuple Object Model Improvements

- Added `w_class` field descriptors for `W_TupleObject` and all specialized tuple variants (`SPECIALISED_TUPLE_II`, `FF`, `OO`), allowing the JIT to access and set the Python-level class of these objects.
- Introduced size descriptors for each tuple object variant, supporting accurate memory layout handling in the JIT.
- Provided new helper functions to retrieve these descriptors in `descr.rs`.

### Concrete Value and Subclass Handling

- Updated `ConcreteValue::from_pyobj` to distinguish between exact `int`, `bool`, and integer subclasses, preserving subclass identity by returning a reference instead of unboxing for non-exact ints.
- Added `is_trace_plain_int` helper to check for exact (non-subclassed) integer objects.
- Added tests to verify correct handling and unboxing of exact ints, bools, and int subclasses.

### JIT Tracing and Opcode Handling Enhancements

- Updated tuple construction tracing to use both the IR and concrete items, enabling precise tracking of tuple contents and class.
- Introduced tracing helpers in `trace_opcode.rs` to:
    - Unbox exact integer payloads and guard for exact Python classes in the trace.
    - Set the `w_class` field for tuple objects during trace construction, maintaining Python-level type information.

These changes ensure that the JIT trace can accurately model Python's dynamic type system, especially for tuples and integer subclasses, improving correctness and compatibility with Python semantics.

Benchmark run:

```
  benchmark                            cpython     pypy             dynasm          cranelift
  ──────────────────────────────────────────────────────────────────────────────────────────────────
  synth/tuple_unpacking                  0.13s    0.01s      0.01s    0.5x      0.02s    1.9x
```

### Prompt & Model

Model: gpt-5.5

Prompt:
```
are available locally.

4. Structural adaptations

“Structural adaptations.”
```

### Answer

```
  1. Regressed PyPy Parity Compared To Main

  pyre/pyre-jit-trace/src/trace_opcode.rs:4525 regresses exact-type parity for future trace executions.

  PyPy’s makespecialisedtuple2 only selects Cls_ii when both args satisfy is_plain_int1, and is_plain_int1 is type(w_obj) is W_IntObject or exact
  W_LongObject that fits int. Main used the opaque tuple helper, so this predicate ran at runtime.

  The patch checks is_plain_int1(lhs/rhs) only while tracing, then emits unbox guards that check only ob_type/layout: pyre/pyre-jit-trace/src/
  trace_opcode.rs:4467 and pyre/pyre-jit-trace/src/state.rs:2142. They do not guard w_class. A later int/long subclass with the same Pyre payload
  layout can incorrectly stay on the Cls_ii trace and lose object identity when unpacked/getitem wraps raw ints.

  pyre/pyre-jit-trace/src/trace_opcode.rs:4575 has the same regression.

  PyPy requires type(w_arg) is W_FloatObject before selecting Cls_ff. The patch checks is_plain_float_strict at trace time, but the emitted pyre/pyre-
  jit-trace/src/state.rs:2166 guard only checks FLOAT_TYPE, not w_class. A future float subclass can incorrectly take the Cls_ff trace and be
  rewrapped as a plain float.

  None found beyond the trace-guard exact-type regressions above. The runtime tuple constructors and object helpers otherwise match the PyPy selection
  rules more closely than main.

  3. Mismatches Already Present Before This Patch

  pyre/pyre-object/src/listobject.rs:299 still use is_float/payload layout where PyPy’s FloatListStrategy.is_correct_type uses exact type(w_obj) is
  W_FloatObject. Similar existing sites include append/setitem strategy decisions in listobject.rs and JIT list append selection at pyre/pyre-jit-
  trace/src/trace_opcode.rs:5133. These can accept float subclasses into float storage, unlike PyPy.

  pyre/pyre-jit-trace/src/state.rs:1268 still unboxes float subclasses with is_float. The patch fixed int subclass preservation, but float subclass
  concrete identity remains weaker than PyPy’s exact W_FloatObject distinction.

  4. Structural Adaptations

  pyre/pyre-jit-trace/src/trace_opcode.rs:250 is a valid Pyre object-model adaptation. PyPy’s specialised tuple classes share the public tuple
  typedef; Pyre models JIT-visible layout with distinct ob_type values, so setting w_class to canonical tuple is necessary.

  The fallback to opaque tuple build when concrete side-channel values are null at pyre/pyre-jit-trace/src/trace_opcode.rs:4509 is also a reasonable
  adaptation. PyPy always has concrete W_Root values at space.newtuple; Pyre sometimes does not.

  Specialised tuple unpack tracing at pyre/pyre-jit-trace/src/trace_opcode.rs:4720 is structurally aligned with PyPy’s specialised tuple tolist()/
  getitem() behavior, with Pyre-specific virtual wrapint/wrapfloat replacing immediate boxed allocation.
```

(My Codex usage limits ran out right after this final assessment, so it might require some more fixes.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve exact int/bool identity so booleans and int subclasses aren’t misclassified; tighter checks avoid incorrect fast-paths.

* **Performance**
  * Faster 2-element tuple construction and unpacking with new specialization fast paths (int/int, float/float, object/object) and stricter class identity guards.

* **New Features**
  * JIT now emits real tuple construction/unpack operations instead of no-op placeholders.

* **Tests**
  * Added/updated unit tests covering int/bool mapping, tuple-specialization selection, and cache behavior.

### DIFF
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -2262,12 +2262,6 @@ impl Backend for DynasmBackend {
                     Some(slot) => {
                         let val =
                             unsafe { crate::llmodel::get_int_value_direct(result_jf, slot) as i64 };
-                        if crate::majit_log_enabled() && i < 10 {
-                            eprintln!(
-                                "[dynasm] fail_arg[{}]: slot={} val={:#018x}",
-                                i, slot, val as u64
-                            );
-                        }
                         raw_values.push(val);
                     }
                     None => raw_values.push(0),
--- a/majit/majit-backend-dynasm/src/x86/assembler.rs
+++ b/majit/majit-backend-dynasm/src/x86/assembler.rs
@@ -1361,13 +1361,21 @@ impl<'a> Assembler386<'a> {
                 dynasm!(self.mc ; .arch x64 ; cmp Rq(r.value), [rbp + f.ebp_loc.value]);
             }
             (Loc::Reg(r), Loc::Immed(i)) => {
-                dynasm!(self.mc ; .arch x64 ; cmp Rq(r.value), i.value as i32);
+                self.emit_cmp_imm64(r.value, i.value);
             }
             (Loc::Frame(f), Loc::Reg(s)) => {
                 dynasm!(self.mc ; .arch x64 ; cmp [rbp + f.ebp_loc.value], Rq(s.value));
             }
             (Loc::Frame(f), Loc::Immed(i)) => {
-                dynasm!(self.mc ; .arch x64 ; cmp QWORD [rbp + f.ebp_loc.value], i.value as i32);
+                if let Ok(v) = i32::try_from(i.value) {
+                    dynasm!(self.mc ; .arch x64 ; cmp QWORD [rbp + f.ebp_loc.value], v);
+                } else {
+                    let scratch = crate::regloc::X86_64_SCRATCH_REG.value;
+                    dynasm!(self.mc ; .arch x64
+                        ; mov Rq(scratch), QWORD i.value
+                        ; cmp QWORD [rbp + f.ebp_loc.value], Rq(scratch)
+                    );
+                }
             }
             _ => {
                 self.regalloc_mov(loc0, &Loc::Reg(crate::regloc::X86_64_SCRATCH_REG));
--- a/majit/majit-ir/src/descr.rs
+++ b/majit/majit-ir/src/descr.rs
@@ -2446,6 +2446,17 @@ pub trait SizeDescr: Descr {
         0
     }
 
+    /// Pyre object-model: the canonical Python class object pointer
+    /// (`PyObject.w_class`) that instances of this type carry, i.e.
+    /// `get_instantiate(vtable_type)`. A `new_with_vtable` virtual of a
+    /// builtin type inherits this value unless the trace stores an
+    /// explicit `w_class` field. OptVirtualize folds `w_class` header
+    /// reads to this constant. `None`/`0` for non-pyre size descrs or
+    /// before the type objects are initialised.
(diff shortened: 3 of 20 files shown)


youknowone:
Static parity/perf follow-up related to this PR scope:

This PR is the right place to verify tuple-unpacking performance parity because it changes trace-visible tuple construction/unpacking specialization.

The specific thing to investigate is whether the strict PyPy-parity guards for exact tuple/int/float class identity still allow the specialized tuple fast paths to be selected in the benchmark cases that PyPy optimizes well. If the strict guards are correct but the trace falls back to the generic tuple path too often, tuple unpacking can remain semantically correct while still missing PyPy-like performance.

Suggested checks for this PR:

- confirm that arity-2 tuple construction records the specialized `ii`/`ff`/`oo` tuple paths when the concrete operands satisfy PyPy exact-type predicates;
- confirm that replay guards reject subclasses or non-exact values before entering those specialized paths;
- confirm that unpacking a known specialized tuple loads inline fields directly rather than routing through the generic sequence/unpack helper;
- compare the emitted trace shape for `synth/tuple_unpacking` against the PyPy intent: exact-type guarded specialized tuple allocation plus direct field loads.

If all of those hold and the benchmark is still slow, the remaining cause is probably outside tuple specialization itself, likely in surrounding trace selection, guard placement, or downstream optimization/regalloc.

PR #70, opened 2026-05-18
Fix synthetic JIT parity failures and improve CI checks

lifthrasiir:
## Summary

This pull request introduces several improvements and bug fixes across the JIT backends, the benchmarking suite, and the command-line interface. The most significant changes address GC ownership queries under reentrant conditions, improve structural compatibility with upstream RPython/PyPy, and refine synthetic benchmark controls and JIT suppression for unsupported bytecode shapes.

**Garbage Collector (GC) Ownership and Reentrancy Handling:**

- Added raw pointer thread-locals (`*_ACTIVE_GC_RAW`) for both Cranelift and Dynasm backends, enabling safe GC ownership checks (`is_managed_heap_object`) even when the main GC lock is already held. This prevents panics during reentrant queries from root walkers and matches the RPython/PyPy GC descriptor behavior.

**JIT Suppression for Unsupported Bytecode Shapes:**

- Introduced a suppression guard and detection for unsupported bytecode patterns (e.g., `WITH_EXCEPT_START`, certain `FOR_ITER` cases). When such patterns are present, the JIT is disabled for the affected frame or region, falling back to the interpreter to match upstream semantics and avoid incorrect execution.

**Structural Compatibility and SSA/Liveness Fixes:**

- Corrected SSA liveness marker selection to use the last `-live-` marker per anchor pair, ensuring correct resume behavior after guard failures and matching upstream RPython/PyPy expectations.
- Added a helper to restore the canonical order of normal and exception edges in control flow blocks, preserving compatibility with upstream exception handling logic.

**Benchmarking and CLI Improvements:**

- Changed the synthetic benchmark CLI flag from `--synthetic` to `--no-synthetic` (defaulting to running synthetic benchmarks unless explicitly skipped), clarified help text, and enforced mutual exclusion with `--synthetic-only`.
- Reduced the default iteration counts in `pyre/bench/synth/comprehensions.py` and `pyre/bench/synth/context_manager.py` to speed up test runs.

**Other Minor Fixes:**

- Updated CLI to disallow argument abbreviation for more robust parsing.

These changes improve robustness, correctness, and upstream compatibility, particularly in GC and JIT edge cases, and streamline synthetic benchmark management.

## Self-review

This PR is intentionally *not* reviewed because its primary purpose is to restore synthetic test parity and enforce it hereafter in the CI. Some tests, notably synth/context_manager and synth/for_range_loop are excluded from JIT in order to restore parity. It's advised to ignore auto-reviews as well because their suggestions often break otherwise working tests.

## Summary by CodeRabbit

* **New Features**
  * Added JIT suppression for frames with unsupported bytecode shapes
  * CLI change: replaced `--synthetic` with `--no-synthetic` and tighter arg validation

* **Bug Fixes**
  * Improved handling of exception-edge/catch initialization during JIT generation
  * Increased GC allocator reentrancy tolerance to avoid borrow panics during teardown

* **Chores**
  * Reduced benchmark iteration counts to speed up synth benchmarks

[Review Change Stack](https://app.coderabbit.ai/change-stack/youknowone/pyre/pull/70?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


youknowone:
A. Aliasing UB (CodeRabbit 2 / Codex 1) — cranelift compiler.rs:1372, dynasm 
  runner.rs:280

  Core problem: The try_borrow_mut() Err branch fires precisely when a &mut borrow
  is already live. Creating unsafe { (&*ptr).is_managed_heap_object(addr) }
  materializes a shared reference that aliases that &mut — strict UB under Rust's
  reference model, regardless of whether the call is read-only.

  Realistic options (smallest → largest):
  1. Snapshot interface (recommended) — Add heap_extents() -> (start, end) to
  GcAllocator. On setter, cache the answer in Cell<Option<(usize, usize)>> alongside
   the raw mirror. The reentrant fallback never builds a &dyn; it just does start <=
   addr < end against the Cell. Works if is_managed_heap_object semantically is
  "addr falls inside the nursery + old-gen extents."
  2. Eliminate the reentrancy window — Refactor callers so root walkers don't run
  while the RefCell is borrowed. Correct but invasive.
  3. Switch to Mutex<Box<dyn GcAllocator>> with recursive locking. Affects every
  caller.
  
  When to fix: Immediately if Miri (cargo +nightly miri test) flags it. Otherwise
  bundle into the next GC-touching PR. For now, leave a one-line // 
  TODO(aliasing-ub): see PR #70 review at the unsafe block so it's not forgotten.

  B. unsupported_jit_shape O(n) rescan (CodeRabbit 5) — eval.rs:2201

  Core problem: The function is queried from try_function_entry_jit,
  maybe_compile_and_run, and jit_merge_point_hook — the last one fires at every
  merge point. Re-walking code.instructions makes the probe O(code_len) on every
  call.

  Options:
  1. Single-slot TLS cache (recommended, minimal change) — In eval.rs: thread_local!
   { static SHAPE_CACHE: Cell<Option<(*const CodeObject, UnsupportedJitShape)>> }.
  Hot path is a pointer compare + hit/miss.
  2. OnceLock field on CodeObject — Cleaner but requires changing
  pyre_interpreter::CodeObject; check FFI implications.
  
  When to fix: Only when profiling shows unsupported_jit_shape is hot. Small/medium
  functions: ignore. Large functions: if a regression appears, ship option (1) right
   away.

  C. restore_canraise_exit_order regression test (CodeRabbit 7, nitpick) — 
  codewriter.rs:2101

  Core problem: The helper now both reorders and prunes exits, but there's no test
  pinning the contract. A unit test should construct a BlockRef with [normal, 
  exception, duplicate_normal] and assert it shrinks to [normal, exception] (length
  + last_exception/last_exc_value shape).

  When to fix: Fold into the next PR that touches codewriter. Cost is trivial — no
  need for a standalone PR.

PR #148, opened 2026-06-06
walker concrete-execute for non-elidable residual_calls + Phase 5.B body wire

### LINKED ISSUE #122: Tagged-int representation for inline Python small ints, with Z2.5 fallback-map cutover as a bundled sub-epic
## Summary

This umbrella tracks two remaining structural alignment items not already covered by other open issues:

1. **Primary** — tagged-int representation for inline Python small ints in the JIT.
2. **Sub-epic** — retire the legacy-walker fallback maps in `cutover.rs`.

Each section below is independently actionable. Cross-references at the bottom list existing issues that cover adjacent work (tracer / codewriter / flatten parity).

> **Scope revision (post-open):** A previously-listed third sub-epic — cranelift `FailDescr` wrapper retirement — was removed after verification showed the work already landed in commit `8d22f4fb36` (May 2026). The current code carries only historical comment references to the deleted struct.

---

### Motivation

PyPy's lltype tagged-int convention (`(i & 1) == 1` for inline-tagged Python ints) permeates `cast_int_to_ptr` / `cast_ptr_to_int`, the heap/array cache, the vtable layout, and `GcRef` semantics. Pyre currently uses an identity-cast design (concrete propagation through the cast op) that diverges intentionally. The identity-cast record arms live at `pyre/pyre-jit-trace/src/jitcode_dispatch.rs:14093` (`cast_int_to_ptr/i>r`) and `:14107` (`cast_ptr_to_int/r>i`); the `pending Phase F` TODO sits on the blackhole identity impls `majit/majit-metainterp/src/blackhole.rs:8139` / `:8145`, and the executor fold is `majit/majit-metainterp/src/executor.rs:758-759`. These four are the complete conversion-boundary set.

This is a runtime data-layout shift, so it should land only after the tracer / codewriter / flatten work tracked by the related issues below is orthodox and bench-green. Otherwise every earlier change would also need to account for tagged-int semantics.

### Scope (in order)

1. **Design doc** — single source of truth describing the representation: which bit patterns mean what, where the conversion boundaries are, how the GC sees a tagged-int (never as a pointer), how the vtable layout accommodates it. Cite `rpython/rlib/objectmodel.py` and `rpython/rtyper/lltypesystem/rffi.py::cast_int_to_ptr`. It must also explicitly distinguish the real `(i & 1) == 1` value tagging from two existing-but-unrelated mechanisms (see Findings below) so the implementation does not false-start on them.
2. **`cast_int_to_ptr` / `cast_ptr_to_int` asserts** — add the upstream assert `(i & 1) == 1` at the cast site, with concrete propagation matching the tagged convention. Cast boundary (complete set): record arms `jitcode_dispatch.rs:14093` / `:14107`, blackhole impls `blackhole.rs:8139` / `:8145`, executor fold `executor.rs:758-759`. **Not independently landable before items 3/4** — `executor.rs:758` currently propagates the untagged value (`r.0 as i64`, with no `<< 1 | 1`), so adding the assert before the representation flip would fire on every even int and red `check.py` immediately. Item 2 is co-dependent on items 3/4, not a standalone slice; item 1 (design doc) is the only independently-landable unit today.
3. **Heap / array cache adjustments** — cache reads must not treat a tagged pointer as an object handle. Touches the optimizeopt heap pass.
4. **Vtable / GcRef wiring** — `is_exception` subclass-range check and per-kind PyType discrimination already work without tagged-int. Tagged-int adds a fast-path skip for tagged values.

### Acceptance criteria

- Tagged-int representation is the default for Python small ints inside the JIT.
- `cast_int_to_ptr` / `cast_ptr_to_int` semantics match the upstream RPython convention line-for-line, with cited upstream source per call site.
- `pyre/check.py` green on dynasm + cranelift, all benches.
- Measurable speedup on int-heavy benches (`int_loop`, `nested_loop`, `synth/int_arithmetic`).

### Size

~30-50 small slices over ~10-15 sessions. Structurally the riskiest item in this umbrella because it shifts runtime data layout.

---

### Scope

Retire the two legacy-walker fallback maps in `majit/majit-translate/src/translator/rtyper/cutover.rs`. (The original line refs `220` and `306-356` predate later edits; `cutover.rs` is now 3092 lines, with the dual-gate machinery at `dual_gate_check`:199, `DualGateOutcome`:301, and `is_known_unported`:734 — re-derive the exact map sites against current code.) These carry the legacy-walker fallback used when the canonical translator does not yet ground a value. Other Z2.5-era carriers have already been removed; these are the remaining tail.

### Blocker

A `no_binding` panic surfaces as the `annotator.annotated[block] is False sentinel` abort at `majit/majit-translate/src/translator/rtyper/rtyper.rs:1717` when the cutover is attempted. (The original `pyre/pyre-jit/src/jit/codewriter.rs:395` reference has drifted — that line is now `variable_slot`.) Failure chain:

```
PyFrame::_check_stack_index
  → self.locals_cells_stack_w          (now *mut FixedObjectArray — pyframe.rs:52, was Vec<PyObjectRef>)
  → classdef-less SomeInstance dispatch
  → annotator False sentinel
  → rtyper "annotated[block] is False sentinel"   (rtyper.rs:1717)
```

Resolving this requires transitive inner-type registration of host-evaluator structures (the `FixedObjectArray` payload, and the broader classdef-less `SomeInstance` dispatch family). A narrow self-resolution patch at `build_function_graph` exposes additional downstream panic sites (the original list `annrpython.rs:432`, `unaryop.rs:3587`, `rtyper.rs:1656`, `codewriter.rs:395` predates later edits and should be re-derived) and is not a viable short-circuit. The blocker is its own multi-session structural effort.

### Acceptance criteria

- `no_binding` chain resolved (may produce its own PR).
- Two `cutover.rs` fallback maps deleted.
- `pyre/check.py` green on dynasm + cranelift.

### Size

3-5 slices for the cutover itself once the `no_binding` chain is resolved; 5-10 sessions total including blocker work.

---

## Sequencing

- The Sub-epic can proceed as soon as its `no_binding` blocker resolves; ordering against the related-issue work below is independent. **Edit-restriction caveat:** the entire sub-epic fix locus (the `cutover.rs` maps, the dual-gate consumer, the legacy walker, the annotator classdef-lift, and the `PyFrame` owner) lives in `majit/majit-translate` + `pyre/pyre-interpreter`. The `no_binding` chain is real but currently dormant — `is_known_unported` absorbs it so the build stays green — and the maps retire only as the terminal milestone of the full Rust-runtime modeling tail, not incrementally.
- The Primary tagged-int work should start only after the tracer / codewriter / flatten work tracked by #50, #73, #97, #112, #115 lands orthodox and bench-green.

## Acceptance criteria (umbrella)

The umbrella closes when:

1. Primary acceptance criteria above are met.
2. The Sub-epic either lands here or is split into its own follow-up issue with this issue closed as superseded.
3. `pyre/check.py` is green on dynasm + cranelift across all benches at every intermediate step. If a slice cannot land green, revert and re-plan rather than gate-flag it.

## Size summary

| Section | Slices | Sessions | Notes |
|---|---|---|---|
| Primary — tagged-int | 30-50 | 10-15 | Long-horizon; starts after the related issues below land |
| Sub-epic — Z2.5 cutover | 3-5 (after blocker) | 5-10 total | Blocked on `no_binding` chain |
| **Total** | **~35-55** | **~15-25** | Independent of #50 / #73 / #97 / #112 / #115 |

## Related existing issues

- **#50** — majit-translate auto-emit SSA per opcode handler. *(OPEN; partially advanced, workarounds intact.)*
- **#73** — Eliminate `pc_map` and per-PC labels by switching the tracer to interpret JitCode bytecode. *(OPEN; the critical-path gate — see Findings.)*
- **#97** — Replace syn-AST front-end with Charon-extracted MIR for the JIT lowering pipeline. *(CLOSED.)*
- **#112** — Port RPython simplify_graph parity and track remaining graph/codewriter parity gaps. *(CLOSED.)*
- **#115** — JIT walker aborts on LoadAttr/StoreAttr (RegisterOutOfRange). *(OPEN; abort symptom no longer reproduces as of #413 — see Findings.)*

## Non-goals

- This issue does not cover `BoxPool` removal — tracked separately.
- This issue does not cover the primary tracer / codewriter / flatten work — see the related issues above.
- This issue does not cover residual mechanical cleanup (no dedicated tracking issue warranted).
- This issue does not cover the deferred aarch64 jump-arg coalescing investigation — confirmed structural ISA limit, no actionable change.
- This issue does not cover cranelift `FailDescr` wrapper retirement — that work landed in commit `8d22f4fb36`; only historical comment references remain in the current code.

---

## Findings & corrections (updated 2026-06-24)

Scoping investigation found inaccuracies (fixed inline above) plus additional detail:

- **Cast-site line numbers were stale** (`4944 / 4973` → the four sites now cited inline). The complete conversion boundary is exactly those four: `jitcode_dispatch.rs:14093` / `:14107` (record), `blackhole.rs:8139` / `:8145` (identity impl), `executor.rs:758-759` (fold). No other cast boundary exists.

- **Scope item 2 is not independently landable.** `executor.rs:758` propagates the untagged value (`r.0 as i64`, no `<< 1 | 1`), so the `(i & 1) == 1` assert reds `check.py` on every even int until the representation flips. It is co-dependent on items 3/4; only item 1 (design doc) is standalone.

- **No runtime-layout tagged-int infra exists yet — two false-start traps.** (a) The `/id>X` `/iXd` getfield/setfield variants at `blackhole.rs:7436` are a *base-pointer-in-int-register* register-class convenience (same `bh_{get,set}field_gc_*` primitive), **not** `(i & 1) == 1` value tagging. (b) The `TAGINT` / `_encode_smallint` encoding in `resume.rs` / `opencoder.rs` tags serialized resume-data numbers (`resume.py` tag()/untag()), never live runtime PyObject pointers. Neither is a head-start on the Primary.

- **Tagged-int is orthogonal to the boxed-int short-circuit/condexpr resume miscompile.** Layer mismatch: tagged-int is an lltype/residual Signed↔Ptr convention; the kept-stack hazard is at the Python-object layer (the kept slot holds a heap `W_IntObject*`; `w_int_new` always malloc's — `intobject.rs:90` — so a large int stays a heap object regardless of lltype tagging). Axis mismatch: the kept-boxed-int decline fires because a loop-invariant-hoisted const has no symbolic resume coordinate to rematerialize on the not-taken arm — a recorder snapshot-box-selection gap (#73), not a representation gap. Tagged-int would not make that case JIT-able-and-correct, and would add its own resume/GC tag-awareness hazard. It is not a shortcut around #73.

- **Gate status / critical path.** #97 + #112 CLOSED. #115's `RegisterOutOfRange` abort symptom no longer reproduces — LoadAttr/StoreAttr are in the production-walker allow-list (via #413) and the attr-heavy repro JIT-compiles to correct output on both backends (issue still open pending formal close). The two real remaining blockers are #50 and #73; **#73 (pc_map / symbolic-valuestack) is the critical path** for the Primary, because tagged-int's runtime data-layout shift re-enters exactly the resume/kept-stack layer #73 governs.

— *opened by Claude; corrections by Claude (2026-06-24)*

youknowone:
Fix https://github.com/youknowone/pyre/issues/122

## Summary

Two structurally-linked epic foundations that together prepare for retiring the M4 SIGBUS workaround in [[project_m4_walker_remaining_unactivated_taxonomy_2026_06_06]]:

**Task #390 — widen walker concrete-execute to non-elidable residual_calls.** PyPy `_opimpl_residual_call{1,2,3}` (`pyjitpl.py:1346/1349/1354`) concrete-executes every `do_residual_call` regardless of `check_is_elidable()` — the EI flag picks the record opcode (`CALL_PURE_*` vs `CALL_*`), not whether execution happens. The pyre walker currently concrete-executes only the elidable path via `try_fold_pure_call_via_executor`. Non-elidable helpers (`store_subscr_fn` / `set_current_exception` / …) are recorded but never executed; `eval_loop_jit:3111`'s walker-dispatched skip then leaves the heap unchanged → SIGBUS in 5 STORE_SUBSCR-hot benches. This PR lands sub-slices 1-5 of the orthodox fix:

- (1) inventory + doc-comment of the gap (`baf59d132f`, `109bccbd6d`)
- (2) `majit_metainterp::executor::execute_residual_call` walker-friendly entry + `try_execute_residual_call_via_executor` for `Call*` / `CallLoopinvariant*` shapes, wired to 3 dispatch sites with `!can_raise` gate (`f7b12284e5`, `e042d274b1`, `ca780d7cb3`)
- (3) walker-side exception propagation: `walker_record_guard_exception` helper, `WalkContext.last_exc_value` seeding, `BH_LAST_EXC_VALUE` restore, `eval.rs` walker-skip path surfaces non-zero exc as `Err(PyError)` (`7dd4238bb8`)
- (4) PyPy-orthodox activation via 47-bit fnaddr-sanity gate `(func_ptr as u64) >> 47 != 0` rejects unpatched `symbolic_fnaddr_for_path` DefaultHasher placeholders; lifts the `!can_raise` gate at all 3 dispatch sites (`963c2d807e`)
- (5) sub-slice 5 foundation: `bh_execute_store_subscr` C-ABI wrapper in `pyre-interpreter::opcode_ops` + `jit_trace_fnaddrs()` entry for path `"execute_store_subscr"`; build-time codewriter now bakes the real address into `constants_i` and `runtime_fnaddr_patch` maps build→runtime correctly (`0a4cfca1bc`)

StoreSubscr walker activation itself is not in this PR — it surfaces a heapcache EffectInfo gap (setitem's list/dict strategy-field writes aren't visible across the trait-dispatch boundary the codewriter analyses). Sub-slice 5b will widen EffectInfo via per-type wrappers (`bh_list_setitem` / `bh_dict_setitem` / …) following the PyPy `bhimpl_setarrayitem_gc_*` precedent.

**Phase 5.B `dispatch_arm_via_blackhole` body wire (workaround scaffolding).** The original workaround path for the same M4 class: run the arm jitcode through `BlackholeInterpreter` to concrete-execute the missing helpers. Foundation work committed here (`dca58bbdbd` → `08bc3d210a`) — JitCode wrapper cache, thread-local BH builder, acquire/release wiring, `bh.run()` + outcome triage, virtualizable wiring, exception → PyError propagation, return-type dispatch, `setposition()` BH register-bank init. `production_blackhole_handles` predicate returns `false` for every opcode today; the wire is unreachable until #390 sub-slice 6 retires it once StoreSubscr activation lands on the orthodox path.

`pyre/check.py` 41/41 green on dynasm + cranelift at HEAD `0a4cfca1bc`.

## Self-review

This patch is not AI-generated. (Assisted by Claude across the session captured in the linked memory note; commits authored by Jeong, YunWon.)

### DIFF
--- a/majit/majit-metainterp/src/executor.rs
+++ b/majit/majit-metainterp/src/executor.rs
@@ -816,6 +816,67 @@ pub fn execute_pure_call(
     }
 }
 
+/// `executor.execute_varargs` parity for the walker layer, simplified
+/// for callers that do not hold a `MetaInterp` (record-time concrete
+/// execution of residual_calls).
+///
+/// PyPy upstream (`rpython/jit/metainterp/pyjitpl.py:1995-2126`
+/// `do_residual_call`) calls `executor.execute_varargs(opnum,
+/// argboxes, descr, exc=can_raise, pure=is_elidable)` for every
+/// residual_call regardless of EI branch — concrete execution always
+/// runs at trace-record time, only the recorded opcode kind and the
+/// post-call guard emission differ.  See `select_residual_call_opcode`
+/// (`pyre-jit-trace/src/jitcode_dispatch.rs`) for the per-EI-branch
+/// inventory.
+///
+/// The walker has no `MetaInterp` to thread; instead it owns a
+/// `WalkContext.last_exc_value` shadow.  This function therefore
+/// returns the BH_LAST_EXC_VALUE seam value via `Result::Err` so the
+/// caller can write it into the walker's exc shadow + emit
+/// `GUARD_NO_EXCEPTION`, mirroring upstream's
+/// `metainterp.execute_raised(bh_exc, constant=False)` +
+/// `handle_possible_exception` flow.
+///
+/// Differences from `execute_varargs`:
+///   * No `MetaInterp` argument; caller handles exception state.
+///   * No `cond_call` / `cond_call_value` dispatch — residual_call
+///     opcodes are not cond-call variants.
+///   * Returns `Result<i64, i64>` instead of side-effecting
+///     `metainterp.execute_raised`.
+///
+/// Differences from `execute_pure_call`:
+///   * No `is_elidable + cannot_raise` debug_assert.
+///   * Clears BH_LAST_EXC_VALUE before dispatch (`execute_pure_call`
+///     does not because elidable_cannot_raise cannot raise).
+pub fn execute_residual_call(
+    descr: &dyn majit_ir::descr::CallDescr,
+    func_ptr: i64,
+    args: &[i64],
+) -> Result<i64, i64> {
+    crate::blackhole::BH_LAST_EXC_VALUE.with(|c| c.set(0));
+    let func_ptr = func_ptr as *const ();
+    let result = match descr.result_type() {
+        majit_ir::Type::Int | majit_ir::Type::Ref => {
+            crate::pyjitpl::call_int_function(func_ptr, args)
+        }
+        majit_ir::Type::Void => {
+            crate::pyjitpl::call_void_function(func_ptr, args);
+            0
+        }
+        // See `execute_varargs`'s Float arm for the i64-bits ABI
+        // rationale: `#[jit_module]` Float helpers expose
+        // `concrete_ptr` as `extern "C" fn(...) -> i64` with the f64
+        // pre-packed via `f64::to_bits`.
+        majit_ir::Type::Float => crate::pyjitpl::call_int_function(func_ptr, args),
+    };
+    let bh_exc = crate::blackhole::BH_LAST_EXC_VALUE.with(|c| {
+        let v = c.get();
+        c.set(0);
+        v
+    });
+    if bh_exc != 0 { Err(bh_exc) } else { Ok(result) }
+}
+
 #[cfg(test)]
 mod execute_pure_call_tests {
     use super::*;
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
@@ -3514,6 +3514,24 @@ fn try_execute_residual_call_via_executor(
         Some(majit_ir::Value::Int(addr)) => addr,
         _ => return None,
     };
+    // Sub-slice 4 safety gate — reject `symbolic_fnaddr_for_path`
+    // placeholder values that escaped runtime patching.  Pyre's
+    // codewriter mints a 64-bit hash of the helper's `CallPath` when
+    // the build-time `pyre_interpreter::jit_trace_fnaddrs()` snapshot
+    // has no entry for it (`majit-translate/src/jit_codewriter/call.rs:
+    // 4926 symbolic_fnaddr_for_path`).  `runtime_fnaddr_patch` rewrites
+    // these to real runtime addresses only when the path appears in
+    // both the build-time and runtime registries; helpers absent from
+    // the runtime registry retain the hash and dereferencing it as a
+    // code address SIGBUSes.  Valid user-space code addresses fit in
+    // 47 bits on macOS/Linux 64-bit (canonical low half); hash values
+    // typically have bits ≥ 47 set.  Reject anything outside that
+    // range — both the symbolic-hash leak class AND any other stray
+    // non-fnptr value (e.g. an int constant mistakenly routed through
+    // the funcbox slot).
+    if (func_ptr as u64) >> 47 != 0 {
+        return None;
+    }
     if allboxes.len() - 1 > majit_translate::jit_codewriter::insns::MAX_HOST_CALL_ARITY {
         return None;
     }
@@ -4167,8 +4185,7 @@ fn walker_record_guard_exception(ctx: &mut WalkContext<'_, '_>, pc: usize) {
     let exc_obj = match ctx.last_exc_value_concrete {
         ConcreteValue::Ref(p) if !p.is_null() => p,
         _ => {
-            ctx.trace_ctx
-                .record_guard(OpCode::GuardNoException, &[], 0);
+            ctx.trace_ctx.record_guard(OpCode::GuardNoException, &[], 0);
             walker_capture_snapshot_for_last_guard(ctx, pc);
             return;
         }
@@ -4533,31 +4550,26 @@ fn dispatch_residual_call_iRd_kind(
         // their heap mutation because `eval.rs:3285-3308`'s walker-skip
         // path bypasses `execute_opcode_step` → SIGBUS on the next read
         // of the un-mutated container (M4 walker unactivated taxonomy).
-        // Task #390 sub-slice 3 activation is gated on `!can_raise`
-        // pending sub-slice 4's helper-side audit.  Empirical probe
-        // (commit history: sub-slice 4 probe) found the
-        // `can_raise=true` SIGBUS class is NOT walker-shadow staleness
-        // but UNPATCHED build-time fnaddrs in `constants_i` for some
-        // non-elidable helpers.  `runtime_fnaddr_patch` covers all
-        // helpers whose path appears in `pyre_interpreter::
-        // jit_trace_fnaddrs()`; helpers outside that registry retain
-        // garbage build-time values (e.g. `0xacc1d48d7993df7d` for
-        // `synth/bytes_ops`'s CallR).  Sub-slice 4 will extend the
-        // patcher registry or add a fnaddr-sanity gate before lifting.
-        let resid_raised = if !can_raise {
-            matches!(
-                try_execute_residual_call_via_executor(
-                    ctx,
-                    call_opcode,
-                    &allboxes,
-                    call_descr,
-                    recorded,
-                ),
-                Some(Err(_))
-            )
-        } else {
-            false
-        };
+        // Task #390 sub-slice 4: PyPy-orthodox activation.  PyPy's
+        // `_opimpl_residual_call*` concrete-executes EVERY residual
+        // call regardless of EI; the `exc` flag only selects the
+        // post-call guard shape (`GUARD_EXCEPTION` vs `GUARD_NO_EXCEPTION`)
+        // in `handle_possible_exception`.  Pyre matches by always
+        // invoking the executor — `try_execute_residual_call_via_executor`
+        // self-gates on a fnaddr-sanity check (rejecting unpatched
+        // `symbolic_fnaddr_for_path` hashes whose bits ≥ 47 are set)
+        // so unregistered helpers degrade gracefully to recording-only
+        // instead of SIGBUSing.
+        let resid_raised = matches!(
+            try_execute_residual_call_via_executor(
+                ctx,
+                call_opcode,
+                &allboxes,
+                call_descr,
+                recorded,
+            ),
+            Some(Err(_))
+        );
         debug_assert!(
             !resid_raised || can_raise,
             "dispatch_residual_call_iRd_kind: helper raised on a \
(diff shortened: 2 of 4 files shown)


youknowone:
Addressed the actionable items from the Codex + CodeRabbit reviews on this round in commit `161f2c1e9c`:

**Fixed in the live residual-call path (`pyre/pyre-jit-trace/src/jitcode_dispatch.rs`):**
- **Codex P1 — "Terminate after raising residual calls"**: 3 dispatchers (`iRd_kind` / `iIRd_kind` / `iIRFd_kind`) now return `DispatchOutcome::SubRaise { exc, exc_concrete }` from the `resid_raised` branch after `walker_record_guard_exception`, instead of falling through to `Continue`. Matches `pyjitpl.py:2156-2168 handle_possible_exception → finishframe_exception`. The outer `walk_loop` already routes top-level `SubRaise` into a `FINISH(exc) + Terminate`, so this terminates the trace correctly without recording dead arm bytecode onto the exception path.
- **CodeRabbit major — "Clear walker exception slot before successful residual call"**: `try_execute_residual_call_via_executor` now resets `ctx.last_exc_value = None` and `ctx.last_exc_value_concrete = ConcreteValue::Null` at the start of the `Ok` arm, mirroring `pyjitpl.py:1685-1690 _opimpl_residual_call*`'s implicit `clear_exception()`. Prevents sticky exception state from leaking past a clean call.

**Fixed in the inventory comment:**
- **Codex Section 2 D — "do_residual_call always calls execute_varargs is an overstatement"**: comment now enumerates the two narrow side branches (`OS_NOT_IN_TRACE` short-circuits via `do_not_in_trace_call` at `pyjitpl.py:2003-2006`; release-gil runs through `do_call_release_gil` at `pyjitpl.py:3671-3681`) instead of claiming literal universality.

**Documented as latent (Phase 5.B body wire is unreachable — predicate `false` everywhere, slated for retirement in #390 sub-slice 6):**
- **Codex Section 2 A** — `dispatch_arm_via_blackhole` has `todo!()` on `BhReturnType::Int|Ref|Float` (only `Void` implemented; opcode arms usually end `ref_return/r`).
- **Codex Section 2 B** — `bh.aborted` ignored; pyre's blackhole `abort/`/`abort_permanent/` (`majit/majit-metainterp/src/blackhole.rs:237`) would be misclassified as normal completion.
- **Codex Section 2 C** — `jitdrivers_sd` not seeded; recursive-call arms would diverge from `rpython/jit/metainterp/blackhole.py:1095`.
- **CodeRabbit major (eval.rs:3108)** — `BH_LAST_EXC_VALUE` not drained on the arm-blackhole path.
- **CodeRabbit major (jitcode_runtime.rs:337)** — `arm_id_for_instruction` does `format!("{instruction:?}")` + O(n) scan; hot-path lookup cost.

All five are now explicitly listed in `dispatch_arm_via_blackhole`'s doc-comment under "Latent gaps before this path can be activated." Body-wire retirement (#390 sub-slice 6) deletes the function entirely, so any activation flip would surface these before crashing in production.

**Section 3 (pre-existing mismatches)** all sit in the `Task #390` epic this PR opened and explicitly does not finish — non-pure residual-call concrete execution (sub-slice 5b heapcache work, next session), eval-loop/walker structural split (not in scope), loopinvariant placeholder concrete, and the deferred residual-call subcases (`OS_JIT_FORCE_VIRTUAL` / libffi / assembler_call / KEEPALIVE / num_live-aware `capture_resumedata`).

**CodeRabbit nitpick (executor.rs:878 — seam tests for `BH_LAST_EXC_VALUE` clear/read/clear)**: deferred — straightforward to add but not blocking; will land alongside the next sub-slice 5b work that exercises the executor's exception arm under activation.

Gate dynasm + cranelift 41/41 × 2 GREEN at `161f2c1e9c`. NOT pushed.

— *commented by Claude*

Issue #47, opened 2026-05-14
Retire BoxPool — box identity via Op/InputArg `_forwarded`


youknowone:
## Investigation outcome

PyPy parity for forwarding state requires `_forwarded` to live on the `AbstractResOpOrInputArg` object itself (`rpython/jit/metainterp/resoperation.py:233-242`), not on a parallel pyre-only Box wrapper. **Box identity = forwarded storage on the Op/InputArg object.**

## Direction (settled 2026-06-04)

A 7-agent RPython-grounded, two-way adversarial parity adjudication settled the operand-representation question: **`BoxRef`-in-`Op.args` is the orthodox direction**, high confidence (details in the status comment below).

- RPython `_args[i]` holds the operand object **directly**, with no integer-position layer (`resoperation.py:785` `self._arg0, = args`; `:793-795` `getarg` → `self._arg0`). pyre realizes this with `Op.args: SmallVec<[BoxRef; 3]>` (`resoperation.rs:1255`, commit `d6d13855`) — a single operand list of identity-bearing references.
- The competing approach — keep `Op.args` as `OpRef` integer indices and route identity through a parallel `arg_handles`/`ArgRef` strong-handle layer (or an `OptContext::get_forwarded(opref)` accessor) — is **not pursued**. It ships parallel operand structures against RPython's single `_args`, and its handle layer has zero production readers. The memory model is identical either way (both route `_forwarded` through the same `Weak`-backed `Forwarded` enum, and `ArgRef::get_box_replacement` delegates to `BoxRef`), so there is no offsetting benefit; `Weak<Op>` is in fact the truer model of a GC-kept `_forwarded` reference, since `optimizer.py:650-652` mutates args in place and liveness comes from the trace root (`history.py:528`), not from `_args` ownership.

Three orthogonal sub-goals, all converging on object identity:

1. **Where `_forwarded` lives** — on the `Op` / `InputArg` object only, no parallel slot. **Done** (S-0).
2. **What handle type production carries** — currently `BoxRef` (a thin `Rc<Box>` routing to the bound `Op` / `InputArg`); final form carries `Rc<Op>` / `Rc<InputArg>` / `Const` directly (S-11).
3. **`OpRef` survives only at the edges** — serialization / backend export / trace persistence; it is removed from in-optimizer identity (keystone `get_box_replacement` → `BoxRef` + the `VecAssoc` rekey).

## Status (2026-06-04)

[...]

PR #153, opened 2026-06-08
Box-identity: retire to_opref crutches + Forwarded::Box, add Operand union (#47)

youknowone:
Part of #47

## Summary

One step in the #47 box-identity program. It removes several long-standing
position-based fallbacks — `to_opref()`-comparison "crutches", non-canonical
`from_opref` box mints, and the `Forwarded::Box` redirect variant — so that
operand identity resolves only through canonical, `Rc`-bound boxes, and it
adds a new `majit-ir::operand::Operand` union as a **dormant** foundation for
the eventual `Op.args` storage flip.

This is **not** the completion of #47: the `Operand` type is added but not yet
wired into storage (`Op.args` is still `RefCell<SmallVec<[BoxRef; 3]>>` at this
PR's tip), and the side-table re-keying and the storage flip itself remain as
follow-up.

A few commits are behavior-changing on a path the commit messages claim was
empirically never hit (probes over the bench corpus + the metainterp lib
tests). Those "0-hit" / "530-hit" figures are commit-message assertions and are
not reproduced in the diffs themselves; weight them accordingly. Where a
previously-silent fallback is removed, the new code either routes through
canonical materialization or `panic!` / `unreachable!()`, so a wrong invariant
surfaces as a crash rather than a silent miscompile.

### `same_box` no longer falls back to position equality
- **`c6ba9c6`** — drops the OR-ed `|| a.to_opref() == b.to_opref()` arm from
  `intbounds.rs` (`autogen_eq_b` plus the four arithmetic postprocessors
  `postprocess_int_add`, `postprocess_int_add_ovf`, `optimize_int_sub_ovf`,
  `postprocess_int_mul_ovf` x+x / x−x checks), leaving the bare `same_box`
  form and removing the stale doc paragraph. The deleted arm could only ever
  return `true` in *more* cases, so dropping it is decisive exactly when two
  distinct (non-`Rc::ptr_eq`, non-both-const) boxes share a position+type;
  safety rests on the probe claim that this arm was never decisive.
- **`316cc62`** — reduces `OptContext::same_box` to the resolved-box case
  only: `true` only when both operands resolve via `get_box_replacement_box`
  and `BoxRef::same_box` agrees, `false` when either side fails to resolve —
  deleting the unresolved-OpRef-equality and const-value fallbacks (the const
  fallback was already largely redundant, since a `Const` OpRef resolves
  through `resolve_to_boxref` into the surviving `(Some, Some)` arm). The
  `pure.rs` edits are `#[cfg(test)]`-only fixture bindings.

### Canonicalization soundness (remove the last non-canonical box mints)
- **`aec52f3`** — extracts a shared `install_canonical_producer` helper and
  routes `supersede_restart_producer`, `register_extra_producer`,
  `emit_extra`, `send_extra_operation`, and `send_extra_operation_after`
  through it. Two real behavior changes: (1) a superseded stand-in's
  accumulated `_forwarded` is now *migrated* onto the new producer (cloned,
  `Rc::ptr_eq`-guarded against self-clone) instead of being dropped by the old
  `swap_remove`; (2) `register_extra_producer` no longer skips when a stand-in
  is already registered, so a queued/dispatched op now *supersedes* it. The
  migration is a blind `*op.forwarded.borrow_mut() = carried` overwrite whose
  correctness rests on a comment-only (not `debug_assert`-enforced) invariant
  that a producer-less stand-in holds only `None`/`Info`, never a redirect.
  The `getnullness` hunk is a comment-only rewrite.
- **`01df77e`** — makes the two non-const lookup-miss arms of
  `recorder.rs::box_for_operand` (InputArg, ResOp) `#[cfg(not(test))] panic!`
  instead of minting a position-only `BoxRef::from_opref`; `#[cfg(test)]`
  builds keep the fallback. This is wired into the production recorder
  (reached via `box_args` from `record_op` / `record_guard` / `record_jump` /
  `finish`), so a non-dense operand now crashes rather than silently
  fabricating a box. The paired `mod.rs` change is a `get_box_replacement`
  **doc-comment-only** rewrite — its runtime `from_opref` fallback (the
  load-bearing ~530-hit peel-boundary path) is explicitly left in place.
- **`69b3fc5`** — normalizes `newop` in `OptContext::make_equal_to` before
  forward dispatch: a non-constant box bound to neither an Op nor an InputArg
  is replaced by `materialize_box_at(newop.to_opref())`, and the former `else`
  arm (which did `op.set_forwarded_box(...)`) becomes `unreachable!()`. This
  retires the **last non-test production writer** of `Forwarded::Box`.

### Deletion of the `Forwarded::Box` variant
- **`a70de80`** — deletes the `Forwarded::Box(BoxRef)` enum variant and its
  sole (now test-only) writer `set_forwarded_box`, removes the
  `get_box_replacement_impl` chain-walk reader arm and the four metainterp
  reader arms (`has_op_forwarding`, `getrawptrinfo`, `getptrinfo`,
  `getnullness`), and migrates the `box_ref` / metainterp fixtures to
  `set_forwarded_op` / `set_forwarded_const`. Behavior-preserving for
  production — the production retirement already happened in `69b3fc5`, so this
  removes already-dead code; the bulk of the diff is comment rewording. Two
  caveats: the migrated dual-write tests drop their `Forwarded::Box`-slot
  mirror assertions rather than replacing them with `Forwarded::Op`
  equivalents (a small test-coverage reduction), and the always-on
  self-cycle / `Const` `assert!`s inside the deleted `set_forwarded_box` go
  with it (they guarded a test-only function).

### S9 measurement probe (diagnostic, behavior-preserving)
- **`7e14c0f`** — restructures `OptContext::get_box_replacement` into an
  explicit `if-let` that, on the `from_opref` fallback path, calls a new
  `&self` classifier (`classify_s9_fallback`) and an env-gated sink
  (`s9_probe_fire`) before returning the **same** `BoxRef::from_opref(opref)`
  as before. Pure instrument: both new methods are dead unless `PYRE_S9_PROBE`
  is set; when set, `s9_probe_fire` appends to a fixed `/tmp` path. The
  reported fire counts are commit-message-only. No re-keying is done here.

### Operand-union foundation (added, **not** wired into storage)
- **`00118ed`** — adds the dormant `majit-ir::operand::Operand` enum (`None` /
  `Op(OpRc)` / `InputArg(InputArgRc)` / `Const`) carrying producers by strong
  `Rc`, with accessors, constructors, `to_opref` / `to_boxref` / `same_box`,
  and unit tests; registers the module in `lib.rs`. **Zero non-test references
  outside its own module** — it changes no runtime behavior. (`same_box`'s
  `Const` branch uses derived `PartialEq` float equality, matching the
  pre-existing `BoxRef`-const behavior, so it is not a regression, though the
  "mirror" doc claim is imprecise for floats.)
- **`71b926b`** — adds the migration `Operand::Box(BoxRef)` catch-all variant
  with delegating accessor arms, a `from_boxref` classifier, an
  `Operand::walk_const_ptr_refs` (acts only on the `Box` arm), and rewrites
  `same_box` to route both sides through `to_boxref().same_box()`. Still
  dormant: `Op.args` remains `RefCell<SmallVec<[BoxRef; 3]>>` at this PR's tip,
  the GC walk still dispatches to `BoxRef::walk_const_ptr_refs`, and
  `from_boxref` / `Operand::walk_const_ptr_refs` have no non-test callers. The
  actual `Op.args` storage flip is a later commit, not included in this PR.

### Prompt & Model

Model: 

Prompt: 

## Summary by CodeRabbit

* **Refactor**
  * Unified forwarding and operand representation across the optimizer and interpreter, standardizing how values and producers are tracked and reducing subtle mismatches.
  * Emitted operations now consistently materialize argument values before use, improving optimization correctness and stability.

* **Public API**
  * Exposes a new operand abstraction module and adds a helper to check whether an argument slot is bound.

* **Tests**
  * Test suite updated to reflect the new forwarding and materialization behavior.

### DIFF
--- a/majit/majit-gc/src/rewrite.rs
+++ b/majit/majit-gc/src/rewrite.rs
@@ -732,7 +732,9 @@ impl RewriteState {
         }
         if let Some(fail_args) = rewritten.fail_args_mut() {
             for arg in fail_args.iter_mut() {
-                *arg = Operand::Box(self.resolve(arg.to_boxref()));
+                // Same shed as `setarg` above: a forwarding target bound to
+                // its producer stays a live-tracking operand.
+                *arg = Operand::from_boxref(&self.resolve(arg.to_boxref()));
             }
         }
         rewritten.pos.set(OpRef::NONE);
--- a/majit/majit-gc/src/rewrite.rs
+++ b/majit/majit-gc/src/rewrite.rs
@@ -12,6 +12,7 @@
 use majit_ir::Type;
 use majit_ir::box_ref::BoxRef;
 use majit_ir::descr::{DescrRef, FieldDescr, SizeDescr};
+use majit_ir::operand::Operand;
 use majit_ir::resoperation::{Op, OpCode, OpRef};
 use majit_ir::{Value, VecAssoc, VecSet};
 
@@ -731,7 +732,7 @@ impl RewriteState {
         }
         if let Some(fail_args) = rewritten.fail_args_mut() {
             for arg in fail_args.iter_mut() {
-                *arg = self.resolve(arg.clone());
+                *arg = Operand::Box(self.resolve(arg.to_boxref()));
             }
         }
         rewritten.pos.set(OpRef::NONE);
@@ -954,7 +955,7 @@ impl GcRewriterImpl {
         // guard for the next iteration to pick up.
         let mut new_guard = op.clone();
         if let Some(fa) = new_guard.fail_args_mut() {
-            fa[idx] = same_pos;
+            fa[idx] = Operand::Box(same_pos);
         }
         // pos is reassigned when emit/emit_result runs on the substituted op.
         new_guard.pos.set(OpRef::NONE);
--- a/majit/majit-ir/src/box_ref.rs
+++ b/majit/majit-ir/src/box_ref.rs
@@ -19,11 +19,10 @@
 //!   the borrow scope immediately after reading.
 //! - `BoxRef`'s `Eq` / `Hash` use `Rc::ptr_eq` / `Rc::as_ptr` — equivalent
 //!   to RPython's use of object identity as a dict key.
-//! - When `Forwarded::Box(BoxRef)` carries a BoxRef whose kind is
-//!   `BoxKind::Const(...)`, that mirrors RPython's
-//!   `box.set_forwarded(constbox)`. Constants are split into
-//!   `Forwarded::Const(Const)` separately so chain walkers terminate on
-//!   the inline value without needing a `BoxKind::Const` carrier.
+//! - A constant forwarding target mirrors RPython's
+//!   `box.set_forwarded(constbox)` via the `Forwarded::Const(Const)`
+//!   variant, so chain walkers terminate on the inline value without
+//!   needing a `BoxKind::Const` carrier.
 
 use std::cell::{Cell, RefCell};
 use std::rc::{Rc, Weak};
@@ -88,7 +87,7 @@ pub enum BoxKind {
     /// RPython counterpart on `AbstractResOpOrInputArg` (which carries
     /// only `_forwarded`); it stores the index where the box lands in the
     /// pool so the chain walker can reconstruct an `OpRef::op_typed(pos,
-    /// type)` when advancing through `Forwarded::Box(target)`. Set by
+    /// type)` when advancing through a `Forwarded::Op` chain step. Set by
     /// `opencoder.py Trace.record()` (and the recorder's `record_op_*`
     /// family in pyre) at construction; updated by `optimizer.rs:2783`
     /// const-pool compaction via `BoxRef::set_position`. RPython has no
@@ -134,11 +133,6 @@ pub enum BoxKind {
 #[derive(Clone, Debug)]
 pub enum Forwarded {
     None,
-
-    /// Forwarding to another `AbstractResOpOrInputArg`. Const targets
-    /// route through [`Forwarded::Const`] instead; `ResOp` targets route
-    /// through [`Forwarded::Op`] once C.5 retires this variant.
-    Box(BoxRef),
 
     /// `resoperation.py:250 AbstractResOp` forwarding — direct
     /// `Weak<Op>` reference, no `BoxRef`/`BoxKind::ResOp` carrier.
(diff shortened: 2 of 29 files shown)


youknowone:
## Parity-note adjudication (empirical)

The static review raised 5 notes. Rather than reason about them only statically, I
reconstructed each *removed* fallback behind an env gate (`PYRE_REV_PROBE`) and ran
the **full bench corpus on both backends (dynasm 44/44 + cranelift 44/44)**, counting
every production case where the old code would have diverged from the new code. This
supplies the reproduction the commit-message "0-hit" claims lacked.

### §1.1 — `OptContext::same_box` no longer falls back to OpRef/const equality — **REFUTED**

Probe placed on the new `_ => false` arm, recomputing the retired `316cc62` fallback
(`get_box_replacement(..).to_opref()` equality, then const-value compare) and logging
every case where it would have returned `true`.

**Fires across the whole corpus × both backends: 0.** The removed fallback was never
decisive in production. Both operands resolve through `get_box_replacement_box`, and
when either fails to resolve the old fallback never recovered a `true` that the new
`(Some, Some) => same_box` path misses.

### §1.2 — intbound self-operand checks depend on canonical box recovery — **REFUTED**

Probe placed inside `BoxRef::same_box`'s non-const `_ => false` arm, reconstructing the
retired `c6ba9c6` crutch (`box1.to_opref() == box2.to_opref()`): it logs any pair where
`same_box` returns `false` yet the two boxes share a position+type.

**Fires: 0.** For the `x + x` / `x - x` patterns (`intbounds.rs:42/1746/2089/2105/2145`),
the two operands referencing one producer are canonicalized to a single `Rc<Op>`
(memoized on `Op::box_cache`), so `same_box` returns `true` by `Rc::ptr_eq` alone — the
`to_opref()` crutch never matched a distinct-identity pair. The self-operand
optimizations still fire.

### §1.3 — `make_equal_to` forwards to a materialized stand-in, not the exact newop — **REFUTED**

Probe placed on the materialize-unbound branch, logging the unbound `newop`'s forwarding
state before `materialize_box_at`.

**Fires: 4 (across corpus), every one with `fwd=None`.** The branch is live but lossless:
an unbound position-only box structurally cannot hold `_forwarded` (`get_forwarded()`
returns `None` by construction for a box with no bound producer), and
`materialize_box_at(newop.to_opref())` mints the canonical box **at newop's position** —
the same box every other resolution of that position yields. So `op.set_forwarded(stand-in)`
establishes exactly the position-aliasing that upstream's `op.set_forwarded(newop)`
(`optimizer.py:392-401`) would. There is no info to transfer (upstream's
`if opinfo is not None: newop.set_forwarded(opinfo)` has no analogue to drop here, since
`opinfo` belongs to `op`, not `newop`). The former `else` arm is now `unreachable!()` and
never panicked across the corpus.

### §2.1 — `getnullness` treats missing producer boxes as `INFO_UNKNOWN` — **PRE-EXISTING (not introduced by this PR)**

The `get_box_replacement_box(opref) → None → INFO_UNKNOWN` guard
(`optimizer.rs` `getnullness(ctx, opref)`) predates this branch: its lines blame to
`ce789c96bab` (2026-05-12) and `d3bfd1724ba`, both ancestors of `main`. The only thing
this PR's `aec52f3` touched there was the **doc comment** (rewritten to describe the
existing `INFO_UNKNOWN` path). The behavior is parity-faithful: `optimizer.py:130-131`
returns `INFO_UNKNOWN` when `getptrinfo(op)` (create=False) is `None`, and a
producer-less position legitimately has no info — a nullness query must not mint a
producer/pollute the registry.

### §2.2 — debug probe `/tmp` side effect — **acknowledged instrumentation, inert by default**

The `7e14c0f` S9 probe is gated on `PYRE_S9_PROBE`; both the classifier and the file
sink are dead unless the env var is set. It is retained as measurement scaffolding for
the box-identity program and can be dropped before merge if preferred. (The probe used
for *this* adjudication was temporary and has been reverted — the tree is unchanged.)

---

**Net:** 0 of the 3 flagged parity regressions reproduce in production; §2.1 is
pre-existing and upstream-faithful; §2.2 is inert instrumentation. Corpus green on both
backends.

— *commented by Claude*

PR #158, opened 2026-06-08
JIT walker: direct dispatch for STORE_SUBSCR, CallMayForce, and Reraise

### LINKED ISSUE #122: Tagged-int representation for inline Python small ints, with Z2.5 fallback-map cutover as a bundled sub-epic

youknowone:
Fix https://github.com/youknowone/pyre/issues/122

## Summary

- STORE_SUBSCR walker activation: entry hook in `dispatch_via_walker_for_opcode` pops 3 values and delegates to `MIFrame::store_subscr_value`, bypassing the arm-jitcode walk that pops from a concrete PyFrame stack the walker never populates.
- `try_execute_residual_call_via_executor` admits `CallMayForce*` when no active virtualizable (matches `do_residual_call`'s force-virtual branch with no vinfo).
- Reraise walker activation delegates to `MIFrame::reraise`, which seeds `err.reraise_lasti` from `concrete_stack` so the dispatcher's lasti extraction works the same on either leg. `dispatch_via_walker_for_opcode` gains an `op_arg` parameter.
- `dispatch_arm_via_blackhole` + `production_blackhole_handles` + `METAINTERP_JITCODE_CACHE` deleted; the walker executor is now the sole non-elidable residual_call mechanism. `build_pyre_production_bh_builder` + `blackhole_control_opcodes` retained for the resume path.
- `try_walker_direct_opcode_dispatch` extracted as the named entry point for opcodes that emit IR directly from the tracer.

Gate: dynasm 44/44 + cranelift 44/44 ×2 backends.

### Prompt & Model

Model:

Prompt:

### Answer

  1. Regressed Parity
  No behavior-level PyPy parity regressions found versus origin/main.

  The new early SubRaise returns in pyre/pyre-jit-trace/src/jitcode_dispatch.rs:4653 move Pyre closer to PyPy: handle_possible_exception() records GUARD_EXCEPTION
  and immediately calls finishframe_exception() in rpython/jit/metainterp/pyjitpl.py:3404-3417, so later arm bytecode should not be walked.

  2. Other Introduced Mismatches
  No new behavior mismatch found.

  Two new comments are misleading:

  - pyre/pyre-jit-trace/src/jitcode_dispatch.rs:3582 says residual-call success clears exception state at pyjitpl.py:1685-1690; upstream clears before execution in
    MIFrame.execute_varargs() at pyjitpl.py:1941-1943, and in the forces branch at pyjitpl.py:2010.

  - The release-GIL comment in the residual-call inventory overstates direct_call_release_gil: upstream concrete execution happens first via
    executor.execute_varargs(... CALL_MAY_FORCE_*) at pyjitpl.py:2023-2041; direct_call_release_gil() later records CALL_RELEASE_GIL_* at pyjitpl.py:3700-3705.

  3. Pre-existing Mismatches

  - pyre/pyre-jit-trace/src/jitcode_dispatch.rs:3493 still does not execute every residual call. It declines non-supported opcodes, non-const funcptrs, unpatched/
    high fnaddrs, over-arity calls, missing concrete args, and NULL refs. PyPy executes residual calls at trace-record time in pyjitpl.py:1995-2126.

  - If that helper returns None, Pyre still does not mirror upstream clear_exception() for that residual call. The patch only clears on Some(Ok(_)).
  - CallMayForce* and CallReleaseGil* concrete execution remain excluded from the walker executor path, while PyPy’s forces branch concretely executes first.
  - Existing documented gaps remain: OS_JIT_FORCE_VIRTUAL, direct_libffi_call, full virtualizable/vref after-call handling, direct assembler-call handling, and
    exact multi-frame guard snapshots.

  4. Structural Adaptations

  - pyre/pyre-interpreter/src/opcode_ops.rs:231 is a Rust ABI bridge, not a line-by-line PyPy shape. PyPy’s STORE_SUBSCR just performs space.setitem and returns
    through interpreter control flow; Pyre needs a C-ABI callable that publishes exceptions through BH_LAST_EXC_VALUE.

  - The bridge returns 1/0, while the opcode-arm JitCode has a Ref-shaped StepResult::Continue return. This is acceptable only because production opcode-entry
    dispatch treats any SubReturn as StepResult::Continue and ignores the returned Ref.

  - The BH_LAST_EXC_VALUE TLS, runtime fnaddr patching, and 47-bit fnaddr sanity gate are implementation-language/runtime adaptations, not RPython structures.
  - Single-frame walker snapshots are a deliberate Pyre adaptation because helper/sub-jitcodes are not blackhole-resumable Python frames.

## Summary by CodeRabbit

* **Performance**
  * Faster store-subscript (item assignment) paths and specialized residual-call dispatch to speed list/dict writes and JIT traces.

* **Bug Fixes**
  * More consistent exception propagation from residual helpers and JIT dispatch; fixed type-attribute deletion to actually remove keys instead of leaving tombstones.
  * Simplified walker-dispatched opcode handling to avoid double-mutating frames.

* **Debugging**
  * Optional environment-gated debug probes for store-subscript and certain heap-cache mismatches.

* **Other**
  * Fixed benchmark time formatting in reporting.

### DIFF
--- a/majit/majit-metainterp/src/trace_ctx.rs
+++ b/majit/majit-metainterp/src/trace_ctx.rs
@@ -854,6 +854,24 @@ impl TraceCtx {
         effectinfo: Option<&majit_ir::EffectInfo>,
         argboxes: &[OpRef],
     ) {
+        if std::env::var_os("PYRE_PROBE_SUBSCR").is_some() {
+            let ei_summary = effectinfo.map(|ei| {
+                format!(
+                    "extraeffect={:?} forces_vorv={} can_raise={} plain_call={} oopspec={:?}",
+                    ei.extraeffect,
+                    ei.check_forces_virtual_or_virtualizable(),
+                    ei.check_can_raise(false),
+                    opnum.is_plain_call(),
+                    ei.oopspecindex,
+                )
+            });
+            eprintln!(
+                "[PYRE_PROBE_SUBSCR] invalidate_caches_varargs opnum={:?} argboxes.len={} ei={:?}",
+                opnum,
+                argboxes.len(),
+                ei_summary
+            );
+        }
         let oracle: &dyn majit_trace::heapcache::SameConstantOracle =
             &crate::history::ConstOprefOracle;
         let const_value = |opref: OpRef| match opref.inline_const_to_value() {
--- a/majit/majit-translate/src/codegen.rs
+++ b/majit/majit-translate/src/codegen.rs
@@ -1077,12 +1077,12 @@ pub fn generated_binary_int_value(
     let lhs_raw = if frame.value_type(a) == majit_ir::Type::Int {
         a
     } else {
-        crate::state::trace_unbox_int_with_resume(frame, ctx, a, int_type_addr)
+        crate::state::trace_unbox_int_with_resume(frame, a, int_type_addr)
     };
     let rhs_raw = if frame.value_type(b) == majit_ir::Type::Int {
         b
     } else {
-        crate::state::trace_unbox_int_with_resume(frame, ctx, b, int_type_addr)
+        crate::state::trace_unbox_int_with_resume(frame, b, int_type_addr)
     };
 
     // The inlined `_ovf_zer` wrapper (`rint.py:429 ll_int_py_div_
@@ -1309,7 +1309,7 @@ pub fn generated_binary_float_value(
             let raw_int = if frame.value_type(obj) == majit_ir::Type::Int {
                 obj
             } else {
-                crate::state::trace_unbox_int_with_resume(frame, ctx, obj, int_type_addr)
+                crate::state::trace_unbox_int_with_resume(frame, obj, int_type_addr)
             };
             let r = ctx.record_op(OpCode::CastIntToFloat, &[raw_int]);
             // Box(value) parity: derive concrete float from the int's
@@ -1405,14 +1405,14 @@ pub fn generated_compare_value_direct(
             } else if let Some(raw) = crate::state::try_trace_const_boxed_int(ctx, a, concrete_lhs) {
                 raw
             } else {
-                crate::state::trace_unbox_int_with_resume(frame, ctx, a, int_type_addr)
+                crate::state::trace_unbox_int_with_resume(frame, a, int_type_addr)
             };
             let rhs_raw = if frame.value_type(b) == majit_ir::Type::Int {
                 b
             } else if let Some(raw) = crate::state::try_trace_const_boxed_int(ctx, b, concrete_rhs) {
                 raw
             } else {
-                crate::state::trace_unbox_int_with_resume(frame, ctx, b, int_type_addr)
+                crate::state::trace_unbox_int_with_resume(frame, b, int_type_addr)
             };
             let truth = ctx.record_op(cmp, &[lhs_raw, rhs_raw]);
             // Box(value) parity: stamp the bool result from the operands'
@@ -1463,7 +1463,7 @@ pub fn generated_compare_value_direct(
                 let raw_int = if frame.value_type(a) == majit_ir::Type::Int {
                     a
                 } else {
-                    crate::state::trace_unbox_int_with_resume(frame, ctx, a, int_type_addr)
+                    crate::state::trace_unbox_int_with_resume(frame, a, int_type_addr)
                 };
                 let cast = ctx.record_op(majit_ir::OpCode::CastIntToFloat, &[raw_int]);
                 // Box(value) parity: derive concrete float from the int's
@@ -1473,7 +1473,7 @@ pub fn generated_compare_value_direct(
                 }
                 cast
             } else {
-                crate::state::trace_unbox_float_with_resume(frame, ctx, a, float_type_addr)
+                crate::state::trace_unbox_float_with_resume(frame, a, float_type_addr)
             };
             // Unbox rhs: same pattern
             let rhs_raw = if frame.value_type(b) == majit_ir::Type::Float {
@@ -1482,7 +1482,7 @@ pub fn generated_compare_value_direct(
                 let raw_int = if frame.value_type(b) == majit_ir::Type::Int {
                     b
                 } else {
-                    crate::state::trace_unbox_int_with_resume(frame, ctx, b, int_type_addr)
+                    crate::state::trace_unbox_int_with_resume(frame, b, int_type_addr)
                 };
                 let cast = ctx.record_op(majit_ir::OpCode::CastIntToFloat, &[raw_int]);
                 if let Some(majit_ir::Value::Int(n)) = ctx.box_value(raw_int) {
@@ -1490,7 +1490,7 @@ pub fn generated_compare_value_direct(
                 }
                 cast
             } else {
-                crate::state::trace_unbox_float_with_resume(frame, ctx, b, float_type_addr)
+                crate::state::trace_unbox_float_with_resume(frame, b, float_type_addr)
             };
             let truth = ctx.record_op(cmp, &[lhs_raw, rhs_raw]);
             // Box(value) parity: stamp the float-compare bool result.
@@ -1542,7 +1542,7 @@ pub fn generated_unary_int_value(
         value
     } else {
         let int_type_addr = &pyre_object::pyobject::INT_TYPE as *const _ as i64;
-        crate::state::trace_unbox_int_with_resume(frame, ctx, value, int_type_addr)
+        crate::state::trace_unbox_int_with_resume(frame, value, int_type_addr)
     };
     // intobject.py int_neg: guard against INT_MIN (overflow to long).
     if matches!(opcode, OpCode::IntNeg) {
@@ -1766,7 +1766,7 @@ pub fn generated_list_setitem_by_strategy(
                 value
             } else {
                 let float_type_addr = &pyre_object::pyobject::FLOAT_TYPE as *const _ as i64;
-                crate::state::trace_unbox_float_with_resume(frame, ctx, value, float_type_addr)
+                crate::state::trace_unbox_float_with_resume(frame, value, float_type_addr)
             };
             crate::state::trace_raw_float_array_setitem_value(ctx, items_ptr, index, raw);
         }
@@ -1899,10 +1899,10 @@ fn unbox_int_or_long_for_int_strategy(
     }
     if unbox_long {
         let long_type_addr = &pyre_object::pyobject::LONG_TYPE as *const _ as i64;
-        crate::state::trace_unbox_long_with_resume(frame, ctx, value, long_type_addr)
+        crate::state::trace_unbox_long_with_resume(frame, value, long_type_addr)
     } else {
         let int_type_addr = &pyre_object::pyobject::INT_TYPE as *const _ as i64;
-        crate::state::trace_unbox_int_with_resume(frame, ctx, value, int_type_addr)
+        crate::state::trace_unbox_int_with_resume(frame, value, int_type_addr)
     }
 }
 
@@ -1964,7 +1964,7 @@ pub fn generated_truth_value_direct(
             {
                 raw
             } else {
-                crate::state::trace_unbox_int_with_resume(frame, ctx, value, int_type_addr)
+                crate::state::trace_unbox_int_with_resume(frame, value, int_type_addr)
             };
             let zero = ctx.const_int(0);
             let truth = ctx.record_op(OpCode::IntNe, &[int_value, zero]);
@@ -1982,7 +1982,7 @@ pub fn generated_truth_value_direct(
                 raw
             } else {
                 crate::state::trace_unbox_int_with_resume_descr(
-                    frame, ctx, value, bool_type_addr,
+                    frame, value, bool_type_addr,
                     crate::descr::bool_boolval_descr(),
                 )
             };
@@ -2002,7 +2002,7 @@ pub fn generated_truth_value_direct(
         if pyre_object::is_float(concrete_val) {
             let float_type_addr = &pyre_object::pyobject::FLOAT_TYPE as *const _ as i64;
             let float_value = crate::state::trace_unbox_float_with_resume(
-                frame, ctx, value, float_type_addr,
+                frame, value, float_type_addr,
             );
             let zero = ctx.const_int(0);
             let zero_float = ctx.record_op(OpCode::CastIntToFloat, &[zero]);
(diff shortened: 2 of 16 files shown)


youknowone:
Review triage round 2 — 6 commits pushed (b659ac35 .. 83528afb).

**Applied**

- chatgpt-codex P1 partial / coderabbit `pyre-interpreter/src/opcode_ops.rs:315` — `bh_store_subscr_fn` now null-checks `value` alongside `obj`/`key`.
- coderabbit `pyre-jit-trace/src/state.rs:2091` + `jitcode_dispatch.rs` — `PYRE_PROBE_SUBSCR` env lookup cached behind a `OnceLock` (`probe_subscr_enabled`); state.rs branch checks `loaded != cached_int` first so the cache-hit path skips the probe entirely.
- coderabbit `pyre-jit-trace/src/jitcode_dispatch.rs:1814` — `dispatch_via_miframe` entry (test/shadow fixture, `outer_jitcode_index = 0`, empty `outer_active_boxes`) now sets `store_subscr_fn_addr: None` so a specialized `generated_store_subscr_value` guard can't attach resume data with the wrong outer-frame info there.
- coderabbit `pyre-jit-trace/src/jitcode_dispatch.rs:4612` — on `bh_store_subscr_fn` raise, drain `BH_LAST_EXC_VALUE` into `ctx.last_exc_value*`, emit `walker_record_guard_exception`, and return `DispatchOutcome::SubRaise` so the caller doesn't fall through to the generic residual-call dispatcher and double-record the call op.
- coderabbit `pyre-jit/src/eval.rs:3181` — `eval_loop_jit_bridge` now mirrors `eval_loop_jit`'s `walker_dispatched_this_opcode` check after `jit_merge_point_hook`: skip `execute_opcode_step` for allow-listed opcodes and drain `BH_LAST_EXC_VALUE` into `PyError`.
- coderabbit `pyre-interpreter/src/jit_fnaddr.rs:375` — regression test `jit_trace_fnaddrs_covers_store_subscr_helpers` asserts the bare `execute_store_subscr` key and both `bh_store_subscr_fn` aliases.
- coderabbit cargo fmt comments (`jitcode_dispatch.rs` + `eval.rs`) — already addressed in 346da16..378375a; `cargo fmt --all -- --check` is clean on the current HEAD.

**Deferred**

- chatgpt-codex P1 main concern (`trace_opcode.rs:7068` — STORE_SUBSCR walker direct-dispatch doesn't mutate the live container) / coderabbit Major #9 (`trace_opcode.rs:7069` — heap-backed frame stack pop sync). Attempted a minimal fix that calls `bh_store_subscr_fn` from `try_walker_direct_opcode_dispatch` after recording IR; that regressed `synth/list_index_update` / `synth/dict_lookup_update` (4 benches turned WRONG in check.py) because the live mutation during recording and the compiled-trace replay double-count the same iteration. The orthodox fix is the rest of Task #390 sub-slice 5c — porting strategy-aware specialization into the walker so the helper is concrete-executed exactly once per opcode along PyPy `_opimpl_*` lines. Reverted the partial attempt; leaving the existing trait-path delegation (records IR, compiled trace owns the heap effect at replay) which check.py validates as correct on all 44 benches.
- coderabbit `pyre-jit-trace/src/jitcode_dispatch.rs:4612` heavy-lift suggestion (rollback / defer IR before helper success) — the partial mitigation above (`SubRaise` on raise) prevents the double-IR symptom. A pre-call rollback or post-call rewrite still requires the sub-slice 5c restructure to determine the right IR shape on the raise path without mutating the heap before the strategy guard fires; kept as part of that epic.
- coderabbit `pyre-jit-trace/src/state.rs:2284` (use `WalkerFrameOps::guard_class`) — declined for this round. The trait helper emits `GuardNonnullClass`, while the three open-coded sites here intentionally emit plain `GuardClass` (matching `pyjitpl.py:1518 opimpl_guard_class` rather than `1542 opimpl_guard_nonnull_class`). Switching opcode shape needs separate empirical validation against trace-shape regressions.

`pyre/check.py` green twice in a row after the push (44/44 dynasm + 44/44 cranelift).

— *commented by Claude*

PR #178, opened 2026-06-14
x86 jit: break cvtsi2sd false dependency with a zeroing idiom

youknowone:
CVTSI2SD preserves the destination's high 64 bits, carrying a false dependency on the register's prior value. In a loop that reuses the same xmm this serialises the int->float conversion against the previous iteration. Emit a pxor zeroing idiom (eliminated at register rename) before the conversion to break it.

float_loop: dynasm 0.27s -> 0.14s (matches cranelift); nbody 0.44 -> 0.17s. Output unchanged: cvtsi2sd overwrites the low 64 bits, so zeroing the rest is semantically inert for scalar doubles.

- rpython/jit/backend/x86/assembler.py: PXOR before CVTSI2SD
- majit dynasm backend: same, both cast_int_to_float emission paths

### Prompt & Model

Model: 

Prompt: 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced performance of integer-to-float conversion operations on the x86_64 architecture by addressing register dependency issues that could cause performance degradation in loops and repeated operations.

### DIFF
--- a/majit/majit-backend-dynasm/src/x86/assembler.rs
+++ b/majit/majit-backend-dynasm/src/x86/assembler.rs
@@ -3170,7 +3170,12 @@ impl<'a> Assembler386<'a> {
                             16
                         }
                     };
-                    dynasm!(self.mc ; .arch x64 ; cvtsi2sd Rx(dst.value), Rq(sr));
+                    // cvtsi2sd preserves the destination's high 64 bits, so it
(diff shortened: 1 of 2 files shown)


youknowone:
upstream patch: https://github.com/pypy/pypy/pull/5508

PR #196, opened 2026-06-16
Z2.5 rtyper cutover: retire legacy-walker fallback (#266) + #164 review fixes + census 674→662

youknowone:
#122

## Summary

Continues the Z2.5 rtyper cutover (retiring the legacy-walker fallback) on top of #164, which landed on `main` as the squash commit `a83cb5cd91`. Five commits: three #164 review-fix folds, two dual-gate census-drop landings, a parity doc note, and the staged (unwired) #277 fmt-recognizer primitives.

### #164 review fixes (`dd32b65726`)

- **Discriminant fold** — `Rvalue::Discriminant` on a local bound directly to its payload by an always-`Ok` decomposed conversion (`try_lower_usize_try_from`) folds to `ConstInt(tag)` from `const_discriminant_locals` instead of reading a `__discriminant` field no runtime enum object carries. Covers the `match try_from(..)` shape.
- **`checked_neg` width gate** — `try_lower_checked_neg` decomposes to a compare against `i64::MIN`, correct only at word width, so it is now gated on a word-sized signed operand (`i64`/`isize`) via the callee's first input atom (`tyref_literal_int_atom`); a narrower `checked_neg` keeps the generic `Call` form.
- **range-list skip** — `SomeValue::List.rtyper_makerepr` reads `range_step` and skips range-backed lists (RangeRepr unported) rather than mislowering them through `FixedSizeListRepr`, whose `getarraysize` / item-load ops would compile to invalid reads off a non-array.

### Census-drop landings

- **Pre-mint unit-only enum variant classes** (`909757ee37`, census 674 → 664) — register every unit-only enum's variant subclasses before the single `assign_inheritance_ids` pass in `PyreCallRegistry::ensure_session`, so each enum-base + variants subtree is numbered as one contiguous bracket. A variant minted lazily at its `ClassesPBCRepr` instantiation (after the base's bracket was baked) was not append-safe, stayed unnumbered, and was Skip-classified. `class-not-numbered` 58 → 0; 10 graphs lift (`unpack_ex`, `PyError::kind_from_exc`, `excobject::is_exception` / `w_exception_get_kind`, `pyobject::get_instantiate` / `ll_isinstance` / `ll_issubclass`, `tupleobject::is_plain_float_strict`, `typeobject::w_type_get_uses_object_setattr`); zero newly-skipping.
- **`num::From` uint widen identity bind** (`4f922cb236`, census 664 → 662) — fold `core::convert::num::<Impl>::from` (Opaque in LLBC) to an identity bind when the source is a smaller-than-word unsigned (U8/U16/U32) and the destination is a word-sized integer carrier (I64/Isize/U64/Usize). Value-preserving, so no op is emitted — the same shape as the always-`Ok` `usize::try_from` payload bind. Lifts `pyopcode::u32_as_i64` and `typeobject::w_type_get_version_tag` from the legacy-walker fallback.

### Parity doc note (`68e87d6ede`)

`space.bool_w(w)` is `bool(int_w(w))` (`baseobjspace.py:1944`) — it unwraps an int and rejects non-ints, not the general `is_true` truth test — so the existing `int_w(...)? != 0` is the exact equivalent. Recorded on both `takewhile` / `dropwhile` `__setstate__` helpers to settle a recurring review misread.

### Staged #277 fmt-recognizer primitives, unwired (`274b19f538`)

Four `#[allow(dead_code)]` helpers in `front/mir.rs` for the deferred fmt-chain recognizer, with unit tests and an `#[ignore]` LLBC IR-dump probe; none are wired into lowering dispatch, so lowering output and the census are unchanged:

- `resolve_to_producer_op` — cross-block back-trace from a `Variable` to the producing `SpaceOperation` (or const), following single incoming `Link`s.
- `decode_packed_format_pieces` — decode the packed format-string template (literal `[len<0x80][bytes]` segments, `0xC0` placeholder, `0x00` terminator) into literal pieces + placeholder count; bails on unrecognized control bytes and non-UTF-8.
- `read_array_literal_elements` — read an `Array` aggregate ctor's `__pos_i` `FieldWrite`s in index order.
- `resolve_const_int` — resolve a `Variable` to its `ConstInt` value.

### Verification

Census 662 on `ssa-repr`. `check.py` 79/79 dynasm + 79/79 cranelift; `cargo test` lib 2503/0 + 251/0.

### Prompt & Model

Model: 

Prompt: 

## Summary by CodeRabbit

* **Improvements**
  * Pre-mints enum variant subclasses before inheritance-id assignment for consistent numbering.
  * Refines MIR lowering of local discriminants and numeric `from`/`checked_neg` conversions with tighter gating and correct masking/widening behavior.
  * Enhances `format_args!` format-chain recognition/recovery across block boundaries (including more end-to-end coverage).
  * Adds `len(range(...))` `RangeRepr` support and improves repr selection for range-list lowering (with clearer errors for unsupported steps).
* **Documentation**
  * Clarified `takewhile`/`dropwhile` iterator state semantics.
* **Tests**
  * Added/updated unit tests for range repr selection, helper graph generation, and fmt-chain reconstruction/output.

### DIFF
--- a/majit/majit-translate/src/annotator/bookkeeper.rs
+++ b/majit/majit-translate/src/annotator/bookkeeper.rs
@@ -634,9 +634,12 @@ impl Bookkeeper {
     /// Spelling MUST match the annotation path
     /// (`flowspace_adapter::translate_op` `SyntheticTransparentCtor` arm)
     /// exactly or the pre-mint is a phantom that never matches the
-    /// lazily-minted class: base = `intern_class_by_qualname(leaf)` (bare
-    /// enum leaf), variant = `intern_class_by_qualname_with_bases(
-    /// "{dotted_owner}.{variant}", [base])`.  `enum_variant_by_discriminant`
+    /// lazily-minted class: base = `intern_class_by_qualname(
+    /// canonical_struct_name(root))` (the same spelling
+    /// `getuniqueclassdef_for_enum_variant` uses, so both resolve to one
+    /// base `HostObject` independent of `STRUCT_ORIGIN_REGISTRY`), variant
+    /// = `intern_class_by_qualname_with_bases("{dotted_owner}.{variant}",
+    /// [base])`.  `enum_variant_by_discriminant`
     /// dual-publishes each enum under both its `::`-qualified path and
     /// its bare leaf; iterate the qualified keys only (the dotted owner
     /// path derives from `key.replace("::", ".")`).  Scoped to unit-only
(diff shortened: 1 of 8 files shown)


majit/majit-translate/src/annotator/bookkeeper.rs
```diff
@@ -583,6 +583,82 @@ impl Bookkeeper {
             .unwrap_or_default()
     }
 
+    /// TODO: no upstream equivalent.  Pre-mint the variant subclasses of
```
youknowone: red flag

PR #209, opened 2026-06-18
ci: save rust-cache on main only; finish LLBC artifact handoff

youknowone:
## Problem

PR #198's macOS CI jobs failed fast at the *Check prepared cache hits* guard: the prepared LLBC entry that `prepare-charon-llbc` saved was **LRU-evicted before the late-scheduled macOS consumers restored it**. The repo's Actions cache had exceeded GitHub's **10 GB limit** (~10–12 GB across `main` + open PR refs), and the dominant consumers are the per-ref Swatinem `rust-cache` target caches (~500–600 MB each × job × ref).

\#198 already moved the main consumers' LLBC handoff from `actions/cache` to `upload`/`download-artifact` (run-scoped, not evictable). This PR cuts the root footprint and finishes that migration.

## Changes

1. **rust-cache: save only on `main`.** `save-if: ${{ github.ref == 'refs/heads/main' }}` on all five Swatinem steps (prepare, cargo-test, pyre-check, cpython-tests, wasm-build). PR jobs become **restore-only** and pull `main`'s default-branch cache via `restore-keys`, so each PR no longer persists its own ~500–600 MB target cache.
   - Cost: a PR's own incremental build state is not carried across its pushes (it recompiles its diff from `main`'s baseline each run); deps and unchanged crates still come warm from `main`.
   - Hits work because GitHub lets any PR restore the **default-branch** cache; this is the pattern Swatinem documents for `save-if`.

2. **wasm-build: LLBC via artifact.** The one consumer #198's migration missed — it still restored LLBC with `actions/cache` and hard-failed on a miss. Switched to `download-artifact` and dropped `.pyre-build/charon-src` from its Charon restore path, matching the other jobs.

## Effect

Removes the per-ref cache multiplication that pushed the repo past the budget, and makes the LLBC handoff eviction-proof across **all** consumers.

### DIFF
--- a/.github/workflows/pyre-ci.yml
+++ b/.github/workflows/pyre-ci.yml
@@ -74,6 +74,10 @@ jobs:
     - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
       with:
         cache-bin: false
+        # Restore-only on PRs (they pull main's cache via restore-keys); save
+        # only on main so per-PR target caches don't multiply across refs past
+        # GitHub's 10 GB cache budget and trigger LRU eviction.
+        save-if: ${{ github.ref == 'refs/heads/main' }}
     - name: Cache Charon binary
       id: charon-cache
       uses: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809 # v4.2.4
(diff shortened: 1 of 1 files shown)


youknowone:
Superseded by #1023 (`83abf4d9f08`), which landed the `save-if` change — including this PR's comment text verbatim — on four of the five `Swatinem/rust-cache` sites, and in the same commit removed the `wasm-build` job that this PR's second commit targeted.

The LLBC artifact handoff is likewise complete, and has moved past the design here: every consumer job (`prepare-charon-llbc-linux`, `cargo-test-linux`, `pyre-check-linux`, `cpython-tests`, `sandbox-build`) downloads the LLBC artifact, no job restores it through `actions/cache` any more, and the producer now writes per-crate caches (`llbc-pyre-object-*`, `llbc-pyre-interpreter-*`, `llbc-pyre-jit-*`) with per-crate fingerprints instead of the single key used here.

One live gap did remain: `sandbox-build` was added by #304 after this PR was opened, so #1023 never gave it `save-if`, even though it shares `shared-key: ${{ runner.arch }}` with the other four. That is fixed on `main` in b925c3e5ead — applied directly rather than by rebasing this branch, since against the current file only those four lines survive.

Closing as superseded.

— *commented by Claude*

PR #287, opened 2026-06-26
#47: strip box_cache memo + retype box-object channels to Operand (E6/E7)

youknowone:
Continues the box-identity endgame (#47 / #9): completes the E6 prerequisite chain (box identity made memoization-independent, then the `Op`/`InputArg` `box_cache` memo deleted) and starts E7 proper — retyping the box-object channels back to the upstream PyPy shape where the channel stores the AbstractValue (`Operand`) directly, and draining `BoxRef` bridge round-trips.

### E6 — box identity made memo-independent, `box_cache` deleted
- `same_box` compares bound producers before the kind match (`ca0d15bc4c`)
- `PartialEq`/`Hash` compare bound producers, memo-independent (`84166ab4a7`)
- `position()` reads live `op.pos` for a bound ResOp box (`70c8cadd35`)
- strip the `box_cache` memo, delete the `Op`/`InputArg` `box_cache` field (`b11979e2a4`)
- `materialize_operand_at` reads the bound host natively (`374335ac99`)

### Map / heap re-keys to Operand identity
- `quasi_immut_cache` re-keyed to Operand identity (`16ec78e67b`)
- heap `cached_structs` re-keyed; drop the dead `available_boxes` param (`cbd20be47f`)
- `DictArgKey::Op` re-keyed to Operand (`74772a74d2`)

### E7 — box-object channels restored to the PyPy shape
- opencoder `_cache` holds `Operand`, drop the `_produced_roots` Weak-upgrade crutch (`e9c20e025f`)
- resume `LiveboxMap` + the box-handle env yield `Operand` (`ccb7b04e67`)
- `ref_regs` holds `OpRef`, forward the inline `ConstPtr` gcref in place during moving collection (`8050090684`)
- flip `force_child` to `Operand`, delete the now-dead `resolve_box_operand` resolver (`4d7c06cdcd`)

### Bridge drains
- drain `resolve_box_operand` `to_boxref` round-trips (`30cfd761fe`)
- drain `BoxRef` bridges via the native Operand twins, rounds 1 and 2 (`bc6ee9159b`, `dae528dc97`)

### jitcode GC-descr (rebase residuals)
- set / dedup `is_gc_managed` on the `bh_new_with_vtable` size descr (`e4c2858606`, `81b202122c`)

Gate: `majit-metainterp` dynasm-lib 1358 / cranelift-lib 1356; `pyre/check.py` 158/158 on both the dynasm and cranelift backends.

Refs #47, #9.

### DIFF
--- a/majit/majit-gc/src/rewrite.rs
+++ b/majit/majit-gc/src/rewrite.rs
@@ -10,7 +10,6 @@
 ///
 /// Reference: rpython/jit/backend/llsupport/rewrite.py GcRewriterAssembler.
 use majit_ir::Type;
-use majit_ir::box_ref::BoxRef;
 use majit_ir::descr::{DescrRef, FieldDescr, SizeDescr};
 use majit_ir::operand::Operand;
 use majit_ir::resoperation::{Op, OpCode, OpRef};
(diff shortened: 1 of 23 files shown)


youknowone:
Adjudication of the Codex parity review (cycle-close for #287):

**§1 (2 findings — heap.rs `export_cached_fields` dropping `available_boxes`): reclassified as non-regressions.** The removed `available_boxes` parameter was dead at the only live call site — `collect_optimizer_knowledge_for_resume` (optimizer.rs) passed `None`, and the sole `Some(...)`-passing wrapper (`export_all_cached_fields`/`export_all_cached_arrayitems`) had zero callers (`git grep export_all_cached_fields <pre-#287>` → definitions only). So the heap-local filter never fired in production. The `serialize_optheap` filter predicate (heap.py:836/845/855/866) is faithfully applied in `bridgeopt::serialize_optimizer_knowledge` (bridgeopt.rs:134-136/153-155), unchanged by this patch: `obj_ok = is_const(obj) || available.contains(obj)`, `val_ok = is_const(val) || available.contains(val)`, keep iff both. pyre computes `available_boxes` from the live-box set at bridge-creation time (after the optimizer pass has collected the cached fields), so the filter sits one layer up in bridgeopt rather than in a heap-local `serialize_optheap` — a pre-existing structural choice, not introduced here. #287 only deleted the never-firing redundant copy.

**§3 (6 pre-existing): deferred (documented).** All six already carry in-code `PRE-EXISTING-DIVERGENCE` annotations with cited convergence paths — info.rs virtual-array/array-struct/raw-buffer force needs an `is_virtual` flag + gate (the `size = -1` sentinel shape); vstring `_int_add`/`_int_sub`/`_strgetitem` need `send_extra_operation` re-dispatch + emission-time `force_box`. Not introduced by #287 and outside the BoxRef-teardown scope; tracked as follow-ups.

**§4 (4): won't-fix (documented).** `Operand`, the `BoxRef` wrapper, the opencoder raw-index/`_fresh` split, and `LiveboxMap` are the intended Rust identity-carrier adaptations of #47/#9 — correct as-is.

No code change required for this cycle.

— *commented by Claude*

PR #294, opened 2026-06-27
#268 bignum nursery/GC representation: kill the malloc_raw leak + bound within-run RSS

### LINKED ISSUE #268: Give bignums a nursery/GC-managed representation (kill malloc_raw leak)
## Problem

`fib_loop` is ~3.4× slower than PyPy, and the gap is **entirely** the bignum path — pyre actually *wins* `int_loop`:

| bench | PyPy3 | pyre-dynasm | pyre/PyPy |
|---|---|---|---|
| int_loop (small int, native i64, no per-iter heap alloc) | 0.117s | **0.099s** | **0.85× (pyre wins)** |
| fib_loop (bignum, per-iter alloc) | 0.227s | 0.782s | **3.4× (pyre slower)** |

(best-of-8, release dynasm vs pypy3)

Because pyre matches/beats PyPy on `int_loop`, the JIT (codegen, trace shape, GuardClass, residual calls) is **not** the bottleneck. The 3.4× is the `W_LongObject`/`BigInt` allocation model.

## Root cause

The `W_LongObject` BigInt payload is allocated via `pyre-object/src/lltype.rs` `malloc_raw` = `Box::into_raw(Box::new(value))` — a bare system-heap allocation with **no GC header** that is **never freed (leaked)**. Every `jit_w_long_add_raw` result BigInt is system-malloc'd and leaked. `fib_loop` leaks ~300k growing bignums (fib(100000) ≈ 1084 limbs ≈ 8.7KB each → GBs accumulated) → allocator/cache/TLB collapse, while PyPy nursery-collects dead bignums (only ~3 live at a time: a, b, t). malachite `BigInt`'s internal limb `Vec` also hits the Rust global allocator each add.

`malloc_typed` (the `W_LongObject` wrapper, `w_int_new`) already goes through `alloc_with_gc_header` (nursery) — it is the bare `malloc_raw` BigInt payload that bypasses GC.

## Goal

Route bignum allocation through the nursery/GC so dead results are reclaimed like PyPy's `rbigint`, instead of the `malloc_raw` leak. Hard because malachite `BigInt` owns a Rust `Vec` the GC cannot trace/free; likely needs a GC-managed bignum representation (GC-array limbs, à la RPython `rbigint`) or an arena/freelist with reclamation. Large architectural change. Measure `fib_loop` vs PyPy after.

## Out of scope / orthogonal (already done)

- The trace-level long-add fast path (`jitcode_dispatch` walker + trait path) is landed.
- The dynasm↔cranelift codegen gap on `fib_loop` is closed.

youknowone:
fix #268

## Summary

`W_LongObject.value` (the rbigint payload) was `malloc_raw`'d (`Box::into_raw` — no GC
header, never freed), so every bignum result leaked. `fib_loop` accumulated ~300k growing
bignums → allocator/cache/TLB collapse (the root of the bignum-path slowdown vs PyPy). This
routes the BigInt payload through the GC so dead bignums are reclaimed.

Two parts (the issue's chosen single combined change):

### Part 1 — majit-gc lightweight destructors (`9cbdbed1a3`)
`TypeInfo.destructor` + young/old object-with-destructor lists +
`deal_with_*_objects_with_destructors` (incminimark.py:2884-2912 parity, `T_HAS_DESTRUCTOR`
bit). Runs `drop_in_place::<BigInt>` on dead bignums so malachite's external limb `Vec` is
freed.

### Part 2/3 — bignum GC/nursery representation (`0ceef909cf`)
- BigInt payload alloc'd via the GC hooks (nursery / old-gen stable), `malloc_raw` only as a
  no-hook fallback; `W_LONG` registered with its `value` field as a gc-pointer offset; a
  `BIGINT` GC type carrying the `drop_in_place` destructor.
- Boxing = inline `NewWithVtable(W_LONG)` + `setfield_gc(value)` behind a
  `GuardFalse(fits_int)` demote guard. The `NewWithVtable` lowers to the collecting
  `CallMallocNursery` — the per-iteration safepoint bigint loops previously lacked.
- The walker reads each operand's immutable `value` via `GetfieldGcPure` and calls the
  elidable rbigint op on the bare `*const BigInt` payloads (`W_LongObject(self.num.add(
  other.num))` parity). Passing the payloads, not the wrappers, keeps the elidable call pure
  on the immutable bigints so the optimizer never reorders it ahead of the boxing setfield.

## Root-caused crash (fixed here)
A JIT-compiled long loop *inside a function* segfaulted on BOTH backends (top-level loops and
no-loop functions were fine → logic/IR bug, not codegen). The elidable op took the
`W_LongObject` *wrappers* and read `.value` opaquely; after the function-loop unroll the
optimizer reordered the elidable `CallR` ahead of the boxing `setfield_gc` that initializes
the fresh result wrapper → it read an uninitialized BigInt pointer. The payload-based
emission above is the fix.

## Verification
- Function-loop crash gone; `fib_loop` + all 8 benchmarks correct on dynasm AND cranelift.
- `majit-gc` 165 / `pyre-object` 194 tests green (4 `jitcode_runtime` failures are
  pre-existing and unrelated — `PopTop#23` arm-numbering, fail on a clean tree).
- Collections now occur (fn `fib(200000)` → 3 minors vs 0 before).
- Leak proof: `fib(100000)`×5 RSS ≈ `fib(100000)`×1 (5× the dead bignums, +14% RSS) →
  reclaimed, no cumulative leak.
- Perf: the machine-independent `fib_loop / int_loop` ratio drops 7.90 → 3.54, i.e. ~2.2×
  faster on the bignum path.

### Part 4 — within-run RSS: collecting alloc + memory pressure (`0af65788ff`)
Part 2/3's bignum alloc was no-collect → a long bignum loop filled the nursery once and
spilled the rest to old-gen unbounded (`fib(100000)` peaked ~316MB, `fib(300000)` ~1.3GB).

- Route the runtime payload helpers (`jit_bigint_*`) through a **collecting** nursery alloc
  (new host-hook chain mirroring the no-collect one, both backends). The wrapper helpers
  (`jit_w_long_*_raw`, record-time / trait path) stay no-collect since the tracer holds their
  operands natively; the shared div/mod/shift logic is extracted into
  `bigint_*_core(a, b, collecting)`.
- **Memory pressure** (RPython `rgc.add_memory_pressure` analog): `charge_memory_pressure`
  adds a `pressure_since_minor` counter and forces a minor when nursery struct-fill + charged
  external bytes reach the nursery size (reset at each minor). The bignum site charges
  `ceil(bits/64)*8` external limb-`Vec` bytes before allocating — so minor cadence reflects
  true footprint, not just the 48-byte struct. Collecting alone did nothing (the nursery only
  counts the struct); the pressure charge is the load-bearing fix.

Results (release, peak working set), both dynasm and cranelift:

| case | minors | peak RSS | vs Part 2/3 |
|---|---|---|---|
| `fib(100000)` | 1 → 105 | 316 → **~59 MB** | **5.4×** |
| `fib(300000)` | 5 → 934 | 1308 → **~157 MB** | **8.3×** |

Also **faster**: the small working set wins cache/TLB locality (`fib_loop / int_loop` ratio
3.54 → 2.46), outweighing the extra memzero. The residual-call gcmap-under-forced-collection
is proven safe on both backends (1000+ forced minors, all benches + function-loops correct).

## Summary by CodeRabbit

* **New Features**
  * Improved bigint handling in the JIT and interpreter, including better support for shifts and division, plus more consistent allocation behavior.
  * Added support for tracking off-heap memory use, which can help the runtime make better GC decisions.
  * Added an optional allocator feature for improved performance on bigint-heavy workloads.

* **Bug Fixes**
  * Improved garbage collection behavior for objects with custom cleanup needs.
  * Reduced unnecessary test job duplication across operating systems in CI.

### DIFF
--- a/pyre/check.py
+++ b/pyre/check.py
@@ -1130,7 +1130,7 @@ def main():
         #             name              script                          timeout  d_vs_cp  d_vs_py  c_vs_cp  c_vs_py
         chk.run_bench("int_loop",       f"{B}/int_loop.py",             5,       None,    2,       None,    2)
         chk.run_bench("float_loop",     f"{B}/float_loop.py",           5,       None,    1.5,     None,    1.5)
-        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       2,       2,       2)
+        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       3,       2,       3)
         chk.run_bench("inline_helper",  f"{B}/inline_helper.py",        5,       None,    1.2,     None,    1.2)
         chk.run_bench("fib_recursive",  f"{B}/fib_recursive.py",        5,       2,       13,      2,       13)
         chk.run_bench("nested_loop",    f"{B}/nested_loop.py",          5,       None,    2,       None,    3)
--- a/pyre/check.py
+++ b/pyre/check.py
@@ -1130,7 +1130,7 @@ def main():
         #             name              script                          timeout  d_vs_cp  d_vs_py  c_vs_cp  c_vs_py
         chk.run_bench("int_loop",       f"{B}/int_loop.py",             5,       None,    2,       None,    2)
         chk.run_bench("float_loop",     f"{B}/float_loop.py",           5,       None,    1.5,     None,    1.5)
-        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       4,       2,       4)
+        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       2,       2,       2)
         chk.run_bench("inline_helper",  f"{B}/inline_helper.py",        5,       None,    1.2,     None,    1.2)
         chk.run_bench("fib_recursive",  f"{B}/fib_recursive.py",        5,       2,       13,      2,       13)
         chk.run_bench("nested_loop",    f"{B}/nested_loop.py",          5,       None,    2,       None,    3)
--- a/.github/workflows/pyre-ci.yml
+++ b/.github/workflows/pyre-ci.yml
@@ -182,7 +182,7 @@ jobs:
     steps: *prepare-charon-llbc-steps
 
   cargo-test-linux:
-    name: cargo test (ubuntu-24.04, ${{ matrix.backend }})
+    name: cargo test (ubuntu-24.04)
     runs-on: ubuntu-24.04
     needs: prepare-charon-llbc-linux
     if: ${{ !cancelled() && needs.prepare-charon-llbc-linux.result == 'success' }}
@@ -191,10 +191,6 @@ jobs:
       # artifact into this workspace path.
       PYRE_SHARED_BUILD: ${{ github.workspace }}/.pyre-build
       CHARON_VERSION: nightly-2026.05.29
-    strategy:
-      fail-fast: false
-      matrix:
-        backend: [dynasm, cranelift]
 
     steps: &cargo-test-steps
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -245,41 +241,43 @@ jobs:
           /opt/hostedtoolcache/CodeQL /usr/local/.ghcup /usr/local/share/boost
         sudo docker image prune --all --force || true
         df -h /
-    - name: Run cargo test for ${{ matrix.backend }}
+    - name: Run cargo tests
       # Pin bash: on windows-latest the default shell is PowerShell, where
-      # "$BACKEND" does not expand the env var (it would need $env:BACKEND),
-      # so cargo test would run with an empty feature set.
+      # POSIX loop/status handling would otherwise need a separate spelling.
       shell: bash
-      env:
-        BACKEND: ${{ matrix.backend }}
-      run: cargo test --all --no-default-features --features "$BACKEND"
+      run: |
+        status=0
+        for backend in dynasm cranelift; do
+          echo "::group::cargo test ($backend)"
+          if cargo test --all --no-default-features --features "$backend"; then
+            echo "::notice title=cargo test::$backend passed"
+          else
+            code=$?
+            echo "::error title=cargo test::$backend failed with exit $code"
+            status=1
+          fi
+          echo "::endgroup::"
+        done
+        exit "$status"
 
   cargo-test-macos:
-    name: cargo test (macos-latest, ${{ matrix.backend }})
+    name: cargo test (macos-latest)
     runs-on: macos-latest
     needs: prepare-charon-llbc-macos
     if: ${{ !cancelled() && needs.prepare-charon-llbc-macos.result == 'success' }}
     env:
       PYRE_SHARED_BUILD: ${{ github.workspace }}/.pyre-build
       CHARON_VERSION: nightly-2026.05.29
-    strategy:
-      fail-fast: false
-      matrix:
-        backend: [dynasm, cranelift]
     steps: *cargo-test-steps
 
   cargo-test-windows:
-    name: cargo test (windows-latest, ${{ matrix.backend }})
+    name: cargo test (windows-latest)
     runs-on: windows-latest
     needs: prepare-charon-llbc-windows
     if: ${{ !cancelled() && needs.prepare-charon-llbc-windows.result == 'success' }}
     env:
       PYRE_SHARED_BUILD: ${{ github.workspace }}/.pyre-build
       CHARON_VERSION: nightly-2026.05.29
-    strategy:
-      fail-fast: false
-      matrix:
-        backend: [dynasm, cranelift]
     steps: *cargo-test-steps
 
   pyre-check-linux:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,7 +611,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -724,7 +724,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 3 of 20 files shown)


youknowone:
## Review triage (Codex + CodeRabbit) + one real, unfixed GC issue

I triaged every automated-bot finding on this PR against HEAD (`44c9824211`) and ran an independent adversarial review. Posting the dispositions, and flagging **one real bug that needs a dedicated follow-up** (Codex P1) rather than a rushed patch.

### Refuted — bot false positives (verified against current code, no action)
- **Pinned destructor objects** (Codex P2 + CodeRabbit *Major*, `collector.rs:1156`): no path ever pins a destructor-bearing object — `gc_pin` lowers to a no-op returning `false`, and the only `pin()` callers are tests on non-destructor types. Unreachable.
- **Uninitialized `w_class` on the inline long box** (Codex P2, `helpers.rs:832`): `w_class` is reliably NULL (zeroed), and `typedef.rs:207` falls back to `gettypefor(ob_type)` → resolves to `int` with its full MRO. No methods lost.
- **Pure-fold unrooted bigint** (Codex P2, `longobject.rs:92`): folded `Ref` constants route through the GC-traced `gc_table` (a movable `ConstPtr` panics in `guard_constptr_immediate` rather than being baked raw); `44c9824211` moved the recorded op onto the collecting payload helper.
- **`IntLshift` double-eval** (CodeRabbit, `jitcode_dispatch.rs:12442`): no double-record — one discarded concrete eval at trace-record time, same as the sibling FloorDiv/Mod/Rshift operand-dependent bails.
- **Stale fallback pointer** (CodeRabbit, `longobject.rs:232`): the stable allocator is non-moving, so `slot == value` always.

### Addressed (fix prepared on the working branch)
- **Missing `note_alloc()` for longs** (Codex P2, `longobject.rs:213`): `w_long_from_raw` now advances `gc_interp::note_alloc()` on its stable wrapper alloc, as `w_int_new`/`w_float_new` do — otherwise a long-dominated interpreter loop under `PYRE_GC_INTERP` never reaches the dispatch-loop safepoint and dead old-gen long wrappers + payloads accumulate.
- **counter-2 direct-old-gen gap** (Codex P2, `collector.rs:824`): `alloc_bigint_stable` now charges its limb-`Vec` bytes through a new `charge_oldgen_external` hook (no minor forced — a memory-pressure charge there would force a moving minor, unsafe on the unrooted interpreter path), so a directly-old-gen bignum's footprint enters the major threshold at alloc time, not only at the next major's `recompute_oldgen_external_bytes`. (`charge_memory_pressure` also switched to `saturating_add`.)
- **fib_loop 2× gate vs mimalloc-off** (review flag, `check.py:1133`): measured on Windows (worst case per the system-heap note) — the GC path alone is **1.44× / 1.42×** PyPy (dynasm/cranelift), comfortably under the 2× gate. Non-issue; no change.

### ⚠️ Real, unfixed: native-mode bignum payload use-after-free (Codex P1, `longobject.rs:199`)

Confirmed real, but the correct fix is a GC-model change, so I recommend a **dedicated follow-up** rather than landing it here.

**The bug (DEFAULT native config: `gc_interp` OFF, JIT initialized).** `alloc_bigint_stable` GC-allocates the BigInt *payload* whenever `bigint_gc_type_id() != 0` (registered unconditionally at JIT init — not gated on `gc_interp`). But `w_long_from_raw` GC-allocates+traces the *wrapper* only when `gc_interp::enabled()`; otherwise the wrapper is `malloc_typed` (GC-invisible). So an interpreter-created long is a **GC-invisible wrapper holding a GC-managed old-gen payload**. The marker's `seed_major_root` / `mark_object` child loop are gated on `is_managed_heap_object` (nursery/oldgen range check), so a `malloc_typed` wrapper is skipped — its `LONG_VALUE_OFFSET` gc-pointer is never read and the payload is never marked. A JIT-triggered minor (`alloc_bigint_nursery_collecting` filling the nursery) drives `run_major_progress_after_minor` → incremental major → oldgen sweep, which drains `old_objects_with_destructors`, finds the payload unmarked, and runs `bigint_destructor` (drop of the malachite limb `Vec`) **while the long is still reachable from Python** → use-after-free. Benches don't hit it (hot longs box via the traced `emit_box_long_inline`); the host `w_long_new` path is interpreter/cold, so it's untested-latent.

**Why the obvious fixes don't work:**
- *Gate the wrapper on `bigint_gc_type_id()!=0`* (the natural one-liner): **unsafe**. Native majors fire at arbitrary points (mid-trace, nursery-full), not safepoints, and the pyframe walker can't see Rust-stack-only temporaries — a GC-managed wrapper reachable only via a Rust-stack temp would be swept (a symmetric wrapper-UAF). This is exactly the documented "no shadowstack pass" hazard (`pyre_object_gc_collect_trampoline`).
- *Per-site `walk_raw_long_roots`* (mirroring `walk_raw_exception_roots`): **incomplete**. Exceptions live only in frame locals (one call site); longs also live in GC-managed containers (list/dict/tuple) and as JIT bignum constants. A container's trace visits the `malloc_typed` long wrapper and stops at `is_managed_heap_object`, so container-held longs' payloads are still swept.

**Parity-faithful fix.** PyPy/RPython incminimark keeps wrapper *and* payload uniformly GC-managed and roots all live pointers (incl. call-stack temporaries) via a **shadowstack**, which is what makes arbitrary-point collection safe. pyre lacks an interpreter shadowstack — that's the real gap. Two viable directions:
1. **Interpreter shadowstack** — then the wrapper is uniformly GC-managed like PyPy and the `walk_raw_*` compensation helpers can be removed. (Largest, most faithful.)
2. **Central marker trace-through hook** — a registered callback the majit-gc marker invokes for non-managed (`malloc_typed`) objects, so any reachable long wrapper (frame, container, cache, const) marks its payload. Avoids the shadowstack but is still a real majit-gc mechanism change.

Either way this is a GC-model change, not a bot-patch; shipping a partial fix (e.g. frame-only `walk_raw_long_roots`) would give false confidence while container-held longs still UAF.

pyre/check.py
```diff
@@ -1130,7 +1130,7 @@ def main():
         #             name              script                          timeout  d_vs_cp  d_vs_py  c_vs_cp  c_vs_py
         chk.run_bench("int_loop",       f"{B}/int_loop.py",             5,       None,    2,       None,    2)
         chk.run_bench("float_loop",     f"{B}/float_loop.py",           5,       None,    1.5,     None,    1.5)
-        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       4,       2,       4)
+        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       2,       2,       2)
```
youknowone:
```suggestion
        chk.run_bench("fib_loop",       f"{B}/fib_loop.py",             5,       2,       3,       2,       3)
```

Issue #73, opened 2026-05-19
Eliminate `pc_map` and per-PC labels by switching the tracer to interpret JitCode bytecode


youknowone:
## tl;dr

Pyre's tracer interprets CPython 3.13 bytecode while RPython's tracer interprets JitCode bytecode. That single divergence forces every JIT'd function to carry a `py_pc → jitcode_pc` translation table (`PyJitCode::metadata.pc_map`) and forces the SSARepr walker to emit one `Label("pc{N}")` per Python bytecode offset — even when those offsets fall inside the same flow block. Both structures are workarounds for the coordinate-system mismatch, not legitimate adaptations. They should disappear together once the tracer is rewritten to interpret JitCode bytecode directly.

## Symptom — three coupled pyre-only artifacts

1. **`PyJitCode.metadata.pc_map: Vec<usize>`** (`pyre/pyre-jit-trace/src/pyjitcode.rs:374`, populated in `pyre/pyre-jit/src/jit/codewriter.rs:9045-9104`). Indexed by Python bytecode offset, yields a JitCode byte offset. Consulted on every guard-resume to translate `frame.pc` → JitCode position before calling `setposition`.

2. **Per-Python-PC labels in SSARepr.** The walker emits `Insn::Label(\"pc{N}\")` at every Python bytecode boundary so the JitCode assembler can record byte offsets per Python PC; those byte offsets are what `pc_map` ends up holding. This is structural — Python opcodes that fall inside the same flow block still each get a label, purely so the side-table has a row.

3. **`ResumedFrame.jitcode_pc: Option<usize>`** (`pyre/pyre-jit/src/call_jit.rs:677-696`). A pre-translated cache of `pc_map[py_pc]` written when resume frames are constructed, read by `resume_in_blackhole` (`pyre/pyre-jit/src/call_jit.rs:898, 1951`). Exists because the resume path performs the translation often enough that the per-read cost matters.

The comment at `call_jit.rs:688-695` documents (1)-(3) explicitly as a NEW-DEVIATION whose proper fix is to rewrite the tracer.

## Root cause

RPython's invariant is that **one coordinate system spans the entire JIT pipeline**:

- Trace records JitCode opcodes (`pyjitpl.py::dispatch_bytecode` reads `miframe.jitcode.code[miframe.pc]`).
- `miframe.pc` IS a JitCode byte offset.
[...]

PR #300, opened 2026-06-28
JIT: migrate resume-liveness readers from flat stack_slot_color_map to per-PC pcdep (#73)

youknowone:
Part of #73 (eliminate `pc_map` and per-PC labels). This batch migrates the
resume-liveness machinery from the flat per-jitcode `stack_slot_color_map`
(operand-stack slot → merge color) to the per-PC `pcdep` color→slot coloring,
and drains two flat-map readers.

## Commits

- **codewriter: extend co-live interference to the resume-depth snapshot** —
  add the resume-depth operand-stack snapshot to the co-live interference set so
  the per-PC coloring stays unambiguous at resume points.
- **jitcode_dispatch: source collect_outer color→slot from pcdep only** —
  `collect_outer_active_boxes` reads the per-PC `pcdep` map instead of cloning
  the flat `stack_slot_color_map`.
- **jitcode_dispatch: drop recovery_complete kept-stack escape and
  branch_resume_stack_colors** — re-source the recovery-complete path off
  `pcdep` and remove the flat `branch_resume_stack_colors` side table.
- **jit: add PyJitCodeMetadata::max_stackdepth and size the bridge frame array
  from it** — record `code.max_stackdepth` (= `co_stacksize`) and size the
  bridge frame array from it, draining the size-calc reader of
  `stack_slot_color_map.len()`. Byte-identical (`code.max_stackdepth ==
  stack_slot_color_map.len()` for every compiled jitcode across the corpus).
- **codewriter: make filter_liveness_in_place pcdep-only, drop flat fallback** —
  delete the production-dead flat else-branch; the sole production caller always
  supplied the per-PC map (only `#[cfg(test)]` callers passed `None`). Drop the
  `depth_at_pc` / `local_color_map` / `stack_slot_color_map` params and make
  `pcdep_color_slots` non-optional. Byte-identical for production.

## Verification

`pyre/check.py` — dynasm 160/160 + cranelift 160/160.

## Summary by CodeRabbit

* **New Features**
  * Added JIT metadata support for `max_stackdepth` (with skeleton defaults), improving runtime frame/stack reconstruction accuracy.
  * Portal-bridge resume logic now derives portal stack colors and live locals from portal-specific inputs for more reliable guard recovery.

* **Bug Fixes**
  * Corrected bridge setup sizing when live-frame length is unavailable.
  * Fixed per-PC color computations for result slots to avoid depth mismatches during guard resumption.

* **Tests**
  * Updated and extended tests to use the new stack-depth metadata and revised resume/bridge behavior.

### DIFF
--- a/pyre/pyre-jit-trace/src/canonical_bridge.rs
+++ b/pyre/pyre-jit-trace/src/canonical_bridge.rs
@@ -224,6 +224,9 @@ pub fn install_portal_for(code_ptr: *const pyre_interpreter::CodeObject) -> Arc<
             after_residual_call_resume_pc: Vec::new(),
             first_jit_pc_by_py_pc: Vec::new(),
             depth_at_py_pc,
+            // Portal bridges have an empty pc_map, so the inline multiframe
+            // capture declines before it reads this; keep it empty.
+            result_color_at_pc: Vec::new(),
             portal_frame_reg,
             portal_ec_reg,
             built_as_portal: true,
@@ -417,6 +420,7 @@ def f(x, y):
                 after_residual_call_resume_pc: vec![None],
                 first_jit_pc_by_py_pc: vec![0],
                 depth_at_py_pc: Vec::new(),
+                result_color_at_pc: Vec::new(),
                 portal_frame_reg: 0,
                 portal_ec_reg: 0,
                 built_as_portal: true,
--- a/pyre/pyre-jit-trace/src/canonical_bridge.rs
+++ b/pyre/pyre-jit-trace/src/canonical_bridge.rs
@@ -190,9 +190,9 @@ pub fn install_portal_for(code_ptr: *const pyre_interpreter::CodeObject) -> Arc<
     // collapsing to `stack_base + 0`. The contract is documented at
     // `PyreJitCode::stack_slot_color_map`. With no regalloc the color of each slot is
     // its own pre-color, hence the identity fill.
-    let (stack_base, depth_at_py_pc, stack_slot_color_map, pyre_color_for_semantic_local) =
+    let (stack_base, depth_at_py_pc, stack_slot_color_map, pyre_color_for_semantic_local, max_stackdepth) =
         if code_ptr.is_null() {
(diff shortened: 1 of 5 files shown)


youknowone:
### TODO — to fully finish #73 (delete `stack_slot_color_map` and per-PC labels)

This PR drains two flat-map readers (the size-calc `.len()` reader and the
`filter_liveness_in_place` flat fallback). The field is still alive. Remaining
flat-map readers, cause-partitioned from a full 160-corpus census on both
backends (dynasm + cranelift):

- [ ] **Bridge inversion flat-else** (`state.rs` `setup_bridge_sym`) — taken only
  when `pcdep_entries` is empty. Empirically inert (reached once across the
  corpus, with no live locals/stack). Collapse to pcdep-only, **gated on a
  pcdep-totality proof** (`validate_pcdep_color_map`, injective + total over the
  corpus), not on the single trivial reach.
- [ ] **`reconstruct_inline_recipe` flat-else + identity-gate** (`state.rs`) —
  cold on the corpus (never reached). Needs a depth-2 nested-inline fixture
  (#68) to exercise the path before the flat readers can be removed safely.
- [ ] **`goto_if_not` `stack_sync` legacy read** (`jitcode_dispatch.rs:8816`) —
  live fallback used when the per-op vstack mirror does not cover a slot.
  Blocked on extending vstack-mirror coverage (#370 / #338) and on positional
  failargs (the symbolic-valuestack work).
- [ ] **C3 inline result-slot** (`jitcode_dispatch.rs` `compute_inline_caller_frame`
  / `compute_nested_inline_caller_frame`) — nulls the not-yet-produced inline
  result slot. This slot is not a live Variable at the fallthrough PC, so it is
  not in `pcdep`; it needs a new `result_color_at_pc` metadata (a port of PyPy's
  `_result_argcode` / `ord(bytecode[pc-1])`), not pcdep.
- [ ] **After all readers drain:** delete the `stack_slot_color_map` field, the
  trait-path (`trace_opcode`) readers (inert under the full-body walker), and the
  remaining per-PC label tables.

— *commented by Claude*

PR #382, opened 2026-07-05
#355: PyFrame as Python-visible frame W_Root (retire sys.namespace stub)

### LINKED ISSUE #355: Make PyFrame a Python-visible W_Root (frame typedef), following PyPy's heap-frame + virtualizable model — retire the sys.namespace frame stub
## Goal

Make `PyFrame` a genuine Python-visible object (`W_Root` analog) with a `frame`
typedef, following PyPy's model exactly, and retire the `sys.namespace` frame
stub in `pyre/pyre-interpreter/src/module/sys/vm.rs`.

Today `sys._getframe()` / `traceback` frames are hand-built `sys.namespace`
attribute-bag stubs. They cover the common surface (`f_lineno`, `f_code`,
`f_back`, `f_locals` read/write, `traceback.extract_stack`) but are eager
per-call snapshots, not the live frame, and diverge from PyPy structurally
(`sys/vm.rs:90,137` ↔ `pypy/interpreter/pyframe.py:767`,
`pypy/interpreter/typedef.py:736`). This blocks faithful `inspect` / `traceback`
/ frame-introspection behavior.

## How PyPy does it (the model to copy)

In PyPy `class PyFrame(W_Root)` — the frame **is** the Python object; there is no
wrapper. The apparent "lifetime problem" (a frame outliving its C stack via
`sys._getframe()` or a traceback) does **not** exist as a special case because in
RPython every frame is a heap GC object from birth, so Python retention just
keeps it alive through the GC like any other object.

`escaped` / virtualizable is a **JIT optimization, not a lifetime mechanism**
(`pypy/interpreter/executioncontext.py:85-109`):

- `enter()`: `self.topframeref = jit.virtual_ref(frame)` — while a trace runs the
  JIT keeps frame fields in registers / the vable and elides writes to the real
  heap object (virtualizable).
- `leave()`: `if frame.escaped or got_exception: frame_vref()` — if the frame was
  exposed to app level (`sys._getframe()` calls `mark_as_escaped()`), the vref is
  **forced**, writing the register/vable values back into the real heap frame so
  it "will be still accessible even after the corresponding C stack died"
  (`mark_as_escaped` docstring, `pyframe.py:176-183`).
- `fget_f_back` = `ExecutionContext.getnextframe_nohidden(self)` — walks
  `f_backref()` skipping hidden frames (`executioncontext.py:79-83`).

So: the frame object always lives on the GC heap; `escaped` only tells the JIT to
de-virtualize (materialize) it so non-JIT / app-level code can read it later.

## What pyre already has

- `GcFramePrefix` / `FrameBox` — nursery-GC-managed frames with a GC header
  (`pyframe.rs:241`, `const _: () = assert!(offset_of!(GcFramePrefix, frame) ==
  GC_HEADER_SIZE)`).
- `escaped: bool` + `mark_as_escaped()` scaffolding (`pyframe.rs:79-80,1756`) and
  the `ec.leave` hook that marks `f_back` as escaped on exception
  (`executioncontext.rs:938-941`) — the direct analog of PyPy's `leave()`.
- The `sys.namespace` stub that this issue replaces.

## What's missing (the actual work)

1. **Every frame GC-managed.** Unify the top-frame `Box::into_raw` (`std::alloc`)
   path onto the GC `FrameBox` path so a frame Python holds stays alive by GC
   reachability. (Callee frames already use the GC path.)
2. **PyObject header + `frame` typedef.** `PyFrame` currently has **no**
   `ob_header` — its first field is `execution_context` (`pyframe.rs:40-44`), so
   it is not dereferenceable as a `PyObject`. Add the header and a `frame`
   `W_TypeObject` with `GetSetProperty` descriptors ported from
   `pypy/interpreter/pyframe.py`: `f_back` (= `getnextframe_nohidden`, skipping
   hidden frames), `f_locals`, `f_globals`, `f_code`, `f_lineno`, `f_lasti`,
   `f_builtins`, `f_trace` (get/set/del), `f_trace_lines`, `f_trace_opcodes`,
   plus `clear()`.
3. **`sys._getframe` / traceback return the real frame**, calling
   `mark_as_escaped()` on exposure (PyPy `pyframe.py:176`), so `ec.leave` forces
   the vref/vable back into the heap frame.
4. **Retire the `sys.namespace` stub** and its eager per-call chain builder in
   `sys/vm.rs`.

## Known cost / blocker

Inserting an `ob_header` at offset 0 shifts the frame layout that the JIT bakes:
`pyre/pyre-jit-trace/src/frame_layout.rs` holds ~33 `offset_of!` / `PYFRAME_*`
constants threaded into the vable and trace IR, and `pyframe.rs` has
`const _: () = assert!` layout asserts. This needs a coordinated offset
migration (and, since it changes `pyre-interpreter` source the JIT prepass reads,
an LLBC re-extraction — `scripts/extract-llbc.py --force`, see AGENTS.md). This
is the main reason the work is an epic rather than a mechanical port.

## Suggested slicing

- **S1** — Unify frame allocation onto the GC path; prove a `sys._getframe()`
  frame retained past its function's return stays valid under a forced GC.
- **S2** — Add `ob_header` + migrate `frame_layout.rs` offsets + re-extract LLBC;
  keep behavior identical (no descriptors yet), gate JIT green (check.py 169×2).
- **S3** — Port the `frame` typedef `GetSetProperty` descriptors from
  `pyframe.py`; wire `mark_as_escaped()` on `sys._getframe` / traceback exposure.
- **S4** — Retire the `sys.namespace` stub; gate on `test_sys`, `test_inspect`,
  `test_traceback` (current baselines: 20 / 34 / 98 failures — measure how many
  are frame-stub-attributable).

## Related

Split out of the `-m test` harness enablement work (`gc.collect` no-op, unpack
subclass, pyexpat, `SimpleNamespace` — commits `8824bf44cc`, `734307c4ff`). The
frame stub was adequate for that harness goal; this issue tracks the faithful
port. Adjacent to the `sys.namespace` → app-level `_structseq.SimpleNamespace`
convergence (`pypy/module/sys/app.py:212`).

— *filed by Claude*

youknowone:
close #355 — make `PyFrame` a Python-visible `frame` W_Root following PyPy's
heap-frame + virtualizable model, and retire the `sys.namespace` frame stub for
the three frame producers (`sys._getframe`, trace/profile callbacks,
`traceback.tb_frame`).

Follows the issue's S1→S4 slicing. All changes are PyPy-faithful in structure
(CPython 3.14 behavior, PyPy/RPython implementation).

## Slices

- **S0** (`850ecb37bb`) — type-directed `PyFrame` GC trace via a custom hook
  (`pyframe_object_custom_trace`), replacing the fixed `gc_ptr_offsets`; closes
  the `w_builtin` / debug-data gaps and makes the locals-array walk
  regime-independent.
- **S1** (`1e1dafce11`) — give `PyFrame` a `PyObject` header (`ob_type` = the new
  `frame` type). Layout offsets in `frame_layout.rs` / vable / descr auto-shift
  off `offset_of!`; LLBC re-extracted.
- **S2** (`20a9ef9e1c`) — executing and generator frames get GC lifetime
  (policy A: no manual free at `leave`; reachability-collected). Three regimes:
  executing chain frames = Gc oldgen (walker greys), tracing snapshots = Box
  (deterministic Drop), generator frames = Gc + traced `frame_ptr` edge.
  type-37 destructor runs `PyFrame::drop` side-effects at sweep.
- **mark_stacks** (`0ad2b3b83a`) — port CPython 3.14's `mark_stacks` fixpoint
  stack-kind analysis so `f_lineno` writes validate debugger line jumps.
- **S3** (`8cf64132a6`) — `frame` typedef + all 13 GetSetProperty descriptors
  (`f_code`/`f_globals`/`f_locals`/`f_back`/`f_lasti`/`f_builtins` +
  `f_lineno`/`f_trace`/`f_trace_lines`/`f_trace_opcodes` + `clear`/`__repr__`);
  `sys._getframe()` returns the live frame + `mark_as_escaped`.
- **S4a** (`fc5f50af9c`) — trace/profile callbacks receive the real frame;
  `trace_frame_type` + `flush_trace_frame_writeback` deleted.
- **S4b** (`e8a0378706`) — `traceback.tb_frame` returns the real frame by making
  `PyTraceback` genuinely GC-managed (custom_trace forwards `w_next`/`w_code` +
  the `frame` edge guarded by `try_gc_owns_object`; oldgen allocation; tid pinned
  after PyCode, net-zero on downstream tids).

## Verification

- `check.py --backend dynasm,cranelift` → **171/171 both backends**.
- GC-stress probes (frames / generators / tracebacks held across `gc.collect()`
  under `PYPY_GC_NURSERY=131072`) green on both backends.

## Notes

- The `sys.namespace` frame stub (`make_traceback_frame_stub`) is kept only as a
  fallback for non-Gc frames (tracer Box snapshots / already-freed arena callees,
  which under `PYRE_JIT=0` are freed by the time a traceback is read) — returning
  such a frame would be a use-after-free. Under the production release+JIT path
  frames are Gc-owned oldgen blocks, so the real frame is returned.
- Arena/callee JIT frames are intentionally left as-is (PyPy virtualizes inlined
  callees to registers with no heap frame; the arena + `walk_jit_callee_frame_roots`
  is the faithful adaptation — making them GC garbage would lower parity).

### DIFF
--- a/pyre/pyre-interpreter/src/eval.rs
+++ b/pyre/pyre-interpreter/src/eval.rs
@@ -1405,6 +1405,16 @@ fn eval_loop(frame: &mut PyFrame) -> PyResult {
                 }
                 return Err(err);
             }
+            // A trace callback may perform a debugger line-jump by setting
+            // `frame.f_lineno` (`PyFrame::fset_f_lineno` → `last_instr =
+            // best_addr`).  Honour it: if a tracer is installed and it
+            // moved `last_instr` off the instruction we were about to
+            // dispatch, resume from the jump target instead of `pc`.  The
+            // `gettrace` null-check keeps this off the no-tracer hot path.
+            if unsafe { !(*ec).gettrace().is_null() } && frame.last_instr as usize != pc {
+                next_instr = frame.last_instr as usize;
+                continue;
+            }
         }
         let (opcode_pc, instruction, op_arg) = decode_instruction_for_dispatch(code, pc)?;
         let fallthrough = opcode_pc + 1;
--- a/pyre/pyre-interpreter/src/executioncontext.rs
+++ b/pyre/pyre-interpreter/src/executioncontext.rs
@@ -8,208 +8,25 @@ use std::sync::OnceLock;
 
 use crate::PyFrame;
 
-fn trace_frame_type() -> PyObjectRef {
-    static TYPE: OnceLock<usize> = OnceLock::new();
-    let raw = *TYPE.get_or_init(|| {
-        let tp = crate::typedef::make_builtin_type("frame", |_| {});
-        // The wrapper wants a per-instance mapdict store; a `__dict__`
-        // rawdict key would instead claim the typedef manages the dict
-        // (typedef.py:40) and suppress the mapdict one
-        // (typeobject.py:253-257), so flip `hasdict` directly — the
-        // `create_dict_slot` flag flip (typeobject.py:1222-1226).
-        unsafe { pyre_object::w_type_set_hasdict(tp, true) };
-        tp as usize
-    });
-    raw as PyObjectRef
-}
-
-/// Copy callback-visible mutations on the trace-frame wrapper back to
-/// the live `PyFrame`'s debug data.
+/// Return the live `PyFrame` as the Python-visible `frame` object passed
+/// to a trace / profile callback.
 ///
-/// PyPy's `_trace` passes the live frame to the callback (`jit.hint(
-/// frame, access_directly=False)`), so `frame.f_trace = local` and
-/// `frame.f_lineno = N` from inside the callback land directly on
-/// `frame.debug`.  Pyre wraps the frame in a `pyre_object` instance
-/// because `PyFrame` is not itself a `PyObject`, so a setattr on the
-/// wrapper would otherwise stay isolated.  After the callback returns,
-/// read the wrapper's mutated attributes and propagate the changes to
-/// the live frame, preserving the user-visible semantics for the
-/// common case (debugger setting `frame.f_trace` to a per-frame
-/// callback while returning `None`).
-///
-/// The `_trace` w_result branch still wins (`executioncontext.py:386-
-/// 391`): a non-None callback return value overrides whichever value
-/// the setattr left behind, matching CPython issue11992 bug-for-bug
-/// compatibility.
-fn flush_trace_frame_writeback(
-    frame: *mut PyFrame,
-    w_frame: PyObjectRef,
-    init_lineno: isize,
-) -> Result<(), crate::PyError> {
-    if frame.is_null() || w_frame.is_null() {
-        return Ok(());
-    }
-    let new_f_trace = match crate::baseobjspace::getattr_str(w_frame, "f_trace") {
-        Ok(v) => v,
-        Err(_) => return Ok(()),
-    };
-    // pyframe.py:785-791 fset_f_trace:
-    //   if space.is_w(w_trace, space.w_None):
-    //       self.getorcreatedebug().w_f_trace = None
-    //   else:
-    //       d = self.getorcreatedebug()
-    //       d.w_f_trace = w_trace
-    //       d.f_lineno = self.get_last_lineno()
-    // The non-None branch also realigns f_lineno to the current
-    // bytecode line so trace events fire from the new tracer's
-    // perspective immediately.
-    unsafe {
-        let d = (*frame).getorcreatedebug(init_lineno);
-        let is_clear = new_f_trace.is_null() || new_f_trace == pyre_object::w_none();
-        if is_clear {
-            d.w_f_trace = pyre_object::PY_NULL;
-        } else if new_f_trace != d.w_f_trace {
-            d.w_f_trace = new_f_trace;
-            d.f_lineno = (*frame).get_last_lineno();
-        }
-    }
-    if let Ok(new_f_lineno_obj) = crate::baseobjspace::getattr_str(w_frame, "f_lineno") {
-        if !new_f_lineno_obj.is_null() && unsafe { pyre_object::is_int(new_f_lineno_obj) } {
-            let new_lineno = unsafe { pyre_object::w_int_get_value(new_f_lineno_obj) };
-            // TODO: pyframe.py:683-764
-            // `PyFrame.fset_f_lineno` is the upstream setter. It runs
-            // line-jump validation against the bytecode (block stack
-            // unwinding through SETUP_LOOP/SETUP_EXCEPT, code address
-            // recomputation via PyCode._signature_addr_to_line, and
-            // last_instr realignment). Pyre's wrapper writes only
-            // debug.f_lineno because PyFrame is not yet a PyObject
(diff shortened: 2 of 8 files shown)


youknowone:
## Parity review (comment 4885303060, commit 8638376) §2/§3 dispositions

That review predates `a536975167`; re-checked each item against the current HEAD source + **CPython 3.14** behavior (authority) and PyPy structure. Acted in `c95aef1a09` + already-done in `a536975167`.

### §2 — mismatches introduced by this patch

- **§2.1 `descr_clear` suspended generator** — already fixed in `a536975167`: raises `"cannot clear a suspended frame"` (CPython 3.14 `frame_clear`; verified on 3.14). Codex cited PyPy's `gen._finalize_()`, which is a pypy-3.11 detail; CPython 3.14 rejects instead. ✅ **fixed**.
- **§2.2 `fset_f_lineno` yield-event jump** — real gap. Fixed in `c95aef1a09`: outside a `line` event, a jump is now permitted when the opcode at `last_instr` is `YIELD_VALUE`, matching pyframe.py:685-689 and CPython 3.14 `frame_lineno_set_impl` (which lists `PY_YIELD` in the allowed events). ✅ **fixed**.

### §3 — pre-existing mismatches

- **§3.2 `_trace` f_lineno restore skip** — real, self-contained. Fixed in `c95aef1a09`: the `d.f_lineno = old_lineno` restore is skipped for generator/coroutine resumptions (`event == "call"` while `last_instr >= 0`), matching executioncontext.py:397-404. pyre already ports PyPy's `_trace` structure, so this completes it. (Not observable yet — pyre emits no `line` events for generators — but a low-cost structural completion, not blocked.) ✅ **fixed**.
- **§3.1 `exception_trace` live operr** — **deferred (blocked)**. `handle_operation_error` holds the live `PyError` but `exception_trace` takes a `(w_type, w_value, w_traceback)` triple and rebuilds a temporary `OperationError`. Threading the live operr through requires reworking the error channel signature across `error.rs` — the OperationError port epic the code comment already names ("The OperationError port … will close that gap"). No observable divergence today (the temp's in-place normalize falls out of scope). Pulling it in would balloon this diff.
- **§3.3 PyTraceback `__reduce__`/`__setstate__`** — **won't-fix (false positive vs CPython 3.14)**. CPython 3.14 tracebacks are unpicklable (`pickle.dumps(tb)` → `TypeError: cannot pickle 'traceback' object`), have no `__setstate__`, and `__reduce__` is only the inherited `object.__reduce__`. pyre already matches (`__reduce__` present via object, `__setstate__` absent, pickle refused). PyPy's `_pickle_support` wiring is pypy-specific; adding it would diverge from 3.14.

### §4 — structural adaptations (confirmed correct, no change)

PyObject header on `PyFrame` (vs `W_Root` subclass), `f_lasti`/`tb_lasti` × 2 byte-offset, `mark_stacks` over 3.14 bytecode, `tb_frame` GC-owned-or-stub — all deliberate CPython-3.14/Rust adaptations, documented in-code. The `tb_frame` stub's full removal is tracked in #383 (#355 Goal 4).

check.py 171/171 dynasm + cranelift.

— *commented by Claude*

PR #390, opened 2026-07-05
majit: field-level IR infrastructure + aarch64 large trace support + observer replay fix

### LINKED ISSUE #363: aheui jit: storage stack ops are opaque residual calls — inline/virtualize the linked-list Node to beat naive
Follow-up to #345 (which closed the aheui-logo jit **compile-cost** gap). With compilation no longer the bottleneck, logo `--jit` sits at **parity** with naive rather than beating it, and profiling shows exactly why: the JIT spends its execution time in the same opaque storage helpers the naive interpreter calls.

## Evidence (post-#345 profile of logo `--jit`)

macOS `sample`, top-of-stack self-weight (idle frames removed):

```
emit_impl                    192   ← compile, now minor (was 725 before #345 fold)
jit_storage_push             139  ┐
aheui_runtime…LinkedList::cmp 130  │
aheui_runtime…Stack::push     122  │  ~1100 samples of runtime
aheui_runtime…LinkedList::mul 105  │  aheui stack execution
jit_storage_dup  82, mul 68,  ...  │
cmp 62, swap 57, pop 46, add 38…   ┘
```

The dominant cost is now `aheui_runtime::storage::linkedlist::*` (push/pop/dup/swap/add/sub/mul/cmp), reached through the `jit_storage_*` residual-call shims. The naive interpreter calls the identical helpers, so jit and naive do the same work — hence parity.

## Root cause: storage ops are opaque residual calls, never inlined or virtualized

Every stack op is modelled as a single residual call (`aheui/aheui-jit/src/lib.rs:394-448`), and the source says so explicitly (`lib.rs:6`, `:496`):

```
// storage = linked list stacks (no virtualizable arrays)
```

```rust
extern "C" fn jit_storage_push(pool_ptr, target, value) { … }   // residual
extern "C" fn jit_storage_pop (pool_ptr, target) -> Val { … }   // residual
extern "C" fn jit_storage_add/sub/mul/dup/swap/cmp(…)           // residual
```

Because these are residual calls, the optimizer cannot see through them: it cannot fold a `push` immediately followed by a `pop`, cannot keep the top-of-stack in a register across ops, and cannot elide the per-`push` `Node` heap allocation. Each op pays full interpreter cost inside the compiled trace.

## Parity target: PyPy virtualizes the linked-list Node

The storage is a faithful port of rpaheui's `class Node` linked list (`Stack`/`Queue`/`Port` over `Node { value, next }`), so the data structure is not the divergence — rpaheui uses the same linked list. The difference is that under PyPy's JIT, a `Node` that does not escape the trace is **virtualized**: escape analysis removes the allocation and turns the push/pop/arith sequence into register-level operations. That is what lets rpaheui-under-PyPy get a real JIT speedup on storage-heavy loops like logo. majit currently gets none of that because the ops are opaque.

*(This is the expected PyPy behavior from its virtualization optimizer; it should be confirmed by inspecting rpaheui's traces, but the observed majit profile — residual calls dominating, jit==naive — is consistent with "no virtualization at all".)*

## Directions

1. **Inline + virtualize the stack ops** (parity path). Express push/pop/dup/swap/arith as inlinable IR over the `Node` linked list (or an oopspec), and make the `Node` allocation virtualizable so a non-escaping push/pop pair folds away and the hot top-of-stack stays in registers. This matches PyPy and keeps the linked-list representation.
2. **Array-backed stack representation** (divergence). Replace the per-push `Node` allocation with an array/vec-backed stack. Faster allocation-wise, but diverges from the rpaheui port and still needs the ops inlined to avoid residual-call overhead, so (1) is the more orthodox lever.

Direction (1) is the RPython-faithful one and the actual "beat naive" lever. It is a substantial aheui-jit + majit-virtualization change, tracked here separately from #345.

Refs #345, #231.

youknowone:
## Summary

Infrastructure for field-level IR (getfield_gc/setfield_gc) in the `#[jit_interp]` macro, aarch64 backend fixes for large traces, and observer replay robustness.

### Commits (13)

**Opcode lowering (1)**
- `d0f26ed027` — Lower CALL_INTRINSIC_1 UnaryPositive instead of abort_permanent

**Optimizer (2)**
- `3c2bacc4f2` — Full same_greenkey close gate for primary traces, drop PYRE_SAME_GREENKEY
- `095d522f1f` — HashMap for unroll assemble_peeled remap tables

**Macro: ref_fields + struct_type (4)**
- `3d3d3a6fe5` — Add ref_fields config and ref-kind getfield_gc_r/setfield_gc_r support
- `50f8be9bc0` — Add struct_type to Binding, lower_ref_binding_getfield/setfield, RefFieldRewriter local ref tracking
- `181ac8148d` — Add call_returns config for ref-returning call struct type
- `c72e82b500` — Add struct_allocs config for concrete-path struct literal rewriting

**Macro: ResidualRef + SizeDescr (3)**
- `d95771af6e` — Add ResidualRef (non-wrapped) call policy for ref-returning plain functions
- `6618081445` — Merge fields in register_struct_layout instead of overwriting
- `97cfcf5a86` — Upgrade SizeDescr cache on fuller layout, patch field parents at finish

**Backend (2)**
- `24a4a8952d` — Replace VecMap/VecSet with IndexMap/IndexSet for O(1) keyed access
- `2459f8b8ea` — aarch64 assembler large frame offset + long branch support

**Observer (1)**
- `e8f3876093` — Scan-match getfield entries in observer replay queue

### Key changes

- **ref_fields config**: Declares which struct fields are ref-kind, so the lowerer emits `getfield_gc_r`/`setfield_gc_r` instead of `_gc_i`. Enables field-level IR for linked-list Node structs.
- **struct_allocs config**: Maps struct literal types to concrete allocator functions on the concrete path (RefFieldRewriter). The JIT path uses `lower_struct_value` (New + SetfieldGc).
- **call_returns config**: Maps ref-returning function calls to their return struct type so subsequent field access resolves correctly.
- **aarch64 large frame offset**: Traces with >4096 frame slots (offset >32760) now use scratch register helpers instead of direct `[x29, #imm]`.
- **aarch64 long branch**: Conditional branches in traces >1MB use inverted-condition skip + unconditional branch pattern (26-bit ±128MB range).
- **Observer replay**: `consume_observed_getfield` scans the queue for matching (obj, offset) instead of strict FIFO, tolerating field access reordering from field-level IR.

### Verification

- check.py: dynasm 180/180, cranelift 180/180, wasm 180/180
- aheui logo JIT: 996310 bytes (correct, `MAJIT_THRESHOLD=20`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AArch64 frame-pointer-relative load/store handling for large offsets and more reliable conditional branching on long traces.
  * Fixed GC/field-size cache upgrades so stale field/size metadata is replaced correctly.
  * Made trace replay for field reads more flexible, and strengthened loop merge detection using full ref/float+int green matching.
* **New Features**
  * Added more struct-aware JIT lowering, including better ref-field handling and residual ref-return typing.
* **Performance**
  * Improved deterministic behavior and lookup efficiency across compilation/codegen/optimization via insertion-ordered maps/sets.


majit/majit-ir/src/vec_map.rs
```diff
@@ -1,44 +1,29 @@
...
-//! [`VecMapExt`] re-exposes the `entry_or_insert_with`, `entry_or_default`
-//! and `iter_entries_mut` shortcut methods majit uses, so caller sites keep
-//! their `map.entry_or_insert_with(k, f)` shape without going through the
-//! intermediate `Entry` value.
+//! Previously backed by `vecmap_rs::VecMap` (linear-scan Vec). Replaced by
+//! `indexmap::IndexMap` for O(1) keyed access while preserving insertion order.
+//! The `VecMap` name is kept as a type alias to minimise churn across callers.
 
-pub use vecmap_rs::VecMap;
+pub use indexmap::IndexMap as VecMap;
```
youknowone: never

Issue #366, opened 2026-07-04
JIT: rewrite the pyre tracer to interpret JitCode bytecode directly (continuation of #73 after the M3 milestone)


youknowone:
## Context

This carries forward the **core mechanism of #73** as a standalone, current-state-verified issue so the work survives #73's closure by PR #365 (which landed the **M3 milestone** — direct-JitCode-pc kept-stack branch-guard resume as the unconditional production default, and deletion of the `m3_jitcode_pc_enabled()` gate + `PYRE_M3_JITCODE_PC` opt-out).

#73's title — *"Eliminate `pc_map` and per-PC labels by switching the tracer to interpret JitCode bytecode"* — names a **rewrite**, not a role-retirement. The M3 milestone retired `pc_map`'s *lossy kept-stack-collapse role* for branch guards; this issue tracked retiring the remaining data-structure artifacts (`pc_map`, the carried `jitcode_pc` side-channel).

## ✅ Resolution (2026-07-08, verified on `rewrite-tracer`, shipped as PR #426)

**The dense `PyJitCodeMetadata::pc_map: Vec<usize>` (Artifact 1) is physically DELETED** (commit `41a6b32dd5`). All three of its roles are retired onto the surviving exact tables:

- **Translation** (py_pc → jitcode): `resume_jitcode_pc_for` now computes the offset from the free fn `derive_resume_marker(first_jit_pc_by_py_pc, block_head_py_by_jit_pc, py)` plus a sparse sidecar `carryfwd_resume_pc: Vec<(u32,usize)>` that captures exactly the non-derivable residual (uncond-jump forward-carry to a jump target, can-raise / branch re-keys). The codewriter build filter is the exact negation of the runtime resolver, so `resume_jitcode_pc_for(py) == old pc_map[py]` for all py by construction, on any corpus. The sidecar is genuinely sparse (140/215 corpus graphs carry zero entries; 363 entries total, vs the dense Vec's one `usize`/PC/graph).
- **Inverse** (jitcode → py, `python_pc_for_jitcode_pc`): re-homed onto the exact `block_head_py_by_jit_pc` + `first_jit_pc_by_py_pc` tables (two dead dense-scan fallbacks deleted).
- **Install-mode discriminator** (`is_populated`/`is_skeleton`/`is_portal_bridge`): re-homed onto `metadata.is_drained` / `jitcode.code.is_empty()` — no `pc_map.is_empty()` test remains.

[...]

PR #407, opened 2026-07-08
jit: decline wasm bridges with CallMayForce re-entry (real fix for the #404 workaround)

youknowone:
## Summary

Real fix for the wasm `float_div_zero_caught_loop` failure that #404 worked around by gating the #366 non-branch direct-pc carry off on wasm.

**Root cause.** Under the m366 carry, the checked-float-truediv deopt (`float_eq(rhs,0.0) -> guard_false`, every 5th iteration) gets a guard side-trace (**bridge**) compiled for it on wasm only. That bridge does `NewWithVtable` + `SetfieldGc` + `CallMayForceR(...1.5-float-const..., v80, 5)` — it re-enters the plain interpreter through a may-force residual call. On wasm the reconstructed interpreter value stack has the `1.5` dividend as NULL, so the re-entered interpreter runs `truediv(NULL, int)` → `TypeError: unsupported operand type(s) for /: '' and 'int'`. **Native never compiles this bridge — it blackholes the deopt** and prints `37500.0`.

The earlier "wasm liveness twin ≠ raw `-live-` decode" (#404) and a later "wasm32 32-bit object-layout corruption" hypothesis were both wrong; the latter was a stale-wasm-module measurement artifact. The blackhole path runs byte-identically on wasm and native — the divergence is native-blackhole vs wasm-bridge.

## Fix

1. **`majit-backend-wasm` `compile_bridge`** — decline any bridge whose ops contain a may-force call, so the deopt routes back through the blackhole interpreter, matching native.
2. **`jitcode_dispatch.rs`** — with the deopt handled, the m366 carry resumes correctly on wasm, so the wasm-only gate in `m366_nonbranch_pc_enabled` (added in #404) and its stale doc-comment are removed. m366 is now ON on all backends (parity).

dynasm/cranelift codegen is unaffected by construction: the lib.rs change is wasm-backend-only, and the removed jitcode gate branch was never taken on native.

## Verification

- `wasm synth/float_div_zero_caught_loop` → `37500.0`
- wasm synthetic suite **173/173**
- dynasm synthetic suite **173/173**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bridge handling to avoid a runtime path that could surface division-related interpreter errors, matching native behavior more closely.
  * Fixed feature flag behavior so the non-branch PC setting now respects the environment variable consistently across platforms.
  
* **Documentation**
  * Updated feature notes to clarify the default behavior and how to disable it.

### DIFF
--- a/majit/majit-backend-wasm/src/lib.rs
+++ b/majit/majit-backend-wasm/src/lib.rs
@@ -1380,6 +1380,20 @@ impl majit_backend::Backend for WasmBackend {
             return Err(BackendError::Unsupported(reason));
         }
 
+        // Decline bridges that re-enter the interpreter through a may-force
+        // residual call. Such a bridge reconstructs the interpreter value stack
+        // from its inputargs before the CallMayForce; on wasm a float-const
+        // dividend materialises as NULL through that inputarg path, so the
+        // re-entered interpreter runs `truediv(NULL, int)` and raises a spurious
+        // `unsupported operand type(s) for /`. Native never compiles this
+        // side-trace — it blackholes the deopt instead. Declining here routes
+        // the deopt back through the blackhole interpreter, matching native.
+        if ops.iter().any(|op| op.opcode.is_call_may_force()) {
+            return Err(BackendError::Unsupported(
+                "wasm backend: bridge with CallMayForce interpreter re-entry".into(),
+            ));
+        }
+
         // The source guard this bridge attaches to. `fail_index` is its index in
         // the source loop's `fail_descrs` / cell array; `trace_id` identifies the
         // owning trace.
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
@@ -7256,24 +7256,7 @@ pub(crate) fn fbw_loop_callee_ca_enabled() -> bool {
 /// avoids that.  Default ON — corpus-wide OFF/ON byte-equality validated on
 /// both backends (171/171 dynasm + cranelift); opt out with
 /// `PYRE_M366_NONBRANCH_PC=0`.
-///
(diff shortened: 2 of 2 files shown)


youknowone:
## ⚠️ Perf regression found — do not merge as-is

A/B/C benchmark of the wasm synthetic suite (WASM column total, per-bench noisy but big ratios reproduce across two independent builds):

| config | bool_arithmetic | is_op | string_ops | suite total |
|---|---|---|---|---|
| **A** = `origin/main` #404 (m366 OFF, no decline) | 2.73s | 0.20s | 1.25s | **125.25s** |
| **B** = this PR (m366 ON + decline) | 15.84s | 1.81s | 3.72s | 139.19s (+11%) |
| **C** = m366 OFF + decline | 16.80s | 2.06s | 3.81s | 144.72s |

Since C ≈ B and both ≫ A, the regression is caused by the **bridge decline**, not m366 (m366 ON/OFF on wasm is noise-level). `bool_arithmetic`/`is_op`/`string_ops` compile *legitimate, fast* CallMayForce bridges on wasm; declining **all** of them kills those speedups. Only the float-truediv deopt bridge is actually broken.

So #404's targeted m366-off was correct **and** had zero perf cost, while this blanket decline is a net regression.

**Real root cause / proper fix direction:** the wasm bridge inputarg/resume path materialises the float const `1.5` as NULL. Lead: `emit_resolve` bakes `constants.get(&opref.raw()).unwrap_or(0)` — a float ConstPtr missing from the bridge resume `constants` table falls back to `0` (NULL). Fixing float-const carriage in the bridge resume data would keep all bridges enabled (perf preserved) and correct — no decline, no m366 gate needed.

Holding this draft pending a decision: (1) real codegen fix [recommended], (2) close and keep #404, or (3) narrow the decline to only the float-const-rebox shape.

Issue #344, opened 2026-07-03
Convert generic majit trace engine from observer/replay two-executor to single authoritative walker (PyPy MetaInterp.interpret parity)


youknowone:
## Summary

The generic `#[jit_interp]` meta-tracing engine in `majit-metainterp` records a
trace with an **observer/replay two-executor** design that diverges from
RPython/PyPy's single-executor model. This issue proposes converting it to a
**single authoritative per-opcode walker**, matching PyPy (`MetaInterp.interpret`
*is* the execution loop). A reference implementation of the target already exists
in-tree (pyre's hand-written `jitcode_dispatch.rs`).

This is a multi-session architecture epic. It is macro-only in blast radius
(does **not** touch `pyre` / `check.py`), and should land behind a flag.

## Motivating smell

`majit/majit-metainterp/src/pyjitpl/dispatch.rs` carries three thread-locals:

- `OBSERVER_MODE` (~L28)
- `OBSERVED_CALLS` (~L29)
- `OBSERVER_REPLAY` (~L36)

These are per-tracer state. In PyPy the equivalent (`MetaInterp.history` and the
tracing mode) lives on the tracer object as a **field**, not thread-local storage.
A thread-local is only forced when the *access site* has no reference to the
tracer — and that is exactly the situation the two-executor design creates:

- the **record** side is `run_one_step(&mut self, ctx: &mut TraceCtx, …)` — it
  *has* the tracer context;
- the **replay** side is the `#[jit_interp]`-generated interpreter mainloop body
  (e.g. `wasm_mainloop`), which runs as the plain concrete interpreter and has
  **no** `TraceCtx` in scope.

Two executors of the same body cannot share a `self`, so they communicate through
an ambient per-thread channel (the TLS). Removing the second executor makes the
TLS collapse into an ordinary field, as PyPy has it.

## PyPy target model (single executor)

In RPython, `MetaInterp.interpret()` **is** the execution loop: the tracer
interprets the jitcode opcode-by-opcode, and that interpretation *is* the
authoritative execution — it mutates VM state while recording trace IR in a single
pass. There is no separate "native mainloop". `self.history` can be a field
because there is exactly one executor with `self` in scope.

This is already documented in-tree (`pyre/pyre-jit/src/eval.rs` ~L3556):

> RPython doesn't see this because `MetaInterp.interpret` IS the execution loop —
> there is no separate `eval_loop_jit`.

## Why the generic majit engine diverged

[...]

PR #415, opened 2026-07-08
majit: retire observer/replay two-executor trace queue (#344 Slice 3)

youknowone:
Completes #344: the generic majit `#[jit_interp]` meta-tracing engine now uses a single authoritative walker, matching RPython's `MetaInterp.interpret` single-executor model. Slices 1-2 landed in #401; this is Slice 3 plus the follow-up cleanup.

## Commits

1. **majit: delete observer/replay two-executor trace queue (#344)** — remove the observer/replay subsystem so the metainterp JitCode walk is the sole executor.
   - Runtime (`dispatch.rs`): delete `OBSERVED_CALLS`/`OBSERVER_REPLAY` thread-locals, the `ObservedCall` enum, `ObserverGuard`, `observer_debug`, `record_observed_{void,int,ref,float,getfield}`, `consume_observed_{void,int,ref,float}_call`, `consume_observed_getfield`, `observed_call_mismatch`, `in_observer_replay`, `drain_observed_calls`, `cancel_observer_replay`, `observer_arg_to_i64`/`observer_i64_to_value`/`observer_raw_word`. Drop the `run_one_step` record gates and the `observer_mode` set/restore in the walk entries.
   - `trace_ctx.rs`: remove the `TraceCtx::observer_mode` field and its two initializers.
   - Macro (`mod.rs`): remove `observer_replay_for_call`/`observer_replay_expr` and the `ObserverReplay*` types; residual calls compile to bare calls. Replace the four `consume_observed_getfield` match arms with a live field read. Drop the now-unused `MarkerRewriter.call_policies`/`io_shims` fields, the `rewrite_body` params that fed them, and `path_segments`.
   - `lib.rs`/`pyjitpl.rs`: drop the removed re-exports.

2. **majit: drop vestigial trace_jitcode_observer_* aliases (#344)** — the three `trace_jitcode_observer{,_with_args,_with_args_and_runtime}` functions were thin passthrough aliases kept to bound the deletion diff. Delete them and point the macro-generated `__trace_*` wrapper directly at `trace_jitcode_with_args_and_runtime`.

3. **check.py: detect the build.rs LLBC preflight failure, not only the translator panic** — the LLBC-missing hint matched only the translator runtime panic string; `pyre-jit-trace/build.rs` fails earlier at compile time with a different string. Match either failure string across stdout+stderr and name both crates in the printed extract command.

## Why the queue was safe to delete

The replay queue was dead for correctness on every exercised path: across all 11 example suites `ObserverGuard::drop` always reported `replay_handover=false`, and in aheui (the only real single-pass consumer) all 1M+ `consume_observed_*` calls were `REPLAY-OFF` passthroughs — zero active replays. A release-active `assert!(!in_observer_replay())` gate at every consume site ran clean across the full surface before deletion. Because the pre-patch wrapper always took the real-call branch when replay was off, the bare call is byte-identical.

## Verification

- `majit-metainterp` test suite + all 11 example crates (tl/tlc/tlr/tla/tiny2/tiny3/tinyframe/braininterp/dualtape/calc/i64env): green.
- aheui: logo `--jit` = 996310 (bit-exact vs `--no-jit`); hello-world / factorial / fibonacci / 99bottles `--jit` == `--no-jit`.
- clippy clean on the touched crates.


youknowone:
End-to-end verification on top of current `main` (LLBC now bootstrapped):

- `python ./pyre/check.py --backend dynasm` → **dynasm 183/183 ALL PASSED** (exit 0), covering the full synth suite (`swap_except_return_resume`, `while_is_none`, unary/type-descr paths, etc.).
- `majit-metainterp` tests + all 11 example crates: green.
- aheui: logo `--jit` = 996310 bit-exact vs `--no-jit`; factorial / fibonacci / 99bottles / hello-world `--jit` == `--no-jit`.
- clippy clean on the touched crates.

— *commented by Claude*

Issue #371, opened 2026-07-04
JIT codewriter: retire remaining pcdep regalloc and walker exception-CFG adaptations


youknowone:
> **STATUS AUDIT (2026-07-23, `main` at `4d652c2dfa6`).** The previous body was stale and internally contradictory. The original walker-slot map/filter work is complete. This issue remains open for two independent structural tails:
>
> 1. pcdep-derived register-allocation compensations that make CPython frame-slot liveness visible to coloring; and
> 2. walker-specific exception-CFG reconstructions in `flatten.rs`.
>
> Draft PR #742 optimizes the cost of the first tail, but intentionally does not claim to delete it.

## Background

This is a follow-up to #267. The runtime flat slot-pinning resume model has been retired, and later work moved runtime lookup to JitCode-keyed metadata:

- `stack_slot_color_map` and the old flat resume readers are gone (#300/#348).
- The py-indexed metadata fields `pcdep_color_slots`, `result_color_at_pc`, and `depth_at_py_pc` are gone (#627). The codewriter still builds temporary py-indexed vectors while constructing the JitCode-keyed `pcdep_trivia_*` / result-color twins; those vectors are not runtime metadata fields.
- Guard-failure resume now carries its Python PC forward rather than recovering it through the old inverse (#727).

The original #371 body continued to describe already-deleted walker machinery and an obsolete `PYRE_NO_XSLOT_FILTER` reproduction. Those descriptions are removed below.

## Completed work

PR #424 and subsequent cleanup retired the mechanisms named in the original title:

- `filter_cross_slot_coalesce_pairs` is deleted.
- `walker_slot_for_variable`, `slot_pre_color`, same-slot pinning, and the post-walk inputarg seeding are deleted from production code.
- Cross-slot candidate rejection now goes through the RPython `RegAllocator._try_coalesce`-equivalent `has_edge` guard on the pre-renaming graph.
- `pcdep_canonical_slot` + `build_slot_disjoint_interference` replaced the old walker-slot post-filter as the source of extra slot-identity edges.

The old four kept-stack/short-circuit failures remain useful regression coverage, but they are no longer a reproduction for a live filter:

- `kept_stack_branch_depths`
- `short_circuit_boxed_int_cross_fn`
- `short_circuit_value_kept_stack`
- `short_circuit_value_local_kept`

## Current pipeline on `main`

`transform_graph_to_jitcode` currently performs the following Ref-bank allocation sequence before canonical flattening:

[...]

PR #424, opened 2026-07-08
jit-regalloc: retire walker-slot cross-slot coalesce filter for pcdep-sourced interference (#371)

youknowone:
Retires the compile-time walker-slot cross-slot coalesce filter (#371), the last load-bearing walker-slot coalescing crutch on the splice path, in favour of RPython-faithful interference-guarded coalescing.

## What changed

`filter_cross_slot_coalesce_pairs` (the walker-slot post-filter over the splice coalesce pairs) is deleted. Cross-slot merges are now rejected through `filter_coalesce_pairs_by_interference`'s `has_edge` guard (`regalloc.py:105 _try_coalesce`), fed by two interference-edge sources:

- **`build_colive_interference`** — co-live outer-frame separations (already present).
- **`build_slot_disjoint_interference`** (new) — an edge between any coalesce candidate whose two endpoints hold **distinct canonical CPython slots**. Canonical slots come from `pcdep_canonical_slot` (new): the slot a Variable occupies at the earliest PC it appears in across the pcdep snapshots, which record each inline frame's slots at that frame's PCs.

Two commits: the `extra_interference` injection hook on `filter_coalesce_pairs_by_interference` (byte-identical when the caller passes `&[]`), then the cutover that wires the two edge sources and deletes the filter.

## Why the second edge source is needed

The co-live edges alone are **not** a sufficient replacement — the filter encodes a slot-identity partition that is *broader* than co-liveness. The failing shape is an inlined callee that keeps a computed value on the operand stack across a conditional-expression guard (`a, b = f(i), g(i); (a + b) + (a if sel else b) - ...`). Instrumenting the compile found the filter dropping two coalesce pairs whose endpoints — the callee's operand-stack temp and the outer merge inputarg — are **never co-live in a single pcdep snapshot** (they live at disjoint PCs). `build_colive_interference` therefore never edges them, the merge is accepted, and coalescing two disjoint-live distinct-slot Variables onto one colour extends that colour's liveness across a box-less region, so a resume reads `OpRef::NONE` in `collect_outer_active_boxes` and panics.

The slot-identity source separates exactly these disjoint-live cross-slot pairs. Because the edges are injected into the coalesce `has_edge` oracle (they drop merges, they are **not** added to `splice_interference`), the resulting colouring is identical to the old filter's — no extra register pressure.

`walker_slot_for_variable` is retained: `walker_slot_for_variable_live` still feeds `slot_pre_color` / `result_color_at_pc`. Migrating those off the walker map (and then deleting it, plus the `insert_exits` walker-compat exception reconstructions) is the remaining #371 work.

## Verification

- `check.py` **dynasm 183/183 + cranelift 183/183**.
- `pyre-jit` unit tests green, including two new tests (`build_slot_disjoint_interference_edges_cross_slot_only`, `pcdep_canonical_slot_takes_earliest_pc`).
- **Perf-neutral** on cranelift (fib_recursive / fannkuch / raise_catch / nbody JIT times equal-or-faster, no timeout) — the injection drops merges only, so the colouring matches the filter.
- Adversarial differential over 43 hostile co-live / kept-stack / swap-unpack / condexpr programs on both backends: the inline-callee regression is fixed and there are **zero new divergences**. The 5 residual failures are pre-existing and orthogonal to coalescing (confirmed by rebuilding the pre-cutover filter baseline): four inline-callee `collect_outer_active_boxes` panics (#124/#68 family) and one heap-int kept-stack SIGSEGV (#225 family).

— *opened by Claude*

## Summary by CodeRabbit

* **New Features**
  * Enhanced JIT register coalescing with stricter cross-instruction slot-aware interference checks, improving correctness around bytecode resume behavior.

* **Bug Fixes**
  * Reduced incorrect coalesce/merge decisions by rejecting merges that would violate per-point slot behavior.
  * Improved result coloring for stack-call results to align with splice/regalloc expectations.

* **Tests**
  * Updated coalescing tests to validate the new slot-disjoint interference logic, including transitive rejection cases.
  * Added coverage for explicitly provided extra interference edges affecting acceptance/rejection.

### DIFF
--- a/pyre/pyre-jit/src/jit/codewriter.rs
+++ b/pyre/pyre-jit/src/jit/codewriter.rs
@@ -4726,9 +4726,8 @@ fn build_slot_disjoint_interference(
     canonical_slot: &[Option<u16>],
 ) -> Vec<(super::flow::VariableId, super::flow::VariableId)> {
     use super::flow::VariableId;
-    let slot_of = |id: VariableId| -> Option<u16> {
-        canonical_slot.get(id.0 as usize).copied().flatten()
-    };
+    let slot_of =
+        |id: VariableId| -> Option<u16> { canonical_slot.get(id.0 as usize).copied().flatten() };
     fn find(parent: &mut HashMap<VariableId, VariableId>, x: VariableId) -> VariableId {
         let mut root = x;
         while let Some(&p) = parent.get(&root) {
@@ -6101,6 +6100,16 @@ impl CodeWriter {
         // precise liveness generation. Stack registers stack_base..stack_base+depth
         // are live at each PC.
         let mut depth_at_pc: Vec<u16> = vec![0; num_instrs];
+        // Per-PC top-of-stack graph Variable at the resume-depth (pre-dispatch)
+        // FrameState, captured at the same program point `depth_at_pc[py_pc]` is
+        // set. `result_color_at_pc` reads the call-result operand slot's colour
+        // from this Variable's canonical splice colour directly, so the capture
+        // never inverts the walker slot map: the call result is not resume-live
+        // (no `pcdep` entry), but the top-of-stack Variable at a call's
+        // fallthrough pc IS the call op's result. `None` where top-of-stack is a
+        // constant / sentinel rather than a Variable.
+        let mut top_of_stack_var_at_pc: Vec<Option<super::flow::VariableId>> =
+            vec![None; num_instrs];
         // Stage 1a (#348, ADDITIVE / gated by `PYRE_PCDEP_VALIDATE`):
         // per-PC snapshot of `slot -> SSA Variable.id` taken from the
         // post-opcode FrameState. The splice regalloc derives the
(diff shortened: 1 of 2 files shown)


youknowone:
Thanks — went through the section-4 adaptations. Disposition:

**Fixed (§4, the splice coalesce oracle machinery):** the amended commit drops the redundant `colive_oracle` from the coalesce filter. The retired `filter_cross_slot_coalesce_pairs` was purely slot-based, so its faithful replacement is the pure-slot `build_slot_disjoint_interference` alone. Co-liveness is not needed to *gate coalescing* — the co-live separations the chordal coloring requires are applied to the interference graph (`splice_interference` via `build_colive_interference`), not to the coalesce filter. Verified identical: check.py dynasm 183/183 + cranelift 183/183, the swap/condexpr adversarial corpus behaves the same as with the co-live oracle (p2 fixed, the pre-existing inline-callee panics unchanged), and perf is neutral. So the coalesce filter is now a single slot-identity edge source.

**Kept as necessary adaptations (the `extra_interference` param, `pcdep_canonical_slot`, slot-identity edges):** pyre mints a fresh `Variable` per `LOAD_FAST` re-read and reuses frame-stack slots across inlined callees, so pyre's SSA `make_dependencies` lacks interference that PyPy's flowspace (one `Variable` per slot per block) has natively. These edges restore exactly that missing interference for the `_try_coalesce` `has_edge` guard, rather than deviating from it — the coalescing still honours `v0 not in dg.neighbours[w0]`.

**Pre-existing, tracked (§4 items on `perform_register_allocation_with_pairs` / the `Kind::Ref`-only external pairs):** the pre-merge-before-`make_dependencies` ordering and the Ref-only external-coalesce restriction predate this patch; they're the remaining #371 endgame (a deeper regalloc refactor that carries the cranelift perf constraint noted in the issue) and are out of scope here.

— *commented by Claude*

Issue #373, opened 2026-07-04
JIT: unported opcode (CALL_INTRINSIC_1 etc.) in a hot loop body declines the whole loop → never compiles


youknowone:
## Summary

A hot loop whose body contains **any** unported opcode (one the JIT codewriter lowers to `abort_permanent`) is **declined as a whole** by the `full_body_walk_trace` up-front check (#125). It never compiles and runs in the pure interpreter forever. `CALL_INTRINSIC_1` (all variants, incl. `UnaryPositive`) is one such unported opcode.

This was found while diagnosing task #6 (`unary_positive_resume`). The task's original "wasm 922×/70×" framing turned out to be a measurement artifact — it is **not** wasm-specific.

## Root cause (general, reusable)

1. The JIT codewriter emits `abort_permanent` for unported opcodes. `CALL_INTRINSIC_1` does this for every variant:
   `pyre/pyre-jit/src/jit/codewriter.rs:10095-10099` → `emit_abort_permanent!(py_pc)`.
   Same for `CallIntrinsic2`, `GetLen`, `LoadSpecial`, `LoadFromDictOrGlobals`, …
2. `full_body_walk_trace` (#73 walker-as-tracer) calls `loop_body_has_abort_permanent(w_code)` (`pyre/pyre-jit-trace/src/trace.rs:1440`): scans jitcode ops after `jit_merge_point`; returns true if any is `abort_permanent`.
3. #125 up-front decline (`trace.rs:1485-1488`): `fbw_decline(key)` + `TraceAction::Abort` — declines the **entire** loop before recording a single op.
4. The trait tracer is retired (gap-10, `trace.rs:207-219`) → a declined key has **no fallback** → the loop **never compiles**, runs interpreter forever. Stats: `loops_aborted=5`, then gives up.

⇒ A single unported op anywhere in a hot loop body kills JIT compilation of the whole loop (not just that op). "Why no bridge on the hot guard" = there is no compiled loop to bridge from.

## Measurement (`unary_positive_resume`, `total += (+i + +(i+1)) & 7`, N=2M)

| runtime | time | note |
|---|---|---|
| CPython 3.14 | 0.13s | baseline |
| pyre-dynasm native | 2.72s | JIT ON == JIT OFF (2.79 vs 2.72) → pure interp, ~20× CPython |
| wasm (wasmtime) | 9.05s | wasm/native = 3.3× = normal wasmtime floor |

No wasm-specific pathology; no missing bridge.

## Fix path

[...]

PR #445, opened 2026-07-09
jit: MAKE_FUNCTION/SET_FUNCTION_ATTRIBUTE residuals + LOAD_ATTR exception-cell fix (#389)

### LINKED ISSUE #389: JIT: hot loops inside a with-block / comprehension never receive a JIT token (no-token cliff behind #373)
> **Status (2026-07-22, merged to `main`) — CLOSED. All three items resolved; the append-fold virtualization now ships default-on and the deferred rooting slices are retired as dead machinery.**
>
> Since the 2026-07-18 note below, the branch rebased onto `main` and pulled in **#344** (retire the trace-time single-executor forks — every production walk is now a full-body walk). That structurally removed the speculative-replay sub-walk the whole "root the nested-comprehension inner-list backing block across a deopt" epic (slices S2a / S2 / S2.2 / S2c) existed to serve: the inner `[[i] …]` backing block is now bound at every guard-exit deopt **without any extra resume-data root**. Consequences:
>
> - **(a) in-loop `MAKE_FUNCTION`, (b) inlined comprehension, (c) `with` cliff — all ✅ RESOLVED** (see the 2026-07-18 note for the per-item root causes / commits).
> - **Nested-list comprehension fold (`[[i] for i in range(n)]`) shipped default-on** (`PYRE_NESTED_LIST_FOLD_VIRT`, #661 / #708). `check.py` green all three backends, `internal_compile_panics=0`, bit-exact vs CPython 3.14, ~27% faster on the acceptance repro. On `wasm32-unknown-unknown` the guest cannot read env vars, so the source default is the only gate-on test.
> - **Rooting slices S2a / S2 / S2.2 = DEAD.** With #344 the backing block needs no extra root, so there is nothing to populate. Both previously-attempted rooting channels (vable-array append; frame liveness boxes) were hard-walled by size/liveness asserts and are moot. **S2c** (the `extra_virtual_roots` resume-data pipe + `GuardCaptureScope::sub_walk_extra_virtual_roots` scaffold) was inert dead code (no writer on any path) and has been **deleted end-to-end in PR #711** (`c358b533303`).
> - **Remaining sub-case `[C(i) for i in range(n)]` (a comprehension body that enters a user Python frame) is NOT tracked here.** It is an adjudicated pyre 2-executor (observer/replay) architectural limitation — force-delivering the in-flight item is the only fix and is unsafe in general (a genuinely-mutating body would double on re-run), and pyre lacks the optimizer-layer object virtualization that would make it safe. It declines cleanly to the interpreter today (byte-correct, un-compiled) and is deferred to single-executor tracing (#73/#34) / object virtualization, not to this issue.
>
> ⇒ #389's three cliffs are all resolved and the follow-on virtualization is shipped. Closing. Any wasm-backend deopt/resume follow-up for the unboxed append fold (currently gated off on wasm via `PYRE_WASM_UNBOXED_APPEND_FOLD`) is a wasm-backend bug, not #389 core.
>
> ---
>
> **Status (2026-07-18, `no-token`) — ALL THREE items now COMPILE with correct results; (b) resolved for call-free comprehension bodies, the remaining sub-case is an adjudicated 2-executor architectural limitation.**
>
> Two commits on `origin/no-token` (`037341694a7`, `cdda4794857`), each zero-regression on both backends (`check.py` 217/217 dynasm + cranelift):
>
> - **(a) in-loop `MAKE_FUNCTION` — ✅ RESOLVED** (unchanged). Repro `loops_compiled=2`, result correct.
> - **(c) `with` cliff — ✅ RESOLVED** (#532, unchanged). Repro `loops_compiled=2`, result correct.
> - **(b) inlined-comprehension `[j for j in range(n)]` — ✅ RESOLVED (compiles).** The issue's exact repro (`[j for j in range(10)]` in `while<3000`) now returns `30000` with `loops_compiled=2`; the standalone `[i for i in range(1000)]` inner loop compiles `loops_compiled=1 loops_aborted=0` (was `0`). Two root causes fixed:
>     1. **`037341694a7` — operand-stack mirror corruption at the comprehension FOR_ITER-entry permutation.** The py3.14 inlined comprehension emits a `SWAP`/`BUILD_LIST`/`SWAP` preamble before its `FOR_ITER`, which the codewriter lowers non-monotonically in jitcode; the JitCode-PC floor pivot maps a later jit_pc back to an earlier py_pc, so `reconcile_vstack_at_boundary` re-applied the `SWAP` effect and reused a stale `vstack_last_ref` across the re-walk, dropping the just-built list from the `vstack_boxes` mirror. The `stack_sync` overlay then copied the corrupt mirror into the resume snapshot, so a deopt reconstructed a frame whose `FOR_ITER` iterator slot held a duplicated iterator and a dropped list (`"not an iterator"`). Fix: detect the backward (non-monotonic) py transition after a `Swap`/`Copy` and reseed the mirror from the virtualizable shadow (`ShadowReseed`) across the permutation region.
>     2. **`cdda4794857` — LIST_APPEND list receiver unbound inside the resident loop.** The codewriter lowered LIST_APPEND's list operand by cloning the compile-time FlowValue, but the comprehension accumulator lives on the operand stack, which is not loop-carried in registers (`jit_merge_point` reds are `[frame, ec]`), so inside the resident loop that slot's register was unbound → the `jit_list_append` residual read an unbound receiver and aborted `ResidualCallArgUnbound` before ever reaching the `#171` append fold (so the comprehension inner loop never compiled and the empty-append fold was dead). Fix: reload the list receiver from its value-stack slot via `getarrayitem_vable_r` each iteration, the same per-iteration re-materialization the `FOR_ITER` iterator reload already uses.
>
>   **Remaining sub-case (guarded, not a correctness bug): a comprehension whose body enters a user Python frame, e.g. `[C(i) for i in range(n)]`.** Such a body's per-element call sets the FOR_ITER in-flight body-effect flag; a mid-body abort then routes through `fbw_foriter_inflight_take`, which refuses in-flight delivery to avoid a double-apply, dropping the trace-attempt iteration's item. `for_iter_bodies_all_jit_safe` admits `LIST_APPEND` only when its `FOR_ITER` body performs no call, so this shape declines cleanly to the interpreter (byte-correct, un-compiled). Empirically, force-delivering the item makes it correct — the refusal is the sole blocker — but delivery is unsafe in general (a body with a genuine external mutation would double on re-run), and pyre cannot cheaply distinguish `C.__init__`'s write-to-fresh-`self` from a global-mutating call without the object virtualization PyPy has at the optimizer layer. This is a genuine pyre 2-executor (observer/replay) limitation, deferred to single-executor tracing (#73/#34) or object virtualization — not an append-fold/gate tweak. Result stays byte-correct throughout.
>
> ---
>
> <details><summary>Earlier status notes (superseded, kept for history)</summary>
>
> **Status (2026-07-15, later update — on the `no-token` branch atop the `rewrite-tracer` base where #73 is landed) — items (a) and (c) RESOLVED (#532, MERGED `ef7d2089134`); (b) advanced through THREE walker layers, now narrowed to a single sub-walk epic. The three landed layers are pushed to `origin/no-token` (`7fdca719f50`), each zero-regression on both dynasm and cranelift (`check.py` 187 pass / 1 pre-existing BASEFAIL `synth/list_insert_pop_index`).**
>
> - **(a) in-loop `MAKE_FUNCTION`/`SET_FUNCTION_ATTRIBUTE` — ✅ RESOLVED.** Per-iteration `def` repro compiles: `loops_compiled=1 loops_aborted=0`. Result correct.
> - **(c) `with` cliff — ✅ RESOLVED by #532** (drop the `WITH_EXCEPT_START` gate). With-loop repro compiles `loops_compiled=2`. The `test_shlex` regression the gate-drop exposed is fixed inside #532 (`read_token` mapdict `LOAD_ATTR` fold dropped the promoted-map `guard_value` → wrong storage slot; fixed by emitting the map guard unless the map op is already constant).
> - **(b) inlined-comprehension `[j for j in range(n)]` — PARTIALLY ADVANCED; now blocked on the sub-walk epic, NOT the operand-materialization gap the prior note named.** On the `rewrite-tracer` base the outer hot loop containing an inlined comprehension IS admitted and traced (no longer declined at the #57 frame gate); the walk reaches the comprehension's `L5` epilogue. Three layers landed, each moving the abort strictly forward and each independently zero-regression:
>     1. **`a4106b87199`** — classify `LIST_APPEND`/`SET_ADD`/`MAP_ADD`/`LIST_EXTEND` as `PopOnlyOrSideStore` in the walk-level operand-stack mirror (`classify_vstack_opcode`), so the mirror stays valid across the comprehension loop instead of latching `vstack_valid=false`. Moved the abort from `ResidualCallArgUnbound` → `JitMergePointGreenKeyUnresolved`.
>     2. **`b546b4f2be8`** — forward-select the nested inner-loop header merge point (delete `header_entry_is_nested`), so the comprehension's inlined `FOR_ITER` gets a concrete pycode green instead of `OpRef::NONE`. Moved the abort to `GuardSnapshotVableUntyped` + `UnfoldableListAppendResidualUnsupported`.
>     3. **`7fdca719f50`** — recover a nested-loop-carried `STORE_FAST` value from the vstack mirror. **RCA (probe-confirmed):** the `L5` epilogue `STORE_FAST xs` / `STORE_FAST j` lower to `setarrayitem_vable_r` whose value register reads `OpRef::NONE` — `xs` (the `BUILD_LIST` accumulator) and saved `j` (`LOAD_FAST_AND_CLEAR`) are operand-stack temps that span the inner `FOR_ITER`, and the loop-header re-seed seeds no operand-stack colors, so their registers are lost. The value survives ONLY in the vstack mirror TOS (`InputArgRef`), not in the vable shadow (cleared to `ConstPtr(0)`). Reading the mirror TOS instead of the stale register avoids writing an untyped box into the virtualizable array (which failed the guard-snapshot buildability precondition, `GuardSnapshotVableUntyped`). Notably this was NOT the `#124` multi-frame-vable case the earlier note guessed: the identity slot was typed; two operand-stack-temp array slots were `OpRef::NONE`.
>
>   **Remaining wall = `UnfoldableListAppendResidualUnsupported` → the sub-walk guard-side-exit resume epic (#62/#73/#34), NOT an append-fold/gate tweak (probe-confirmed, superseding the prior "blocked on operand materialization / #73 seeding" framing).** With the three layers in place the sole remaining abort is the list-append fold declining. Drilled with staged experiments (all reverted): the comprehension's FIRST traced append is **`strategy=Empty`, `value=int`**, so two barriers stack — (1) the `Empty→Integer` strategy SWITCH (the `#171` fold only recognizes already-typed Integer/Float/Object storage; `Empty` matches none), and (2) the first alloc (Empty has no backing array). Bypassing the `recognize` gate to admit `Empty`+int makes the commit **sub-walk** abort `OrthodoxSubWalkTraceUnsupported`: descending the `w_list_append` body hits `switch_to_correct_strategy` + `grow_list_items_block` (`#[dont_look_inside]`), whose in-body helper carries a `symbolic_fnaddr` (`>>47` hash, e.g. the `SyntheticTransparentCtor "Tuple"` unit ctor) that a compiled sub-walk trace must not bake as a code address (→ SIGSEGV). So (b) is genuinely gated behind either (A) a realloc-safe append journal / retiring the legacy replay-from-snapshot rollback (the `FBW_APPEND_JOURNAL` is pure-length-rewind only; a realloc frees/relocates the block irreversibly), or (B) a correct sub-walk residual + guard-side-exit reconstruction for the `dont_look_inside` grow. Being on the `rewrite-tracer` base (#73 landed) did not dissolve either wall — they are downstream of #73. Multi-session; correctly deferred. Result stays byte-correct (`30000`) throughout — (b) is a missing-speedup cliff, never a correctness bug.
>
> ---
>
> **Correction (2026-07-08, second refresh) — the first refresh note above mis-diagnosed the comprehension case; this supersedes it.**
> Re-measured on a clean build at the current tip (after #344 Phase B / #426 / #427). Two claims in the first refresh note are **wrong** and are corrected here.
>
> **Correction 1 — pyre DOES inline comprehensions (PEP 709).** The first note claimed "RustPython compiles a comprehension as a separate `<listcomp>` code object (no PEP 709 inlining)". That is false: `[sys._getframe().f_code.co_name for _ in range(1)]` returns `'main'`, not `'<listcomp>'` — the same as CPython ≥ 3.12 and pyre's 3.14 compat target. This is interpreter-level (holds under `PYRE_JIT=0`), so it is not a JIT artifact. The listcomp's `FOR_ITER` + `LIST_APPEND` are inlined directly into the enclosing frame.
>
> **Correction 2 — the two repros have DIFFERENT root causes; the `0/5` abort does not belong to the comprehension.** The first note conflated them. Measured separately:
>
> - **comprehension `[j for j in range(10)]` in a hot loop** → `loops_compiled=0 loops_aborted=0`, **`main` never starts tracing** (0 `start tracing` lines even at 200k iters). Because the listcomp is inlined, `main` is a `FOR_ITER` frame whose body contains `LIST_APPEND`, which is **not** in the `for_iter_body_op_is_jit_safe` allow-list (`eval.rs`). So `unsupported_jit_shape` classifies `main` as `CurrentFrameOnly` and declines it **before the counter ticks** (`eval_with_jit_inner`) — the #57 non-journalable-mutator soundness gate. **Direct control:** an explicit inlined `for j in range(10): out.append(j)` in the hot loop (no comprehension, no `def`) declines *identically* (`traces_started=0`) — confirming the comprehension is incidental; the cause is the inlined-append #57 gate.
>
> - **a per-iteration inner `def` whose body is `for`+`list.append`** (the first note's "minimal repro") → `loops_compiled=0 loops_aborted=5`, and `main` **does** start tracing then aborts. But the abort is **not** "walker fails to residualize the declined callee": a per-iteration `def` that is **defined but never called** still aborts `0/5`. The real cause is `MAKE_FUNCTION` → `emit_abort_permanent!` (`codewriter.rs:9280`, and its partner `SET_FUNCTION_ATTRIBUTE` at :10647), which lands after the loop header `jit_merge_point`, so the up-front `loop_body_has_abort_permanent` scan (`trace.rs:1638`) declines the whole outer loop. The trait tracer already residualizes this op (`trace_make_function` → `jit_make_function_from_globals`, `helpers.rs:558`); only the full-body-walk codewriter aborts. The named-module-helper control (identical loop, `def` hoisted out) compiles (`loops_compiled=1`), isolating the in-loop `MAKE_FUNCTION` as the sole difference.
>
> **Revised remaining work (three independent items, all correctness-neutral cliffs — every repro produces the correct result today):**
> - **(a) in-loop `MAKE_FUNCTION`/`SET_FUNCTION_ATTRIBUTE`** — port the `newfunction` residual into the full-body-walk codewriter (the trait tracer already has it; MAKE_CELL at `codewriter.rs:11454` is the residual template). This is the tractable next step and is **not** blocked by #57 or task#50. Fixes any hot loop that builds a function/closure per iteration.
> - **(b) inlined-comprehension / inlined `for`+`append`** — the `LIST_APPEND` (and `STORE_SUBSCR`/`SET_ADD`/`MAP_ADD`) #57 gate. Requires a realloc-aware journaled append (a comprehension grows an empty list through every capacity boundary, so the spare-capacity `orthodox_list_append_commit` fold can never apply) plus the call-body side-effect-double fix (task#50). Multi-session.
> - **(c) `with` cliff** — the `WITH_EXCEPT_START` exception-link lowering (#373/#395). Declines cleanly and census-visibly today (`FrameShape::StructuralRegion`); removing the gate miscompiles real stdlib `with`-frames (the reverted S1 / PR#410 CI regression).
>
> The original "invisible cliff" framing (tracer never starts / empty census / gate upstream) was already corrected in the first note and remains fixed by PR #410. Everything below is the original report, kept for history.

---

> **Refresh (2026-07-08, after PR #410) — the original diagnosis below is superseded.**
> The "no-token cliff" as originally framed (*"the tracer never starts; `loops_compiled=0` **and** `loops_aborted=0`; empty FBW census; the gate is upstream of the tracer"*) **no longer holds**. PR #410's codewriter walker fall-through-into-joinpoint fix removed the invisibility: the tracer now **does** start on both repros, reaches the frame-shape gate, and declines **visibly** (non-empty census). Re-measured on the current `no-token` tip:
>
> **`with` case** — `loops_compiled=0` but census is now populated: `FrameShape::StructuralRegion` + `FrameShape::CurrentFrameOnly` entries. This is the **restored `StructuralRegion` gate** (S1 was reverted: removing it miscompiled `with`-frames — SIGSEGV in the exception path + guard storms, the PR #410 CI regression). The loop declines cleanly because its `WITH_EXCEPT_START` exception-link lowering is still residualized. **Compiling it needs the exception-link lowering — that is #373 work, correctly deferred here.**
>
> **comprehension case** — `loops_compiled=0 loops_aborted=5`, census shows `FullBodyWalk::LoopBodyAbortPermanent: 5` + `Trait::DeclinedAbort: 5` on the outer `main` loop, and `FrameShape::CurrentFrameOnly` on the comprehension frame. Isolation (repros below) pins the real cause: it is **not** "the tracer never starts" and **not** the comprehension per se. RustPython compiles a comprehension as a **separate `<listcomp>` code object** (no PEP 709 inlining) containing `FOR_ITER` + `LIST_APPEND` → the #57 non-journalable-mutator gate declines that callee as `CurrentFrameOnly` (correct). The remaining defect is that the **outer `while` loop aborts (`abort_permanent`) when the walker tries to inline that declined callee, instead of emitting it as a residual call and compiling the outer loop.** This is the **same family as task#38** (walker aborts on a callee it should residualize). Minimal reproduction that is *not* a comprehension but reproduces `0/5` exactly:
>
> ```python
> def main():
>     total = 0; i = 0
>     while i < 3000:
>         def lc():                 # per-iter MAKE_FUNCTION + FOR_ITER + list.append
>             out = []
>             for j in range(10):
>                 out.append(j)
>             return out
>         xs = lc()
>         total += len(xs)
>         i += 1
>     print(total)
> main()
> ```
>
> Controls that isolate it (all on the current tip):
> - `while` loop, no call → **compiles** (`loops_compiled=1`).
> - `while` loop calling a plain `helper()` returning a constant → **compiles**.
> - `while` loop calling a **named** helper with `for`+`list.append` → **compiles** (`loops_compiled=1`): a declined-mutator callee reached via a *named* function is residualized fine.
> - `while` loop with a **per-iteration** inner function (or comprehension) whose body is `for`+`list.append` → **`loops_compiled=0 loops_aborted=5`**: the walker aborts the outer loop instead of residualizing.
>
> **Revised remaining work:** (a) comprehension/`while` cliff = make the walker **residualize a declined callee and keep compiling the outer loop** (task#38 family — the walker's abort-on-uncompilable-callee should be a residual, not an outer-loop `abort_permanent`); (b) `with` cliff = the `WITH_EXCEPT_START` exception-link lowering (#373). The invisibility bug (original framing) is fixed by PR #410. Everything below is the original report, kept for history.

---

## Summary

A hot loop whose body sits inside a `with` block (or another
exception-table / block-stack region), and a hot loop that rebuilds a
comprehension each iteration, **never receive a JIT token at all**. The tracer does
not start, no loop is recorded, and the loop runs in the pure interpreter forever.

This is a distinct failure mode from both:

- **#373** (`abort_permanent` declines a loop the tracer *did* start), and
- **#152** (a loop that *did* compile fails to stay resident).

Here the tracer is never even invoked — `loops_compiled=0` **and** `loops_aborted=0`,
with an empty FBW census (the `abort_permanent` up-front decline is never reached). It is
the "no-token cliff" that masks the latent-but-portable `abort_permanent` opcodes noted in
#373 (`WITH_EXCEPT_START`, `LOAD_SPECIAL Enter/Exit`, the comprehension accumulators):
their ports are correct, but no bench can demonstrate a win because the loops that contain
them never compile in the first place.

## Reproduction (dynasm/cranelift, `84261b23a5`)

**`with` block in a hot loop — never compiles:**

```python
class CM:
    def __enter__(self): return self
    def __exit__(self, *a): return False

def main():
    total = 0
    i = 0
    while i < 200000:
        with CM():
            total += 1
        i += 1
    print(total)

main()
```

```
$ MAJIT_STATS=1 pyre-dynasm with_loop.py
200000
[jit-stats] loops_compiled=0 bridges_compiled=0 loops_aborted=0 guard_failures=0 internal_compile_panics=0
```

**Control — identical loop without `with` — compiles:**

```python
def main():
    total = 0
    i = 0
    while i < 200000:
        total += 1
        i += 1
    print(total)
main()
```

```
[jit-stats] loops_compiled=1 bridges_compiled=0 loops_aborted=1 guard_failures=1 internal_compile_panics=0
```

**Comprehension rebuilt each iteration — never compiles:**

```python
def main():
    total = 0
    i = 0
    while i < 3000:
        xs = [j for j in range(10)]
        total += len(xs)
        i += 1
    print(total)
main()
```

```
[jit-stats] loops_compiled=0 bridges_compiled=0 loops_aborted=0 guard_failures=0 internal_compile_panics=0
```

## Evidence that the tracer never starts

- `MAJIT_STATS`: both cliff cases show `loops_compiled=0` **and** `loops_aborted=0`. An
  aborted trace would bump `loops_aborted`; a declined `abort_permanent` body would show
  in the FBW census. Neither happens.
- `PYRE_FBW_DEBUG_ABORT=1` census is **empty** for both cases — the `abort_permanent`
  up-front decline (`FullBodyWalk::LoopBodyAbortPermanent`, trace.rs) is never reached.
- `MAJIT_LOG=1` prints **zero** `start tracing` / `loop_header` / `jit_merge_point` lines
  for the `with` case, versus one `start tracing` for the identical `with`-free control.

So the gate is *upstream* of the tracer and of the `abort_permanent` decline: the hot loop
inside the exception-table region is never detected as a compilation candidate (or its
`can_enter_jit` back-edge is never counted toward the compile threshold).

`emit_loop_header` (the lowered `can_enter_jit`, codewriter.rs:2224, emitted
unconditionally at each `JumpBackward`, codewriter.rs:7994-8015) is present in the
jitcode, so the miss is not "no loop_header op emitted". The next suspect is the runtime
hot-loop threshold accounting (warmstate) or how a back-edge inside a nonempty block-stack
region (`blocklist`, codewriter.rs:261) is treated — root cause to be pinned in
investigation.

## Why it matters

1. `with` blocks and comprehensions are common in real Python; any hot loop using them
   silently falls off the JIT.
2. It blocks closing #373: the latent-but-portable opcodes (`WITH_EXCEPT_START`,
   `LOAD_SPECIAL`, `SET_ADD`/`MAP_ADD`/`DICT_MERGE` and friends) are ported and correct,
   but this cliff prevents any hot loop containing them from ever compiling, so the ports
   cannot be demonstrated and #373 cannot be fully verified against real loops.

## Investigation / fix path

1. Pin the exact gate: instrument the warmstate hot-loop counter / `can_enter_jit`
   back-edge accounting for a back-edge whose `blocklist` is nonempty (inside a `with` /
   `try` region), and for the comprehension inner function. Determine whether the back-edge
   is never counted, counted under a different green key, or the candidate is filtered
   before tracing.
2. Compare against RPython `can_enter_jit` / `jit_merge_point` semantics for a
   loop nested inside a block-stack region — RPython compiles these.
3. Once a loop inside a `with` compiles, re-check that the ported `WITH_EXCEPT_START` /
   `LOAD_SPECIAL` residuals behave (they were only latent-verified in #373).

## Related

- #373 — the `abort_permanent` opcode-decline issue this cliff sits behind; its
  latent-B opcode ports are masked by this cliff.
- #152 — residency of *already-compiled* loops (the opposite end: those loops do get a
  token; these never do).
- The comprehension case matches the earlier "list-comprehension retrace" observation
  (a comprehension-built list on a hot path); consolidated here as the same
  no-token-cliff family.

</details>

youknowone:
Closes #389 item (a), the "no-token cliff": a hot loop that builds a function per iteration previously declined to compile because the function-definition opcodes aborted in the full-body-walk (FBW) codewriter. This branch residualizes them and fixes an independent pre-existing LOAD_ATTR correctness bug found along the way.

## Commits

- **detect closure self-recursion (LoadDeref) in static walker decline** — `code_is_self_recursive` only matched `LoadGlobal <own-name>`, missing a nested function that recurses via its closure cell (`LoadDeref` / `LoadFromDictOrDeref` of a free/cell var named after itself). Such a 2+-arg self-recursive callee slipped into the full-body walker and hit the concrete-append-then-abort double-mutation. Now matches `LoadDeref`/`LoadFromDictOrDeref` against `code.obj_name` via `deref_name_and_kind`; the `named_function_code` test helper recurses into nested code objects and a closure-self-recursion unit test was added.

- **residualize MAKE_FUNCTION in the full-body-walk codewriter (#389)** — the FBW codewriter aborted MAKE_FUNCTION with `emit_abort_permanent!`, so the up-front `loop_body_has_abort_permanent` scan declined the whole enclosing hot loop. Lowers MAKE_FUNCTION to `make_function_value(globals, code)` → `residual_call_r_r(make_function_fn, ListR[globals, code])`, mirroring the MAKE_CELL residual (`globals` is the code's `w_globals` baked as a `Signed(ptr) + Kind::Ref` constant, `code` is the popped TOS, the result replaces the code object on the shadow stack). Reuses `jit_make_function_from_globals` (Plain — allocates, runs no user code, never raises), registered as `make_function_fn`.

- **publish LOAD_ATTR residual raises into the backend exception cells** — independent pre-existing correctness bug. `bh_load_attr_fn` published a getattr raise into only `BH_LAST_EXC_VALUE`, not the backend `_store_exception` cells that a compiled trace's `GUARD_NO_EXCEPTION` reads — so the guard saw a stale 0, passed, and let the NULL result flow to the consumer, silently swallowing the exception (`try/except` caught nothing). Routes the error arm through `publish_residual_call_exception`, which writes both cells, matching every other raising `bh_*` residual. Isolated repro: a `@property` raising `ValueError` read in a hot `try/except` loop yielded 203 instead of 100000; proven independent of any function-definition opcode.

- **residualize SET_FUNCTION_ATTRIBUTE in the full-body-walk codewriter (#389)** — the def-time partner of MAKE_FUNCTION still aborted, so a loop with a closure- or default-bearing `def` compiled the outer loop but declined at the attribute op. Lowers SET_FUNCTION_ATTRIBUTE to `set_function_attribute(func, attr, flag)` → `residual_call_ir_r(set_function_attribute_fn, ListI[flag], ListR[func, attr])` → `jit_set_function_attribute(func, attr, flag)`, mirroring the IMPORT_NAME int+ref residual. `flag` is the `MakeFunctionFlag` `#[repr(u8)]` discriminant baked as a compile-time int; the helper matches on it and applies the same setter as `set_function_attribute_with_flag`. Plain — stamps a typed field, runs no user code, never raises. A closure-per-iteration hot loop now compiles fully.

## #389 status

The MAKE_FUNCTION + SET_FUNCTION_ATTRIBUTE work completes #389 item (a). Items (b) comprehension realloc-append and (c) `with` WITH_EXCEPT_START remain, blocked on task#50 and #373 respectively.

## Verification

- check.py: 155/155 on both dynasm and cranelift backends.
- pyre-jit lib unit tests pass.

### DIFF
--- a/pyre/pyre-interpreter/src/runtime_ops.rs
+++ b/pyre/pyre-interpreter/src/runtime_ops.rs
@@ -61,6 +61,48 @@ pub extern "C" fn jit_make_function_from_globals(globals: i64, code_obj: i64) ->
     make_function_from_code_obj_with_globals_obj(code_obj as PyObjectRef, w_globals) as i64
 }
 
+/// SET_FUNCTION_ATTRIBUTE residual: stamp one attribute on `func` per the
+/// compile-time `flag` discriminant, then return `func` so the caller pushes
+/// the same function back onto the stack (net stack effect -1).  `flag` is the
+/// `MakeFunctionFlag` bit-position discriminant (`oparg.rs`
+/// `MakeFunctionFlag`): `Defaults = 0`, `KwOnlyDefaults = 1`,
+/// `Annotations = 2`, `Closure = 3`, `Annotate = 4`, `TypeParams = 5`.  The
+/// match must stay in sync with `eval.rs`
+/// `set_function_attribute_with_flag` and the codewriter's
+/// `SetFunctionAttribute` arm.  Sets a typed field on the function; runs no
+/// user code and never raises → `Plain`.
+#[majit_macros::dont_look_inside]
+pub extern "C" fn jit_set_function_attribute(func: i64, attr: i64, flag: i64) -> i64 {
+    let func = func as PyObjectRef;
+    let attr = attr as PyObjectRef;
+    match flag {
+        0 => unsafe { crate::function_set_defaults(func, attr) },
+        1 => unsafe { crate::function_set_kwdefaults(func, attr) },
+        2 => {
+            // `function.py:553-559 fset_func_annotations` — the eager
+            // annotations dict is stamped on the typed `Function.w_ann`
+            // slot via `function_write_barrier`, so `f.__annotations__ is
+            // f.__annotations__` holds through the getattr arm.
+            unsafe { crate::function::function_set_annotations(func, attr) };
+        }
+        3 => unsafe { crate::function_set_closure(func, attr) },
+        4 => {
+            // PEP 649: `attr` is the `__annotate__` callable the runtime
+            // `__annotations__` getter evaluates lazily; stored on the
+            // typed `w_annotate` slot.
+            unsafe { (*(func as *mut crate::function::Function)).w_annotate = attr };
+            // The direct field store bypasses `function_write_barrier`;
+            // record it for the prebuilt-root minor-collection skip, exactly
+            // as the interpreter does.
+            pyre_object::gc_roots::mark_prebuilt_roots_dirty();
+        }
+        // `TypeParams = 5` carries a PEP 695 type-parameter tuple; pyre has
+        // no PEP 695 surface, so the operand is accepted silently.
+        _ => {}
+    }
+    func as i64
+}
+
 #[majit_macros::dont_look_inside]
 pub extern "C" fn jit_load_name_from_namespace(
     frame_ptr: i64,
--- a/pyre/pyre-jit-trace/src/trace.rs
+++ b/pyre/pyre-jit-trace/src/trace.rs
@@ -121,14 +121,17 @@ fn start_pc_is_loop_header(code: &pyre_interpreter::CodeObject, start_pc: usize)
     false
 }
 
-/// Heuristic static test: does `code` call a global whose name is its own
-/// (`co_name`)?  That is the self-recursion shape in bytecode — `LOAD_GLOBAL
-/// <own-name>` feeding a `CALL`.  The name-index low bit is the `push_null`
-/// flag (`pyopcode.rs` load-global decode), so the real `co_names` index is
-/// `namei >> 1`.  A shadowed same-name global is a false positive, but the only
-/// cost is declining a walker trace that would abort anyway.
+/// Heuristic static test: does `code` load its own name (`co_name`) feeding a
+/// `CALL`?  That is the self-recursion shape in bytecode.  A module-level
+/// function loads itself with `LOAD_GLOBAL <own-name>` (the name-index low bit
+/// is the `push_null` flag, so the real `co_names` index is `namei >> 1`); a
+/// nested/closure function loads itself with `LOAD_DEREF`/`LOAD_FROM_DICT_OR_DEREF`
+/// of the cell/free var whose name is its own.  A shadowed same-name binding is
+/// a false positive, but the only cost is declining a walker trace that would
+/// abort anyway.
 fn code_is_self_recursive(code: &pyre_interpreter::CodeObject) -> bool {
     use pyre_interpreter::Instruction as I;
+    let own_name: &str = code.obj_name.as_ref();
     let mut arg_state = pyre_interpreter::OpArgState::default();
     let mut self_name_loaded = false;
     let mut has_call = false;
@@ -137,8 +140,14 @@ fn code_is_self_recursive(code: &pyre_interpreter::CodeObject) -> bool {
         match instr {
             I::LoadGlobal { namei } => {
                 let idx = (namei.get(op_arg) as usize) >> 1;
-                let own_name: &str = code.obj_name.as_ref();
                 if code.names.get(idx).map(|n| -> &str { n.as_ref() }) == Some(own_name) {
+                    self_name_loaded = true;
+                }
+            }
+            I::LoadDeref { i } | I::LoadFromDictOrDeref { i } => {
+                let idx = i.get(op_arg).as_usize();
+                let (name, _is_free) = pyre_interpreter::deref_name_and_kind(code, idx);
+                if name == own_name {
                     self_name_loaded = true;
                 }
             }
@@ -2057,16 +2066,21 @@ mod tests {
     }
 
     fn named_function_code(source: &str, name: &str) -> pyre_interpreter::CodeObject {
-        let module = compile_exec(source).expect("test code should compile");
-        module
-            .constants
-            .iter()
-            .find_map(|constant| match constant {
-                pyre_interpreter::ConstantData::Code { code } if code.obj_name.as_str() == name => {
-                    Some((**code).clone())
+        fn find_in(code: &pyre_interpreter::CodeObject, name: &str) -> Option<pyre_interpreter::CodeObject> {
+            for constant in code.constants.iter() {
+                if let pyre_interpreter::ConstantData::Code { code: inner } = constant {
+                    if inner.obj_name.as_str() == name {
+                        return Some((**inner).clone());
+                    }
+                    if let Some(found) = find_in(inner, name) {
+                        return Some(found);
+                    }
                 }
-                _ => None,
-            })
+            }
+            None
+        }
+        let module = compile_exec(source).expect("test code should compile");
+        find_in(&module, name)
             .unwrap_or_else(|| panic!("test source should contain function {name}"))
     }
 
@@ -2090,6 +2104,26 @@ mod tests {
             "fill",
         );
         assert!(!super::code_is_self_recursive(&code));
+    }
+
+    #[test]
+    fn code_is_self_recursive_detects_closure_self_call() {
+        // A nested function loads its own name via LOAD_DEREF (closure cell),
+        // not LOAD_GLOBAL — the scanner must still recognize the self-call.
+        let code = named_function_code(
+            "def outer():\n    def rec(xs, n):\n        xs.append(n)\n        if n > 0:\n            return rec(xs, n - 1) + 1\n        return 1\n    return rec\n",
+            "rec",
+        );
+        assert!(
+            super::code_is_self_recursive(&code),
+            "closure self-recursion (LOAD_DEREF) must be detected"
+        );
+        assert_eq!(code.arg_count, 2);
+        // 2-arg closure self-recursion traced from entry must decline.
+        assert!(
+            decline_predicate(&code, 0),
+            "2-arg closure self-recursion must be declined"
+        );
     }
 
     /// Mirror of the `static_walker_should_decline` predicate on a raw
--- a/pyre/pyre-jit-trace/src/trace.rs
+++ b/pyre/pyre-jit-trace/src/trace.rs
@@ -2066,7 +2066,10 @@ mod tests {
     }
 
     fn named_function_code(source: &str, name: &str) -> pyre_interpreter::CodeObject {
-        fn find_in(code: &pyre_interpreter::CodeObject, name: &str) -> Option<pyre_interpreter::CodeObject> {
+        fn find_in(
+            code: &pyre_interpreter::CodeObject,
+            name: &str,
+        ) -> Option<pyre_interpreter::CodeObject> {
             for constant in code.constants.iter() {
                 if let pyre_interpreter::ConstantData::Code { code: inner } = constant {
                     if inner.obj_name.as_str() == name {
--- a/pyre/pyre-jit/src/call_jit.rs
+++ b/pyre/pyre-jit/src/call_jit.rs
@@ -4021,8 +4021,15 @@ pub extern "C" fn bh_load_attr_fn(obj: i64, w_code_ptr: i64, name_idx: i64) -> i
     match pyre_interpreter::baseobjspace::getattr_str(obj as pyre_object::PyObjectRef, name) {
         Ok(attr) => attr as i64,
         Err(err) => {
+            // Publish the raise into BOTH the blackhole `BH_LAST_EXC_VALUE` and
+            // the backend `_store_exception` cells (`publish_residual_call_exception`).
+            // The LOAD_ATTR residual runs under the blackhole interpreter AND
+            // inside a compiled trace; writing only `BH_LAST_EXC_VALUE` leaves a
+            // compiled trace's `GUARD_NO_EXCEPTION` reading a stale 0, so the
+            // guard wrongly passes and the NULL result flows to the consumer (a
+            // raising property getter in a compiled loop is silently swallowed).
             let exc_obj = err.to_exc_object();
-            majit_metainterp::blackhole::BH_LAST_EXC_VALUE.with(|c| c.set(exc_obj as i64));
+            publish_residual_call_exception(exc_obj as i64);
             0
         }
     }
--- a/pyre/pyre-jit/src/call_jit.rs
+++ b/pyre/pyre-jit/src/call_jit.rs
@@ -4486,8 +4486,13 @@ pub extern "C" fn bh_load_name_fn(frame_ptr: i64, w_name: i64, namei: i64) -> i6
     match frame.load_name_checked_value(name, namei as usize) {
         Ok(w_value) => w_value as i64,
         Err(err) => {
-            let exc_obj = err.to_exc_object();
-            majit_metainterp::blackhole::BH_LAST_EXC_VALUE.with(|c| c.set(exc_obj as i64));
+            // Publish into BOTH the blackhole `BH_LAST_EXC_VALUE` and the
+            // backend `_store_exception` cells: LOAD_NAME lowers into the
+            // full-body-walk compiled trace (a handler-bearing body skips the
+            // cell-fold), where the following `GUARD_NO_EXCEPTION` reads the
+            // backend cells — writing only `BH_LAST_EXC_VALUE` lets the guard
+            // pass on a stale 0 and a raising LOAD_NAME is silently swallowed.
+            publish_residual_call_exception(err.to_exc_object() as i64);
             0
         }
     }
@@ -4517,8 +4522,12 @@ pub extern "C" fn bh_store_name_fn(frame_ptr: i64, w_name: i64, value: i64) -> i
     match frame.store_name_value(name, 0, value as pyre_object::PyObjectRef) {
         Ok(()) => 1,
         Err(err) => {
-            let exc_obj = err.to_exc_object();
-            majit_metainterp::blackhole::BH_LAST_EXC_VALUE.with(|c| c.set(exc_obj as i64));
+            // Publish into both exception cells: STORE_NAME lowers into the
+            // full-body-walk compiled trace, whose `GUARD_NO_EXCEPTION` reads
+            // the backend `_store_exception` cells; writing only
+            // `BH_LAST_EXC_VALUE` would let a raising `__setitem__` on a
+            // mapping-locals namespace be silently swallowed.
+            publish_residual_call_exception(err.to_exc_object() as i64);
             0
         }
     }
@@ -4547,8 +4556,12 @@ pub extern "C" fn bh_store_global_fn(frame_ptr: i64, w_name: i64, value: i64) ->
     match frame.store_global_value(name, 0, value as pyre_object::PyObjectRef) {
         Ok(()) => 1,
         Err(err) => {
-            let exc_obj = err.to_exc_object();
-            majit_metainterp::blackhole::BH_LAST_EXC_VALUE.with(|c| c.set(exc_obj as i64));
+            // Publish into both exception cells, matching the STORE_NAME arm.
+            // `store_global_value` is infallible today (this arm is dead), but
+            // STORE_GLOBAL lowers into the compiled trace with a following
+            // `GUARD_NO_EXCEPTION` that reads the backend cells, so should a
+            // fallible path appear the raise must not be swallowed.
+            publish_residual_call_exception(err.to_exc_object() as i64);
             0
         }
     }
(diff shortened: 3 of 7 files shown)


youknowone:
Added a follow-up commit: **jit: publish LOAD_NAME/STORE_NAME/STORE_GLOBAL residual raises into the backend cells** (`654426b05a8`).

An adversarial investigation of the three remaining raw-`BH_LAST_EXC_VALUE` sibling residuals found the same exception-swallow class as the LOAD_ATTR fix in this PR:

- **`bh_load_name_fn`** and **`bh_store_name_fn`** — both lower into the full-body-walk compiled trace (a handler-bearing body skips the LOAD_NAME cell-fold; STORE_NAME is a can-raise void call), where the following `GUARD_NO_EXCEPTION` reads the backend `_store_exception` cells. Their `Err` arms wrote only `BH_LAST_EXC_VALUE`, so a raise was silently swallowed. Isolated repros: a LOAD_NAME of a removed global in a compiled `try/except` loop swallowed one `NameError` (`caught2 399999` → `400000`); a STORE_NAME to a mapping-locals namespace whose `__setitem__` raises ran to completion instead of breaking (`300000` → `150000`). Both now match the interpreter.
- **`bh_store_global_fn`** — same compiled path, but `store_global_value` is infallible today (dead `Err` arm); fixed defensively for consistency.
- **`bh_getattr_fn`** — left as-is; it is dead-wired (no lowering arm emits it, LOAD_ATTR routes through `bh_load_attr_fn`), so its raw write is unreachable.

Verification: check.py 159/159 on both dynasm and cranelift; the two swallow repros are bit-exact against the interpreter on both backends.

Separately, #389 items (b) comprehension realloc-append and (c) `with` WITH_EXCEPT_START were re-confirmed as multi-session work (blocked on a realloc-aware journaled append + task#50, and #373/#395 respectively) — no tractable slice this cycle.

— *commented by Claude*

Issue #215, opened 2026-06-19
fib_recursive major  performance blockers


youknowone:
- [x] Add min_heap_size to GcConfig and floor the major-collection threshold (incminimark-style) instead of gating on bare last_major_bytes × 1.82, eliminating the major-cycle thrash. — **done**: `min_heap_size` field + incminimark-style `set_major_threshold_from` floor (`collector.rs`).
- [ ] Virtualize the callee frame and unbox the result int across the CALL_ASSEMBLER boundary to remove the per-call heap-frame/box allocation. — **partially done**: #320 (merged) landed the sound multiframe bridge — all fib sizes correct — but the per-call heap-frame/box allocation is not yet eliminated. The deep-unroll lever that would remove it is clamped to `FBW_MAX_MULTIFRAME_DEPTH = 1`, leaving a ~4.3x gap vs pypy (fib32 0.129s vs 0.03s). Remaining work tracked in #343 (depth≥2 middle-frame virtual PyFrame rematerialization / walker-arch rework) — see also progress comment below.
- [x] Port estimate_best_nursery_size (L2-cache-based nursery sizing) so the default nursery matches PyPy per machine. — **done**: `estimate_best_nursery_size()` (`collector.rs:141`) drives the default nursery when `PYPY_GC_NURSERY` is unset.
- [x] Replace oldgen.contains's linear scan (oldgen.rs:144) with an O(1) membership lookup to cut the per-root seeding cost. — **done**: `contains` is now O(1) via the `payloads` index (`oldgen.rs:158`).
- [x] Drop the per-object Vec<GcRef> allocation in mark_object (collector.rs:1267) to reduce residual allocation during marking. — **done**: `mark_object` reuses the `incr_state.mark_scratch` scratch buffer (`collector.rs:1722`); no per-object Vec.

youknowone:
## item 2 (virtualize callee frame + unbox result across CALL_ASSEMBLER) — progress

Landed via #320 (merged 2026-07-03). check.py 169/169 × dynasm+cranelift green.

### ✅ Done
- **L1** portal-red color collision in the compiled multiframe-bridge guard resume — EC red (color 2) and a real frame slot (delivered call result) shared a color, so blackhole rebuilt `ADD(EC, fib(n-2))` → SEGV. Fixed with encoder scratch-gate + decoder ec-seed gate + trait-side gate.
[...]

Issue #57, opened 2026-05-16
`FOR_ITER` user-defined iterator: `__next__` inlining blocked by frozen root-frame concrete and threshold alignment


lifthrasiir:
tl;dr: further speed-up of `synth/iteration_protocol.py` benchmark requires a deep architectural rework.

### Benchmark

```python
# pyre/bench/synth/iteration_protocol.py
N = 40000

class CountDown:
    def __init__(self, n):
        self.n = n
    def __iter__(self):
        return self
    def __next__(self):
        if self.n <= 0:
            raise StopIteration
        self.n = self.n - 1
        return self.n

def main():
    i = 0
    acc = 0
    while i < N:
        for x in CountDown(5):
            acc = acc + x
        i = i + 1
    print(acc)

main()
```

| runtime | time | ratio |
|---------|------|-------|
| CPython | 0.04s | 1x |
| **PyPy** | **0.01s** | **0.25x** |
| pyre-dynasm | 0.31s | 7.8x |

### What PyPy does

PyPy's FOR_ITER calls `space.next(w_iterator)` → `__next__`. The tracing JIT traces *into* `__next__`, recording `getfield(self,
n)`, `int_le`, `guard_false`, `int_sub`, `setfield(self, n)`, `return`. StopIteration becomes a `guard_no_exception`. The entire
inner loop compiles down to a handful of integer operations and field reads/writes.

### What pyre does today

Infrastructure is in place (`for_iter` override on `OpcodeStepExecutor`, `PendingInlineFrame` with `for_iter_exhaust_target`,
StopIteration abort handler in metainterp). However, the inlined `__next__` always captures the **exhausted** path (n≤0 →
StopIteration guard) instead of the hot continues path. The unrolled second FOR_ITER falls back to the residual
`jit_user_iter_next_or_null` call. Net result: no speedup from inlining.

### Blocker 1: threshold always fires on exhausted iterator

The inner for-loop's `JumpBackward` calls `close_loop(FOR_ITER_PC)`, which ticks the jit counter under the `green_key(pycode,
FOR_ITER_PC)`. CountDown(5) produces 5 inner `JumpBackward` ticks per outer iteration. `JIT_THRESHOLD = 200`, which is a multiple
of 5, so the counter always reaches threshold on the *last* inner iteration — at which point `self.n = 0` and the concrete iterator
 is exhausted.

```
outer iteration 40 (i=39):
  inner iter 1: __next__→4, JumpBackward → tick 196
[...]

PR #448, opened 2026-07-09
jit: admit direct heap stores in abort-free FOR_ITER bodies (#57)

youknowone:
## What

Widen the JIT `FOR_ITER` loop-body allow-list (lever A / #57) so that a
**straight-line** `FOR_ITER` frame whose body writes the heap is JIT-compiled
instead of running the whole frame in the interpreter. Admits the direct
heap-store opcodes `STORE_SUBSCR` / `STORE_ATTR` / `STORE_NAME` / `STORE_GLOBAL`
and the `LOAD_NAME` that reads module globals — but **only** in an *abort-free*
body (no call, branch, or nested loop).

`unsupported_jit_shape` → `for_iter_bodies_all_jit_safe` previously rejected any
`FOR_ITER` frame whose loop body contained a heap store, sending the entire
frame to the interpreter. That frame-level rejection — not call-inlining
coverage — is the dominant JIT perf gap: `while` loops hit ~10.7×, function-local
`for` loops ~14.75×, but any `for` loop that writes the heap (`d[i]=`, `o.x=`,
module global) got **0× (never JIT)**.

## Why abort-free only

This gate is a conservative adaptation, not an upstream mechanism. The FBW
single-pass walker recovers from a mid-body walk abort by **refusing to
re-deliver** the in-flight `FOR_ITER` item (Option-C deliver/refuse), which drops
the rest of that iteration. The register-machine metainterp instead resumes
exactly: it reads the precise `(jitcode, pc)` from the guard's resume data,
`setposition`s a blackhole interpreter there and runs the body to its end
(`blackhole_from_resumedata`, `resume.py:1312`), so it never drops or doubles.

An **abort-free** body cannot reach the deliver/refuse path at all: with no call,
branch, or nested loop there is no abort after the store, so the store commits
exactly once and its result coincides with the exact-resume result. That makes
abort-free stores the observable-deviation-free subset.

Admitting store-bearing bodies that *can* abort (calls, branches, method loads)
was explored and **reverted**: when a body effect commits and a later op aborts,
the refuse-drop skips the rest of that iteration — a partial-iteration wrong
answer (`for x: d[k]=v; l.append(x)` → `l` short by the number of aborts), not a
clean iteration drop. Converging those bodies requires porting the exact-resume
path (#215 P2 / #73 / task#50); until then they stay interpreted and correct.

## Validation

- `check.py` **dynasm 159/159 + cranelift 159/159** (2/2 backend runs).
- `foriter57` `run_matrix` (JIT vs interp oracle) **MATRIX OK** on both backends.
- `for_iter` gate unit tests **9/9**: straight-line `STORE_SUBSCR`/`STORE_ATTR`
  bodies admitted; append / store+branch / store+call bodies rejected.
- Partial-iteration-drop repros (`d[k]=v; l.append`, `s.add; l.append`,
  `h(x); l.append`) now match the interpreter exactly under JIT.

## Perf (all previously 0× / never-JIT)

| shape | speedup |
|---|---|
| module-global `g = g + i` | ~4.75× |
| list `buf[i] = i` (subscript store) | ~2.6× |
| dict `d[k] = i` | ~1.4× |

## Summary by CodeRabbit

* **New Features**
  * Expanded JIT support for more `for`-loop bodies that are simple and branch-free.
  * Direct assignments inside loops, including item updates and attribute writes, are now more likely to run with JIT acceleration.
  * Module-global reads used in straight-line loop bodies are also supported.

* **Bug Fixes**
  * Loops with branches or function calls remain correctly excluded from JIT safety, preserving expected behavior.

### DIFF
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -3613,13 +3613,58 @@ fn for_iter_body_op_is_jit_safe(instr: pyre_interpreter::Instruction) -> bool {
     )
 }
 
-/// True iff every `FOR_ITER` loop body in `code` contains only
-/// `for_iter_body_op_is_jit_safe` opcodes. A nested `FOR_ITER` appears as a body
+/// True when the `FOR_ITER` body in `[body_start, exit)` is straight-line with
+/// respect to walk aborts: it contains no call, no nested-loop iterator setup,
+/// and no conditional or forward branch. Such a body has no reachable trigger
+/// for the mid-loop walk abort that hands control to the Option-C
+/// deliver/refuse path, so any eager heap store performed inside it is always
+/// committed exactly once and can never be dropped or doubled (#57).
+/// `JumpBackward` is the loop's own back edge and does not introduce an abort.
+fn for_iter_body_is_abort_free(
+    instructions: &[pyre_interpreter::bytecode::CodeUnit],
+    body_start: usize,
+    exit: usize,
+) -> bool {
+    use pyre_interpreter::Instruction as I;
+    let mut body_state = pyre_interpreter::OpArgState::default();
+    let mut body_pc = body_start;
+    while body_pc < exit && body_pc < instructions.len() {
+        let (body_instr, _) = body_state.get(instructions[body_pc]);
+        if matches!(
+            body_instr,
+            I::Call { .. }
+                | I::CallKw { .. }
+                | I::ForIter { .. }
+                | I::GetIter
+                | I::PopJumpIfFalse { .. }
+                | I::PopJumpIfTrue { .. }
+                | I::PopJumpIfNone { .. }
+                | I::PopJumpIfNotNone { .. }
+                | I::JumpForward { .. }
+        ) {
+            return false;
+        }
+        body_pc += 1;
+    }
+    true
+}
+
+/// True iff every `FOR_ITER` loop body in `code` is admissible. The BASE rule is
+/// `for_iter_body_op_is_jit_safe`. A nested `FOR_ITER` appears as a body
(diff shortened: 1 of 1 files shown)


youknowone:
**P1 confirmed — reverted.** This is a real correctness regression, and the reproductions show it is not specific to the store opcodes:

| body | JIT | interp / CPython |
|---|---|---|
| `cnt[0]=cnt.get()+1; seen.append(x)` | `(500, 498)` | `(500, 500)` |
| `d[k]=v; seen.append(x)` | `(8, 498)` | `(8, 500)` |
| `s.add(x); seen.append(x)` | `(500, 498)` | `(500, 500)` |
| `h(x)` (mutates a global)`; seen.append(x)` | `(500, 498)` | `(500, 500)` |

The FBW refuse-drop drops the *rest* of an iteration once any body effect has committed and a later op aborts — a partial-iteration wrong answer (`seen` short by the number of aborts), not the clean whole-iteration drop the gate comment claimed. The register-machine metainterp never does this: it resumes exactly at the guard's `(jitcode, pc)` (`blackhole_from_resumedata`, `resume.py:1312`) and runs the body to its end. So the deviation is the refuse-drop, not the allow-list widening.

Remediation: reverted the `LOAD_ATTR` admission and the unconditional-store admission, keeping only the abort-free store slice — an abort-free body (no call, branch, or nested loop) cannot reach the deliver/refuse path at all, so its store commits exactly once and coincides with the exact-resume result. The gate doc comment is rewritten to state this and to record that converging the callful/branchy/method-load bodies requires porting the exact-resume path (#215 P2 / #73 / task#50). check.py dynasm 159/159 + cranelift 159/159; the four repros above now match the interpreter under JIT.

— *commented by Claude*

PR #473, opened 2026-07-10
Builtin method and slot dunder arity guards: TypeError parity, panic fixes

youknowone:
Two slices continuing the argument-validation parity work (follow-up to #451).

## list/slot arity messages (`5c1a0cbaa82`)

- New helpers in `type_methods.rs`: `arity_exact_unpack` (PyArg_UnpackTuple min==max form, e.g. `insert expected 2 arguments, got 1`), `arity_slot` (check_num_args slot-wrapper form `expected 1 argument, got 0`, no method name), `arity_no_args` (METH_NOARGS form).
- `list.append`/`list.extend` messages gain the `list.` type prefix; `list.insert` switches to the UnpackTuple form; `list.pop` checks arity before the empty-list check (`[].pop(1, 2)` now raises TypeError, not IndexError).
- Every `init_list_type` slot closure indexing `args[1]` is guarded — `[].__add__()` and `[].__mul__()` previously **panicked** with `index out of bounds`.
- The shared `cmp_dunder!`/`complex_fail_cmp!` macros gain a single `arity_slot` guard covering the comparison dunders of int/float/complex/str/list/tuple/bytes/bytearray (all previously panicked, e.g. `(1).__eq__()`).

## slot dunder arity guards across types (`3d177dcadf8`)

- New `arity_pow` helper (`expected 1 or 2 arguments, got M`).
- The six shared binop macros (`int/float/complex_binop_fwd/rev`) gain one guard each, covering every generated binary arithmetic and reflected dunder — all previously **panicked** (e.g. `(1).__add__()`, `(1.0).__radd__()`, `(1j).__sub__()`). The five macro-generated pow/rpow dunders become standalone fns with the ternary-pow form.
- tuple/str/dict/mappingproxy/bytearray closures: panics (`().__getitem__()`, `bytearray().__setitem__(1)`), silent successes (`().__contains__()` returned False, `{}.__or__()` returned self), name-only messages (`TypeError: __add__`), and ad-hoc `X requires N args` messages all replaced with the CPython-verified templates.

## Verification

- 49-case (#19) + 82-case (#20) probes byte-identical to CPython 3.14.2 on both backends (one pre-existing mappingproxy value-equality divergence excluded and tracked separately).
- `pyre/check.py`: dynasm 160/160, cranelift 160/160, on both commits.
- `cargo fmt --check` clean.

Known follow-ups tracked locally: missing bytes dunder exposure (`b''.__add__` is AttributeError), `object.__eq__/__ne__` inherited-slot arity, mappingproxy value equality.

### DIFF
--- a/pyre/pyre-interpreter/src/objspace/descroperation.rs
+++ b/pyre/pyre-interpreter/src/objspace/descroperation.rs
@@ -1070,6 +1070,45 @@ pub(crate) unsafe fn str_repeat(s: PyObjectRef, n: PyObjectRef) -> PyResult {
     Ok(w_str_from_wtf8(buf))
 }
 
+pub(crate) unsafe fn bytes_concat(a: PyObjectRef, b: PyObjectRef) -> PyResult {
+    let Some(b_src) = crate::typedef::buffer_as_bytes_like(b)? else {
+        return Err(PyError::type_error(format!(
+            "can't concat {} to {}",
+            crate::baseobjspace::object_functionstr_type_name(b),
+            crate::baseobjspace::object_functionstr_type_name(a)
+        )));
+    };
+    let a_data = pyre_object::bytesobject::bytes_like_data(a);
+    let b_data = pyre_object::bytesobject::bytes_like_data(b_src);
+    let mut result = a_data.to_vec();
+    result.extend_from_slice(b_data);
+    Ok(if pyre_object::bytesobject::is_bytes(a) {
+        pyre_object::bytesobject::w_bytes_from_bytes(&result)
+    } else {
+        pyre_object::bytearrayobject::w_bytearray_from_bytes(&result)
+    })
+}
+
+pub(crate) unsafe fn bytes_repeat(s: PyObjectRef, n: PyObjectRef) -> PyResult {
+    let data = pyre_object::bytesobject::bytes_like_data(s);
+    let count = repeat_count(n, "repeated bytes are too long")?;
+    let cap = data
+        .len()
+        .checked_mul(count)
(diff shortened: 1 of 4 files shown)


youknowone:
### Disposition of the Codex parity review (section 2)

Both "section 2" findings judge the diff against PyPy's descriptors, but this repo tracks **CPython 3.14** behavior for these dunders, and CPython diverges from PyPy at exactly these two points. Verified live on both backends against `python3.14` (3.14.2):

**1. `bytes.__add__` for a non-buffer RHS** — Codex expected `NotImplemented`; CPython raises:

```
>>> b'a'.__add__('x')
TypeError: can't concat str to bytes
>>> b'a'.__add__(1)
TypeError: can't concat int to bytes
```

pyre (both backends) matches CPython. The operator path still reflects correctly: `b'a' + U()` with `U.__radd__` → `42`.

**2. `bytes.__rmod__` with a `bytearray` left operand** — Codex expected it to accept `bytearray`; CPython declines:

```
>>> bytes.__rmod__(b'%s', bytearray(b'x'))
NotImplemented
>>> b'%s'.__rmod__(bytearray(b'x'))
NotImplemented
```

pyre returns `NotImplemented` here too — matching CPython.

Since fixing either to the PyPy shape would regress away from CPython 3.14, both are **won't-fix (CPython-faithful)**.

Section 1 (`codewriter` block-head rewind) is the already-merged #478 and is acknowledged as a valid bytecode-model adaptation in section 4. Section 3 items are pre-existing and tracked as follow-ups.

— *commented by Claude*

Issue #339, opened 2026-07-03
codewriter: numeric boxing NewWithVtable panics at assemble when fuse_boxing_alloc fails to capture ob_type


youknowone:
The two-phase rtyper prepass (`slow_generated_jitcodes_keep_opcode_dispatch_shape`, the census harness) terminates with a fail-loud panic when a numeric boxing struct that `fuse_boxing_alloc` rewrote to `NewWithVtable` reaches the assembler with an unresolved `ob_type` pointer.

### Symptom
```
thread '...' panicked at majit/majit-translate/src/codewriter/assembler.rs:1513:
new_with_vtable: boxing owner "W_IntObject" has no resolved type pointer
(fuse_boxing_alloc could not capture the ob_type address); refusing to emit a
null-ob_type allocation
```
The owner is order-sensitive — observed as both `W_IntObject` and `W_FloatObject` across runs (same bug class). Only the four numeric boxing structs (`W_Float`/`W_Int`/`W_Complex`/`W_Long`) get a `NewWithVtable` lowering; the guard at `assembler.rs:1504-1519` (`OpKind::NewWithVtable`) refuses to emit a null-`ob_type` allocation when `vtable == 0`, which is correct (silently leaving `ob_type` null would be worse) — the real gap is upstream: `fuse_boxing_alloc` did not capture the `ob_type` address for this owner.

### Scope note
This is a **Path-2 / census-pipeline** codegen gap (the release two-phase prepass), not proven reachable in the production FBW tracer. It is the standing terminating panic that the dict-census A/B runs harvest their log *before* — every `#140` slice worked around it by reading the histogram emitted prior to this panic. Resolving it would let the census prepass complete instead of aborting.

### Investigation starting points
- `fuse_boxing_alloc` (the malloc→`NewWithVtable` fusion + `ob_type` capture) — why the vtable/ob_type address is not captured for these owners in some graphs.
- Related unfused-path guard: `#144`'s "`lltype::malloc_typed` survived fuse_boxing_alloc unfused" fail-closed (also visible in the same log for `function_new_impl`, `alloc_dict_object`).

— *filed by Claude*

youknowone:
## Already resolved by #403 — recommend closing

Verified on branch `pc-map` (HEAD `645fcf91669`): this panic is no longer reachable. `fuse_boxing_alloc` (`majit/majit-translate/src/model.rs`) now **skips fusion when the vtable cannot be resolved**:

```rust
[...]

PR #479, opened 2026-07-10
cranelift: demote invariant deopt references to GC-root slots (#339)

youknowone:
## Bug

`#339` family, cranelift only: a callee whose returns are wrapped in `try/finally`, called twice per iteration in a hot loop, produced **empty stdout with exit 0** (dynasm and CPython print the correct value).

```python
def f(x):
    try:
        if x % 7 == 0:
            return x
        return x + 1
    finally:
        pass
def g(x):
    return f(x) + f(x + 1)
s = 0
for i in range(300000):
    s = (s + g(i)) % 1000003
print(s)
```

## Root cause

The cranelift backend kept loop-invariant, deopt-only GC references as SSA loop phis. After a collecting call inside the loop, a guard exit could rebuild the frame from a stale pre-move reference; the failure manifested in minor GC freeing corrupted `PyFrame` contents. Disabling the JIT or enlarging the nursery masked it.

## Fix

`majit/majit-backend-cranelift/src/compiler.rs`: eligible loop LABEL references are demoted to GC-root frame slots; guard/deopt and loader paths reload those slots instead of consuming the phi value. The transform is restricted to invariant, non-input GC-root references. A unit test covers the classification.

## Verification

- Repro: CPython = dynasm = cranelift = `244285` (was: cranelift empty stdout).
- Adversarial variants byte-exact on both backends: finally-return, nested finally, raised/caught finally, non-inlined callee, single call, loop-carried accumulator.
- `check.py`: cranelift 161/161; dynasm full run had only the documented `nested_loop` perf flake, which meets its threshold when rerun isolated.
- Cranelift crate tests: 163 passed, 2 ignored; dynasm `cargo check` clean.

Also includes a documentation-only commit recording the Phase 4c census resolution in `flatten-path4-closure-roadmap.md` (recipe and framestack paths are sequential decode→compile, not competitors; the leftover `color==semantic` gate fires zero times corpus-wide).

Implemented by Codex (gpt-5.6-terra), orchestrated and verified by Claude.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of loop-carried values during deoptimization, helping preserve correct reference values when execution exits optimized loops.
  * Reduced redundant value materialization and improved frame reconstruction reliability.

* **Documentation**
  * Added a detailed roadmap covering remaining work for graph flattening, path handling, program-counter mapping, and related cleanup.
  * Documented current blockers, completed milestones, planned phases, dependencies, and known follow-up issues.

* **Tests**
  * Added coverage verifying correct treatment of deoptimization-only loop values.

### DIFF
--- a/flatten-path4-closure-roadmap.md
+++ b/flatten-path4-closure-roadmap.md
@@ -1,0 +1,372 @@
+# Closure roadmap: flatten-graph residual + Path 4 epic (#238) + pc_map (#73), PyPy-faithful
+
+Status snapshot (2026-07-02, branch `flatten-graph`, HEAD `3c5191088ea` atop issue62 P2
+framestack commits). Every phase below is sliced so each slice lands gate-green
+(cargo test + check.py ×2 backends + 599-prog corpus sweep + adversarial refute WF for
+resume-zone slices). One green slice per session; never push.
+
+## Ground truth (censused this session)
+
+Runtime flat-map (`stack_slot_color_map`) readers remaining:
+
+| reader | site | gate | class |
+|---|---|---|---|
+| hazard flat fallback | jitcode_dispatch.rs:9034 (`kept_stack_has_boxed_int_hazard`) | `!mirror_covers_kept` | bounded, DEAD on CI, 22 safe-smallint fires on 9 adv progs |
+| overlay legacy leg | jitcode_dispatch.rs:9644 (`stack_sync`) | mirror-primary/legacy-fallback per slot | same trigger population |
+| encoder snapshot | trace_opcode.rs:1599, 1666 (`emit_live_refs`, via `semantic_ref_slot_for_reg_color`) | none | color→slot inversion, PyPy never inverts |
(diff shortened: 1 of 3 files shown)


youknowone:
Review findings drained in follow-up PR #481:

- Codex parity items 1–2 (last-LABEL-only demotion; preamble uses counted as body reads) were confirmed as residual correctness holes via classifier repros and are fixed there.
- Parity item 3 is intentionally kept: inputarg refs are already rooted/spilled/reloaded across collections; int/float demotion is excluded because carried scratch slots can be overwritten by failarg spills. Item 4 (per-box Lifetime representation) skipped as disproportionate for this backend.
- CodeRabbit items are all applied: thread-local `DEMOTED_FAILARG_SLOTS` replaced by explicit parameter threading, shared `is_demoted_failarg` predicate, and in-tree end-to-end regression tests for the miscompile class.

— *commented by Claude*

PR #481, opened 2026-07-10
cranelift: demote invariant refs at every label (review follow-up to #479)

youknowone:
Drains the review findings on #479 (Codex parity review + CodeRabbit).

## Parity fixes (potential residual correctness holes, each proven by a classifier repro that failed before the change)

1. **Every LABEL, not just the last** (`regalloc.py` `consider_label` runs per LABEL): demotion now analyzes every LABEL; multi-LABEL traces previously kept stale-ref-exposed SSA phis across earlier labels.
2. **Positional last-use criterion** (`real_usages[-1] <= position`): pre-LABEL (preamble-only) uses no longer block demotion; that shape previously stayed an SSA phi and remained exposed to the original stale-ref bug.

## Skipped with rationale

3. Extending demotion to ints/floats/inputargs: inputarg refs already receive ref-root slots, LABEL synchronization, and collection spill/reload (covered by existing collection tests); int/float demotion is excluded because carried scratch slots can be overwritten by failarg spills.
4. Per-box `Lifetime` representation parity: no direct Cranelift equivalent of the forwarded-box lifetime; a rewrite would be disproportionate.

## Maintainability (CodeRabbit)

5. Removed the `DEMOTED_FAILARG_SLOTS` thread-local + RAII guard; the demotion map is now threaded explicitly through guard resolution and root handling.
6. Extracted a shared `is_demoted_failarg` predicate for the spill/reload paths.
7. Added in-tree end-to-end tests: nursery-moving allocation, loop-invariant deopt-only ref failargs, guard exit, preamble use, backedge, and an earlier-LABEL case.

## Verification

- #339 repro: `244285` on both backends (strict JIT).
- `check.py`: dynasm 10/10 bench + 151/151 synth; cranelift 10/10 + 151/151.
- Cranelift tests: 132 unit + 35 integration + 1 tail-call passed (2 ignored); dynasm `cargo test` passed.

Implemented by Codex (gpt-5.6-terra), orchestrated and verified by Claude.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -5185,6 +5185,34 @@ fn resolve_failarg_opref(
     resolve_opref(builder, constants, opref)
 }
 
+/// Resolve a local JUMP argument that is still a block parameter at its
+/// target.  `regalloc.py` keeps this distinction at `consider_label` /
+/// `force_spill_var`, then `_pop_all_regs_from_frame` restores every live
+/// reference after a collecting call before the next control-flow transfer.
+///
+/// The per-LABEL classifier can demote this raw OpRef at an earlier LABEL,
+/// while a later LABEL keeps it because that later header is only reached by a
+/// terminator.  In that case the global demoted-slot map suppresses the normal
+/// SSA reload, but the local JUMP must still forward the collector-updated root
+/// value into the kept target parameter.
+fn resolve_local_jump_arg(
+    builder: &mut FunctionBuilder,
+    constants: &indexmap::IndexMap<u32, i64>,
+    jf_ptr: CValue,
+    demoted_failarg_slots: &IndexMap<u32, i32>,
+    opref: OpRef,
+) -> CValue {
+    if !opref.is_none()
+        && !opref.is_constant()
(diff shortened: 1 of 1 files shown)


youknowone:
The P1 finding is confirmed and fixed in `92f01c53601` ("cranelift: reload demoted refs for kept jump args").

Reachability was confirmed with a moving-GC repro: the classifier excludes LABEL/JUMP from real uses, so LABEL A demoted a raw ref while a later LABEL B kept that same ref as a block parameter; the global demotion map then suppressed the post-GC SSA reload and B's local JUMP forwarded the pre-move pointer. A temporary CLIF dump confirmed the demoted ref being resolved as a JUMP arg, and the repro returned the original (stale) pointer before the fix.

Fix: `resolve_local_jump_arg` (used by local JUMP lowering) reloads a globally demoted ref from its forwarded root slot when the target keeps that argument, citing the `consider_label` / `force_spill_var` / `_pop_all_regs_from_frame` analog. Added a bridge-to-later-LABEL moving-GC regression test (`demoted_ref_reloads_before_kept_local_jump_arg`).

Verification: regression test fails before / passes after; #339 oracle 244285 on both backends; check.py cranelift 161/161 (dynasm's only miss was the known nested_loop timing flake, 10/10 isolated); cranelift crate tests 133 + 35 integration passed.

— *commented by Claude*

Issue #126, opened 2026-05-30
fib_recursive: recursive CALL_ASSEMBLER lands (now faster than CPython); residual ~4-5x-vs-pypy floor needs flat/virtualized jitframe


youknowone:
> **STATUS UPDATE (2026-07-22).** Original symptoms **resolved** — self-recursive CALL_ASSEMBLER folding shipped (default-on), both backends faster than CPython, check.py green. **Remaining:** this issue's own lever — a flat/virtualized jitframe that eliminates the per-call heap PyFrame allocation — is not landed; the residual ~4–5x-vs-PyPy floor persists. The sibling depth-unroll lever is tracked in #343.

---

## Status (2026-07-06, HEAD `04fc5ebc0b`)

The original symptoms are **resolved**: the self-recursive Python call is now detected on the bytecode-JIT side and lowered to the recursive `CALL_ASSEMBLER` path, fib compiles and hits the compiled code on essentially every call, and **both backends are now faster than CPython**. The residual gap is a ~4-5x-vs-pypy floor caused by the orthodox per-call `PyFrame` construction. Two independent upstream-orthodox levers remove that per-call allocation: **bounded recursion unroll** (#343, `_opimpl_recursive_call` / `max_unroll_recursion`, which folds recursive callees away before the `CALL_ASSEMBLER`) and a **flat / virtualized jitframe** representation (this issue, which changes the frame layout so the heap `PyFrame` is never allocated). They are siblings — either can move the floor; neither blocks the other.

## Bench

`pyre/bench/fib_recursive.py` — naive recursive Fibonacci, fib(0..34):

```python
def fib(n):
    if n < 2:
        return n
    return fib(n - 1) + fib(n - 2)

for i in range(35):
    print(fib(i))
```

```
cargo build --release --no-default-features --features cranelift --bin pyre-cranelift
cargo build --release --no-default-features --features dynasm    --bin pyre-dynasm
./target/release/pyre-cranelift pyre/bench/fib_recursive.py
./target/release/pyre-dynasm    pyre/bench/fib_recursive.py
```

## Current timing (aarch64 macOS, release, best of 3)

| Runtime | Time | vs CPython | vs pypy |
|---|---|---|---|
| CPython 3.14 | 0.63 s | 1.00x | — |
| PyPy3 | 0.22 s | 0.35x | 1.00x |
| pyre-dynasm | 0.96 s | **1.52x** | 4.4x |
[...]

PR #485, opened 2026-07-10
jit: #73 entry-carry — codewrite-time trace-entry coordinates for the walk entry

youknowone:
## #73 entry-carry: codewrite-time trace-entry coordinates for the walk entry

Next phase of #73/#366 after #455: retire the runtime `resume_jitcode_pc_for` py_pc→jitcode translation from the **walk-entry family** of its callers.

### Why this shape

Scoping established that the plain-portal `start_pc` is a Python bytecode pc at every hop of its producer chain (`close_loop` target → green key → merge point → `trace_bytecode`) and must stay py_pc for green-key/re-entry matching — so the coordinate cannot be carried down from the driver. PyPy's trace-entry coordinate is instead born at codewriting: tracing always starts at the portal jitcode's `jit_merge_point`. The analog here is a codewrite-time sidecar keyed by trace-entry greens.

### Commits

1. **`8e57c0f4fbd` — green→entry sidecar (E1).** `PyJitCodeMetadata.merge_entry_by_green`: per-loop-header `(green py_pc → entry JitCode offset)`, built in the codewriter finalize block with `resume_jitcode_pc_for`'s exact precedence (carryfwd sidecar, then `derive_resume_marker`), restricted to `find_loop_header_pcs`. `run_perfn_walk` selects it for plain-portal loop-header entries. Gates: `PYRE_M73_ENTRY_AUDIT` (default off) compares sidecar vs derived; `PYRE_M73_ENTRY_CARRY` (default **on**, `=0` opt-out).
2. **`caff209d7ec` — function-entry green + carried bridge offset (E2).** Sidecar keys extended with function entry (py_pc 0), covering recursive function-entry portals (fib). A bridge resume that carries a jitcode coordinate (`sym.bridge_walk_entry_pc`) now uses it directly instead of deriving an unused entry marker — an unencodable guard py_pc no longer declines such a walk. New `M73EntryAudit::BridgeNoCarry` census: **zero** across the corpus (every bridge resume carries a coordinate post-#455).
3. **`0ffaa684e3f` — P2 carrier recipes ride the decoded `jitcode_pc` (E3).** `ReconstructRecipe.jitcode_pc` / `BridgeInlineCarrier.root_jitcode_pc` carry the resume frame's guard-carried offset; the three P2 carrier entry sites prefer it (resolved via `resolve_bridge_walk_entry_at`) over the derivation, through a shared `select_recipe_entry` helper with audit census. Derivation retained as fallback (the path is `PYRE_P2_DRAIN`-gated experimental).

### Verification

- `check.py` **161/161 on dynasm and cranelift** at every step, including carry default-ON and `PYRE_M73_ENTRY_CARRY=0` opt-out.
- Audit certificate: zero `M73EntryAudit::*` mismatches across the 161-program corpus (dynasm) and on direct stderr capture of fib_recursive / nbody / fannkuch / raise_catch_loop / the four kept-stack & nested-loop synth programs.
- `PYRE_P2_DRAIN=1 PYRE_P2_COMPILE=1` fib harness byte-identical before/after E3.
- `cargo test -p pyre-jit-trace` (265+11+1) and `-p majit-metainterp` (1380+3+3) green.
- Codex parity review (E1): sections 1–3 **None**; section 4 = the intended structural adaptations, documented in-code.

### Remaining `resume_jitcode_pc_for` callers (out of scope here)

- Bridge resumes without a carried coordinate (corpus-cold, census-instrumented) and the `PYRE_M73_ENTRY_CARRY=0` opt-out legs.
- The encode-side resume-marker family in `jitcode_dispatch.rs` (4 sites) + `trace_opcode.rs:1475` — the A2 jitcode-native-resume front.
- Diagnostics (`trace.rs` P2_DIAG / perfn dump).

## Summary by CodeRabbit

* **New Features**
  * Improved bridge and inline-frame resume handling across nested JIT frames.
  * Added accurate entry-point mapping for trace merges and portal transitions.
  * Added optional audit and carry controls for entry-coordinate verification.

* **Bug Fixes**
  * Corrected resume entry selection when carried JITCode offsets are available.
  * Improved handling of loop headers, merge points, and bridge walks.
  * Added mismatch detection and diagnostic metrics for inconsistent resume mappings.


youknowone:
### #73 encode-side (A2) caller scoping — ultracode pass

An 8-agent read-only workflow + adversarial verification classified the remaining `resume_jitcode_pc_for` callers after the entry family was carried. Latest commit deletes the two cold diagnostic callers; the rest are characterized below.

| Caller | Class | Disposition |
|---|---|---|
| jitcode_dispatch `carry_audit` assert | diagnostic, cold (`PYRE_PCMAP_CARRY_AUDIT`, 0/162) | **deleted this commit** |
| jitcode_dispatch `M73_FLIP` recon | diagnostic, cold (`PYRE_M73_FLIP_AUDIT`, 0/162) | **deleted this commit** |
| jitcode_dispatch nonbranch marker (`#366` arm) | encoder carries the same block-head `-live-` marker the decoder derives ("identical to that fallback by construction") | blocked on the jitcode-space resume-marker twin |
| jitcode_dispatch after-residual-call (`None` arm) | cheap-rekey to `ctx.live_after_jit_pc`, but its setter has a staleness gap (fires only under `emit_guard_not_forced`, not on the `GuardNoException` capture) | needs a codegen setter first |
| trace_opcode:1475 `.or_else` | hot loop-close hits a loop-header green (carriable), but the call also serves mid-body guards whose `live_pc` is not a sidecar key | fallback stays until the marker twin |
| pyjitcode:698 | decode/blackhole-resume time, no jitcode coord in hand; consumes a py coordinate with no jitcode-native producer | structural multi-week |

Adversarial verification caught one real trap: the design's first candidate (carry `pc_map_entry` keyed on `start_pc` into the loop-close snapshot) is coord-equality **UNSOUND** for bridge closes (nbody), where `bridge_walk_entry_pc` is a mid-op offset ≠ the loop-header marker — it would have shipped a wrong resume coordinate.

**Net:** the terminal deletion of `resume_jitcode_pc_for` is gated on a single structural keystone — authoring the block-head resume marker in JitCode space (the jitcode-keyed resume-marker twin) — which the four remaining callers all reduce to. That is the genuine next front; no further byte-identical caller retirement exists on the current branch.

— *commented by Claude*

youknowone:

- **§1 (regression, 1)** — *fixed this push* (`db32ac7f5f2`). The carried recipe entry was resolved against `jitcodes[recipe.jitcode_index]` while the walk drives `callee_pjc` selected by code identity; with multiple JitCode bodies per code object these can be different bodies. Added a `body_index` guard so the carried coordinate is used only when `jitcode_index == callee_pjc.jitcode.index()`, else the body-consistent `resume_jitcode_pc_for` derivation. 161/161 both backends; P2 fib harness unchanged.
- **§2 (other mismatches)** — None.
- **§3 (pre-existing, 1)** — the `carrier.recipes.len() == 1` depth-1-only continuation (deeper inline-frame stacks abort to a correct no-JIT re-interpret). Pre-existing P2 limitation, part of the separate multiframe-depth epic (#126); not adjacent to this patch. **Deferred** to that epic.
- **§4 (structural adaptations, 2)** — `ReconstructRecipe.jitcode_pc` dual-coordinate and `merge_entry_by_green` compile-time cache are the intended entry-carry design, documented in-code with upstream citations (`resume.py:1050`, `pyjitpl.py:1567`). **Won't-fix (documented).**

— *commented by Claude*

PR #503, opened 2026-07-12
interp(generic_alias): fix abort on non-tuple __typing_subst__/prepare_subst (GH-138497)

fregataa:
## Summary

`test_typing.GenericTests.test_return_non_tuple_while_unpacking` aborted the whole test run with a Rust panic:

```
pyre/pyre-object/src/object_array.rs:107: misaligned pointer dereference
```

Root cause: `_pypy_generic_alias::subs_parameters` called `w_tuple_len()` on the `__typing_prepare_subst__` result without checking it was a tuple. An evil `prepare` returning `None` (a tagged singleton) was dereferenced as a tuple → abort. Separately, the unpack path spliced the `__typing_subst__` result via `collect_iterable` (any iterable) instead of requiring a tuple, so it never produced CPython's TypeError.

## Fix

Mirror CPython's `genericaliasobject.c` `_Py_subs_parameters`:

- A non-tuple `item` is treated as a single argument (`nitems = 1`, `argitems = &item`) instead of having `len()` taken on it.
- When unpacking, `__typing_subst__` must return a tuple; otherwise raise
  `expected __typing_subst__ of <T> objects to return a tuple, not <T>`
  rather than iterating an arbitrary object.

The vendored PyPy reference `lib_pypy/_pypy_generic_alias.py` is kept in sync with the Rust port.

## Test

`GenericTests.test_return_non_tuple_while_unpacking` now passes:

```
test_return_non_tuple_while_unpacking ... ok
```

and the run no longer aborts, so the tests that previously got cut off after this point can run again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed generic type substitution handling when prepared arguments are provided as a single value instead of a tuple, including correct argument count validation.
  * Improved substitution error handling for unpacked type variables, raising a `TypeError` when the result is not a tuple.
  * Enhanced error output with clearer Python type names for easier debugging.

### DIFF
--- a/lib_pypy/_pypy_generic_alias.py
+++ b/lib_pypy/_pypy_generic_alias.py
@@ -214,15 +214,10 @@ def subs_parameters(self, args, params, items):
         prepare = getattr(param, '__typing_prepare_subst__', None)
         if prepare is not None:
             items = prepare(self, items)
-    # __typing_prepare_subst__ is not required to return a tuple; a non-tuple
-    # result is treated as a single argument (nitems == 1) rather than having
-    # len() taken on it.
-    is_tuple = isinstance(items, tuple)
-    nitems = len(items) if is_tuple else 1
+    nitems = len(items)
     if nparams != nitems:
         direction = 'many' if nitems > nparams else 'few'
         raise TypeError(f"Too {direction} arguments for {self}; actual {nitems}, expected {nparams}")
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Thank you! and welcome to Pyre project

PR #524, opened 2026-07-13
jit: #22 tagged-int Dir3 trace-boundary machinery + 4 gated deref guards (flag-false inert)

youknowone:
## Summary

Direction-3 tagged-int trace-boundary machinery plus tag-immediate guards at four object-deref boundaries. All changes are gated on `pyre_object::tagged_int::CAN_BE_TAGGED` (default `false`) and are dead code at the flag's default, so flag-false trace emission and interpreter behavior are byte-identical to before.

Groundwork toward enabling tagged small ints (`#22`): representing small `int` as an immediate `(v<<1)|1` value to eliminate int-boxing malloc. This branch closes the four flag-true SIGSEGV crashes that blocked the enablement flip. The flip itself (`CAN_BE_TAGGED=true`) is **not** in this branch.

## Commits

1. **Dir3 tagged-int trace-boundary machinery** — a hot-loop-carried tagged int enters the JIT in the flag-false heap representation (never carries a tagged pointer in a trace):
   - `untag_tagged_frame_locals` (eval.rs): convert a frame's tagged-immediate local slots to heap `W_IntObject` before a compiled loop reads them.
   - record-time seed (state.rs): convert tagged locals on the `snapshot_for_tracing` copy so the recorded body matches the runtime-converted frame.
   - `wrapint` class-record, `box_int_payload` heap-only, `walker_unbox_int_typed` two-arm split, bridge ec-color seed fix.

2. **Gated tag-immediate guards at 4 deref boundaries** — each fixes one flag-true crash class:
   - majit-gc `get_actual_typeid`: return `None` for a tagged immediate before the header read (mirrors `can_move`).
   - interpreter `builtin_str`: stringify a tagged int to decimal before the `ob_type` derefs (mirrors `py_str_wtf8`/`py_repr_obj`).
   - jit-trace `loop_inlines_abort_permanent_callee`: skip a tagged immediate in the globals-dict and cell scans.
   - jit assembler `untag_const_ref_value`: re-realize a tagged immediate as a heap `W_IntObject` (`w_int_new_unique`) before it enters the GC-root-walked jitcode `constants_r` pool, at each `ConstRef` pool-bake site.

## Verification

- **flag-false** (shipped default) `check.py --backend dynasm,cranelift`: cranelift 163/163, dynasm 162/163 — the one red is the `raise_catch` vs-pypy 1.5x timing gate, a boundary jitter flake (re-ran 3×: 1.5x/1.4x/1.0x), not a correctness regression. Slice is gated dead-code so the tree is byte-identical to base.
- **flag-true** (`CAN_BE_TAGGED=true`, local only) `check.py --backend dynasm,cranelift`: 163/163 both backends. All four formerly-crashing tests (fannkuch, nested_for_varying_trip, polymorphic_slot_retype, str_fstring) pass; fannkuch still JITs (`loops_compiled=5`).

## Follow-ups (documented, for the eventual flip, not in this branch)

- `builtin_str` tag-guard sits after the `encoding`/`errors` block that derefs `ob_type`; `str(tagged, encoding=)` would still deref tagged at flip-true (no test reaches it). Move the guard earlier at flip time.
- Why the majit optimizer folds a loop-carried const to `Operand::ConstRef(tagged)` at record time instead of a heap-box const — the assembler-boundary normalization covers it; the optimizer-side root is the single-root ideal for cleanup.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tagged integer values across interpretation and compiled execution.
  * Fixed integer string conversion to consistently produce decimal representations.
  * Prevented tagged integers from being misidentified as objects, function references, or closure cells.
  * Improved JIT tracing, bridge resumes, constant handling, and deoptimization reliability.
  * Reduced risks of invalid memory access and guard failures when processing small integers.

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -3543,6 +3543,12 @@ impl GcAllocator for MiniMarkGC {
         if gcref.is_null() {
             return None;
         }
+        // gc/base.py:380 `is_valid_gc_object`: a tagged immediate carries no
+        // header/classptr — reading offset 0 would deref the value bits.
+        // Mirrors `can_move`'s guard.
+        if self.is_tagged_immediate(gcref.0) {
+            return None;
+        }
         if self.is_managed_heap_object(gcref.0) {
             let header_addr = gcref.0.wrapping_sub(crate::header::GcHeader::SIZE);
             let header: crate::header::GcHeader = unsafe { *(header_addr as *const _) };
--- a/majit/majit-ir/src/resoperation.rs
+++ b/majit/majit-ir/src/resoperation.rs
@@ -188,6 +188,16 @@ impl OpRef {
         matches!(self, Self::None)
     }
 
+    /// Mirrors RPython `isinstance(value, AbstractInputArg)` — the operand is
+    /// one of the trace's typed input args (a loop/function entry value), not a
+    /// recorded op result, constant, or temp.
+    pub fn is_input_arg(self) -> bool {
+        matches!(
+            self,
+            Self::InputArgInt(_) | Self::InputArgFloat(_) | Self::InputArgRef(_)
+        )
+    }
+
     /// Extract the raw u32 payload. For `None` returns `u32::MAX` to
     /// preserve pre-Phase-3 round-trip semantics.
     ///
--- a/pyre/pyre-interpreter/src/builtins.rs
+++ b/pyre/pyre-interpreter/src/builtins.rs
@@ -4796,6 +4796,15 @@ pub(crate) fn builtin_str(args: &[PyObjectRef]) -> Result<PyObjectRef, crate::Py
         }
         return crate::typedef::bytes_method_decode(&decode_args);
     }
+    // A tagged `int` immediate stringifies to its decimal value; format it
+    // before `is_str` / `unwrap_cell` / `ob_type` touch it as a pointer.
+    // Mirrors `py_str_wtf8` / `py_repr_obj`. Gated on `CAN_BE_TAGGED`.
+    if pyre_object::tagged_int::CAN_BE_TAGGED && pyre_object::tagged_int::is_tagged_int(obj) {
+        return Ok(w_str_new(&format!(
+            "{}",
+            pyre_object::tagged_int::untag_int(obj)
+        )));
+    }
     unsafe {
         if is_str(obj) {
             // A `str` subclass keeps `ob_type` at STR_TYPE but carries the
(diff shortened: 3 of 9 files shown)


youknowone:
## Codex parity-review disposition (team-lead)

Ran the codex parity review against `upstream/main`. Note: the current
`upstream/main` (37cc318) lags `origin/main` (aa36172), so the diff base pulls in
`#523` (always-portal) and `#447` (gc-hook) on top of this slice's two commits
(`ba2dac7110d` Dir3 machinery, `d4f5e51826b` 4 gated guards). Findings adjudicated
per-section:

**§1 Parity regressions (3) — all on the tagged-int machinery, all conditioned "when tagging enabled"; the committed slice ships flag-FALSE (dead code, byte-identical), so none affect the default.**

- **§1-1** `box_int_payload` heap-only (`trace_opcode.rs`) → **reclassified §4 (intentional, won't-fix documented).** This is Direction-B by design: we deliberately rejected Direction-A (in-trace tagged pointer per `rtagged.py`) because it would force `GcConfig.taggedpointers` ON and make the JIT carry tagged pointers in traces — something the mainline int path never does — and it regressed hot int loops. Dir-B: the JIT never carries a tagged int; it untags once at the concrete boundary to a virtualizable heap box, trace stays byte-identical to flag-false, box virtualizes away.
- **§1-2** frame-local untag → distinct heap object → identity/`is`/hash (`eval.rs`, `state.rs`) → **false-positive, empirically refuted.** With tagging live, `is`/identity compares the immediate bits `(v<<1)|1`, identical for equal values — value-identity holds without an address compare. The heap box the untag creates is consumed only by the in-trace `GuardClass` deref; it is never exposed to a Python `is`.
- **§1-3** const-pool `untag_const_ref_value` (`assembler.rs`) → **false-positive, same refutation.** The const-pool box exists solely so GC-walked `constants_r` holds a real gcref for a `GuardClass` deref; Python `is` still compares the original tagged immediate.

Verified live on a flag-true build vs CPython 3.14, all matching: `a=1000;b=1000; a is b`→True; `256 is 256`→True; `big=10**18; a is copy`→True; loop-accumulator `is` literal→True; `[7,7,7]` elements identical; `d[42]` lookup correct.

**§2 Other mismatches (1)** — codewriter always-Portal (`codewriter.rs:5745`) → **out of scope.** This is `#523` (`aa36172bc1f`, already merged to `origin/main`), leaked in by the stale `upstream/main` base. Neither commit in this slice touches `codewriter.rs`.

**§3 Pre-existing (1)** — single-frame inline collapse (`jitcode_dispatch.rs:14729`) → **defer.** Pre-existing on `upstream/main`, tracked under the multiframe/walker-arch work.

**§4 Structural adaptations (4)** — free-threading GC (`gc_sync`, `shadow_stack`, `collector`, `stack_check`) → **won't-fix documented.** STW/per-thread adaptations, orthogonal to this slice.

**Net:** no §1/§2 fix lands on the tagged-int files — 1 intentional design (reclassified), 2 refuted false-positives, 1 out-of-scope. The committed slice stands as-is (flag stays OFF).

— *commented by Claude*

PR #526, opened 2026-07-13
interp: reject instantiation of abstract classes in object.__new__

fregataa:
## Summary

Pyre had no abstract-type mechanism at all: assigning a non-empty `__abstractmethods__` never marked the type abstract, and `object.__new__` never checked it. So `C()` on an ABC with unimplemented abstract methods silently produced an instance instead of raising `TypeError` — and even a plain class with a manually-set `__abstractmethods__` instantiated freely.

This ports PyPy's `flag_abstract` faithfully:

- **`pyre-object` `W_TypeObject`** gains a `flag_abstract: AtomicBool` with `w_type_is_abstract` / `w_type_set_abstract` accessors (`typeobject.py` `is_abstract` / `set_abstract`), initialized `false` in both constructors.
- **`pyre-interpreter` setattr/delattr** mirror `descr_set/del___abstractmethods__`: assigning `__abstractmethods__` sets the flag to the value's truthiness (empty `frozenset` → concrete, non-empty → abstract); deleting it clears the flag.
- **`object.__new__` (`object_descr_new`)** raises `TypeError` for an abstract class, using the CPython 3.12+ message form: `Can't instantiate abstract class C without an implementation for abstract method 'foo'`.

## Test results

- `test_abc`: failures **33 → 24** (the pure-Python `_py_abc` path is now fully fixed; no new failures).
- `test_enum`: **154 before and after** (no regression; verified against a `git stash` baseline rebuild).
- Verified inheritance semantics (abstract base / still-abstract subclass rejected, concrete subclass instantiable), `del __abstractmethods__` clearing the flag, and that normal / builtin-subclass instantiation is unaffected.

## Out of scope (tracked separately)

The remaining `test_abc` failures are two independent defects:
1. The C `_abc` path — `_abc_init` does not compute `__abstractmethods__` (missing `compute_abstract_methods`).
2. `classmethod` / `staticmethod` subclass type collapse — `abstractclassmethod` / `abstractstaticmethod` lose their `__isabstractmethod__`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Abstract classes can no longer be instantiated.
  * Abstract status now updates correctly when `__abstractmethods__` is set or deleted.
  * Instantiation now raises a clearer error naming the abstract class and listing the missing method implementations.

### DIFF
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -7284,8 +7284,20 @@ pub fn object_setattr(obj: PyObjectRef, name: &str, value: PyObjectRef) -> PyRes
             }
             let dict_ptr = w_type_get_dict_ptr(obj) as *mut crate::DictStorage;
             if !dict_ptr.is_null() {
+                // typeobject.py:1258-1261 descr_set___abstractmethods__ —
+                // evaluate truthiness BEFORE mutating the dict so a raising
+                // `__bool__` leaves the type unchanged (CPython
+                // type_set_abstractmethods).
+                let abstract_flag = if name == "__abstractmethods__" {
+                    Some(is_true(value)?)
+                } else {
+                    None
+                };
                 crate::dict_storage_store(&mut *dict_ptr, name, value);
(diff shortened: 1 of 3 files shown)


youknowone (approved):
👍

PR #530, opened 2026-07-13
cranelift: run the entry-prologue frame-depth check for loops, not only bridges

youknowone:
## Problem

A cranelift-only, deterministic GC crash `bogus frame field get: index=33 >= frame_lgt=33` (`jitframe_trace` assert in `majit-backend/src/jitframe.rs`) fires for N≥10000 in a dict-across-allocation loop. dynasm is correct at all N.

## Root cause

The cranelift entry-prologue frame-depth check (`_check_frame_depth` / `cranelift_realloc_frame`) was gated on `source_guard.is_some()` — **bridges only**. The comment assumed "main loops skip the check because `run_compiled_code_inner` allocates the frame at exactly the size the body needs."

That assumption is false: a loop reached by a closing-jump or a **bridge-dispatch tail-call** runs on the *caller's* frame. A bridge reallocates the JITFRAME to its own `max_output_slots + num_ref_roots` and then jumps back to a wider parent loop; the narrower frame is delivered to the loop entry, the loop's per-call gcmap references a slot past `jf_frame_length`, and the GC trace aborts.

This mirrors RPython's `check_frame_before_jump` (`aarch64/regalloc.py` `prepare_op_jump` → `assembler.py`), whose comment describes exactly this case: *"a frame coming from god knows where that's passed to a jump to another loop. Make sure it has the correct depth."* pyre had this only in `emit_attached_loop_dispatch` (source side); the loop **entry prologue** (cranelift's destination-side tail-call adaptation) lacked it.

## Fix

Compute `expected_size = max_output_slots + num_ref_roots` for loops (matching the published `target_frame_depth` and what `run_compiled_code_inner` allocates) and run the existing load/compare/realloc for **both loops and bridges** (bridges keep `precompute_max_output_slots`). Reallocs only grow, and the pinned frame register changes only via tail-call (re-enters the entry block) or realloc — so a destination-side entry check catches every delivery. A fresh loop entry already matches its size, so the added check is a no-op there.

## Verification

- cranelift repro all N (3000/5000/8000/**10000/20000**): correct output, no crash
- dynasm repro all N: correct (this change is cranelift-only; `cargo tree -p pyrex --features dynasm -i majit-backend-cranelift` = "did not match" — dynasm does not link the changed crate)
- `cargo test -p majit-backend-cranelift --features cranelift`: pass
- `check.py`: **cranelift 180/180**; dynasm 179 + pre-existing `nested_loop` perf-margin flake (correctness passes)

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -8695,22 +8695,28 @@ impl CraneliftBackend {
         builder.switch_to_block(stack_check_continue);
         builder.seal_block(stack_check_continue);
 
-        // aarch64/assembler.py:927 `_check_frame_depth` parity — bridge
-        // prologue.  Bridges may require a wider JITFRAME than the
-        // failing parent guard's frame supplies (the parent was sized
-        // for its own `max_output_slots + num_ref_roots`, the bridge
-        // for potentially more).  PyPy emits this check in every bridge
-        // prologue; on `jf_frame.length < expected_size` it calls
-        // `_frame_realloc_slowpath` (assembler.py:118 / aarch64:434),
-        // which `cranelift_realloc_frame` line-by-line ports.  Main
-        // loops skip the check because `run_compiled_code_inner`
-        // (compiler.rs:6191) allocates the frame at exactly the size
-        // the body needs.  Source-guard presence is the bridge
-        // discriminator (`compile_bridge` is the only do_compile caller
-        // that supplies it; `compile_loop` passes `None`).
-        let mut jf_ptr = if source_guard.is_some() {
-            let expected_size =
-                (precompute_max_output_slots(inputargs, ops) + ref_root_slots.len()) as i64;
+        // `_check_frame_depth` parity — entry prologue.  A trace can be
+        // entered on a JITFRAME another trace allocated: a loop reached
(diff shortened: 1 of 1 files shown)


youknowone:
Audited the dynasm backend for the same class of bug — it is **not** affected, by a different (correct) mechanism, which is worth recording since it explains why the check belongs at a different place in each backend:

- dynasm's loop prologue also skips `emit_check_frame_depth` (bridge-only, `aarch64/assembler.rs`), and it has no runtime `check_frame_before_jump`.
- Instead, at an external JUMP it accumulates `jump_target_frame_depth = max(.., descr.target_frame_depth())` and then grows the trace's own frame at compile time: `self.frame_depth = self.frame_depth.max(self.jump_target_frame_depth)`. A bridge that jumps to a wider parent loop therefore pre-sizes its **own** frame_depth to cover the target, and its prologue reallocs the incoming frame to that — so it never delivers a narrower frame downstream.

cranelift lacked that compile-time source pre-sizing: the bridge's `expected_size` was just `precompute_max_output_slots + ref_roots`, missing the jump target's depth — hence the narrow frame reaching the loop. This PR closes the gap on the destination side (loop entry re-checks and reallocs); an equivalent alternative would have mirrored dynasm's source pre-sizing.

— *commented by Claude*

Issue #493, opened 2026-07-11
Epic: FBW always-commit protocol — retire replay-from-entry


youknowone:
> **STATUS UPDATE (2026-08-23).** Multi-slice epic, partially landed. **Shipped:** always-commit bridge slices (PR #508 — CA adopted-frame completion, orphan-stamp default-on, merge-point cut), nested-loop vstack / non-flush bridge resume (PR #499), the #495 B1 forward-commit + exception forward-delivery chain (PR #513), and **S1 in full** — the `ResidualDecline` cause split, the `fbw_bump_executed_effect` counters, and the `PYRE_UNJOURNALED_SITE` census line. Flush coverage has also reached S4's shape incrementally: **eleven named `WalkEndCommitLeg` legs**, not the two end-variants the audit below counted. **Still open:** the core replay-from-entry retirement — the decline marker `fbw_mark_unjournaled_effect` still fires at **8 sites**, and S5–S6 (the no-replay rule, deleting the refusal machinery + journal rollback + sticky flags) are untouched; nothing has been deleted.
>
> ⚠ **The 2026-07-11 audit census below is stale in every line.** Re-census with current numbers, file locations and sub-issue states: https://github.com/youknowone/pyre/issues/493#issuecomment-5382904311

---

# Epic: FBW always-commit protocol — retire replay-from-entry

## Charter

Quoting #467: "The replay-from-entry path is the FBW adaptation, and it is the defect."

RPython invariant (the spec):

- Tracing IS execution: the metainterp executes every operation exactly once while recording it (`do_residual_call`).
- On successful compile, control returns to the interpreter with the **current** red boxes — `raise_continue_running_normally` (pyjitpl.py:3072-3085) unpacks the boxes tracing reached, not trace-entry state.
- On ANY abort, `run_blackhole_interp_to_cancel_tracing` (pyjitpl.py:2949-2956) converts the **current** metainterp framestack into blackhole interpreters "filled with the same values" and runs FORWARD (`convert_and_run_from_pyjitpl`). Nothing ever replays.

Pyre's FBW walk instead has a mixed protocol: walk-end/abort flush legs (trace.rs) try to adopt the walked state (`flush_walk_end_state_to_frame` — pyre's `convert_and_run` analog), but are gated; when any gate declines, the portal replays the region from the entry snapshot (portal) or guard pc (bridge), re-executing what the walk already did, with only an **incomplete** store journal rolled back.

Two complementary failure shapes, one root:

[...]

PR #544, opened 2026-07-14
jit: multi-param self-recursive CALL_ASSEMBLER fold + dynamic double-apply gate

youknowone:
Three commits extending the self-recursive `CALL_ASSEMBLER` fold in the full-body walker.

### `decline self-recursive fold after a foreign body write`
A non-pure residual (e.g. `events.append(n)`) executed concretely before a self-recursive call could double-apply on replay: folding terminates the walk symbolically, a later value-unavailable decline leaves it uncommittable, and the interpreter replays the region. A dynamic per-walk latch `FBW_EXECUTED_BODY_RESIDUAL` declines the fold when a foreign (non self-recursive) non-pure residual has already run. The self-recursive call itself runs as a residual during warmup and is excluded via a callee-code == walk-code check at the executor, so pure recursion (`fib`) still folds.

The decline is dynamic, not static: a body write *before* the self-call must decline, but a write *after* the self-call is fine to fold, so an order-insensitive "body has a non-pure call" gate would be wrong.

### `generalize self-recursive fold to N int params`
`try_walker_call_assembler_self_recursive` now accepts N dense all-int positional parameters (capped at 8) instead of exactly one, building the callee frame through `emit_new_pyframe_inline_with_params`. Every self-recursive call routes to the fold, or to the plain residual call when the fold declines. Removes `static_walker_should_decline` / `code_is_self_recursive`; the multi-param body double-apply they statically guarded is now covered by the dynamic gate above.

Multi-param accumulator/linear recursion (`rs(n, acc)`, `fib2(n, a, b)`, `gcd`) compiles instead of falling back to the interpreter — `mp_perf` 11.08s → 1.96s (dynasm), ~5.6×. `fib` fold and speed are preserved.

### `decline self-recursive fold on bool arguments`
The per-parameter `is_int` check also accepted `bool`, whose payload reads through a different accessor than the int unbox path, corrupting the box and crashing (SIGSEGV) once the multi-parameter fold could carry a bool through a separate parameter. Excludes `is_bool` so a bool argument declines to the plain residual call.

### Verification
- `check.py` dynasm 180/180 + cranelift 180/180
- multi-param probes and 8 double-apply / edge repros: oracle == JIT on both backends
- corpus Terminate-leak scan: 0/169
- fold engagement preserved: `mp_perf` EMIT=1, `fib` EMIT=2

A session-close codex parity review flagged six findings; the bool crash was the one real bug (fixed here), and the other five were empirically refuted against oracle output (non-foldable recursion routes to the residual by design; `*args`/keyword-only signatures already decline the fold; the callable/globals resolution is per-function and correct under reassignment; the body-residual latch scenario does not reproduce).

— *commented by Claude*

### DIFF
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
@@ -5983,9 +5983,8 @@ fn residual_callee_is_walk_self_recursive(
         if sym.jitcode.is_null() {
             return false;
         }
-        let caller_code = pyre_interpreter::live_code_wrapper(
-            (*sym.jitcode).raw_code() as *const (),
-        ) as *const ();
+        let caller_code = pyre_interpreter::live_code_wrapper((*sym.jitcode).raw_code() as *const ())
+            as *const ();
         w_code as usize == caller_code as usize
     }
 }
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
@@ -5945,6 +5945,51 @@ fn probe_resid_decline_ctx(
     );
 }
 
+/// Whether a residual call is a self-recursive call to the walk's own code —
+/// the `CALL_ASSEMBLER` fold target running as a plain residual because the
+/// fold declined (no compiled token yet, a non-concrete argument during a
+/// bridge resume, etc.).  Mirrors the callee/self resolution in
+/// `try_walker_call_assembler_self_recursive`.  Keeps the recursion itself out
+/// of the foreign-body-residual latch so pure recursion (`fib`) still folds.
+fn residual_callee_is_walk_self_recursive(
+    ctx: &WalkContext<'_, '_>,
+    allboxes: &[OpRef],
+    helper: majit_ir::PyreHelperKind,
+) -> bool {
+    if helper != majit_ir::PyreHelperKind::CallFn {
+        return false;
+    }
+    let sym_ptr = ctx.fbw_mode.snapshot_sym;
+    if sym_ptr.is_null() {
+        return false;
+    }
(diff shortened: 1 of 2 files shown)


youknowone:
## Review follow-up

Rebased onto current `main` (was based on a slightly older tip); the diff is unchanged (2 files, self-recursive fold only).

**Codex P2 — `jitcode_dispatch.rs` self-recursive residual latch exemption (`residual_callee_is_walk_self_recursive`):**
Valid observation in the abstract. Investigated empirically and could not construct any program that triggers the double-apply, including a maximally-adversarial case built to hit the exact described path — a `bool`-argument same-code residual whose body mutates via an `IntEq`-gated `append` (so no foreign `CallFn` arms the latch first), followed by an `int` self-call that folds with a value-dependent `+ 1` (forcing an uncommittable walk). That repro and a `float`-argument variant both match the `PYRE_JIT=0` oracle on dynasm and cranelift.

The exemption is broader than its strict purpose (sparing the pure warmup/bridge recursion so `fib` still folds), but no observable double-apply results: a foreign body op arms the latch before any fold, and the pure warmup recursion is the only same-code residual that reaches the executor ahead of a fold. A provably-minimal narrowing (exempt only fold-eligible-argument residuals, or distinguish an actually-mutating residual) belongs to the broader always-commit / #493 double-apply work — two earlier narrowing attempts on this exact gate regressed `fib` ~7×, so I am not shipping an unvalidated tightening for a concern I cannot make observable.

**CI `pyre/check.py` failure (`cranelift fib_recursive 2.91s > cpython x2`):**
Perf-threshold flake. `fib_recursive` is a short, high-variance bench; locally cranelift runs it in 1.17–1.29s across 8 runs (~x1.1 vs cpython), well under the 2.22s bound, and `fib` still folds (`EMIT=2`). The compiled loop is byte-identical to base — this change only adds one-time trace-side cost — so steady-state perf is unchanged. check.py is 181/181 on both backends locally; re-running on the rebased base.

— *commented by Claude*

Issue #518, opened 2026-07-13
majit-gc shadow_stack: mutator registry invariants (paired unregister, valid extra-area data, STW-only foreign walks) are not enforced by the API


youknowone:
## Context

Follow-up from the PR #447 multi-thread root registry (issue #452). Raised by the coderabbit review on PR #447 (`majit/majit-gc/src/shadow_stack.rs:337` thread); triaged as real API-hygiene debt but too invasive for that PR.

## Problem

The `MUTATOR_REGISTRY` API in `majit/majit-gc/src/shadow_stack.rs` upholds its soundness invariants only by caller discipline; several of them are reachable from safe code:

- `register_mutator()` / `unregister_mutator()` (`shadow_stack.rs:259`, `:328`) are a manual pair. A thread that registers and exits without unregistering leaves a dangling TLS entry that a later foreign walk dereferences.
- `register_mutator_extra_area(walk, data)` (`shadow_stack.rs:285`) is a safe `fn` accepting an arbitrary `*const ()` plus a walker that will dereference it from another thread.
- The foreign walkers `walk_all_extra_areas` / `walk_all_roots` / `walk_all_jf_roots` / `walk_all_bh_regs` / `walk_all_resume_ref_roots` (`shadow_stack.rs:296`, `:420`, `:610`, `:769`, `:837`) dereference other threads' TLS areas but are safe functions with no check that the caller owns STW (`gc_sync::mutators_quiesced()`).

All current call sites uphold the invariants (registration is paired in the mutator-thread lifecycle, walks run only from the collection drivers under STW), so this is latent, not a live bug.

## Direction

- Make `register_mutator` return an RAII registration guard (unregister on drop), or add a debug-mode leak check on thread exit.
- Mark `register_mutator_extra_area` `unsafe` with a documented contract for `data` lifetime/validity.
- Make the `walk_all_*` foreign walkers `unsafe`, or have them `debug_assert!(gc_sync::mutators_quiesced() || single-thread fast path)` before touching foreign areas.

Keep the RPython framing in mind: upstream's `root_walker` has no Rust-safety notion, so this is purely a Rust-boundary hardening — the walk structure itself should not change.

— *filed by Claude*

youknowone:
Implemented and merged via #551 (squash `9ba9913623e`).

Pure Rust-boundary hardening, zero release-behavior change (the `debug_assert` compiles out; the `unsafe` is compile-time):

[...]

Issue #517, opened 2026-07-13
gc_sync STW: compiled traces have no safepoint poll — a non-allocating compiled loop can stall quiesce_mutators() indefinitely


youknowone:
## Context

Follow-up from the PR #447 STW quiescence protocol (issue #452). Raised by the codex review on PR #447 (P1, `pyre/pyre-jit/src/eval.rs:2999` thread); triaged as real but out of that PR's scope.

## Problem

The STW protocol parks mutators at **entry-only safepoints**: the single `safepoint_poll()` call sits in `PyPyJitDriver::dispatch` (`pyre/pyre-jit/src/eval.rs:2999`), and `gc_op` itself acts as a park point for threads that enter GC trampolines.

A registered mutator executing a compiled trace that neither returns to `dispatch` nor calls a GC trampoline — e.g. a pure unboxed arithmetic loop with inline allocation disabled or no allocation at all — has no point where it can park. If another thread triggers `collect_nursery`/`collect_full`, `quiesce_mutators()` waits for the RUNNING count to drop and the collection hangs until the compiled loop happens to exit.

## Direction

Emit an STW check on compiled back-edges (or an equivalent compiled-code safepoint), in all three backends (dynasm, cranelift, wasm). The check should be a single load-and-branch on `stw_requested` in the fast case, with the slow path routing into the existing park (`gc_sync` RUNNING decrement + wait + re-entry).

RPython reference: with a GIL this cannot arise (the collector owns the only running thread); the free-threaded STW protocol is a pyre deviation, so the poll placement should mirror how JIT guard/counter checks are already emitted rather than any upstream structure.

## Notes

- Today the hang is theoretical for the check.py corpus (hot loops allocate, and allocation routes through `gc_op`), but it is trivially reachable with a hand-written unboxed loop plus a second registered thread.
- Cost matters: the poll must not measurably regress hot loops (see #22 back-edge sensitivity). Gate/measure accordingly.

— *filed by Claude*

youknowone:
Implemented and merged via #551 (squash `9ba9913623e`).

[...]

Issue #174, opened 2026-06-13
Statically enforce the one-can-raise-op-per-block invariant in JIT flattening (currently only a debug_assert)


youknowone:
## Summary

Pyre's JIT flattener relies on an invariant that each basic block contains **at most one operation that can raise**, and that the block's `catch_exception` directly follows that raising op's trailing `-live-`. In RPython this invariant is *structural* — flowspace closes a block at every can-raise operation, so a violating block cannot be constructed. In Pyre the same invariant is currently only *maintained by construction* (a hoist) and *checked by a `debug_assert!`* (compiled out in release), not statically enforced. A future change that places two can-raise ops in one block, or serializes an op between the raising op's `-live-` and the `catch_exception`, would pass flattening in release builds and surface only as a silent blackhole resume corruption at runtime.

## Why the invariant matters

The blackhole exception-resume path and the after-residual-call resume anchor both assume the adjacency:

- `handle_exception_in_frame` (`rpython/jit/metainterp/blackhole.py:396`) skips exactly one `-live-` and then expects the `catch_exception`.
- `derive_after_call_indices_from_sparse` (Pyre's after-residual-call resume anchor) is keyed on `insns[catch - 1].is_live()`.

If a vable-mirror store (`setfield_vable_i` / `setarrayitem_vable_r`) is serialized between the raising op's `-live-` and the `catch_exception`, both mechanisms silently mis-resume.

## How RPython enforces it structurally

- Flowspace closes a block at each can-raise op: `rpython/flowspace/flowcontext.py` (`do_op` → `guessexception(op.canraise)`, ~line 379).
- The codewriter flattener then emits exception links off that block boundary, walking back over trailing `-live-` ops: `rpython/jit/codewriter/flatten.py:200-220`.

Because the block is *closed* at the can-raise op, both "one can-raise op per block" and "`catch_exception` adjacent to the trailing `-live-`" are guaranteed by construction.

## Current Pyre state

Pyre does not split blocks at the can-raise op. Instead it:

1. Computes `block_can_raise` and early-returns for non-raising blocks — `pyre/pyre-jit/src/jit/flatten.rs:1858-1872`.
2. Hoists `catch_exception` to directly follow the raising op's trailing `-live-`, moving any trailing vable-mirror stores after it — `pyre/pyre-jit/src/jit/flatten.rs:1887-1915`.
[...]

PR #551, opened 2026-07-14
jit/gc: eval-breaker back-edge fold (#517) + registry hardening (#518) + catch-adjacency (#174)

youknowone:
Three JIT/GC hardening commits from the #517-519 STW epic, on `ssa-repr` ahead of `main`.

## 1. `jit: fold ticker+STW back-edge polls into one word` (#517)

Compiled-loop back-edges polled two separate process-global words — the async-action ticker address and a GC stop-the-world flag. This folds both into one `AtomicUsize` bitmask (bit0 async mirrors `_ticker < 0`, bit1 STW mirrors `stw_requested`), baked into one reserved register, so the back-edge is **one load + one nonzero branch** (matching the pre-STW baseline — the two-poll version regressed non-allocating hot loops).

- New `majit-ir/eval_breaker_word.rs`: the bitmask word, its published address, and a test-only per-thread address override.
- `executioncontext`/`signalstate`: mirror the ticker sign into bit0 (`reset_ticker`/`decrement_ticker`/`rearm_ticker`).
- `gc_sync`: set/clear bit1 next to `stw_requested` under the quiesce lock.
- `eval.rs`: publish the address at GC init and call `safepoint_poll()` at the `eval_loop_jit` head so a deopted non-allocating loop actually parks (closes a latent park gap where a back-edge STW deopt re-entered the interpreter without a park point).
- Backends: single load/cmp/branch arm — aarch64 hoists the address into a reserved callee-saved register, x86 immediate, cranelift condition-only.
- Removes the now-superseded `gc_stw` and `eval_breaker` address holders → single-word design.

## 2. `majit-gc: harden mutator registry API safety boundary (#518)`

Encodes the `MUTATOR_REGISTRY` foreign-walk / extra-area soundness invariants at the API boundary (no release-build behavior change): `debug_assert!(mutators_quiesced())` on the five `walk_all_*` foreign walkers, `register_mutator_extra_area` marked `unsafe` with a documented lifetime contract, and the `register_mutator` RAII pairing documented.

## 3. `jit: enforce catch_exception -live- adjacency in release` (#174)

Promotes the `insert_exits` catch_exception/`-live-` adjacency guard from `debug_assert!` to `assert!` so a vable-mirror store serialized between them cannot silently mis-resume the blackhole in release builds.

## Verification

- `python pyre/check.py --backend dynasm` → **180/180**; `--backend cranelift` → **180/180**.
- Backend unit tests pass in parallel (`--features dynasm`/`--features cranelift`); the one global-STW multi-thread park test is `#[ignore]`d (drives process-global STW, runs serially).
- Parity/correctness review of the fold: codex parity review reported no regression introduced by this patch; a focused adversarial verification (4 lenses × 2-skeptic refutation) surfaced three candidate divergences, **all refuted** — the bridge-loop park gap targets dead code, the `_ticker`/bit0 mirror window is closed by the interpreter's direct-`_ticker` gate before any bit0-polling trace is entered, and the reserved-register cross-trace hazard is unreachable because address publication precedes all trace compilation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved responsiveness to asynchronous signals and stop-the-world requests in JIT-compiled loops, with more reliable deoptimization/pausing and correct resumption behavior.
  * Guard behavior is now consistent across compiled traces and remains inert until the required runtime state is published.
* **Reliability**
  * Strengthened garbage-collection safepoint coordination and added an additional STW poll inside the JIT dispatch loop.
  * Added/updated tests covering signal-triggered deoptimization and mutator parking/resumption.


youknowone:
Ran the codex parity review cycle on this PR (15 files, base = merge-base with `main`).

**One real finding, fixed** (commit `16c2a240ff2`): `ActionFlag` is per-execution-context, so the fold's `reset_ticker`/`decrement_ticker` async-bit mirror fired for *every* context's ticker. A non-main context clearing its own ticker could clear the shared `EB_ASYNC` bit while the signal-registered ticker was still negative, dropping a pending async from a compiled loop polling that bit. Pre-fold this couldn't happen — compiled traces baked only the registered (main) ticker's address. The fix gates both mirror sites on the registered ticker cell (`std::ptr::eq(&self._ticker, registered_ticker_ptr())`), so only that single cell drives the shared word — behaviorally identical to the pre-fold poll. Both gates re-run green afterward (dynasm 181/181, cranelift 181/181).

**No other in-scope findings.** The remaining structural-adaptation items (free-threaded async/STW word, the reserved poll-address register, `unsafe` foreign-TLS root registration, the release-profile `-live-`/`catch_exception` adjacency assert) are intended Rust/free-threading adaptations. A separate focused adversarial pass over the fold (async-mirror, STW-park, backend-arm, equivalence lenses) surfaced three further candidates — all refuted (the bridge park gap targets dead code, the same-thread ticker/bit window is closed by the interpreter's direct-`_ticker` gate before any trace entry, and the reserved-register cross-trace hazard is unreachable because address publication precedes all trace compilation).

— *commented by Claude*

Issue #346, opened 2026-07-03
Retire the rtyper legacy walker: terminal PyPy-faithful cutover plan (#131 goal G)


youknowone:
Successor to #131. Goal **G** is unchanged: remove pyre's transitional rtyper legacy walker and make the ordinary PyPy-shaped pipeline authoritative for every graph in the portal closure:

```text
interpreter source → flow graph → one whole-program annotator fixpoint
                   → one whole-program rtyper specialization
                   → CodeWriter.make_jitcodes()
```

This issue was closed in July 2026, but its terminal condition was not reached. The old counts and the old A–E plan are now stale. This body supersedes them with the current census and the implementation rules needed for a PyPy-faithful cutover.

## Current technical state

Fresh measurement on branch `rtyper-legacy` at `748004e1c67`, after re-extracting the default LLBC set from the current sources:

```bash
python3 scripts/extract-llbc.py
touch pyre/pyre-jit-trace/build.rs
PYRE_RTYPER_VERBOSE=1 cargo build --release -p pyre-jit-trace
```

Single-run reading:

| Metric | Count |
|---|---:|
| CodeWriter JitCodes considered | 2,866 |
| real-rtyper Match | 926 |
| legacy fallback Skip | 1,940 |
| phase A failures | 1,695 |
| phase B failures | 316 |
| direct dual-gate divergences | 6 |

Approximately 68% of the current closure still takes the legacy fallback. This is a directional snapshot, **not a stable regression baseline**: #1139 proves that the prepass census can vary at a fixed commit and fixed LLBC corpus. Until #1139 is closed, every before/after claim must repeat the same arm against one fixed corpus and compare subject sets, not only counts.

The code's own completion criterion remains the authoritative one:

- `cutover::unported_category` matches nothing;
- no `[PREPASS phaseA fail]`, `[PREPASS phaseB fail]`, or `[PYRE_RTYPER skip]`;
- `DualGateOutcome::Skip` and its legacy baseline are dead;
- `legacy_annotator.rs` and `legacy_resolve.rs` can be deleted.

## PyPy/RPython design constraints

1. **One translator-owned annotator and one rtyper.** RPython does not independently annotate/rtype each graph and recover from failures. The per-subject `catch_unwind`, poisoned-session repair, and Skip isolation are migration scaffolding; they retire at terminal cutover.
[...]

PR #553, opened 2026-07-14
#346: retire rtyper legacy walker — vec!→NewList, GC-safepoint residuals, f64::abs lowering

youknowone:
## Summary

#346 rtyper-legacy retirement — four census slices on top of `main`, each lifting JIT-prepass graphs that previously fell back to the legacy walker. The metric is the distinct `[PREPASS phaseA fail] CallPath` count (what `cutover::is_known_unported` counts); every slice was verified with a census delta + `check.py` gate.

## Commits

1. **`vec![..]` → `OpKind::NewList` with cross-block box sweep** — recognizes the `box_assume_init_into_vec_unsafe(box [e0..eN])` form the `vec!` macro lowers to, emits `NewList`, and sweeps the dead cross-block `Box::new_uninit` remnant via `prune_dead_boxing_remnants` (dependency-flow liveness).

2. **GC-safepoint global-state readers residual** — six zero-arg void/bool cannot-raise readers of runtime-mutable process globals (`try_gc_collect_oldgen`, `try_gc_jitframe_empty`, `at_outermost_activation`, `collect_enabled`, `bump_frame_entry_count`, `itemsblock_gc_enabled`) marked `#[dont_look_inside]` and registered in `jit_trace_fnaddrs`, mirroring the `gc_interp::enabled` / `note_alloc` precedent.

3. **safepoint entry + dict store-through hook residual** — `gc_interp::safepoint` (reads `ALLOC_SINCE_GC` inline + dispatches the collection hook) and `dictmultiobject::maybe_sync_dict_storage_store` (stores through `DICT_STORAGE_STORE_HOOK`); dominoes `typing_intrinsic_1/2`.

4. **f64::abs → `float_abs`, float `/` → truediv** — recognizes `core::f64::<Impl>::abs` and emits an abstract `abs` `UnaryOp` (rtyper lowers via `FloatRepr::rtype_abs` → `float_abs`, `rfloat.py:44-46`), teaches the closed `normalize_unary_op_name` allowlist the `"abs"` opname, and relabels a float-result `floordiv` to `truediv` (the type-blind `canonical_binop_label` maps every `Div` to the integer `floordiv`, but `FloatRepr` has only `rtype_truediv`). Lifts `complex_truediv`.

## Verification

`python3 ./pyre/check.py` — **dynasm 181/181, cranelift 181/181, wasm 180/180** at the branch tip.
Each slice's census delta was confirmed with no path regression (set-diff of the PREPASS-fail path set before/after).

### DIFF
--- a/pyre/pyre-interpreter/src/call.rs
+++ b/pyre/pyre-interpreter/src/call.rs
@@ -152,7 +152,12 @@ pub fn frame_entry_count() -> u64 {
 /// ([`FRAME_ENTRY_COUNT`]).  Called once at every `eval_loop` /
 /// `eval_loop_jit` entry, i.e. each time a user Python frame begins
 /// executing bytecode.
-#[inline(always)]
+///
+/// Touches the runtime-mutable `FRAME_ENTRY_COUNT` thread-local, not a
+/// build-time constant, so the JIT residualizes the call instead of tracing
+/// into it (`@dont_look_inside`, the `note_alloc` sibling). A `()` return has
+/// no discriminant to erase and it cannot raise.
+#[majit_macros::dont_look_inside]
 pub fn bump_frame_entry_count() {
     FRAME_ENTRY_COUNT.with(|c| c.set(c.get().wrapping_add(1)));
 }
--- a/pyre/pyre-interpreter/src/jit_fnaddr.rs
+++ b/pyre/pyre-interpreter/src/jit_fnaddr.rs
@@ -721,6 +721,28 @@ pub fn jit_trace_fnaddrs() -> Vec<(&'static str, i64)> {
         "pyre_interpreter::call::bump_frame_entry_count",
         crate::call::bump_frame_entry_count as *const (),
     );
+    // The dispatch-loop safepoint entry itself reads `ALLOC_SINCE_GC` inline and
+    // dispatches to the collection hook, and `maybe_sync_dict_storage_store`
+    // stores through the `DICT_STORAGE_STORE_HOOK` fn-pointer cell — both
+    // runtime-mutable, `()`-returning, cannot-raise, word-argument residuals
+    // (`#[dont_look_inside]`), bound by qualified path.
+    push_alias_pair(
+        &mut entries,
+        "pyre_object::gc_interp::safepoint",
+        "pyre_object::safepoint",
+        pyre_object::gc_interp::safepoint as *const (),
+    );
+    let maybe_sync_dict_storage_store: unsafe fn(
+        *mut u8,
+        pyre_object::PyObjectRef,
+        pyre_object::PyObjectRef,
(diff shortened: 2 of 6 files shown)


youknowone:
Added a fifth slice on top: **f64::is_finite → `(f - f) == 0.0` float-op lowering** (`ad054a26e07`).

`core::f64::<Impl>::is_finite` was an Opaque `core` leaf walling `reject_float_coercion_overflow` (the int→float coercion guard shared across the float-arithmetic family). Lowered to the `ll_math.py:106-111` isfinite identity — `ConstFloat(0.0)` + `float_sub(f, f)` + `float_eq(diff, 0.0)`; `inf`/`NaN` minus self is `NaN`, `NaN == 0.0` false, so only finite yields true. Census 288 → 284 (`reject_float_coercion_overflow` + `float_add`/`float_sub`/`float_truediv` cleared); check.py 181/181/180.

— *commented by Claude*

youknowone:
## Rebased onto `main` (`9ef026d637e`) — dropped 3 of 5 commits

Rebased onto current `main` (which now includes #534 f64-intrinsic MIR lowering and #552 objspace operator parity). Three commits were dropped after test-driving each against the new base:

1. **`f64::is_finite` lowering** — #534 landed the identical `(x - x) == 0.0` shape upstream. Pure duplicate.
2. **`f64::abs` + truediv relabel** — #534 **deliberately removed** its own `f64::abs` lowering ("abs mis-kinds to `int_abs`/`f>f` — an unwired divergence"). Re-adding it would re-introduce the bug they removed.
3. **`vec![..]` → `OpKind::NewList`** — this was the **actual cause of the failing `cargo test`** on this PR: `jitcode_runtime::tests::default_bh_builder_unwired_set_matches_task_85_snapshot` drifted `["newlist/r>r"]`. The graph drops to the legacy walker on the dead cross-block `Box::new_uninit`, so `rtype_newlist` never runs and the raw `NewList` flattens with kind `newlist/r>r` (no `bhimpl_n` handler) instead of `newlist/idddd>r`. The build artifact `jit_metadata.json` confirmed the only `newlist` insns entry was the malformed `newlist/r>r`. Deferred until the vec! graph genuinely two-phase-lifts.

**Remaining (2 commits):** the GC-safepoint residual slices — `dont_look_inside` on the dispatch-loop global-state readers + the safepoint entry / dict store-through hook. Both are orthodox residuals (runtime-mutable GC-interp state with no PyPy-traced counterpart).

**Verification after rebase:**
- `cargo fmt --all --check` clean (the earlier failure was a line-length wrap in the covers-test, folded into the readers commit).
- `cargo test -p pyre-jit-trace --lib`: **263 passed, 0 failed** (was 262/1 on the previous tip).
- The previously-failing snapshot test now passes (`newlist/r>r` gone from the fresh artifact).

— *commented by Claude*

PR #554, opened 2026-07-14
interp: compute __abstractmethods__ in _abc_init and propagate __isabstractmethod__

fregataa:
## Summary

`_abc_init` only set up `_abc_registry` and never computed `__abstractmethods__`, so ABCs were freely instantiable and `inspect.isabstract` always returned False. This PR completes the ABC abstract-method machinery.

- **`_abc_init` computes `__abstractmethods__`** — new `compute_abstract_methods` (mirrors `_abcmodule.c` / `app_abc.py`): the abstract members in `cls.__dict__` plus those inherited from bases and not overridden with a concrete member.
- **`property.__isabstractmethod__`** getset descriptor (`isabstract(fget) or fset or fdel`).
- **`staticmethod`/`classmethod` `__new__` honour the subtype**, so the legacy `abstractstaticmethod`/`abstractclassmethod` subclasses behave correctly (subtype instance returned, `__init__` runs, class-level `__isabstractmethod__ = True` is visible).
- **`w_type_get_flags` reports `Py_TPFLAGS_IS_ABSTRACT`** from the existing `flag_abstract`, which `inspect.isabstract` reads.

## Test

`python -m test test_abc`: **24 failures + 8 errors → 11 failures, 0 errors.**

Remaining `test_abc` failures are unrelated to this change:
- `test_abstractproperty_basics` / `test_descriptors_with_abstractmethod` are blocked by a separate `super().<property>` bug (property `__get__` is not invoked through `super`), reproducible independently of `abc`.
- `test_issubclass_bad_arguments` is a pre-existing `issubclass()` argument-validation gap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subclass behavior for `staticmethod` and `classmethod`, ensuring type checks correctly reflect custom subtypes.
  * Added support for identifying abstract properties based on their accessors.
  * Updated type flags so abstract classes are correctly recognized by runtime checks.

### DIFF
--- a/pyre/pyre-interpreter/src/typedef.rs
+++ b/pyre/pyre-interpreter/src/typedef.rs
@@ -8420,12 +8420,25 @@ fn init_staticmethod_type(ns: PyObjectRef) {
             "__new__",
             make_builtin_function("__new__", |args| {
                 // staticmethod(func) — args[0] is cls (staticmethod type), args[1] is func
+                let cls = args.first().copied().unwrap_or(pyre_object::PY_NULL);
                 let func = if args.len() > 1 {
                     args[1]
                 } else {
                     pyre_object::w_none()
                 };
-                Ok(pyre_object::function::w_staticmethod_new(func))
+                let sm = pyre_object::function::w_staticmethod_new(func);
+                // function.py `allocate_instance(StaticMethod, w_subtype)` —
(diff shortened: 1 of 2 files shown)


youknowone:
@fregataa Thank you! could you please run `cargo fmt --all` for formatting?

PR #562, opened 2026-07-15
Exception-path JIT: fold exception construction/attrs/str, lower DELETE_FAST/DEREF

youknowone:
Removes a JIT trace abort on the exception hot path, and fixes an exception-reload
defect that kept two exception loops interpreter-only.

## 1. `DELETE_DEREF` no longer `abort_permanent`

When a nested function closes over an `except E as e:` name, the local becomes a
cell and the handler's implicit `del e` compiles to `DELETE_DEREF`. That opcode
reached `abort_permanent`, and `loop_body_has_abort_permanent` is a *static* scan —
one occurrence declines the whole loop.

It is now lowered by composing existing graph ops — no new residual:
`load_deref_value` (raise-if-unbound) + `store_deref_value(cell, NULL)` (clears the
cell contents, keeps the cell in the slot) + `store_local_value`. The live
interpreter impl is `delete_deref`. Added a guard bench
`exception_as_cell_cleanup.py`, which compiles (`loops_compiled=1`,
`bridges_compiled=2`) and matches the `PYRE_NO_JIT=1` oracle.

## 2. Cleanup-landing exception reloaded from the frame's stack slot

A `Copy` handler landing reached by an explicit raise seeded its top-of-stack
exception with a `last_exc_value` read. A residual call between the raise landing
and that read clears the metainterp exception transient (`execute_varargs` clears
before a call that may itself raise), so the read found no value and the walk
aborted with `LastExcValueWithoutActiveException`.

Those landings now reload the propagated exception from the frame's durable top
stack slot via `getarrayitem_vable_r` — catch dispatch has already stored it there.
Reraise-entered landings keep the `last_exc_value` read, which still carries the
live payload from the exception edge. The per-thread current-exception slot is
deliberately **not** the source: at this landing it can still name the older
exception whose handler raised this one, which would mislink `__context__`.

## Results

- `exception_bare_reraise_restore` and `exception_context_chain_inhandler` now
  compile (`loops_compiled` 0 → 1), output byte-identical to the oracle.
- Wins: `swap_except_return_resume` 2.79 → 0.06s, `finally_bare_raise` 0.86 → 0.05s,
  `exception_oserror_fields` 1.12 → 0.51s.

## Withheld from this PR

**`DELETE_FAST` → #586.** The lowering in the earlier revision of this PR
**miscompiled**: it dropped the arm's `ptr_iszero` guard and emitted the slot clear
unconditionally after the can-raise `load_fast_check`, breaking the
`pyopcode.py:998` shape where the raise *prevents* the clear. `del x; del x` in a
hot loop then swallowed exactly one `UnboundLocalError` (oracle 200000, JIT 199999
— loss is exactly 1 at every N, so it is a trace/compile transition artifact).
`origin/main` is correct here because it declines the loop. Caught by the Codex
parity review; full repro and characterisation in #586.

**Exit-frame raise FINISH → #581.** It made
`synth/exception_inlined_callee_caught` compile a bridge for the first time, which
exposed a pre-existing GC hole and crashed wasm. Root cause is not in that commit:
the wasm backend has no `gcreftracer` / `LoadFromGcTable` path and bakes
`OpRef::ConstPtr(GcRef)` into emitted code as a raw address immediate, where the GC
can neither find nor forward it — dynasm and cranelift both run
`remove_constptrs_in` → `GcTable::from_gcrefs` (RPython's `GCREFTRACER`).

## Gates

`check.py` dynasm and cranelift; lib tests `pyre-jit`, `pyre-jit-trace`,
`majit-metainterp`. wasm's one failing fixture,
`synth/residual_raise_except_resume`, **fails identically on `origin/main`**
(0.21s vs pypy 0.01s = 21.4x): #576 added `# pyre-check: max-pypy-ratio=10` to it
and the wasm backend does not meet it, while unannotated fixtures in the same sweep
run at 13–33x and pass ungated. This branch adds no new wasm failure.

### DIFF
--- a/majit/majit-ir/src/resumedata.rs
+++ b/majit/majit-ir/src/resumedata.rs
@@ -399,7 +399,13 @@ pub struct RebuiltFrame {
     pub values: Vec<RebuiltValue>,
 }
 
-fn decode_tagged(
+/// Decode one tagged resume value outside the numbered frame stream.
+///
+/// Pending fields carry their target and value through the same
+/// TAGBOX/TAGCONST/TAGVIRTUAL encoding as frame slots
+/// (`resume.py:_add_pending_fields`).  Bridge setup uses this entry point so
+/// its consume-side view is identical to `rebuild_from_numbering`.
+pub fn decode_tagged_value(
     tagged: i16,
     num_failargs: i32,
     rd_consts: &[Const],
@@ -499,7 +505,7 @@ pub fn rebuild_from_numbering(
             break;
         }
         let tagged = reader.next_item() as i16;
-        vable_values.push(decode_tagged(
+        vable_values.push(decode_tagged_value(
             tagged,
             num_failargs,
             rd_consts,
@@ -516,7 +522,7 @@ pub fn rebuild_from_numbering(
             break;
         }
         let tagged = reader.next_item() as i16;
-        vref_values.push(decode_tagged(
+        vref_values.push(decode_tagged_value(
             tagged,
             num_failargs,
             rd_consts,
@@ -548,7 +554,7 @@ pub fn rebuild_from_numbering(
                 break;
             }
             let tagged = reader.next_item() as i16;
-            values.push(decode_tagged(
+            values.push(decode_tagged_value(
                 tagged,
                 num_failargs,
                 rd_consts,
--- a/majit/majit-translate/src/tool/algo/unionfind.rs
+++ b/majit/majit-translate/src/tool/algo/unionfind.rs
@@ -32,6 +32,7 @@
 //! preserve the upstream absorb-skip semantics.
 
 use indexmap::IndexMap;
+use std::collections::HashMap;
 use std::hash::Hash;
 
 /// RPython `info1.absorb(info2)` (unionfind.py:76) — called by
@@ -54,15 +55,18 @@ where
     K: Eq + Hash + Clone,
 {
     /// RPython `self.link_to_parent` (unionfind.py:8).
-    /// `IndexMap`, not `HashMap`: `keys()`/`infos()` iterate these to
-    /// drive commonbase folding and access-set numbering downstream, and
-    /// upstream stores Python dicts (insertion order).  A `HashMap` would
-    /// make that order — and the resulting classification — run-varying.
+    /// `IndexMap`, not `HashMap`: `keys()` iterates this to drive commonbase
+    /// folding and access-set numbering downstream, matching the upstream
+    /// dictionary's insertion order.
     link_to_parent: IndexMap<K, K>,
     /// RPython `self.weight` (unionfind.py:9).
-    weight: IndexMap<K, usize>,
+    /// Upstream dictionary deletion in `union` is O(1); ordering is carried
+    /// by `link_to_parent`, so hash storage preserves both properties.
+    weight: HashMap<K, usize>,
     /// RPython `self.root_info` (unionfind.py:11).
-    root_info: IndexMap<K, V>,
+    /// Upstream dictionary deletion in `union` is O(1); ordering is derived
+    /// from `link_to_parent`, so hash storage preserves both properties.
+    root_info: HashMap<K, V>,
     /// RPython `self.info_factory` (unionfind.py:10). See module doc
     /// for why the Rust port makes this mandatory.
     info_factory: Box<dyn Fn(&K) -> V>,
@@ -81,8 +85,8 @@ where
     {
         UnionFind {
             link_to_parent: IndexMap::new(),
-            weight: IndexMap::new(),
-            root_info: IndexMap::new(),
+            weight: HashMap::new(),
+            root_info: HashMap::new(),
             info_factory: Box::new(info_factory),
         }
     }
@@ -98,9 +102,13 @@ where
         self.link_to_parent.keys()
     }
 
-    /// RPython `UnionFind.infos(self)` (unionfind.py:31-32).
+    /// RPython `UnionFind.infos(self)` (unionfind.py:31-32). Roots enter all
+    /// three dictionaries together; filtering `link_to_parent` insertion
+    /// order to surviving roots matches upstream `root_info.values()`.
     pub fn infos(&self) -> impl Iterator<Item = &V> {
-        self.root_info.values()
+        self.link_to_parent
+            .keys()
+            .filter_map(move |k| self.root_info.get(k))
     }
 
     /// RPython `UnionFind.__getitem__(self, obj)` (unionfind.py:13-20).
@@ -189,15 +197,13 @@ where
         // take both infos out, absorb, and re-insert the merged one
         // onto the chosen new root below. `()` infos no-op via the
         // trait impl, matching the upstream None branch.
-        // `shift_remove`, not `swap_remove` (IndexMap's `remove`): keep the
-        // insertion order of surviving entries stable for the `infos()` walk.
-        let mut info1 = self.root_info.shift_remove(&rep1).expect("rep1 info");
-        let info2 = self.root_info.shift_remove(&rep2).expect("rep2 info");
+        let mut info1 = self.root_info.remove(&rep1).expect("rep1 info");
+        let info2 = self.root_info.remove(&rep2).expect("rep2 info");
         info1.absorb(info2);
 
         // upstream: weighted union, smaller tree under larger.
-        let w1 = self.weight.shift_remove(&rep1).expect("rep1 weight");
-        let w2 = self.weight.shift_remove(&rep2).expect("rep2 weight");
+        let w1 = self.weight.remove(&rep1).expect("rep1 weight");
+        let w2 = self.weight.remove(&rep2).expect("rep2 weight");
         let w = w1 + w2;
 
         let (new_rep1, new_rep2) = if w1 < w2 { (rep2, rep1) } else { (rep1, rep2) };
--- a/pyre/bench/synth/exception_as_cell_cleanup.py
+++ b/pyre/bench/synth/exception_as_cell_cleanup.py
@@ -1,0 +1,30 @@
+# An `except E as e:` handler whose exception variable `e` is captured by a
+# nested function makes `e` a cell variable, so the implicit handler cleanup
+# emits DELETE_DEREF (clear the cell contents) rather than DELETE_FAST. The
+# traced DELETE_DEREF dereferences the cell to raise if unbound, then clears
+# its contents; this bench pins that a captured-exception handler in a hot
+# loop compiles and returns byte-identically.
+N = 60000
+
+
+def run(n):
+    acc = 0
+    i = 0
+    while i < n:
+        try:
+            if (i % 5) == 0:
+                raise ValueError(i)
+            acc += 1
+        except ValueError as e:
+            def grab():
+                return e          # captures e -> e becomes a cell
+            acc += grab().args[0] & 7
+        i += 1
+    return acc
+
+
+def main():
+    print(run(N))
+
+
+main()
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -7992,6 +7992,138 @@ pub fn object_setattr(obj: PyObjectRef, name: &str, value: PyObjectRef) -> PyRes
     Err(raiseattrerror(obj, name))
 }
 
+/// A direct `W_BaseException` reference slot whose hard-coded getattr/setattr
+/// arm has no value conversion on the selected branch.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum ExceptionAttrSlot {
+    Args,
+    Code,
+    Errno,
+    Strerror,
+    Filename,
+    Filename2,
+}
+
+/// Ingredients for the full-body walker's mirror of the typed exception-slot
+/// attribute arms.  This helper deliberately lives beside `getattr_str_impl`
+/// and `object_setattr`, whose branch order it audits.
+///
+/// PyPy exposes these fields through `readwrite_attrproperty_w` in
+/// `interp_exceptions.py`; `GetSetProperty.descr_property_get/set` in
+/// `typedef.py` reaches the tiny slot accessor.  Pyre's equivalent arms run
+/// only after the receiver type lookup.  Therefore a class-dict hit (a user
+/// property or plain shadowing value) declines, while the returned nonzero
+/// version tag lets the JIT pin that miss.  The exception instance dictionary
+/// is consulted only after these whitelisted arms, so it is not a branch input
+/// and must not shadow them.
+///
+/// `GuardClass` on the returned kind-specific `ob_type` pins the `ExcKind`;
+/// heap exception subclasses share that physical layout, so callers must also
+/// guard the exact `w_class` and its version tag.  Loads select only a concrete
+/// non-NULL slot, pinning the stored-value branch instead of the `args_w`
+/// fallback.  `args` returns the raw `args_w` storage as an ingredient only;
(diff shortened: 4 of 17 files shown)


youknowone:
Rebased onto `origin/main` (`49ec51c3b9c`) and resolved both P1 review findings in `7ddf9729537`.

### Both findings traced to one root: the disposition was conflated with the no-replay value

`FBW_FINISH_PAYLOAD` (the symbolic OpRef the compile records) and `FBW_FINISH_CONCRETE` (the concrete value the no-replay portal exit returns) have **different lifetimes**. `879676a211a` used the concrete stash as the only raise-vs-return discriminator, which is what both findings caught:

1. **"Avoid stashing non-concrete raises"** — confirmed. `fbw_finish_raise_set(exc_concrete)` recorded `Raise(Null)` when the concrete shadow was absent, and the no-replay epilogue only checks `.is_some()`, so `call_jit.rs` / `eval.rs` reached `unreachable!("FinishConcrete::Raise must hold a concrete Ref")`. Notably the two pre-existing `!fbw_raise_enabled()` sites **already had** the `ConcreteValue::Ref(p)` + `!p.is_null()` guard — `879676a211a` simply dropped it at the new site. The guard now lives inside `fbw_finish_raise_set`, so the invariant is enforced at the single producer and those two callers collapse to a bare call.

2. **"Preserve raise disposition after no-replay is declined"** — confirmed, and it was a real miscompile. `run_perfn_walk` clears the concrete stash when the shortcut is declined while the payload stays live, so a declined raising walk built `FINISH(exc, done_with_this_frame_descr_ref)` — the trace **returned** the exception object.

### The fix went further than "a separate disposition channel"

The first attempt added a sibling `FBW_FINISH_DISPOSITION` cell. That was wrong: only 1 of the 3 payload writers would have written it, making it a one-way latch whose correctness rested solely on the shared reset — and the previous design at least self-corrected (`fbw_finish_concrete_set` overwrote `Raise` with `Return`). It is also the parallel store `AGENTS.md` rule 2 forbids by name.

So the disposition is carried **in** the payload: `Option<(OpRef, Type, FinishDisposition)>`. Every writer names its disposition, drift is unrepresentable, and a `Raise` payload without a value cannot exist — which retires the `Terminate::RaiseNoFinishPayload` / `P2Framestack::OuterRaiseNoFinishPayload` census keys. This also matches how the sibling channel and upstream both work: `FinishConcrete` fuses disposition and value in one enum, and `finishframe` / `finishframe_exception` raise `DoneWithThisFrameRef` / `ExitFrameWithExceptionRef` (`pyjitpl.py:2513-2561`, `jitexc.py`) — the disposition rides with its value, never beside it.

Auditing that fix surfaced a third defect in the same commit: in `drive_outer_continuation_and_map` both return arms retarget the root green key before emitting `Finish`, but the raise arm `879676a211a` added did not — so the outer frame's exit-with-exception loop was attached/redirected under the bridge's trace-start key (`finish_and_compile` reads `ctx.green_key` for `compiled_loops.swap_remove` + `attach_procedure_with_redirect`). The retarget is now hoisted once to the top of the `Terminate` arm, covering all three outcomes.

### ⚠️ The DELETE_FAST arm intentionally overrides merged upstream #533

#533 landed its own `DeleteFast` walker lowering gated on `fbw_delete_fast_enabled()`, but it **still emits `abort_permanent`** (when `local_value_at(idx).is_none()`, and as the unbound-path block). `loop_body_has_abort_permanent` is a static scan, so it declines the *whole* loop — every `except X as e` (implicit `del e` → DELETE_FAST) stays `comp=0`. A/B proven: skipping our commit regressed `exception_context_chain_inhandler` loops 1→0, abrt 0→5 (`FullBodyWalk::LoopBodyAbortPermanent`). Ours is the orthodox port (`load_fast_check` = raise-if-unbound + clear, per `pyopcode.py:998-1004`; no trace abort, no env flag), so the rebase keeps our arm and deletes the now-orphaned gate. Flagging it explicitly since it replaces already-merged code.

### Verification

- `check.py` cranelift **189/189**; dynasm **188/189** — the single failure is `raise_catch` at `0.17s > pypy 0.11s ×1.5`, a **perf-margin** gate, not correctness. A/B against this branch's own base shows the change is not the cause (with: median 0.172s, min 0.166; without: median 0.202s, min 0.166 — the gate is 0.165s, so both sides straddle it).
- Exception benches on **both** backends, JIT output byte-identical to the `PYRE_JIT=0` oracle: `exception_bare_reraise_restore` (`15000 15000`), `exception_context_chain_inhandler` (`30000`, so the `__context__` chain is intact), `exception_as_cell_cleanup` (`90000`), `exceptions` (`1125000`), `exception_args_virtual` (`1140629 1954`). Compile stats hold (`exceptions`: brdg=1, abrt=5, gf=200).
- Units: pyre-jit-trace 263, pyre-jit 306, majit-metainterp 1399.

— *commented by Claude*

youknowone:
Force-pushed a **reduced** stack, rebased onto `6552f3cd55`. The exit-frame raise
FINISH commit and its review follow-up are withheld; what remains is
DELETE_FAST/DELETE_DEREF plus the cleanup-landing exc reload.

**Why the wasm job was red.** `synth/exception_inlined_callee_caught` aborted with
`GC BUG: invalid type_id=44289240 ... site=object_total_size`
(`majit-gc/src/collector.rs:1511`). Bisected: `origin/main` PASS →
+DELETE_FAST/+DELETE_DEREF PASS → +exit-frame FINISH **PANIC**. dynasm and
cranelift passed throughout.

**It is not a bug in that commit.** The commit's only sin is that it made this
raise trace compile a bridge for the *first* time (`bridges_compiled` 0 → 1);
previously the trace aborted and was thrown away, so it never reached the compile
at all. What it stepped on:

> The wasm backend has no `gcreftracer` / `LoadFromGcTable` path. `emit_resolve`
> (`majit-backend-wasm/src/codegen.rs`) lowers `OpRef::ConstPtr(GcRef)` to
> `i64.const <raw address>` — baked into the emitted bytecode, where the GC can
> neither find nor forward it. `codegen.rs` panics on `LoadFromGcTable` with
> *"the GC rewrite must not run for wasm"*, and `set_constants_pool` flattens the
> typed `ConstMap<Const>` into `IndexMap<u32, i64>`, erasing `Const::Ref` so even
> a walker could not tell refs from ints.
>
> dynasm (`runner.rs`) and cranelift (`compiler.rs`) both run `gc_rewriter()` →
> `rewrite_for_gc_with_constants` → `remove_constptrs_in` (`majit-gc/src/rewrite.rs`,
> `rewrite.py:105-126` parity), pulling every non-null ref constant into
> `GcTable::from_gcrefs`, which self-registers a root walker
> (`gcreftracer.rs install_gc_table_walker`) that both keeps the constant alive
> and forwards it in place for the loop's whole life — RPython's `GCREFTRACER`
> reached through `asmmemmgr_gcreftracers`.

So **wasm is the parity deviation**, and it is latent for any trace that carries a
nursery-resident ref constant — not specific to the raise path. Filed separately;
the FINISH work lands on top of that fix rather than being rushed in beside it.

A secondary, backend-symmetric hole was also confirmed and is in the same issue:
`walk_active_trace_refs` gates the op-graph walk on `self.tracing`, which
`finish_and_compile` `take()`s before recording the FINISH — and never restores,
since the ops then move into `compiled_loops`. That window has no end.

**`residual_raise_except_resume` (the remaining wasm failure) is pre-existing.**
Measured on pristine `origin/main` @ `6552f3cd55`: 0.21s vs pypy 0.01s = **21.4x**,
same FAIL. #576 added `# pyre-check: max-pypy-ratio=10` to that fixture three
hours before this push, and the wasm backend does not meet it; unannotated
fixtures in the same sweep run at 13–33x pypy and pass ungated. This branch adds
no new wasm failure.

Gates on the pushed stack: dynasm 192/192, cranelift 192/192, wasm 190 passed /
1 failed (the fixture above). Lib: pyre-jit 311, pyre-jit-trace 263,
majit-metainterp 1399, 0 failed.

— *commented by Claude*

youknowone:
The wasm GC parity gap described above is now tracked as #581 (three layered defects, all cited to source). The withheld exit-frame FINISH commits are preserved locally at tag `perf-exc-full-stack` and land on top of that fix.

— *commented by Claude*

youknowone:
⚠️ **Do not merge yet** — the Codex parity review's §1 finding is **real**, reproduced against the oracle.

```python
x = 1
del x
del x        # must raise UnboundLocalError
```

In a hot loop (200000 iterations), `PYRE_NO_JIT=1` prints `200000`, the JIT prints **`199999`** — one `UnboundLocalError` is swallowed. Same for a never-bound local (`del y` where `y` was only assigned under a false branch). `loops_compiled=2`. This is a **silent wrong answer**, not a perf issue.

Codex's mechanism reads correct: commit `12db975068` removed the arm's two abort paths (`local_value_at(idx).is_none()` and the `ptr_iszero` branch), leaving `load_fast_check`'s result ignored. After a prior delete the codewriter's `FrameState` for that local is `None`, and the can-raise residual's guard can resume its normal continuation instead of raising — `pyopcode.py:998` requires `DELETE_FAST` to raise before clearing an already-empty slot.

The PR body's claim that both lowerings are *"byte-identical to the interpreter path by construction"* is therefore **wrong for `DELETE_FAST`**; I'm retracting it. A/B against `origin/main` is running to confirm this is a regression rather than pre-existing, then I'll fix it in-session and re-push. Flagging now so nobody merges on the earlier green report.

— *commented by Claude*

youknowone:
## ⛔ Do not merge — this PR introduces a JIT miscompile (retracting my previous green claim)

My earlier comment said this stack was verified green. That was wrong, and I am retracting it. The failing `CPython suite (gate)` check on this PR is **not** the known `test_strftime` flake — it is a regression introduced by these two commits.

### The A/B, both sides rebuilt with the canonical command, binaries hash differently

`cargo build --release -p pyrex --bin pyre-dynasm --no-default-features --features dynasm`

| build | binary sha256 (16) | `TZ=UTC` + JIT on | `TZ=UTC` + `PYRE_NO_JIT=1` |
|---|---|---|---|
| `6564959c41` (origin/main — **this PR's exact base**) | `828087dba9be6d48` | `Ran 4 tests` → **OK** | OK |
| `6311e8a5c0` (this PR head) | `599f6118e6be9c58` | **`TypeError: not an iterator`** → FAILED | OK |

Deterministic local repro — no CI, no timing, 3/3 reproductions:

```
TZ=UTC ./target/release/pyre-dynasm lib-python/3/test/test_strftime.py
```

JIT off passes, JIT on fails (`loops_compiled=5`), so this is a compiled-trace defect, not a `time`-module or formatting problem.

### Why I initially misread it as a pre-existing flake

Two plausible signals lined up and both were misleading:

1. The failure is recorded in our notes as a *"wall-clock-dependent flake — do not attribute it to whatever patch is in flight."* It is not wall-clock dependent at all: the trigger is simply **whether `TZ` is set** (`TZ=UTC`, `TZ=Asia/Seoul`, and even empty `TZ=` all fail; unset passes). That is 100% deterministic, which is why it looked intermittent across environments rather than broken.
2. `main` really does show `test_strftime: PASS -> FAIL` on several recent commits (`2b6b42c6e1`, `9d79802222`, `d405c6b83a`). But those are *other* commits — **this PR's base `6564959c41` is green**, both in CI and locally. The right control is the base this branch actually sits on, not "does main fail somewhere".

`TypeError: not an iterator` is a familiar miscompile symptom in this codebase (it was the #533 signature fixed by PR#558, where codegen picked the last label instead of the JUMP-descr target), so the iterator/label path is the place to start.

Bisecting the two commits now. I will report which one is responsible and either fix it or drop it, as with the `DELETE_FAST` commit earlier in this branch (#586).

— *commented by Claude*

youknowone:
## ✅ Retracting the do-not-merge hold — my accusation was wrong, these commits are innocent

My previous comment claimed these two commits introduce a `TypeError: not an iterator` miscompile. **That was wrong, and the A/B I based it on was invalid.** Correcting the record.

### What was wrong with my experiment

I compared this branch against `git checkout origin/main` — but `origin/main` had **advanced during the session**. So my "main" binary was not this PR's base; it was the newer tip, which carries three commits my base lacks:

```
eca75827fe  #346: union keystone (#590)
e297968a0b  jit: walker-native range FOR_ITER-next specialization with polymorphic demotion (#545)
091a7953fc  jit: virtualize BUILD_LIST on the FBW walker (#580)
```

So I was measuring **"my base without #545" vs "newer main with #545"** and attributing the whole delta to my two commits. The tell was there and I explained it away: the two "main" builds produced different binary hashes (`828087db` → `86de731c`) and I dismissed that as non-reproducible builds.

### The control I never ran, run now — my base ALONE

`red2.py` is a clock-pinned reduction of the failure (the bug is wall-clock dependent — only ~11 of 40 sampled clock values trigger it, which is why it reads as a flake).

| build | contents | `TZ=UTC` + pinned clock ×3 values |
|---|---|---|
| `828087dba9be6d48` | `6564959c41` = **this PR's base, without my commits** | **`not an iterator`** ✘ |
| `599f6118e6be9c58` | base + these 2 commits | **`not an iterator`** ✘ — identical to base |
| `86de731c17a56ee5` | origin/main tip (**has #545**) | **DONE** ✔ |

Removing my commits does not change the outcome. The defect is **pre-existing on the base** and **already fixed on main**, almost certainly by #545 — `not an iterator` is raised by the specialized iterator helpers `range_iter_has_next` / `range_iter_next_or_null` (`pyre/pyre-interpreter/src/runtime_ops.rs:1395`, `:1468`), which reject anything that is not a range/long-range/seq iterator, and #545 is precisely the range FOR_ITER-next specialization with **polymorphic demotion**.

Independent confirmation from a parallel investigation: for the only loop whose trace differs, the per-CodeObject jitcode dump is **byte-identical between the two binaries** (same md5, 596/596 lines), and the `abort_permanent` census is identical (zero hits on both). These commits do not touch that loop at all. My earlier theory that removing `abort_permanent` made the branch compile more, and thereby expose a latent bug, was also refuted by that census.

### Next

Rebasing this branch onto current `origin/main` to pick up #545, then re-verifying. The CI failure on this PR should resolve with the rebase; it reflects the base predating the fix, not this patch.

Apologies for the noise — the earlier hold was posted on a confounded measurement.

— *commented by Claude*

youknowone:
CI failure diagnosis for the latest run (29510274730):

**`cargo test` (ubuntu + macos) — real bug, fixed on this branch by `f1d9ba36d1`.** The flaky `index out of bounds: the len is 23 but the index is 23` panics inside `indexmap` (hitting different tests each run: `argument::tests::raise_type_error_with_function`, `baseobjspace::tests::object_functionstr_scalar_fallback` on CI; various `pyre-jit eval::tests` locally) are one race: `init_typeobjects` ran its five post-init `patch_*` passes *outside* the `TYPEOBJECT_CACHE` `get_or_init` closure on every call, so a second test thread returning from the already-initialized `OnceLock` re-entered the passes while the first thread was still inserting `object.__class__` (the 24th entry) into the 23-entry object type dict — a concurrent IndexMap read/write. The fix wraps the passes in `std::sync::Once` (they cannot move inside the closure because `patch_typeobject_descriptor_names` reads `TYPEOBJECT_CACHE.get()`), plus an 8-thread regression test. Local stress: ~1 failure per 5–10 suite runs before, 0/30 after.

Note this race predates the branch — `main`'s CI has been flaking with the same panic since 7/15, so the `cargo test` legs there should also stabilize once this lands.

**`pyre/check.py` (ubuntu) — perf-gate flakes, no action.** `int_loop 0.40s > pypy 0.18×2` and `fib_recursive 2.98s > cpython 1.31×2` are the known shared-runner load flakes (int_loop is already on record as such); correctness was 207/209 with only those two timing gates failing.

— *commented by Claude*

PR #565, opened 2026-07-15
#346: suffix the Discriminant read's classdef root per instantiation

youknowone:
## What

`Rvalue::Discriminant` keyed the synthetic `__discriminant` FieldRead's `owner_root` off `tyref_adt_name_path` (the bare template name, e.g. `Option`), while the constructor, receiver-type and field-read sites suffix reference-payload enum roots via `adt_head_instantiation_suffix` (`Option<Result<*mut PyObject,PyError>>`). The two spellings minted distinct `ClassDef`s, so a `match` reading the tag and the constructor that wrote it landed on different classes and failed to union — surfacing as `setbinding: new value does not contain old` PANICs and `SomeInstance` union failures.

This routes the Discriminant `owner_root` through a new `tyref_adt_class_root` helper (which shares the `Deduplicated`/`HashConsedValue` unwrap via the extracted `tyref_adt_body`) so it carries the same per-instantiation suffix the other three projection sites use. The `owner_id` (layout `StructId`) stays on the bare `strip_instantiation_suffix` path so the tag layout is still shared across instantiations.

## Why

Part of gh#346 (retire the rtyper legacy walker). The census `GENERIC-ADT / UNION` cluster is the single largest rtyper-stage root cause; this is the first enum slice — aligning the fourth projection site (discriminant) that was still emitting the bare spelling.

The suffix machinery is the RPython-orthodox container-style per-site model (RPython forbids class specialization; `Option`/`Result` are modelled like built-in containers whose payload repr is per-site). This change makes the discriminant read consistent with that existing model rather than adding new machinery.

## Verification

- **census**: `typing_intrinsic_1` / `typing_intrinsic_2` fully lift (stable across two builds); `PREPASS phaseB` 8→6 (two `setbinding` PANICs cleared). The `getdict`/`getdict_backing` flicker is a pre-existing `SomeInstance.setattr` union nondeterminism, unrelated.
- **check.py**: 186/186/185 (dynasm/cranelift/wasm). The sole failure is `synth/list_insert_pop_index`, a pre-existing **BASEFAIL** (cpython vs pypy reference output differ, before any pyre backend runs) — independent of this change.

Design scoping for the rest of the epic is in `majit/majit-translate/docs/design-346-generic-adt-classdef-consistency.md`.

— *commented by Claude*

### DIFF
--- a/majit/majit-translate/docs/design-346-generic-adt-classdef-consistency.md
+++ b/majit/majit-translate/docs/design-346-generic-adt-classdef-consistency.md
@@ -1,0 +1,172 @@
+# Design: #346 generic-ADT ClassDef consistency (C-cluster)
+
+**Status:** scoping / design. Not yet implemented.
+**Branch:** `rtyper-legacy` (fixed per project rule).
+**Arbiter:** census set-diff (`PREPASS histogram phaseA`) + `python3 ./pyre/check.py` (bit-exact, all 3 backends).
+
+## 1. What this epic closes
+
+gh#346 retires the rtyper legacy walker (`cutover::is_known_unported`, cutover.rs:1165, must match
+nothing). Census on merged main (`38173e3cde6`) = **277 PREPASS-fail paths**. The census's own orthodox
+histogram buckets the single largest *rtyper-stage* root cause as the generic-ADT collapse:
+
+| bucket | count | note |
+|---|---|---|
+| UNION-PAIR-PORT | 44 | `setattr(field)` heterogeneous union, single collapsed classdef field |
+| GENERIC-ADT-SPECIALIZE | 9 | `cannot unify instances with no common base` / `don't know how to convert` |
+| UNION-ERROR (generic-ADT phi / pair) | 4 | mergeinputargs + `setbinding` monotonicity PANIC |
+
+= **57 directly bucketed**, but the actual reach is wider: **58 distinct roots** touch a
+UnionError/setbinding/no-common-base failure (store_global/store_name, the whole unary/binop
+descroperation family, call_function, setattr_str, w_dict_new, Range build, PyError::new). A large share
+of the 28 `BLOCKED-BLOCK` (downstream dead-end) paths are the hydra tails of these same union failures,
+so the domino payoff plausibly exceeds 57.
+
+## 2. Root cause (source-verified, three-agent cross-check)
+
+### 2a. The failure is INCONSISTENT splitting, not absent splitting
+
+pyre already implements *partial* per-instantiation ClassDef splitting (retrofit under #100/#312/#346).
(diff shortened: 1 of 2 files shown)


youknowone:
CI is green except **`pyre/check.py (macos-latest)`**, which failed solely on a **performance threshold**, not correctness:

```
nested_loop   dynasm SLOWER  pyre 0.73s > pypy 0.35s x2   (0.03s over the x2 gate)
              cranelift PASS 0.43s
```

Evidence this is the known macOS `nested_loop` perf flake, not a regression from this diff:

- **Correctness passes everywhere**: `pyre/check.py` ubuntu-24.04 ✅ and windows-latest ✅; macOS itself reports `187 passed` (outputs correct); `cargo test` on all three OSes ✅; CPython suite gate ✅; wasm + sandbox e2e ✅.
- The only failure is a **timing** comparison (`SLOWER`), macOS-only, dynasm-only, exceeding the x2 gate by 0.03s. The same benchmark passes on cranelift from the same tree.
- This diff is purely annotation-stage (front-end `Rvalue::Discriminant` classdef-root spelling); it does not touch runtime codegen, so it cannot affect `nested_loop` runtime.
- This matches the previously-recorded marginal macOS `nested_loop` flake (~0.71s > 0.35s x2) seen on prior PRs under loaded runners.

A re-run of the macOS job should clear it. Flagging rather than force-merging — happy to re-run or investigate further if preferred.

— *commented by Claude*

youknowone:
Re-run reproduced the identical macOS-only failure (`nested_loop` dynasm 0.71s > 0.35s x2), confirming it is the **known pre-existing macOS perf flake**, not a regression. Local measurement settles it:

```
nested_loop, target/release/pyre-dynasm, 5 runs:
  0.343s  0.311s  0.314s  0.303s  0.293s
  output 39996000100000000 (correct every run)
```

All well under the 0.70s gate (pypy 0.35 x2) — less than half. The CI 0.71–0.73s is a marginal overshoot on a loaded macOS runner.

Why this diff cannot cause it:
- `pyre/bench/nested_loop.py` is a pure nested-int `while` loop (`s += i*j`); it has no `Option`/`Result`/enum/`classdef` surface, so the front-end `Rvalue::Discriminant` classdef-root spelling change in this PR does not touch its trace.
- Correctness passes everywhere: `check.py` ubuntu ✅ + windows ✅ (macOS itself: 187 passed), `cargo test` ×3 ✅, CPython gate ✅, Codex parity ✅, wasm ✅, sandbox e2e ✅.
- This same `nested_loop` macOS x2 marginal flake was triaged as non-blocking on prior rtyper-legacy PRs (#445, #446).

Not blocking on this. Deferring to you on whether to merge as-is.

— *commented by Claude*

PR #573, opened 2026-07-15
jit: make MaJIT portal translation driver-generic

XiNiHa:
## Summary

Replace MaJIT’s Pyre-specific opcode-arm synthesis with explicit JIT drivers
and ordinary portal-reachable graph discovery.

Pyre now selects the exact `eval::eval_loop_jit` portal. MaJIT compiles that
portal and its callees through the canonical `make_jitcodes` pipeline, without
synthetic `__opcode_dispatch__` graphs or a runtime opcode-arm side table.

## Self-review

## Summary by CodeRabbit

* **New Features**
  * JIT compilation now uses explicitly configured driver portals and compiles the reachable code graph into canonical JIT artifacts.
  * Runtime loading now uses dedicated JIT driver metadata for reliable portal resolution.

* **Bug Fixes**
  * LLBC extraction now handles bootstrap dependencies more reliably and ensures normal builds regenerate complete production artifacts.
  * Source deletions are correctly reflected in extraction fingerprints.

* **Documentation**
  * Updated build, translation, extraction, and runtime documentation to describe the revised JIT workflow and artifact generation.


scripts/llbc_extract.py
```diff
@@ -252,7 +252,14 @@ def source_fingerprint(eng: Engine, crates: list[str], cargo_features: str) -> s
...
         digest.update(b"\0")
-        digest.update((eng.root / path).read_bytes())
+        full_path = eng.root / path
+        if full_path.is_file():
+            digest.update(full_path.read_bytes())
+        else:
+            # `git ls-files` includes tracked paths deleted in the working
+            # tree. A deletion is part of the source state and must change the
+            # fingerprint instead of making extraction unusable until commit.
+            digest.update(b"<deleted>")
```
youknowone, line 262: 👍

pyre/README.md
```diff
@@ -97,7 +97,13 @@ pyre is a structural port of PyPy's interpreter (`pypy/interpreter/` and `pypy/o
 
 ## Key differences from PyPy
```
youknowone, line 99: 👍

pyre/check.py
```diff
@@ -601,10 +601,9 @@ def build_backend(self, backend):
                 print(red("LLBC artefacts are missing under build/llbc/."))
                 print("Run the extractor first, then re-run this script:")
                 # No crate arguments: extract the full DEFAULT_CRATES set
-                # (pyre-object, pyre-interpreter, pyre-jit). Naming only the
-                # mandatory pair would omit pyre-jit.ullbc, leaving check.py to
-                # validate the degraded 2-crate frontend without the
-                # eval_loop_jit portal.
+                # (pyre-object, pyre-interpreter, pyre-jit). The exact
+                # eval::eval_loop_jit portal lives in pyre-jit.ullbc, so a
+                # production JIT build requires all three artifacts.
```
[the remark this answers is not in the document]
XiNiHa: unrelated to the change

youknowone (approved):
Thank you so much! This is looking very important change for generalized MaJIT



auto-generated review
-----

  ## 1. Regressions versus main

  None found.

  The replacement now follows upstream’s CodeWriter.make_jitcodes() shape: register explicit JitDrivers, discover the portal-reachable closure, allocate portal JitCodes,
  drain pending graphs, then use the single allocation-ordered JitCode store. This corresponds to rpython/jit/codewriter/codewriter.py:74-89 and call.py:145-150.

  ## 2. Other mismatches introduced by this patch

  None found.

  In particular, deleting the synthetic instruction-match-arm JitCodes is correct: PyPy discovers ordinary graphs from JitDriver portal graphs; it does not fabricate a
  JitCode per branch of a language-level opcode match.

  ## 3. Mismatches that already existed before this patch

  - The runtime still supports exactly one portal despite the translator now modelling multiple JitDrivers. pyre/pyre-jit-trace/src/jitcode_runtime.rs:137 asserts
    drivers.len() == 1; upstream permits one mainjitcode per JitDriverStaticData. This was already effectively true on main, which scanned for at most one marked portal
    JitCode. A multi-driver runtime needs a driver-qualified portal lookup rather than portal_jitcode().

  ## 4. Structural adaptations

  - Rust uses an exact qualified CallPath where RPython has graph/function-pointer identity. The patch’s explicit portal resolution is the appropriate equivalent and
    avoids the previous leaf-name heuristic.

  - LLBC is a frozen extracted input, and build.rs runs in a separate process. The new extraction-mode placeholder artifacts and runtime address patching are necessary
    Rust/build-process adaptations; RPython’s translated binary does not cross that serialization/ASLR boundary.

  - Serialized JitCode/descriptor/instruction tables are required by the build-time/runtime split. Upstream retains these objects directly in translated static data.
  - The build-time virtualizable-info factory remains intentionally empty until runtime, because field offsets are runtime Rust facts. This pre-dates the patch and is a
    Rust layering adaptation.

  Validation: cargo check -p majit-translate passed; git diff --check origin/main was clean.

PR #577, opened 2026-07-15
jit: compile bridges for guard failures inside an inlined callee

youknowone:
## Problem

A hot loop that inlines a callee whose body branches never compiled a bridge for the rare arm — the guard deopted to the blackhole on every failure, leaving the JIT barely ahead of the interpreter.

Probe (3M-iteration loop over a branching helper):

```python
def helper(i):
    if i % 7 == 0:
        return i * 2
    return i + 1

def main():
    s = 0; i = 0
    while i < 3000000:
        s = s + helper(i)
        i = i + 1
    print(s)
main()
```

`bridges_compiled=0, loops_aborted=2142, guard_failures=428544` — 3.75s against 9.06s for the interpreter.

This is **not** the multi-frame epic (gh#343): `n_recipes=1`, and the depth-1 reconstruction machinery already works. It was three bounded defects in the guard-reconstruction path, each masking the next.

## Changes

**`setup_reconstructed_callee_frame`** — the reconstructed callee's local boxes are the paused root's OpRefs, whose trace concrete can read as a stale NULL, while the value captured at guard failure lives in `concrete_r`. Restamp each local from its capture, skipping Void/NULL captures the way `setup_bridge_sym` already skips them when seeding the root frame.

**`getarrayitem_vable_via_metainterp`** — an indexed read hands back the *stored* box, so stamping the read result writes onto a box that may already carry a live concrete. A frame whose array slot is an unmaterialized hole reads NULL while the real Ref is still in the guard's register file; stamping that NULL clobbered the live value and folded a downstream may-force residual's Ref arg to NULL (`MayForceNullRefArgUnsupported`). Skip **only** that clobber — a NULL read onto a box with no live Ref still stamps, so slots that genuinely hold NULL keep it.

**`jit_merge_point`** — the loop-callee `CALL_ASSEMBLER` branch used a non-empty framestack as its "inside a sub-walk" test. A bridge's outer-frame continuation runs the trace root forward with a frame still on the stack and reaches the loop it exited, so that test also matched a bridge which must close with a JUMP, routing it to a `CALL_ASSEMBLER` request no caller consumes (`OuterNonTerminate`). Compare the header's frame against the trace root instead.

All three are required; dropping any one restores `bridges_compiled=0`.

## Results

Probe, both backends — output identical to `PYRE_NO_JIT=1`:

| | before | after |
|---|---|---|
| bridges_compiled | 0 | 1 |
| loops_aborted | 2142 | 1 |
| guard_failures | 428544 | 201 |
| wall clock | 3.75s | 0.04s |

Bench corpus JIT stats are **byte-identical to baseline on all 13 benches** (measured as an explicit A/B against a pristine build, isolated from other load), including `nbody` 5/5/0/1286, `fannkuch` 5/40/0/8202 and `spectral_norm` 3/2/0/441.

`check.py --backend dynasm,cranelift`: **dynasm 189/189, cranelift 189/189**.

## Note for reviewers

An earlier iteration of the middle change used a blanket "never stamp a NULL read" rule. It unlocked the same probe but regressed `nbody` ~8x (0.26s → 2.17s, `loops_aborted` 0 → 5): the may-force check fires only on `Some(Value::Ref(GcRef(0)))`, so a blanket skip left the box *unknown* and merely dodged the safety check, while also suppressing genuine NULL stamps other traces depend on. The narrow `clobbers_live_ref` rule here fixes the actual defect and leaves every genuine NULL stamp intact. Worth knowing that `check.py` cannot see that class of regression — its `nbody` gate is 10s and the regressed build ran 2.17s green on both backends.

### DIFF
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch.rs
@@ -4857,7 +4857,21 @@ fn getarrayitem_vable_via_metainterp(
     // Mirror `getfield_vable_via_metainterp`: stamp the read result's
     // concrete so `concrete_of_opref(result)` honors the Box.value
     // contract for downstream consumers; `Value::Void` = no live concrete.
-    if !matches!(shadow_value, Value::Void) {
+    //
+    // An indexed read hands back the STORED box, so the stamp is a write onto a
+    // box that may already carry a live concrete. A frame whose array slot is
+    // still an unmaterialized hole reads NULL while the real Ref lives in the
+    // guard's register file, so stamping that NULL would clobber the live value
+    // (last write wins) and fold a downstream may-force residual's Ref arg to
+    // NULL -> `MayForceNullRefArgUnsupported`. Skip ONLY that clobber: a NULL
+    // read onto a box with no live Ref still stamps, keeping NULL a real
+    // concrete for the slots that genuinely hold one.
+    let clobbers_live_ref = matches!(shadow_value, Value::Ref(majit_ir::GcRef(0)))
+        && matches!(
+            ctx.trace_ctx.box_value(result),
+            Some(Value::Ref(majit_ir::GcRef(addr))) if addr != 0
+        );
+    if !matches!(shadow_value, Value::Void) && !clobbers_live_ref {
         ctx.trace_ctx.set_opref_concrete(result, shadow_value);
     }
     let dst = code[op.pc + 7] as usize;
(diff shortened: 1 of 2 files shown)


youknowone:
## Correction: this change is inert on `main`

The performance win in the description is **not real for `main`**, and I want that on the record rather than discovered later.

**What happened.** I verified this branch on base `38d1a23e5c`. While the session ran, `main` advanced by four commits, among them #538 (`9d79802222`), which flipped `p2_drain_enabled()` to default-ON. That routes the branchy-inlined-callee shape to `drive_bridge_carrier_walk` instead of the `drive_bridge_framestack_walk` this PR improves. The squash merge was textually clean and **semantically inert**. All three sites are still present in `HEAD`; they are simply never reached.

**Measured on the merge commit `6552f3cd55`** (both backends, machine idle), probe = 3M-iteration loop calling a helper with an `i % 7 == 0` arm:

| config | bridges | aborts | guard failures | time |
|---|---|---|---|---|
| `main` default (drain ON) | **0** | 2142 | 428544 | **11.43s** |
| `PYRE_P2_DRAIN=0` (this PR’s path) | **1** | 1 | 201 | **0.04s** |

`brdg=0 / abrt=2142 / gf=428544` is the exact pre-fix baseline.

**The path cannot simply be re-enabled.** `trace.rs` says the framestack-walk bridge is "correctness-buggy for a branchy inlined-callee continuation", and #538’s own regression guards confirm it — with `PYRE_P2_DRAIN=0` they miscompile on both backends:

| config | `inline_multiframe_module_branch_deopt` (ref `7215477652`) | `inline_multiframe_drain_journaled_store` (ref `7215477652 120000`) | probe bridge |
|---|---|---|---|
| default | OK | OK | none |
| `PYRE_P2_DRAIN=0` | **7215474836** | **7215474836** | yes |
| `PYRE_P2_COMPILE=1` | OK | **`120299` ≠ `120000`** | none (`RunPerfnWalkNone`) |

No configuration is both correct and fast. **This PR does not fix the underlying defect** — it only makes an already-disabled buggy path faster.

**Why this was not caught.** The A/B was honest but its base went stale mid-session; the corpus A/B and 189/189×2 were run against `38d1a23e5c`, not the merge commit. The lesson is to re-measure the probe on the merge commit, not the branch tip.

**Kept, not reverted**, per maintainer decision: the three sites were bisected as all-required and are correct for that path, so they are worth preserving for whoever lands the orthodox reconstruction. Tracking issue and an acceptance test to follow.

— *commented by Claude*

PR #632, opened 2026-07-17
jit: fix stepped-slice panic and nested-break miscompile

youknowone:
Two JIT correctness fixes, each with a synth regression bench.

## `newslice` HLOp had no lowering arm
`BUILD_SLICE` with a step operand records a `newslice(start, stop, step)` HLOp
that had no arm in the flatten lowering dispatch, so it reached the SSARepr
assembler un-lowered and panicked with `unimplemented opname "newslice"`
whenever a stepped slice (e.g. `lst[::-1]`) sat in a JIT-compiled loop.

Added `lower_newslice_hlop_to_insn`, lowering the three-ref HLOp to
`residual_call_ir_r(ConstInt(build_slice_fn_idx), ListI([3]), ListR([start,
stop, step]), Descr)` against the existing `bh_build_slice_fn` helper. The
lowering flavor is `Plain`, matching the bind site. Synth bench
`newslice_step_hot.py`.

## Nested-loop secondary-edge `break` mis-maps the outer iterator
A nested inner `for` loop `break` (`POP_TOP` popping the inner iterator then a
`JUMP_BACKWARD` to the enclosing `FOR_ITER`) miscompiles when the break lands on
the secondary edge of its inner-loop guard: the break edge becomes a guard
side-exit whose bridge resume snapshot mis-maps the outer iterator slot, so the
resumed `FOR_ITER` dereferences a non-iterator (SIGSEGV, or `TypeError: not an
iterator`).

Detect the two hazardous guard shapes (a `POP_JUMP_IF_TRUE` whose true edge
continues the loop, and a `POP_JUMP_IF_FALSE` that jumps forward to the break)
and route the frame through the interpreter via a new
`UnsupportedJitShape::NestedBreakBridgeResume`. The plain `if cond: break` idiom
(a `POP_JUMP_IF_FALSE` whose fall-through is the break — the primary edge) still
JIT-compiles. Synth bench `nested_break_not_hot.py`.

## Verification
- Both stepped-slice and nested-break repros clear on dynasm and cranelift.
- Discriminator: hazardous nested-break shape declines (`loops_compiled=0`, no
  panic) while the plain-break control still compiles (`loops_compiled=2`).
- `pyre/check.py`: dynasm 219/219, cranelift 219/219, wasm 218/218 — all passed.

### DIFF
--- a/pyre/bench/synth/nested_break_not_hot.py
+++ b/pyre/bench/synth/nested_break_not_hot.py
@@ -1,0 +1,52 @@
+# A nested `for` loop whose inner `break` lands on the secondary edge of its
+# guard — reached from `if not cond: break` (a POP_JUMP_IF_TRUE fall-through) or
+# from a break-check that is the last statement in the loop body (a
+# POP_JUMP_IF_FALSE that jumps forward to the break) — must reconstruct the
+# operand stack correctly when the inner iterator is popped and control resumes
+# at the outer FOR_ITER. Both the break-check-first and accumulate-first forms,
+# plus a compound `and` guard, are exercised. Output is verified against
+# CPython/PyPy.
+N = 9000
+
+
+def break_check_first(n):
+    t = 0
+    for i in range(n):
+        for j in range(2, 4):
+            if not i % 3 != 0:
+                break
+            t += j
+    return t
+
+
+def accumulate_first(n):
+    t = 0
+    for i in range(n):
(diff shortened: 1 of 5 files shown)


youknowone:
Superseded by #633 (merged), which lands the same five hunt-2 JIT correctness fixes. Closing this duplicate. — *commented by Claude*

PR #667, opened 2026-07-19
interpreter: align stdlib state and expose public ASTs

youknowone:
## Summary

- keep builtin names out of fresh module globals and rely on the live `__builtins__` fallback, matching PyPy module namespaces
- consolidate class construction on the active execution context and remove the duplicate class-specific TLS slot
- clamp negative sequence-iterator pickle state to zero and validate `chain.from_iterable` arity
- implement `compile(..., PyCF_ONLY_AST)` as a compiler-native Ruff AST to public `_ast` heap-object conversion
- define AST fields, location attributes, constructors, f-string nodes, pattern-matching nodes, and AST identity for `PyCF_ONLY_AST`

## Root cause and PyPy parity

The `_ast` module previously exposed empty type stubs and `compile` ignored `PyCF_ONLY_AST`, so `ast.parse()` returned a code object. PyPy parses to its native compiler AST and then calls `Node.to_object(space)` at the public boundary. This change adds that same boundary conversion while continuing to use the pinned RustPython compiler API; no RustPython patch or local RustPython dependency is needed.

The earlier commits also follow PyPy in keeping builtins behind module `__builtins__`, clamping sequence iterator state, and using the active execution context for class construction.

AST-object-to-bytecode conversion remains separate work; this PR covers source-to-public-AST conversion and the `PyCF_ONLY_AST` identity case.

## Validation

- `target/debug/pyre pyre/extra_tests/snippets/builtin_compile.py`
- targeted `ast.parse` coverage for functions, comparisons, imports, comprehensions, async/await, f-strings, and structural pattern matching
- `traceback` and `asyncio` import smoke tests
- `python3 pyre/check.py --backend dynasm --no-synthetic` (all 10 benchmarks plus loop-reentry passed; 11/11)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,6 +2175,7 @@ dependencies = [
  "rustpython-compiler-core",
  "rustpython-host_env",
  "rustpython-literal",
+ "rustpython-ruff_text_size",
  "rustpython-sre_engine",
  "rustpython-unicode",
  "rustpython-wtf8",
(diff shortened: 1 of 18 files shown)


youknowone:
Retiring this stacked PR. Its patch is already represented on main; ongoing unmerged stdlib compatibility work is consolidated in #682.

PR #669, opened 2026-07-19
interpreter: implement PyPy-shaped FileIO basics

youknowone:
## Summary

- expose `_io.FileIO` as a working `_RawIOBase` subclass backed by pyre's existing path/fd file operations
- restore the PyPy IO base hierarchy and provide IOBase close/context-manager/isatty behavior
- add a minimal delegating `BufferedReader` surface used by the standard library
- make path-backed seek/tell and closed-file checks match FileIO behavior
- honor `closefd=False` for externally owned file descriptors
- replace the `_io` setup HashMap with direct hierarchy construction
- replace the file wrapper's TLS type cache with one process-global `OnceLock`, preserving cross-thread type identity

## Root cause and PyPy parity

`_io.FileIO` and the buffered classes were empty heap-type stubs, while `builtins.open` already had usable path/fd read and write operations on a private wrapper. The standard `io` module could import, but direct `FileIO(...)` construction failed before any I/O.

PyPy models `FileIO` as a `_RawIOBase` instance with per-object mode/fd/closed state, and its IO base classes form a real inheritance chain. This change connects pyre's existing per-instance file state to that public hierarchy and removes the unrelated per-thread type singleton. No RustPython patch or local RustPython dependency is used.

## Validation

- `target/debug/pyre pyre/extra_tests/snippets/stdlib_io.py`
- related `stdlib_io_bytesio.py`, `stdlib_io_stringio.py`, `stdlib_subprocess.py`, and `builtin_open.py`
- `python3 pyre/check.py --backend dynasm --no-synthetic` (all 10 benchmarks plus loop-reentry passed; 11/11)

### DIFF
--- a/pyre/extra_tests/snippets/stdlib_io.py
+++ b/pyre/extra_tests/snippets/stdlib_io.py
@@ -4,7 +4,16 @@ from io import BufferedReader, BytesIO, FileIO, RawIOBase, StringIO, TextIOWrapp
 from testutils import assert_raises
 
 fi = FileIO("README.md")
+assert isinstance(fi, RawIOBase)
+assert issubclass(FileIO, RawIOBase)
+assert fi.mode == "rb"
+assert fi.readable()
+assert not fi.writable()
 assert fi.seekable()
+assert fi.tell() == 0
+assert fi.seek(2) == 2
+assert fi.tell() == 2
+assert fi.seek(0) == 0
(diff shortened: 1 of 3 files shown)


youknowone:
Retiring this stacked PR. Its patch is already represented on main; ongoing unmerged stdlib compatibility work is consolidated in #682.

Issue #171, opened 2026-06-13
Trace the real list builtin callee — retire #143 name-based arms, dm143 advance, and synthetic-callee snapshot (depends on #73)


youknowone:
## tl;dr

The `#143` fix (PR #159, merged) makes the JIT handle `list.append` / `list.pop` / `list.reverse` in hot loops by **trace-time special-casing in the opcode tracer** rather than by tracing the real list callee the way RPython/PyPy does. Three pyre-only mechanisms implement that workaround:

1. **Name-based builtin-method arms** in the opcode tracer (`call_callable_value`).
2. **`dm143` live-frame advance** — an after-the-fact "if the heap was mutated during trace, copy live locals into the frame after `CloseLoop`" side channel.
3. **Synthetic-callee guard snapshot** + a **synthetic runtime-helper jitcode** for the `append` resize path (a null-`W_CodeObject` jitcode with an identity `pc_map`).

These are correct on every tested case (52/52 both backends; all `#143` benches match CPython exactly), but they are structural divergences from PyPy's meta-tracing model. PyPy reaches `list.append`/`pop` through the **translated list implementation + rtyper helpers**, traced via **oopspec**, so there is no name-based special-casing, no "advance live locals" fallback, and no fabricated callee frame. This issue tracks converging onto that model and deleting the three mechanisms.

**This does not block anything.** `#143`/`#159` is merged and correct. This is parity/architecture debt, surfaced by a reviewer note and confirmed against the code.

## Background — why the workaround exists

Pyre's production tracer is a hand-written **CPython-bytecode walker**. The Rust list helpers (`jit_list_append`, `jit_list_pop_at`, `jit_list_reverse`, …) are `#[dont_look_inside]` (`pyre/pyre-object/src/listobject.rs:1123-1156`), so they have **no jitcode** for the tracer to step into. Faced with `xs.append(i)` in a loop, the tracer cannot trace the real callee, so it hand-emits a folded fast path instead. The `#143` saga (resize-guard resume into a NULL callable, double-mutation, indexed-pop length overlay) is the cost of making that hand-emitted path correct without a real callee.

### 1. Name-based builtin-method arms
`pyre/pyre-jit-trace/src/trace_opcode.rs`:
- `call_callable_value` (~`:6158`) dispatches on `canonical_list_method("append"|"pop"|"reverse")` (10 sites) for both the bound-method `W_Method` branch (~`:6230`) and the unbound builtin-form branch (~`:6320-6440`).
[...]

PR #675, opened 2026-07-20
jit: fix seven differential-fuzz miscompiles/crashes (DELETE_FAST, unary int, neg-slice del, nested FOR_ITER, bignum-through-local, yield-from)

youknowone:
## Summary

JIT differential fuzz-hunt on `main`: run each program under the JIT and under `PYRE_NO_JIT=1`
on the same binary (objspace-immune — any output difference is unambiguously a JIT bug), then fix
each divergence. This branch closes seven miscompiles/crashes across both backends (dynasm and
cranelift).

## Fixes (one commit each)

| Commit | Symptom fixed |
|---|---|
| `Fix conditional DELETE_FAST walker merge` | conditional `del x` then reuse: walker merge produced an orphan flatten block (`make_return got N` panic / miscompile) |
| `close the LOAD_FAST_CHECK unbound-local bailout block` | `del x; …; del x` / unbound-local bailout left an exit-less block, panicking in flatten |
| `carry forwarded unary int inputs across loops` | `-s` / `abs(s)` of a loop-carried int froze the value at its compile-in point; large-magnitude unary was low-bits wrong. Forwarded unary int reads were missing from the short-preamble ownership contract |
| `Fix omitted slice step lowering` | `del seq[-3:]` (a negative-bound slice with an omitted step) crashed / deleted the wrong count — the omitted slice step used a null-ref sentinel instead of the `None` singleton, so the slice was never materialized before `DELETE_SUBSCR` |
| `Fix nested FOR_ITER bridge scratch seeding` | reading an outer local after an inner `for j in range(m)` loop returned a stale/garbage value (crash on dynasm, garbage int on cranelift) |
| `preserve local operands for binary overflow resume` | an integer `*`/`+` whose operand is loaded from a local (recursive-call result, tuple unpack) and overflows to a bignum was miscomputed; inlining or non-overflowing values were correct |
| `decline tracing before yield-from delegation` | `yield from <range>` in a hot-loop-consumed generator dropped a bounded number of delegated values at the compile-in transition; declines tracing before SEND delegation can mutate a shared iterator |

## Verification

- Each fix was reproduced (JIT vs `PYRE_NO_JIT` divergence) and confirmed resolved on **both**
  backends, with a full differential re-sweep after each commit showing no new regressions.
- Guard/effect-touching fixes (unary forwarding, binary-overflow resume) were A/B perf-checked: the
  affected hot loops still JIT-compile (unary loop ~34×, bignum loop ~3.7× over the interpreter).
- `pyre/check.py` green on dynasm and cranelift.

## Out of scope (deeper, tracked separately)

The hunt also surfaced three deep items left for follow-up:
- a `list += tuple` abort/replay double-count whose real root is a **nested-loop forward-resume**
  inconsistency (the decline settles the outer walk from an inconsistent resume point), not the
  in-place journaling — needs forward-resume/journal-lifetime rework;
- a select-kept-stack abort variant;
- a pre-existing **wasm32 bignum interpreter** defect (wasm JIT == wasm NOJIT, both wrong for
  bignum arithmetic — not a JIT-lowering bug).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1565,6 +1565,7 @@ dependencies = [
  "smallvec",
  "wasm-bindgen",
  "wasm-encoder 0.245.1",
+ "wasmi",
  "wasmparser 0.225.0",
 ]
 
--- a/majit/majit-backend-wasm/Cargo.toml
+++ b/majit/majit-backend-wasm/Cargo.toml
@@ -34,3 +34,4 @@ wasm-bindgen = { workspace = true, optional = true }
 majit-ir = { workspace = true, features = ["test-support"] }
 wasmparser = { workspace = true }
 smallvec = { workspace = true }
+wasmi = "1.1"
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -4478,6 +4478,36 @@ fn find_label_args(ops: &[Op], jump: &Op) -> Vec<OpRef> {
     Vec::new()
 }
 
+/// A legacy pool-indexed const that is absent from the constants pool at emit
+/// time is an optimizer-seeding invariant violation — panic loudly, matching
+/// `collect_constants_from_ops`' `missing_legacy_const`, instead of emitting a
+/// silent `0`. On native a null Ref traps on the first dereference; wasm's
+/// offset 0 is valid linear memory, so a silent `0` is read as garbage and
+/// miscompiles quietly rather than crashing.
+#[cold]
+#[inline(never)]
+fn missing_emit_const(opref: OpRef) -> ! {
+    panic!(
+        "wasm emit_resolve: legacy pool-indexed const OpRef (raw={}) is absent \
+         from the constants pool — the optimizer producer must seed it (or mint \
+         an inline Const) instead of emitting a silent 0.",
+        opref.raw()
+    );
+}
+
+/// Resolve a constant operand's i64 bits: the inline `Const` value if the
+/// variant carries one (`history.py:227/268/314`), else the legacy pool entry.
+/// A pool miss panics via [`missing_emit_const`] rather than falling back to a
+/// silent `0`.
+fn resolve_const_bits(constants: &indexmap::IndexMap<u32, i64>, opref: OpRef) -> i64 {
+    opref.inline_const_bits().unwrap_or_else(|| {
+        constants
+            .get(&opref.raw())
+            .copied()
+            .unwrap_or_else(|| missing_emit_const(opref))
+    })
+}
+
 fn emit_resolve(
     sink: &mut InstructionSink<'_>,
     constants: &indexmap::IndexMap<u32, i64>,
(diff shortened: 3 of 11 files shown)


youknowone:
Pushed the rebase onto `origin/main` plus four wasm-backend commits. The headline one root-causes the wasm-only `synth/sre_pattern_methods` wrong answer.

### Root cause

Constant folding and the short preamble can leave a folded constant under its original op position: a plain (non-`Const`) `OpRef` whose only definition is a **constants-pool entry keyed by `raw()`**, with no producing op left in the trace.

dynasm already handles this — `RegisterManager::loc` takes a constants-map fallback once it finds no register and no frame binding, and its comment says so verbatim. The wasm backend had no counterpart: `emit_resolve` consults the pool only when `opref.is_constant()`, otherwise emitting `local.get(1 + raw)`. **A never-written wasm local is zero-initialized**, so the folded constant silently became `0`.

On the `re._parser.parse_template` trace:

```
60:NewArrayClear(v83)->v60            # v83 undefined; constants[83] == 1  (capacity)
61:SetarrayitemGc(v60,v84,v110)->v85  # v84 undefined; constants[84] == 0  (index)
```

A length-0 array was allocated and then written at index 0, with `W_ListObject.length` already `1` — an out-of-bounds store into whatever the nursery hands out next. Native receives the same trace as `NewArrayClear(Int(1))` / `SetarrayitemGc(v60, Int(0), ...)`.

### Fix

`unbound_pool_const_seeds` collects the value ids read as a plain `OpRef` that have no producing op or inputarg and do have a pool entry, and binds each in the function prologue ahead of every block, so the binding dominates the body including a resume-at-LABEL entry. Binding in the prologue rather than patching `emit_resolve` covers every reader at once — operands, label/jump args and guard fail-arg spills all go through `local.get(1 + raw)`. An id with neither a producer nor a pool entry now declines the trace instead of emitting a zero, matching the unhandled-opcode arm's policy; it never fires in the suite.

The #171 orthodox append fold stays fully enabled — nothing is declined or disabled to get this green.

The other three commits are hardening picked up along the way: missing pool-indexed consts, three silent-wrong-value codegen arms, and missing layout descrs now fail loud instead of emitting a wrong value.

### Verification

| gate | result |
|---|---|
| dynasm | 244/244 |
| cranelift | 244/244 |
| wasm | 242 passed, `sre_pattern_methods` green |

### Known remaining wasm failure

`synth/delete_negative_open_slice_hot` is still wrong on wasm (242980541 vs 242981142). A control build with the new seeding disabled produces the identical wrong value, and the trace has no undefined operands at all, so it is a different defect — not a regression from this change. Narrowed: only `del items[:k]` and `list.clear()` reproduce (`del items[-3:]`, `pop(0)` and `items[:-2] = []` are all correct); both native backends are correct; the loss saturates at 4 iterations regardless of N, and the skipped `i` values land on JIT compile/retrace events, so it is a lost-iteration defect rather than a value miscompile.

— *commented by Claude*

PR #677, opened 2026-07-20
interpreter: implement marshal module

youknowone:
## Summary

- replace the empty `marshal` builtin stub with a native module implementing `dump`, `dumps`, `load`, and `loads`
- map PyPy-style `W_Root` object dispatch onto the shared RustPython compiler-core marshal wire codec
- support None, booleans, integers and big integers, floats, complex values, strings, bytes-like values, tuples, lists, dicts, sets, frozensets, slices, StopIteration, Ellipsis, and code objects
- preserve ordinary shared references with a moving-GC-safe shadow-stack-backed VecMap
- enforce recursive `allow_code=False` checks and expose the shared format version
- expand the marshal stdlib snippet from 11 to 14 tests, including file APIs, shared identity, bigint/complex/bytes, and nested code rejection

No RustPython2 source change was needed.

## Root cause

`marshal` was registered as an empty builtin module, so stdlib consumers imported it successfully but every API access failed with `AttributeError`. The new module registers the actual implementation and reuses the compiler's authoritative code-object marshal format instead of inventing a pyre-only encoding.

## Validation

- `PYRE_RTYPER_VERBOSE=1 cargo build --release -p pyre-jit-trace`
- `target/debug/pyre pyre/extra_tests/snippets/stdlib_marshal.py` — 14/14 passed
- CPython 3.14 → pyre primitive/container wire compatibility
- pyre → CPython 3.14 primitive/container wire compatibility

## Follow-up

Recursive container decoding remains unsupported by the shared compiler-core `MarshalBag` construction API: a self-reference is resolved before the container value can fill its reserved reference slot. This is also a RustPython-side limitation and is being reported separately rather than patched locally in pyre.

### DIFF
--- a/pyre/extra_tests/snippets/stdlib_marshal.py
+++ b/pyre/extra_tests/snippets/stdlib_marshal.py
@@ -66,6 +66,28 @@ class MarshalTests(unittest.TestCase):
             bytearray(b"\x01\x02"),
         )
 
+    def test_marshal_additional_atoms(self):
+        self._test_marshal(None)
+        self._test_marshal(Ellipsis)
+        self._test_marshal(StopIteration)
+        self._test_marshal(2**100)
+        self._test_marshal(-(2**100))
+        self._test_marshal(1 + 2j)
+        self._test_marshal(b"marshal")
+
+    def test_shared_reference(self):
+        shared = [1, 2]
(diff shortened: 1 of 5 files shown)


youknowone:
Consolidated into #682, which now targets main and contains the complete unmerged stdlib compatibility series.

PR #681, opened 2026-07-20
struct: preserve format encoding errors

youknowone:
## Summary
- encode text struct formats using strict ASCII
- preserve bytes format values without lossy UTF-8 conversion
- keep `UnicodeEncodeError` for non-ASCII text distinct from `struct.error` for invalid byte formats

## Root cause
The previous backslash-replacement and lossy conversion changed both format bytes and exception semantics before parsing.

## Validation
- `target/release/pyre-dynasm extra_tests/snippets/stdlib_struct.py`
- `python3 scripts/extract-llbc.py`
- `PYRE_RTYPER_VERBOSE=1 cargo build --release -p pyre-jit-trace`

No RustPython2 patch is required.


youknowone:
Consolidated into #682, which now targets main and contains the complete unmerged stdlib compatibility series.

PR #683, opened 2026-07-20
jit: heal nested for-range degenerate loop; virtualize range GET_ITER/FOR_ITER

youknowone:
## Summary

wasm-jit correctness fixes. The first three commits fix a hang/miscompile of nested
`for _ in range(N): for x in range(P): total += x`, which stormed (a 0-progress infinite
bridge) for any OUTER ≳ 50. The fourth is an independent wasm self-recursive
`CALL_ASSEMBLER` exception-unwind fix.

**virtualize range FOR_ITER items with guarded exhaustion**
Replaces the range-iterator pointer mask with a snapshot-backed continuation guard so a locally consumed item stays a removable virtual `New`. Takes the exhaustion edge *before* the continuation guard (returns the loop-exit null for the trailing `GuardNonnull`), so a nested inner loop run to completion does not record a `GuardTrue` against a concretely-false condition.

**heal nested for-range degenerate loop; virtualize range GET_ITER**
- `optimizeopt/unroll.rs`: `assemble_peeled_trace_with_jump_args` no longer filters base label args through the backend-constants map. `import_state` already drops `LEVEL_CONSTANT` virtual-state slots at `make_inputargs` time; re-deriving the slot set post-hoc drops a still-live loop-carried box that a phase-2 guard postprocess const-forwarded (the exhaust guard's `remaining`), desyncing the label from the positionally-aligned jump and orphaning the head guard's operand.
- `jitcode_dispatch`: `FbwWalkMode.bridge_entry_merge_pc` carries a guard-failure bridge's own resume python-pc; the `jit_merge_point` arm skips exactly the first crossing that lands on it. A guard's resume coordinate lies past the dispatch-top `jit_merge_point` (`generate_guard resumepc=orgpc`), so an RPython bridge never re-crosses the loop-header merge point at position zero; the walker resumes at the opcode boundary and would otherwise close a 0-op no-op bridge → storm.
- range `GET_ITER` on an exact `range` emits the virtual `W_IntRangeIterator` allocation shape (new `GetIter` helper kind, `W_Range` field descriptors/offsets, `is_exact_w_range`).

Net effect: a Loop + healing Bridge sharing one target token, matching PyPy's structure for this nested loop.

**register _collections deque iterators as GC roots**
Adds `_deque_iterator` / `_deque_reverse_iterator` to the immortal `register_pyre_class_offsets` block so their managed `deque` child is traced.

**route CALL_ASSEMBLER callee ExitFrameWithException exits through the deopt helper**
`FailDescr::is_finish()` is true for *both* `DoneWithThisFrame` and `ExitFrameWithException`
FINISH exits. The self-recursive `CALL_ASSEMBLER` arm picked its "clean callee finish"
`fail_index` (`bridge_finish_fi` in codegen, `loop_finish_fi` in the published dispatch
metadata) as the first `is_finish` guard, which could be an `ExitFrameWithException`. The
guest arm then short-circuited that exit to output slot 0, banking the raised exception
object as the recursive-call return value (surfaced as `int + <exc>`). Native has no CA arm
(it calls the compiled portal directly), so its callee exception flows through the normal
portal-return → `GUARD_NO_EXCEPTION` channel and never hits this `is_finish`-classified
fail-index set — the miscompile was wasm-only.

Fix: exclude `ExitFrameWithException` from both finish-index selectors
(`meta_descr_is_exit_frame_with_exception`) so an exception finish routes to
`wasm_ca_resume_deopt`; a new `FinishedException` outcome there publishes the exception via
`store_jit_exception` and returns garbage so the caller's `GUARD_NO_EXCEPTION` fires — parity
with the outer Finished arm and `handle_blackhole_result`'s `ExitFrameWithExceptionRef` arm.
Only exception finishes route to the host, so the #501 CA fast-path for non-raising recursion
is preserved. Fixes `synth/selfrec_tail_exception_unwind` (was `int + ZeroDivisionError`
TypeError, now 4250).

## Verification
- `nfr.py` sweep 25/25 (primary `10 200 → 9000`, regressions `v1_single_for`/`nfr_const`/`nfr_var`, P×OUTER sweep {3,7,10,11,13,17,100}×{200,700,2000})
- `check.py --synthetic-only`: dynasm 230/230, cranelift 230/230, wasm 230/230

## Summary by CodeRabbit

- **Performance**
  - Improved optimization of `range` iteration and iterator creation for faster execution.
  - Enhanced list-append and loop handling for more efficient traced code.

- **Bug Fixes**
  - Improved resumption after bridge guard failures to avoid redundant loop crossings.
  - Corrected exception propagation during compiled loop and function transitions.
  - Improved handling of loop completion and iterator exhaustion.

### DIFF
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -1367,7 +1367,9 @@ pub fn build_wasm_module(
     // home-slot writes overflow into the next arena slot.
     let bridge_finish_fi = guards
         .iter()
-        .find(|g| g.is_finish)
+        .find(|g| {
+            g.is_finish && !crate::failguard::meta_descr_is_exit_frame_with_exception(&g.meta_descr)
+        })
         .map(|g| g.fail_index)
         .unwrap_or(0);
     // CA frames execute the source loop and this bridge on the same frozen
--- a/majit/majit-backend-wasm/src/failguard.rs
+++ b/majit/majit-backend-wasm/src/failguard.rs
@@ -25,6 +25,21 @@ pub struct WasmFailDescr {
     pub meta_descr: Option<DescrRef>,
 }
 
+/// `compile.py:658-662` ExitFrameWithExceptionDescrRef parity: whether a FINISH
+/// exit is an ExitFrameWithException (the callee raised; slot 0 holds the
+/// exception) rather than a DoneWithThisFrame. `is_finish` alone is true for
+/// both, so the self-recursive CALL_ASSEMBLER arm must exclude the exception
+/// variant when it picks the "clean callee finish" `fail_index` — an
+/// ExitFrameWithException must route to `wasm_ca_resume_deopt`, which propagates
+/// the exception, not be short-circuited to its output slot.
+pub fn meta_descr_is_exit_frame_with_exception(meta_descr: &Option<DescrRef>) -> bool {
+    meta_descr
+        .as_ref()
+        .and_then(|d| d.as_fail_descr())
+        .map(|fd| fd.is_exit_frame_with_exception())
+        .unwrap_or(false)
+}
+
 impl Descr for WasmFailDescr {
     fn index(&self) -> u32 {
         self.fail_index
--- a/majit/majit-backend-wasm/src/lib.rs
+++ b/majit/majit-backend-wasm/src/lib.rs
@@ -1783,7 +1783,10 @@ impl majit_backend::Backend for WasmBackend {
             .fail_descrs
             .borrow()
             .iter()
-            .find(|descr| descr.is_finish)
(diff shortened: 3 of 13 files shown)


youknowone:
Pushed an additional independent commit (\`119c6405270\`): **route CALL_ASSEMBLER callee ExitFrameWithException exits through the deopt helper** — a wasm-only miscompile where a self-recursive `CALL_ASSEMBLER` callee that raises had its exception banked into the recursive-call return slot (`int + ZeroDivisionError` instead of the expected result), because `is_finish()` does not distinguish `DoneWithThisFrame` from `ExitFrameWithException`. Fixes `synth/selfrec_tail_exception_unwind`; 3-backend synthetic suite 230/230. Body updated to cover it.

— *commented by Claude*

PR #689, opened 2026-07-20
perf-exc: exception JIT perf (attr fold + loop-header dominance)

youknowone:
## Summary

Exception-handling JIT performance work, stacked on `perf-exc`.

### Commits

1. **guard inlined-op arg by ob_type + w_class** — a P1 rhs-guard passed a
   `W_TypeObject` to `GuardClass` (which needs a physical `ob_type`); a
   builtin-int RHS was then provably-false and the loop never compiled. Emit a
   physical `ob_type` `GuardClass` plus a conditional `w_class` `GuardValue`.

2. **decline str-override inline on nested call** — a str `__str__`-override
   inline admitted nested-call bodies, whose abort poisoned the whole loop. Add
   `exception_string_override_has_nested_call` so only the single-invoke body
   inlines.

3. **fold `__context__`/`__cause__` exception attr reads** — extend
   `exception_attr_slot_fold` so `LOAD_ATTR` of `__context__`/`__cause__` on an
   exception folds to a guarded field read (`GetfieldGcR`) instead of a residual
   getset-descriptor call. Stores decline (`descr_setcause` sets
   `suppress_context`; `descr_setcontext` type-validates). Removes ~3 residual
   calls/iter from raising-in-`except` loops.

4. **classify loop backedges by target dominance** — the loop-header predicate
   excluded a genuine loop header whenever a loop body contained a `try/except`:
   the loop's own backedge is laid out after the handler block, so the prior
   landing-pad heuristic never saw it, and same-frame raise-catch loops stayed
   uncompiled. Decide by dominance instead — build a CFG (fall-through, forward/
   backward jumps, exception-table edges); a backward jump is a loop backedge
   iff its target dominates its source. Validated against an iterative-dominator
   oracle across 8 try/except-in-loop shapes.

### Effect

Same-frame raise-catch loops now compile: `vbase_big` @3M JIT **0.04s** vs
interp **12.86s** (was JIT slower than interp); `context_chain` compiles. This
also unblocks commit 3's benefit (a set `__context__` requires raise-in-`except`
= the previously non-compiling shape).

### Verification

`pyre/check.py` **241/241** on both backends (dynasm + cranelift). Two new
`codewriter` unit tests (`loop_with_try_except_body_keeps_its_loop_header`,
`handler_rejoin_backward_jump_is_not_a_loop_header`).

— *commented by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly handle `__context__` and `__cause__` exception attributes, preserving expected setter side effects.
  * Prevent unsafe specialization for exception attribute stores and avoid incorrect folding during execution.
  * Improved JIT loop detection in the presence of nested try/except handlers.
  * Avoid inlining exception-string overrides when they contain nested calls.
* **Performance**
  * Faster exception attribute reads and broader safe `for`-loop tracing when try/except appears in the loop body.
  * More targeted operator inlining with improved type/guard handling.

### DIFF
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -8464,6 +8464,8 @@ pub fn object_setattr(obj: PyObjectRef, name: &str, value: PyObjectRef) -> PyRes
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum ExceptionAttrSlot {
     Args,
+    Context,
+    Cause,
     Code,
     Errno,
     Strerror,
@@ -8508,6 +8510,8 @@ pub unsafe fn exception_attr_slot_fold(
     let kind = unsafe { pyre_object::w_exception_get_kind(obj) };
     let slot = match name {
         "args" => ExceptionAttrSlot::Args,
+        "__context__" => ExceptionAttrSlot::Context,
+        "__cause__" => ExceptionAttrSlot::Cause,
         "code" if kind == pyre_object::interp_exceptions::ExcKind::SystemExit => {
             ExceptionAttrSlot::Code
         }
@@ -8549,6 +8553,18 @@ pub unsafe fn exception_attr_slot_fold(
         }
         _ => return None,
     };
+    // Only loads of `__context__`/`__cause__` fold to a raw slot read.
+    // `descr_setcontext` (`interp_exceptions.py:183-190`) type-validates and
+    // `descr_setcause` (`:166-174`) also flips `suppress_context` to True, so a
+    // direct slot write would drop those effects.
+    if is_store
+        && matches!(
+            slot,
+            ExceptionAttrSlot::Context | ExceptionAttrSlot::Cause
+        )
+    {
+        return None;
+    }
     let w_type = crate::typedef::r#type(obj)?;
     // Any hit precedes the hard-coded exception arm.  The version-tag guard
     // below pins this miss across later heap-subclass mutations.
@@ -8564,6 +8580,10 @@ pub unsafe fn exception_attr_slot_fold(
             ExceptionAttrSlot::Args => {
                 pyre_object::interp_exceptions::w_exception_get_args_storage(obj)
             }
+            ExceptionAttrSlot::Context => {
+                pyre_object::interp_exceptions::w_exception_get_context(obj)
+            }
+            ExceptionAttrSlot::Cause => pyre_object::interp_exceptions::w_exception_get_cause(obj),
             ExceptionAttrSlot::Code => pyre_object::interp_exceptions::w_exception_get_code(obj),
             ExceptionAttrSlot::Errno => pyre_object::interp_exceptions::w_exception_get_errno(obj),
             ExceptionAttrSlot::Strerror => {
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -8557,12 +8557,7 @@ pub unsafe fn exception_attr_slot_fold(
     // `descr_setcontext` (`interp_exceptions.py:183-190`) type-validates and
     // `descr_setcause` (`:166-174`) also flips `suppress_context` to True, so a
     // direct slot write would drop those effects.
-    if is_store
-        && matches!(
-            slot,
-            ExceptionAttrSlot::Context | ExceptionAttrSlot::Cause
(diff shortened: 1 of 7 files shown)


youknowone:
## Rebased onto `origin/main` (#692) + a correctness fix

Force-pushed `0747916e5c3` → `7620f6fd27b`, rebased onto `e7fd9b8f852` (#692), no conflicts.

### New commit: `jit: decline FOR_ITER frames with a raising named handler`

Fuzzing same-frame exception shapes surfaced a **miscompile introduced by this branch's own `jit: allow exception opcodes in FOR_ITER loop bodies`**, which CI did not catch:

```python
for i in range(N):
    for j in range(M):
        try:
            try:
                if j > 2: raise TypeError('x')
            except TypeError as e:
                raise e            # named reraise
        except TypeError:
            acc += 9
```

Before that commit the frame was declined (interpreted, correct). After it, the frame compiles and the result is **short by exactly one handler visit per trace abort** (measured: `loops_aborted=54`, deficit `486 = 54 × 9`); some variants ran to a 60s timeout. A/B confirmed it: commenting the five added opcodes out restores `loops_compiled=0` and the correct answer.

**Mechanism.** `except ... as name:` emits an exceptional cleanup tail `STORE_FAST name; DELETE_FAST name; RERAISE 1`, reached only when the handler body itself raises. That tail lowers to a `last_exc_value` jitcode op which a bridge walk cannot answer, so the walk aborts with `LastExcValueWithoutActiveException` (abort census is 1:1 with `loops_aborted`, and scales linearly with iteration count — 13/53/213 at N=8k/32k/128k). With a `FOR_ITER` in the frame an abort can land while an item is in flight, and that iteration is dropped. A bare `raise` emits no such tail and never aborts — the clean discriminator.

**Fix.** A frame-level gate beside `for_iter_frame_is_finally_duplicated`, firing only when a handler binds a name *and* a `RAISE_VARARGS` lies between that binding and the cleanup that clears it. A body-range predicate cannot work: for-loop handlers are emitted in the cold section after the loop, outside the scanned `FOR_ITER` body range. The cleanup tail alone is not a usable signal either — it is emitted for *every* `except as e:`, raising or not.

Verified: 6/6 previously-miscompiling programs correct; **19/21** for-loop exception shapes still compile (the only declines are the bug shape and a pre-existing unrelated one); 3 new unit tests.

### CI failure triage — none of the three is caused by this branch

| job | verdict | evidence |
|---|---|---|
| `cargo test (ubuntu-24.04)` | pre-existing on `main` | `mapdict.rs` `cache_attrs` is a `RefCell` on process-global interned map nodes; libtest runs each test on its own thread, so two threads writing attributes on the same global type race. Reproduced locally at `--test-threads=32` (6/40). `main`'s own run fails the same way across 10 tests. No commit here touches `mapdict.rs`, and `pyre-interpreter` does not even link `pyre-jit`/`pyre-jit-trace`. |
| `pyre/check.py (ubuntu-24.04)` | pre-existing on `main` | `FAIL wasm synth/delete_negative_open_slice_hot wrong output`, byte-identical (`1 failed, 239 passed`) on the **merge-base commit this PR sits on**. dynasm/cranelift are 241/241 in the same job. |
| `pyre/check.py (windows-latest)` | timer quantization, not a regression | check.py subtracts a measured per-interpreter startup; Windows quantizes it to 15.625 ms. `raise_catch` raw times are identical on `main` and here (pypy 0.19s / dynasm 0.23s) yet the ratio reads 1.0x vs 2.4x purely from that subtraction. `raise_catch_loop.py` is a `while` loop with no `as e` and no `FOR_ITER`, so the new gate cannot touch it. The merge-base run fails this job too. |

### Known follow-ups (not in this PR)

- The same shape in a **`while`** frame stays correct but is **Θ(N²)** — 0.98s/4.05s/16.21s at N=200/400/800 vs a linear interpreter. The compiled trace leaves `ExecutionContext.sys_exc_value` stale after the handler exits, so the `__context__` chain grows without bound and 92.5% of runtime lands in `_break_context_cycle`.
- Root cause of the abort itself: `reads_last_exc_before_next_catch` is a flat byte-stream scan that does not follow `goto/L` and does not stop at a block terminator, so it can read a physically-adjacent unrelated block's `catch_exception` and clear an exception the real successor still needs. Fixing that would remove the aborts outright and make this gate unnecessary.
- `LIST_EXTEND` is missing from the `FOR_ITER` body allow-list, so a ≥3-element constant list literal or any `[*a]` declines the whole frame (unrelated to exceptions).

— *commented by Claude*

PR #692, opened 2026-07-21
jit: attribute + cut the #394 driver residual — hold PyFrame ptr across eval_loop_jit steps, re-seed at collection points

### LINKED ISSUE #394: JIT: compile_loop pipeline is ~7.6x slower than PyPy per-function warmup — optimizer + codegen dominates
> **STATUS UPDATE (2026-07-20).** This issue is CLOSED. The body below described the state at close (2026-07-09) and named two remaining structural levers; the **Progress since close** ledger at the bottom tracks what has happened since, with a **2026-07-20 addendum** that resolves the last-named prerequisite (the per-pass optimizer profile) and records that every warmup lever this issue names is now either shipped, at its orthodox floor, or source-refuted as non-orthodox. A **2026-07-26 addendum** closes the final open item — the ~170 μs driver residual — via merged PRs #692/#698. Outdated inline claims are ~~struck through~~. Stale bench counts (`187/187`) are now `225/225`.

## Summary

Per-function JIT warmup overhead in pyre is ~7.6x higher than PyPy (1.12 ms vs 0.15 ms per function). The gap decomposes into two roughly equal halves:

1. **Interpreter speed (~50% of warmup)**: reaching the 1039-iteration trace threshold costs ~550 μs in pyre vs ~36 μs in PyPy, because pyre's interpreter runs this loop at ~531 ns/iter vs PyPy's ~35 ns/iter (**~15x per-iteration gap**, measured with `PYRE_JIT=0` vs `pypy3 --jit off`).
2. **Trace+compile pipeline (~50% of warmup)**: ~570 μs in pyre vs ~112 μs in PyPy (**~5x gap**). Within pyre's pipeline, `compile_loop` (optimizer + backend codegen) dominates at ~340 μs (~60% of the pipeline half), the walk/tracer is ~58 μs (~10%), and driver/bookkeeping residual is ~170 μs (~30%).

## Measurement

Benchmark: 30 distinct code objects, each a `while n < 5000: total += n + K; n += 1` loop. First pass = warmup, second pass = steady.

| | Pyre | PyPy | Ratio |
|---|---|---|---|
| Per-function warmup overhead | 1.12 ms | 0.15 ms | **7.6x** |
| Interpreter per-iteration (JIT off) | 531 ns | 34.7 ns | **15.3x** |

## Cost Decomposition (Pyre, per function)

Instrumented timing (`Instant::now` around `full_body_walk_trace` and `compile_loop`), cross-checked against the measured total:

| Phase | First function | Steady (2–30 avg) | Share of total warmup |
|---|---|---|---|
| Pre-threshold interpretation (1039 iters × ~531 ns) | ~550 μs | ~550 μs | **~49%** |
| **compile_loop** (optimizer + codegen) | 3536 μs | ~260–340 μs | **~27%** |
| Driver residual (warm-counter ticks, CRN re-entry, guard-exit blackhole, bookkeeping) | — | ~170 μs | ~15% |
| walk (run_perfn_walk) | 889 μs | 43 μs | ~5% |
| setup (init_symbolic etc.) | 85 μs | 2 μs | <1% |

Arithmetic check: 550 + ~300 + 170 + 58 ≈ 1080–1120 μs ✓ (matches the measured 1.12 ms/fn). The `compile_loop` steady figure ranges ~260 μs (inner `compile_loop_body`, per the decomposition below) to ~340 μs (outer `compile_loop`, including `clear_trace_session` and measurement noise); the decomposition table below uses the inner ~260 μs breakdown.

The first function additionally pays a one-time JIT infrastructure cost (~3.5 ms first compile, ~950 μs first walk).

Notes:

- `MAJIT_STATS` reports `loops_aborted=30` for this run, but that is bookkeeping: `abort_trace(false)` in the post-successful-close cleanup bumps `loops_aborted` via `aborted_tracing(Generic)` (pyjitpl.rs `abort_trace`). All 30 walks ended in `CloseLoop` and all 30 compiles succeeded — **each function traces exactly once**; there is no wasted abort/re-trace cycle in this benchmark.
- `guard_failures=60` = one compiled-loop exit guard (`n < 5000` fails) per pass per function — normal loop exits, not deopts.

### Half 1 — interpreter speed (separate axis, not a JIT-pipeline problem)

The threshold cost is `threshold × interp_ns_per_iter`. The threshold (1039, rlib/jit.py:588 parity) is identical to PyPy's, so the entire 550 μs vs 36 μs difference is the ~15x interpreter per-iteration gap. Fixing this means making the interpreter faster (dispatch, boxing, refcount/GC-barrier costs per op) — but it bounds warmup latency no matter how fast the compile pipeline gets.

#### UPDATE — a large, self-contained slice of Half 1 found and fixed (~3.2x on the numeric loop)

Profiling the pre-threshold interpreter loop (macOS `sample`) put `str::from_utf8` at the top of the leaderboard, and iterative re-profiling after each fix uncovered three independent per-iteration costs, all now removed. Benchmark: `while i < N: total += i + 1; i += 1`, `PYRE_JIT=0`, 3M iters, macOS aarch64. Result byte-identical (`4500001500000`) throughout.

| stage | ns/iter | commit |
|---|---|---|
| baseline | ~700–808 | — |
| + numeric binop dunder short-circuit | ~287–300 | `4fa11c8398` |
| + in-place (`+=`) special short-circuit | ~240–257 | `043b6fa17f` |
| + cached `npure_cellvars` | ~227–229 | `91645a1546` |

Cumulative **~3.2x**. The three costs:

1. **Numeric binop dunder lookup.** Every `int + int` (and `- * // % / ** << >> & | ^`, unary) ran `needs_numeric_binop_dispatch` → `numeric_operand_overrides` → `dunder_overridden` → `lookup_where_with_method_cache(type, "__add__")`, which **interns the dunder name via `box_str_constant(Wtf8::new(name))`** (`baseobjspace.rs:6127`, a UTF-8 validation) up to **4× per binop**, even when both operands are exact `int`/`float`. The comparison path (`i < n`) already guarded this with `is_exact_builtin_instance` at the top of `subclass_special_override` (`baseobjspace.rs:739`); `is_true` too. The numeric binop/unary path was simply missing that established guard — added it.

2. **In-place (`+=`) special lookup.** `total += …` routes through `try_inplace_special`, which ran `lookup_type_special(lhs, "__iadd__")` — the same string-keyed probe — for every augmented numeric assignment. int/long/float/complex/bool define no in-place special, so gated on `is_exact_numeric_builtin` (exact builtin + numeric storage). Builtin sequences (list/bytearray *do* define `__iadd__`/`__imul__`) and subclasses fall through unchanged.

3. **`npure_cellvars` recompute.** `PyFrame::ncells()` — called on every `pop_value`/`peek_at` (twice per binop) — recomputed `npure_cellvars`, an O(cellvars × varnames) overlap count. It is code-invariant, so cached as a `u32` on the `PyCode` wrapper (computed once at construction).

Parity verified against py3.14 and JIT-on/off for: exact int/float/bigint/bool, mixed int+float, overriding int/float subclasses (forward/reflected/unary), `list`/`bytearray` `+=`/`*=` (identity-preserving), `__iadd__`-overriding subclasses, and a closure loop (non-empty cellvars). `check.py` dynasm 187/187 + cranelift 187/187 green at each stage.

This does not close the residual ~15x interpreter gap. Re-profiling after these fixes shows the remaining per-iteration cost is now pure opcode dispatch (`execute_opcode_step`, ~~`decode_instruction_for_dispatch` re-decoding each instruction~~ [see update], `eval_loop`) and int-boxing `malloc` — both of which are larger structural changes (a predecoded instruction stream; a small-int cache with GC/identity implications) rather than missing fast-path guards. But these three fixes remove the dominant per-iteration costs the profile identified, and together they are the biggest warmup lever found in this investigation — larger than anything inside `compile_loop`.

> **Update (2026-07-19):** the per-step `decode_instruction_for_dispatch` (backward-scan + forward-replay each step) has since been replaced on the hot path by a forward-only single-pass `decode_instruction_forward` (miframe, PR #548). This addresses the *re-decode* half of Lever 1; a fully predecoded instruction stream is still not built. int-boxing `malloc` (Lever 2) was attempted via small-int tagging and reverted as non-orthodox; a 2026-07-19 source review then **closed it as having no orthodox interpreter-level path** (pyre's interpreter already parities PyPy's per-result `malloc`; box-elision is JIT-trace-only and pyre already has it). See **Progress since close**.

### Half 2 — trace+compile pipeline (~5x vs PyPy)

1. **PyPy's optimizer is natively translated** RPython code → C-level speed with translate-time specialization. Pyre's optimizer is ordinary Rust.
2. **Allocation-heavy compile pipeline**: each trace compile allocates fresh `Vec`s for IR, optimizer state, register allocation, and backend emit. PyPy reuses preallocated/arena structures.
3. **Backend codegen**: pyre's dynasm backend is a straightforward emitter; PyPy's x86 backend is heavily tuned.
4. **Per-trace pass chain**: every `compile_loop` runs the full optimizer chain (simplify → heap → virtualizable → rewrite → unroll) fresh.
5. **Driver residual (~170 μs)** is un-attributed: warm-counter ticking on ~1039 back-edges, ContinueRunningNormally re-entry, guard-exit blackhole decode, and warmstate bookkeeping. Needs finer instrumentation before optimizing.

## compile_loop internal decomposition (measured)

Instrumented `compile_loop_body` sub-phases and the `UnrollOptimizer` two-phase split (`PYRE_WARMUP_DIAG`, steady-state avg over the 30-function benchmark):

| compile_loop sub-phase | Steady time | Share of compile_loop |
|---|---|---|
| prep (recorder→trace, deep-clone, snapshot map build) | ~5 μs | ~2% |
| **optimize (UnrollOptimizer)** | **~205 μs** | **~78%** |
| ├ Phase 1 (`optimize_preamble`) | ~78–97 μs | ~40% |
| ├ Phase 2 (`optimize_peeled_loop`) | ~65–78 μs | ~35% |
| └ setup/handoff (consts+ops clone, TraceIterator, remap, ExportedState root) | ~13–25 μs | ~10% (see UPDATE below — earlier "~57 μs" was noise) |
| post_opt (vectorize gate, inputarg normalize) | ~8 μs | ~3% |
| backend codegen (dynasm) | ~50 μs | ~19% |
| **compile_loop total** | **~260 μs** | 100% |

**UPDATE (refined by finer instrumentation — the earlier "inter-phase ~57 μs is wasteful bookkeeping" reading was WRONG).** Re-measuring the optimize region with per-segment `Instant`s (dropping the first cold call) decomposes the ~205 μs `optimize` as:

| optimize segment | Steady time | Share of optimize |
|---|---|---|
| Phase 1 `run_optimize_from_inputs` | ~97 μs | ~48% |
| Phase 2 `optimize_with_constants_and_inputs_at` | ~78 μs | ~39% |
| Phase 2 setup (consts clone + `ops` deep-clone + `TraceIterator` build/drain + snapshot remap + the snapshot `.clone()`s) | **~13 μs** | ~6% |
| Phase-1-done → Phase-2-setup gap (ExportedState build + `root_all_gcrefs` + phase1_out backup) | remainder | ~7% |

**The two optimizer passes are ~87% of `optimize`; the snapshot/clone bookkeeping is only ~13 μs total, not ~57 μs.** The earlier ~57 μs "inter-phase" figure was measurement noise between coarse instrumentation points, not real clone cost. This benchmark's traces are small (a single `while` loop), so the snapshot vectors have few elements and their clones are cheap.

Empirical confirmation: converting the six Phase 2 `self.snapshot_*` / `phase1_emit_ops` / `call_pure_results` `.clone()`s to `std::mem::take` moves (verified GC-safe by three adversarial lenses + the InvalidLoop `simple_opt` move precedent) produced **no measurable warmup change** (optimize ~224 μs before vs ~232 μs after — within ±10 μs noise). The move is retained as a strictly-cheaper, safer form, but it is **not** a warmup lever. `self.snapshot_*` is confirmed not read after the Phase 2 move point (in-function or by the caller).

### High ROI within compile_loop (~260 μs) — REVISED
The real cost is the **two optimizer passes themselves (~175 μs / 87% of optimize)**, not setup/bookkeeping. Clone-elimination is a dead end here.

- **Profile the pass pipeline per-pass** (simplify → heap → virtualize → rewrite → force) inside `run_optimize_from_inputs` to find the dominant pass before touching it. This is the prerequisite investigation for any real win — the passes run twice (preamble + peeled loop), so a per-pass hot spot is paid twice.
- **Arena/pool allocator** for pass-internal Vecs (OpRef lists, type arrays, box vectors) to cut per-trace malloc/free — but only after the per-pass profile shows allocation is the bottleneck (it may instead be dominated by the pass's own logic, e.g. heap-cache lookups or virtual-state matching).
- **Backend codegen (~50 μs)**: per-op overhead in the dynasm emitter (register encoding, label resolution) — smaller than the passes but self-contained.
- ~~Kill the inter-phase clone/remap~~ — **REFUTED as a lever** (measured ~13 μs, move produced no warmup change). The clone→move change landed anyway for hygiene, not speed.

### Medium ROI
- **Attribute the ~170 μs driver residual** (finer instrumentation), then trim the biggest item.
- **Interpreter per-iteration cost** — biggest single lever for warmup latency overall (~49%). Three slices of it (numeric-binop dunder lookup, in-place `+=` lookup, `npure_cellvars` recompute) are now fixed for a cumulative ~3.2x on the numeric loop (see the Half 1 UPDATE above). The residual gap (opcode dispatch/redecode, int-boxing malloc, GC-barrier) remains interpreter-perf work.

### Not worth it
- Walk/trace path optimization — ~5% of warmup, already fast (43 μs steady).
- Threshold tuning — the threshold matches PyPy (1039); lowering it would trade steady-state quality for warmup and diverge from upstream.
- prep / post_opt sub-phases — ~13 μs combined, negligible.

## Reproduction

```python
# /tmp/warmup_many.py
import time

N_FUNCS = 30
N_ITERS = 5000

src = []
for i in range(N_FUNCS):
    src.append(f"""
def fn_{i}():
    total = 0
    n = 0
    while n < {N_ITERS}:
        total += n + {i}
        n += 1
    return total
""")

fns = []
for s in src:
    ns = {}
    exec(s, ns)
    fns.append(list(ns.values())[-1])

t0 = time.perf_counter()
for f in fns:
    f()
t1 = time.perf_counter()
for f in fns:
    f()
t2 = time.perf_counter()

first_total = (t1 - t0) * 1000
second_total = (t2 - t1) * 1000
per_fn_warmup = first_total / N_FUNCS
per_fn_steady = second_total / N_FUNCS
print(f"First pass (warmup):  {first_total:.1f} ms total, {per_fn_warmup:.3f} ms/fn")
print(f"Second pass (steady): {second_total:.1f} ms total, {per_fn_steady:.3f} ms/fn")
print(f"Warmup overhead/fn:   {per_fn_warmup - per_fn_steady:.3f} ms")
```

```bash
MAJIT_STATS=1 cargo run --release --features dynasm -- /tmp/warmup_many.py
# Compare with: pypy3 /tmp/warmup_many.py
# Interpreter per-iteration: PYRE_JIT=0 vs pypy3 --jit off on a 1M-iteration loop
```

— *created by Claude*

---

## Progress since close (2026-07-19)

The issue closed COMPLETED on 2026-07-09 after the Half-1 investigation, but the two **structural levers** it named were left open. Ledger of what has since happened:

### ✅ Achieved (landed on `origin/main`)
- **Half-1 fast-path slices** (numeric binop dunder short-circuit, in-place `+=` special short-circuit, cached `npure_cellvars`) — cumulative ~3.2x on the numeric interpreter loop. Present on main (the numeric guards live in `objspace/descroperation.rs` after a rebase; `npure_cellvars` cached on the `PyCode` wrapper).

### 🟡 Partially achieved (on `miframe`, unmerged — PR #548)
- **Lever 1 — opcode dispatch re-decode.** The per-step `decode_instruction_for_dispatch` (backward-scan to the logical instruction start + forward `OpArgState` replay, every step) is replaced on the three hot dispatch callers by a forward-only single-pass `decode_instruction_forward`. This removes the per-step backward scan/replay. **Not** a fully predecoded instruction stream — decode still runs per step, just O(1) on the common no-`EXTENDED_ARG` path. A follow-up fix backs the decoder up over an `EXTENDED_ARG` prefix so a JIT loop-header/resume coordinate (normalized through `skip_python_trivia_forward`, which names the real opcode past its prefix) recovers the full oparg — this was a real miscompile (`getwidth` FOR_ITER → `test_fnmatch`/`test_strtod`).

### ❌ Not achieved
- **Lever 2 — int-boxing `malloc`.** Attempted as small-int tagging (`CAN_BE_TAGGED`) on `miframe`, then **reverted as non-orthodox**: PyPy elides int allocation via *virtualization / escape analysis*, not tagging (`taggedpointers` defaults off; `W_IntObject` is a plain heap box, not an `UnboxedValue`). **Update (2026-07-19): there is no orthodox interpreter-level form of this lever.** Source review of `pypy/` + `rpython/` confirms the PyPy *interpreter* (JIT off) allocates a fresh `W_IntObject` per arithmetic result — `wrapint` (`objspace/std/intobject.py:904-905`) returns `W_IntObject(x)` under the default `withprebuiltint=False` (`config/pypyoption.py:207`; the prebuilt cache covers only `[-5,100)` and is off by default). Pyre's `w_int_new` (`pyre-object/src/intobject.rs:74-122`) is a line-by-line port with `WITHPREBUILTINT=false`, so pyre's interpreter is **already `malloc` bit-parity with PyPy's**. Box-elision exists only in the RPython *JIT trace optimizer* (`OptVirtualize.optimize_NEW_WITH_VTABLE`, `optimizeopt/virtualize.py:207`) — never in the interpreter — and pyre **already has** the equivalent trace pass (`majit/majit-metainterp/src/optimizeopt/virtualize.rs` `optimize_new_with_vtable`). So "teach the interpreter to elide short-lived int boxes" is a PyPy-absent optimization (the same class as the reverted tagging); the interpreter int-box cost is at its orthodox floor and this lever is **closed as having no orthodox path**, not "untouched."
- **Half-2 — trace+compile pipeline (~5x).** The decomposition (two optimizer passes = ~87% of `optimize`) was measured but **no pipeline optimization was implemented**: no per-pass profile of `run_optimize_from_inputs`, no arena/pool allocator, no backend-codegen trimming.
- **Driver residual (~170 μs)** — still un-attributed; no finer instrumentation landed.
- **Predecoded instruction stream** — not built (Lever 1 mitigated the re-decode cost but kept per-step decode).

### Net
Of the two structural levers the issue named, Lever 1 is mitigated (not fully realized) on an unmerged branch. Lever 2 is now **closed as having no orthodox interpreter-level path** (see the ❌ update above: pyre's interpreter already parities PyPy's per-result `malloc`, and the only box-elision either has is the JIT trace optimizer, which pyre already ships). The entire Half-2 pipeline work remains open. If a future warmup push is wanted, the remaining orthodox targets are: (a) a **per-pass profile of the optimizer chain** (`run_optimize_from_inputs`: simplify → heap → virtualize → rewrite → unroll — the two passes are ~87% of `optimize`), which the issue names as the prerequisite investigation for any Half-2 win and which is pure measurement (no orthodoxy judgment); and (b) attributing the ~170 μs driver residual. Interpreter warmup (Half-1) is otherwise at its orthodox floor.

---

## Addendum (2026-07-20) — Half-2 per-pass profile done; all named levers resolved

The prerequisite investigation the **Net** above named — a per-pass profile of the optimizer chain — was carried out (`miframe`, reverted-HEAD `65a72c64888`, macOS `/usr/bin/sample`). It resolved into a specific sub-target (short-preamble construction), which was then attacked, measured, and closed. Both the compile-pipeline (Half-2) sub-target and the steady-state execute path were source-checked against PyPy/RPython **before** any code, per the repo's verify-orthodoxy-first rule. Result: no remaining warmup lever has an orthodox path that also pays off.

### Half-2 optimizer sub-target — EXHAUSTED

- **Per-pass profile.** Inside `optimize`, short-preamble construction (`ShortBoxes::create_short_boxes` → `produced_ops` → `produce_arg`/`materialize_one` recursion) was the dominant optimizer self-time bucket (~40% of optimizer self-time on a compile-amplified bench; ~12% of `compile_loop`, ~3% of wall). virtualize / heap / rewrite / intbounds each measured <5%.
- **Root of the short-preamble cost.** pyre's `OpRef` is a *positional* index, so every `ShortBoxes` identity-map lookup calls `materialize_operand_at(opref)` (→ `find_producer_op`, up to 4 `FxHashMap` probes + `Rc` clone + `Operand` build) to convert `OpRef` → canonical `Operand`. RPython keys the same maps on box *objects* and pays zero per-lookup resolution (`shortpreamble.py:283-296`), resolving forwarding eagerly once at emit via in-place arg-slot rewrite (`optimizer.py:644,651-652 op.setarg`) — a mechanism pyre's positional `OpRef` cannot replicate.
- **Lever A — build-scoped `OpRef→Operand` memo (attempted, DROPPED).** A memo consulted only during the frozen `create_short_boxes` window (correct by construction; that window inserts nothing into the producer stores). Implemented, and correctness-green: `cargo test -p majit-metainterp` (1402 passed), `check.py` dynasm 225/225 + cranelift 225/225 **byte-identical**, `internal_compile_panics=0`, `loops_compiled` unchanged. But the interleaved best-of-N A/B (EPOCHREALTIME, loaded-machine rule) was **perf-negative**, not neutral: `−1.2%` on a 120-loop compile-heavy bench (base won 5/9) and `−3.2%` on a smaller-body variant (base won **8/9**, consistent direction). The memo saves the conversion but adds an `FxHashMap` probe + insert + `Operand` clone per call; at real trace sizes the profiled cost is dominated by the recursive walk + `Rc`/`SmallVec` churn *underneath* the conversion, not the conversion itself. Per the perf-neutral-kills-added-complexity rule, the three commits were reverted (`git reset --hard 65a72c64888`).
- **Lever B — cut the `produce_arg` walk depth / per-op alloc churn (REFUTED before code).** pyre's `produce_arg` (`shortpreamble.rs:793`) is line-for-line RPython (`shortpreamble.py:283-296`): the **same** `produced_short_boxes` + `boxes_in_production` memo guards (so the walk is already O(distinct-ops), not a re-walk bug), and the **same** per-op allocation (`arglist = [produce_arg(a) for a in getarglist]` → `copy_and_change` ≡ `.map(produce_arg).collect::<SmallVec>()` → `copy_and_change`). RPython allocates a fresh list + ResOp per produced op too. The deep (~120-frame) recursion on the compile-heavy bench is the source program's genuine linear dependency chain, not redundant work. Cutting it would deviate from RPython = non-orthodox. And the payoff ceiling is ~3% of wall regardless.
- **Verdict:** the only real pyre-vs-RPython delta in short-preamble is the `OpRef→Operand` conversion tax — which is exactly what Lever A targeted and lost on. No orthodox short-preamble-local lever remains; the Half-2 optimizer sub-target is exhausted.

### Steady-state execute path — AT PyPy PARITY (no lever)

Profiled the execute path (which is ~72% of wall on the compile-heavy bench, but that bench is execution-bound; it is the *steady-state* axis, not warmup). Net-of-startup best-of-5 vs `pypy3`:

| bench | pyre | pypy | ratio |
|---|---|---|---|
| int_loop (pure int) | 0.103 s | 0.097 s | **1.05x** (parity) |
| fib_loop (bignum) | 0.090 s | 0.102 s | **0.88x** (pyre faster) |
| nbody (float+list) | 0.212 s | 0.133 s | **1.46–1.60x** (only gap) |

Execute self-time is almost entirely **JIT-emitted machine code** (the profiler's `??? unknown binary` = dynasm output), not removable runtime calls — int_loop shows ~5600/5715 work-thread samples in JIT code with **zero** malloc. nbody's largest single frame is `pow` (libm, **43%**) from `x ** 0.5`; but both PyPy `_pow` (`floatobject.py:865-940`) and pyre `_pow` (`descroperation.rs:2989`) special-case only `y == 2.0 → x*x` and fall through to libm `pow` for `0.5` — so the pow cost is *shared* with PyPy, not a gap, and lowering `**0.5 → sqrt` would be a pyre-only, non-orthodox deviation. nbody's 1.46x residual is the float/list JIT-codegen constant-factor floor already scoped as near-floor by prior epics. No single orthodox lever.

### Lever 1 (opcode re-decode) — confirmed SHIPPED

`decode_instruction_forward` is present (`pyre-interpreter/src/pyopcode.rs`) with a full oracle + malformed-chain + cross-check test suite, and the hot dispatch callers are rewired (interpreter `eval_loop` and both JIT `eval_loop_jit` sites use it; the remaining `decode_instruction_at` callers are the intended cold ones). Landed in **PR #548** (`1ad9c56d625`), with the `EXTENDED_ARG` high-byte fix applied afterward. The forward-only single-pass decode is realized; a fully predecoded instruction stream is still not built (deliberately out of scope — it changes code-object representation).

### Remaining #394 tasks (all no-go under current orthodoxy, or out of scope)

Everything the issue named as a warmup lever is now resolved:

| Task | Status |
|---|---|
| Half-2 per-pass optimizer profile | ✅ **done** — short-preamble was the sub-target; both its levers (memo / walk-churn) closed above |
| Short-preamble compile-speed lever | ⛔ **no orthodox payoff** — memo dropped (perf-negative); walk/churn refuted (= RPython) |
| Steady-state execute path | ⛔ **at PyPy parity** (int 1.05x / fib 0.88x / nbody 1.46x constant-floor); no lever |
| Lever 1 — opcode re-decode | ✅ **shipped** (PR #548) |
| Lever 2 — int-boxing malloc | ⛔ **closed** — no orthodox interpreter-level path (already `malloc` bit-parity with PyPy) |
| Half-1 fast-path slices | ✅ **shipped** (on main) |
| Arena/pool allocator for pass-internal Vecs | ⚪ **out of scope** — was gated on "per-pass profile shows allocation is the bottleneck"; the profile instead shows the walk logic + the (RPython-equivalent) alloc, not pyre-excess allocation, so there is no orthodox arena lever here. A general per-compile arena is a large, separate structural change, not a #394 lever. |
| Driver residual (~170 μs) attribution | ✅ **done** (PR #692 + #698) — attributed to `majit_gc::shadow_stack::get` (the #1 non-idle warmup self-time frame) and cut; see the 2026-07-26 addendum below. |

**Net (2026-07-20):** no code change this session (the one attempted lever, the short-preamble memo, was reverted after an honest perf-negative A/B). The session's product is a source-confirmed *negative* result: the optimizer sub-target and the execute path have no orthodox lever, Lever 1 is already shipped, and Half-1 is at its orthodox floor. The single remaining open item with no orthodoxy obstacle is attributing the ~170 μs **driver residual** (measurement only).

— *addendum by Claude*

---

## Addendum (2026-07-26) — driver residual attributed + cut; all #394 warmup items now resolved

The single open item the 2026-07-20 Net named — the ~170 μs **driver residual** ("pure measurement, needs finer instrumentation") — was attributed and closed by two merged PRs.

### PR #692 — attribute + cut the driver residual (merged, `e7fd9b8f85`)

- **Attribution.** Warmup-dominated profiling (macOS `/usr/bin/sample`, many fresh compiles) put `majit_gc::shadow_stack::get` as the **#1 non-idle warmup self-time frame**, above every compile/optimizer frame. `FrameRoot::frame()` reloaded the `PyFrame` pointer from the thread-local `SHADOW_STACK` on every call, and `eval_loop_jit` (the JIT-mode loop that runs pre-threshold + during tracing = warmup) called it ~27× per opcode step.
- **Fix.** Hold the resolved `PyFrame` pointer in a loop-local, seeded once after the top-of-loop safepoints and re-seeded after each of the 8 collection points; ~27 reloads/step → ~3, `FrameRoot`'s move-safety contract unchanged. Orthodox toward `interp_jit.py` `dispatch(self, …)` holding the frame as `self` across `handle_bytecode`.
- **Result.** check.py 241/241 dynasm + cranelift byte-identical, gc_stress 9/9, `MAJIT_STATS` exact match; warmup microbench **+16.6–17.0%**, nbody **+8.5%**, bands disjoint.

### PR #698 — warmup follow-ons (merged, `56644af7`)

A re-profile of the #692 result surfaced two more:
- **Commit 1 — cache the shadow-stack cell in `FrameRoot`.** `shadow_stack::get` was *still* the #1 warmup frame: the remaining ~3 seeds/step each paid a thread-local resolution (`_tlv_get_addr`) via `SHADOW_STACK.with()`. Resolve the cell once per activation (`ShadowStackSlot`) and re-read `entries[depth]` through the cached pointer. Orthodox: RPython resolves the root-stack base once per function and caches it in a register (`assembler.py`, `shadowcolor.py`). **+3.5–3.9%, bands disjoint, every round faster.**
- **Commit 2 — correct `valuestackdepth` on the super-instruction walk-end handback.** The non-flush handback set the resume pc but not the operand depth; `eval_loop_jit` patched it by recomputing `instruction.stack_effect` **every opcode step** — the **#2 non-idle warmup self-time frame**. Making the handback symmetric (`correct_resume_vsd`) lets that per-step reconcile block (and its `stack_effect` recompute) be **removed** — the depth table becomes the single source of truth. Orthodox: PyPy never computes a per-step stack effect (`valuestackdepth` is an incrementally-maintained counter; `stack_effect` is compile-time only). Perf-neutral-to-positive; lands as machinery-removing cleanup.

### Net (2026-07-26)

Both non-idle warmup self-time frames the profiling named — **#1 `shadow_stack::get`** and **#2 per-step `stack_effect`** — are now measured GONE. The ~170 μs driver residual, the last #394 warmup bucket without an orthodoxy obstacle, is attributed and cut. Every lever the issue named is resolved: Lever 1 shipped (#548), Lever 2 closed (no orthodox interpreter path), the optimizer sub-target exhausted, the execute path at PyPy parity, and the driver residual closed (#692/#698). The remaining warmup frontier is the meta-tracing interp+recorder loop and the int-box `malloc` floor — both already source-confirmed at their orthodox floor, with no single lever.

— *addendum by Claude*

youknowone:
close #394

Closes the last open item of #394 (the ~170 μs "driver residual" the issue named as the only remaining warmup bucket with no orthodoxy obstacle — "pure measurement, needs finer instrumentation").

## Root cause (attributes the #394 driver residual)

Profiling the warmup-dominated bench (macOS `/usr/bin/sample`, many fresh compiles) put `majit_gc::shadow_stack::get` as the **#1 non-idle warmup self-time frame** — above every compile/optimizer frame. Source: `FrameRoot::frame()` reloads the `PyFrame` pointer from the thread-local `SHADOW_STACK` (`.with()` + `RefCell::borrow()` + bounds-checked index) on **every** call, and `eval_loop_jit` (the JIT-mode interpreter loop that runs pre-threshold + during tracing = warmup) called it ~27× per opcode step.

## Change

Hold the resolved `PyFrame` pointer in a loop-local `f`, seeded once after the top-of-loop safepoints and re-seeded after each of the 8 collection points (the only places the frame can move under a moving collection); every other frame touch reborrows `f` via `unsafe { &mut *f }` / `&*f`. Cuts ~27 reloads/step → ~3 on the common path. `FrameRoot` itself is unchanged (still the root-registration + reload authority).

This converges toward PyPy's orthodox shape: `interp_jit.py` `dispatch(self, …)` holds the frame as `self` across the whole `handle_bytecode` (frame is `reds=['frame','ec']` + `virtualizables=['frame']`) and does not reload per field access. Collapsing pyre's redundant reloads *between safepoints* is exactly that hold-self, without changing `FrameRoot`'s move-safety contract.

## Correctness

- `check.py --backend dynasm,cranelift`: **241/241 dynasm + 241/241 cranelift, byte-identical** both backends.
- `MAJIT_GC_STRESS=1` gc_stress **9/9** — the UAF gate: a moving collection is forced at every allocation window, so a missed re-seed across any collection point crashes here. No crash ⇒ the re-seed discipline holds.
- `cargo test -p pyre-jit` 328 tests incl. both eval-loop twins.
- `MAJIT_STATS` exact match vs baseline (`loops_compiled`/`bridges_compiled`/`loops_aborted`/`guard_failures` identical, `internal_compile_panics=0`) — transformation altered no compile behavior.

## Performance (A/B, interleaved best-of-N, EPOCHREALTIME)

| bench | ratio (base/branch) | delta |
|---|---|---|
| warmup microbench (run 1, 15 rounds) | 1.170x | **+17.0%** |
| warmup microbench (run 2, 11 rounds) | 1.166x | **+16.6%** |
| nbody (has a warmup phase) | 1.085x | **+8.5%** |

Warmup bands have zero overlap (base min > branch max) — reproducible, far beyond loaded-machine noise.

— *opened by Claude*

### DIFF
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5544,12 +5544,15 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
             if needs_trace {
                 if let Err(err) = unsafe {
                     (*ec_ptr).bytecode_trace(
-                        frame_root.frame() as *mut PyFrame,
+                        f,
                         pyre_interpreter::executioncontext::TICK_COUNTER_STEP,
                     )
                 } {
                     return LoopResult::Done(Err(err));
                 }
+                // bytecode_trace may allocate (tracer callback) → the frame may
+                // have moved; re-seed before reading it again.
+                let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
                 // A trace callback may perform a debugger line-jump by
                 // setting `frame.f_lineno` (`fset_f_lineno` → `last_instr
                 // = best_addr`).  The opcode for this iteration was
@@ -5562,11 +5565,9 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                 // (= `last_instr + 1`) at the top, so rebase the target
                 // through `set_last_instr_from_next_instr` for the next
                 // iteration to land on it rather than one past it.
-                if frame_root.frame().last_instr as usize != opcode_pc {
-                    let jump_target = frame_root.frame().last_instr as usize;
-                    frame_root
-                        .frame()
-                        .set_last_instr_from_next_instr(jump_target);
+                if unsafe { &*f }.last_instr as usize != opcode_pc {
+                    let jump_target = unsafe { &*f }.last_instr as usize;
+                    unsafe { &mut *f }.set_last_instr_from_next_instr(jump_target);
                     continue;
                 }
             } else {
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5648,27 +5648,36 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                     driver.blackhole_if_trace_too_long();
                     continue;
                 }
+                // execute_opcode_step (above) is a collection point and this arm
+                // re-reads the frame; seed a fresh pointer for the compile path.
+                let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
                 // ── can_enter_jit (RPython interp_jit.py:114) ──
                 // RPython interp_jit.py:114 → warmstate.py:446
-                let marker_ec = frame_root.frame().execution_context as *const PyExecutionContext;
-                let marker_pycode = frame_root.frame().pycode as pyre_object::PyObjectRef;
-                let marker_profiled = frame_root.frame().get_is_being_profiled();
+                let marker_ec = unsafe { &*f }.execution_context as *const PyExecutionContext;
+                let marker_pycode = unsafe { &*f }.pycode as pyre_object::PyObjectRef;
+                let marker_profiled = unsafe { &*f }.get_is_being_profiled();
                 pypyjitdriver.can_enter_jit(
-                    frame_root.frame(),
+                    unsafe { &mut *f },
                     marker_ec,
                     loop_header_pc,
                     marker_pycode,
                     marker_profiled,
                 );
-                let green_key = make_green_key(frame_root.frame().pycode, loop_header_pc);
+                // can_enter_jit is a lowered no-op / merge point; re-seed
+                // conservatively before the next frame reads.
+                let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+                let green_key = make_green_key(unsafe { &*f }.pycode, loop_header_pc);
                 if let Some(loop_result) = maybe_compile_and_run(
-                    frame_root.frame(),
+                    unsafe { &mut *f },
                     green_key,
                     loop_header_pc,
                     driver,
                     info,
                     &env,
                 ) {
+                    // maybe_compile_and_run compiles and may allocate → re-seed
+                    // before handle_exception reads the frame.
+                    let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
                     // warmspot.py:998-1005 handle_jitexception: an
                     // `ExitFrameWithExceptionRef` from a direct compiled-code
                     // exit is re-raised into the interpreter loop.  Offer it to
@@ -5677,13 +5686,13 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                     // caught, otherwise propagate it out as a plain `Done(Err)`.
                     if let LoopResult::ExitFrameWithException(mut err) = loop_result {
                         if pyre_interpreter::eval::handle_exception(
-                            frame_root.frame(),
+                            unsafe { &mut *f },
                             &mut err,
                             &mut next_instr,
                         ) {
-                            frame_root
-                                .frame()
-                                .set_last_instr_from_next_instr(next_instr);
+                            // handle_exception may allocate → re-seed.
+                            let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+                            unsafe { &mut *f }.set_last_instr_from_next_instr(next_instr);
                             continue;
                         }
                         return LoopResult::Done(Err(err));
@@ -5694,14 +5703,17 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
             Ok(StepResult::Return(result)) => return LoopResult::Done(Ok(result)),
             Ok(StepResult::Yield(result)) => return LoopResult::Done(Ok(result)),
             Err(mut err) => {
+                // execute_opcode_step (above) is a collection point and this arm
+                // re-reads the frame; seed a fresh pointer.
+                let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
                 if pyre_interpreter::eval::handle_exception(
-                    frame_root.frame(),
+                    unsafe { &mut *f },
                     &mut err,
                     &mut next_instr,
                 ) {
-                    frame_root
-                        .frame()
-                        .set_last_instr_from_next_instr(next_instr);
+                    // handle_exception may allocate → re-seed before the write.
+                    let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+                    unsafe { &mut *f }.set_last_instr_from_next_instr(next_instr);
                     continue;
                 }
                 return LoopResult::Done(Err(err));
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5619,6 +5619,11 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
         let raw_arg: u32 = op_arg.into();
         let delta = instruction.stack_effect(raw_arg);
         if delta > 0 {
+            // `frame` is a shared reborrow used only in this block's reads and
+            // the `if` condition below; its last use ends before the `&mut *f`
+            // reborrow in the taken branch, so the two never alias. Keep any
+            // future `frame` use above that write — the raw pointer means the
+            // borrow checker will not catch an overlap introduced here.
             let frame = unsafe { &*f };
             let pushed_top = frame.valuestackdepth.saturating_add(delta as usize);
             let next_pc = opcode_pc + 1;
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5586,25 +5586,24 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                     )
                 };
                 if ticker < 0 {
-                    if let Err(mut err) =
-                        unsafe { (*ec_ptr).perform_actions(frame_root.frame() as *mut PyFrame) }
-                    {
-                        // Deliver the action's exception (e.g. a signal
-                        // handler's KeyboardInterrupt) as if raised at the
-                        // current opcode so the frame's try/except can
-                        // catch it — CPython runs the eval-breaker
-                        // exception through the same `goto error` path.
-                        // `frame.last_instr` was set to `pc` above, so
+                    if let Err(mut err) = unsafe { (*ec_ptr).perform_actions(f) } {
+                        // perform_actions may allocate → re-seed before reading
+                        // the frame. Deliver the action's exception (e.g. a
+                        // signal handler's KeyboardInterrupt) as if raised at
+                        // the current opcode so the frame's try/except can catch
+                        // it. `frame.last_instr` was set to `pc` above, so
                         // `handle_exception` finds the covering handler.
-                        let mut next_instr = frame_root.frame().next_instr();
+                        let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+                        let mut next_instr = unsafe { &*f }.next_instr();
                         if pyre_interpreter::eval::handle_exception(
-                            frame_root.frame(),
+                            unsafe { &mut *f },
                             &mut err,
                             &mut next_instr,
                         ) {
-                            frame_root
-                                .frame()
-                                .set_last_instr_from_next_instr(next_instr);
+                            // handle_exception may allocate; re-seed before the
+                            // final frame write.
+                            let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+                            unsafe { &mut *f }.set_last_instr_from_next_instr(next_instr);
                             continue;
                         }
                         return LoopResult::Done(Err(err));
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5611,11 +5611,15 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                 }
             }
         }
-        let mut next_instr = frame_root.frame().next_instr();
+        // The ec block above may have run bytecode_trace / perform_actions
+        // (collection points) on a fall-through path; re-seed before the
+        // stack-effect reads and the opcode dispatch.
+        let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+        let mut next_instr = unsafe { &*f }.next_instr();
         let raw_arg: u32 = op_arg.into();
         let delta = instruction.stack_effect(raw_arg);
         if delta > 0 {
-            let frame = frame_root.frame();
+            let frame = unsafe { &*f };
             let pushed_top = frame.valuestackdepth.saturating_add(delta as usize);
             let next_pc = opcode_pc + 1;
             // A JIT handoff can arrive with the stack depth for the point just
@@ -5627,12 +5631,12 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
                 && pyre_jit_trace::state::depth_based_vsd_for_wcode(frame.pycode as usize, next_pc)
                     == Some(frame.valuestackdepth)
             {
-                frame_root.frame().set_last_instr_from_next_instr(next_pc);
+                unsafe { &mut *f }.set_last_instr_from_next_instr(next_pc);
                 continue;
             }
         }
         let step_result =
-            execute_opcode_step(frame_root.frame(), code, instruction, op_arg, next_instr);
+            execute_opcode_step(unsafe { &mut *f }, code, instruction, op_arg, next_instr);
         match step_result {
             Ok(StepResult::Continue) => {
                 // pyjitpl.py:2843 blackhole_if_trace_too_long — check after
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5476,26 +5476,38 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
         // note above), so this is a walkable safepoint.
         majit_gc::gc_sync::safepoint_poll();
 
-        if frame_root.frame().next_instr() >= code.instructions.len() {
+        // Seed the frame pointer once after the two top-of-loop safepoints.
+        // The frame is GC-managed and can move only at a collection point; this
+        // pointer stays valid until the next such point, so the reads below reuse
+        // it instead of re-resolving the shadow-stack slot each time. Re-seeded
+        // after every collection-point call in this loop (bytecode_trace,
+        // perform_actions, execute_opcode_step, handle_exception, can_enter_jit /
+        // maybe_compile_and_run, jit_merge_point).
+        let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
+
+        if unsafe { &*f }.next_instr() >= code.instructions.len() {
             return LoopResult::Done(Ok(w_none()));
         }
 
-        let pc = frame_root.frame().next_instr();
+        let pc = unsafe { &*f }.next_instr();
 
         // interp_jit.py:85-87 — source-level marker declaration.  Its
         // untranslated body is a no-op; source translation
         // recognizes this method call and lowers it to JitCode
         // `jit_merge_point` rather than leaving a residual call.
-        let marker_ec = frame_root.frame().execution_context as *const PyExecutionContext;
-        let marker_pycode = frame_root.frame().pycode as pyre_object::PyObjectRef;
-        let marker_profiled = frame_root.frame().get_is_being_profiled();
+        let marker_ec = unsafe { &*f }.execution_context as *const PyExecutionContext;
+        let marker_pycode = unsafe { &*f }.pycode as pyre_object::PyObjectRef;
+        let marker_profiled = unsafe { &*f }.get_is_being_profiled();
         pypyjitdriver.jit_merge_point(
-            frame_root.frame(),
+            unsafe { &mut *f },
             marker_ec,
             pc,
             marker_pycode,
             marker_profiled,
         );
+        // jit_merge_point is a lowered no-op / merge point and does not itself
+        // collect, but is treated as a collection boundary conservatively.
+        let f: *mut PyFrame = frame_root.frame() as *mut PyFrame;
 
         let (opcode_pc, instruction, op_arg) = match decode_instruction_forward(code, pc) {
             Ok(decoded) => decoded,
@@ -5504,10 +5516,8 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
 
         // ── handle_bytecode (RPython interp_jit.py:90) ──
         trace_jit_bytecode(pc, "");
-        frame_root.frame().last_instr = pc as isize;
-        frame_root
-            .frame()
-            .set_last_instr_from_next_instr(opcode_pc + 1);
+        unsafe { &mut *f }.last_instr = pc as isize;
+        unsafe { &mut *f }.set_last_instr_from_next_instr(opcode_pc + 1);
         // pyopcode.py:170-176 dispatch_bytecode parity: fire
         // `ec.bytecode_trace(self)` each opcode while warming up,
         // with the default `TICK_COUNTER_STEP` decrement.  This is
@@ -5528,7 +5538,7 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
         // PyPy's `actionflag.decrement_ticker(decr_by)` invariant);
         // the `action_dispatcher` slow path itself is still a stub
         // pending the actionflag port.
-        let ec_ptr = frame_root.frame().execution_context as *mut PyExecutionContext;
+        let ec_ptr = unsafe { &*f }.execution_context as *mut PyExecutionContext;
         if !ec_ptr.is_null() {
             let needs_trace = unsafe { !(*ec_ptr).w_tracefunc.is_null() };
             if needs_trace {
--- a/pyre/pyre-jit/src/eval.rs
+++ b/pyre/pyre-jit/src/eval.rs
@@ -5543,10 +5543,8 @@ fn eval_loop_jit(frame: &mut PyFrame) -> LoopResult {
             let needs_trace = unsafe { !(*ec_ptr).w_tracefunc.is_null() };
             if needs_trace {
                 if let Err(err) = unsafe {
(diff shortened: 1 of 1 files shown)


youknowone:
### CI note: the one red check is pre-existing broken-main, not this PR

`pyre/check.py (ubuntu-24.04)` is red on a single line:

```
FAIL wasm synth/delete_negative_open_slice_hot  wrong output
FAILED: wasm 1 failed, 239 passed
```

The same run reports `ALL PASSED: dynasm 241/241` and `ALL PASSED: cranelift 241/241`, and `check.py` on **macOS-latest and windows-latest passes** (both run wasm too). So this is a **wasm-only, Linux-x86_64-only** failure.

It is **not introduced by this PR** — proven by the identical failure on three base commits, each from that commit's own `pyre CI` run:

| commit | role | `delete_negative_open_slice_hot` |
|---|---|---|
| `9da9b5f53` | current `main` | `WRONG` (run 29799105127) |
| `5275a6390` | this PR's base | `WRONG` (run 29787837608) |
| `b159a2fd9` | older base | `WRONG` (run 29760668596) |

This PR's diff is `pyre/pyre-jit/src/eval.rs` only (the backend-agnostic JIT interpreter loop); it touches no wasm/trace/IR code, and both native JIT backends are byte-identical (241/241). The red line is inherited from a broken base via the merge-with-main CI build.

The underlying latent bug is a Linux-x86_64 wasm miscompile of the negative-open-slice `del` (delta 601 in the checksum), pre-existing since #646 and orthogonal to this change.

— *commented by Claude*

PR #698, opened 2026-07-21
jit: gh#394 warmup follow-ons — cache FrameRoot shadow-stack cell + correct walk-end handback vsd

youknowone:
Follow-on to #692, carrying two independent gh#394 warmup cleanups that a re-profile of the #692 result surfaced. Each is separately gated and A/B-measured.

---

## Commit 1 — cache the shadow-stack cell in `FrameRoot`; const-init `SHADOW_STACK`

#692 cut the per-opcode-step `FrameRoot::frame()` reload count in `eval_loop_jit` (gh#394 driver residual); a re-profile showed `majit_gc::shadow_stack::get` still the #1 non-idle warmup self-time frame — the remaining ~3 seeds/step each pay a thread-local resolution (`_tlv_get_addr` on macOS) because `SHADOW_STACK.with()` runs on every access.

- **majit-gc**: add `ShadowStackSlot` (a cached `*const RefCell<ShadowStack>`) plus `shadow_stack_slot()` / `unsafe slot_get()`. The cell address is stable for the owning thread's life — the same invariant `MutatorEntry` already relies on for STW walks.
- **pyre-jit**: `FrameRoot::new` resolves the thread-local cell once and stores the slot; `frame()` re-reads `entries[depth]` through the cached slot instead of resolving the thread-local on every call. `FrameRoot` is a same-thread stack local, so the cached pointer can neither outlive nor cross its thread. `slot_get` takes the same transient borrow the old `get()` did, and re-reads still index the live `entries` Vec, so a Vec realloc between reads is unaffected.
- Also `const`-initialize the `SHADOW_STACK` thread-local (`ShadowStack::new` is now `const` with an empty Vec instead of `Vec::with_capacity(64)`), dropping the per-access lazy-init guard.

**Orthodoxy.** RPython does not re-resolve the root-stack base per access: it lives in a plain global (`gcdata` singleton, `framework.py`), the C backend resolves it once per function (`gc_enter_roots_frame`, `funcgen.py`), the x86 JIT caches the root-stack top in a register across call boundaries (`assembler.py:_load_shadowstack_top_in_ebx`), and `shadowcolor.py` even hoists root saves out of loops. pyre's per-access `SHADOW_STACK.with()` is the divergence; caching the resolution once per activation and reusing it between GC safepoints brings the access cost back in line with upstream. The cache never spans a thread-switch/GIL boundary (single stack-local `FrameRoot` per activation).

**Perf (A/B, interleaved best-of-N, EPOCHREALTIME):** run 1 (15 rounds) +3.66% best-of-N / +3.85% median, bands disjoint; run 2 (12 rounds) +3.71% / +3.52%, disjoint. Every round faster.

---

## Commit 2 — correct `valuestackdepth` on the super-instruction walk-end handback

The non-flush walk-end handback in `compile_and_run_once` set the resume pc but **not** `valuestackdepth`: when a loop-header marker sits inside a super-instruction, the walk closes the loop at `loop_header_pc + 1` and leaves `valuestackdepth` advanced through the super-instruction, while the handback wrote only `last_instr`/`next_instr`. The frame was handed back with pc and operand depth inconsistent, and `eval_loop_jit` patched it downstream by recomputing `instruction.stack_effect` **every opcode step** and, when the projected push overflowed the frame array and the depth table agreed, advancing the pc instead of dispatching. That per-step `stack_effect` was the #2 non-idle warmup self-time frame.

- Make the handback symmetric with the flush leg (`restore_resume_state_from`) and the blackhole legs (`apply_blackhole_crn_handoff`): after `set_last_instr_from_next_instr(restart_pc)`, call `correct_resume_vsd(frame, restart_pc)` to re-derive the operand depth from the resume pc via the depth table.
- The handed-back frame is then self-consistent, so the per-step reconciliation block and its `stack_effect` recompute are **removed**. `valuestackdepth` is now only ever *compared* against the precomputed depth table (`depth_based_vsd_for_wcode` / liveness), never re-derived per dispatch step — the depth table is the single source of truth, and rustpython's `stack_effect_info` leaves the hot path.

**Orthodoxy.** PyPy's dispatch loop never computes a per-step stack effect: `valuestackdepth` is an incrementally-maintained counter (`pyframe.py` push/popvalue), `stack_effect` is compile-time only (co_stacksize, `astcompiler/assemble.py`), and the JIT frame is a virtualizable restored *positionally* from resume data in one shot (`rpython/jit/metainterp/virtualizable.py`), so pc and vsd are never mutually inconsistent at a merge point. Fixing the seam so the frame is self-consistent — rather than patching the inconsistency downstream — is the orthodox shape.

**Perf (A/B, interleaved best-of-N, 21 rounds):** +1.39% best-of-N / +0.17% median, bands overlap. Perf-neutral-to-slightly-positive; this commit lands as a correctness/orthodoxy cleanup that *removes* machinery (the per-step reconciliation), with the A/B confirming no regression.

---

## Correctness (whole branch)

- `check.py --backend dynasm,cranelift`: **241/241 dynasm + 241/241 cranelift, byte-identical** both backends (measured on both the pre- and post-commit-2 base).
- `MAJIT_GC_STRESS=1` gc_stress **12/12** — the UAF gate: a moving collection at every allocation window.
- `cargo test -p majit-gc` 191, `cargo test -p pyre-jit` 320+12 (both eval-loop twins).
- Golden warmup output byte-identical (`3004320000`) for both commits.

— *opened by Claude*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved garbage-collection root tracking during JIT execution.
  - Fixed frame and operand-stack state restoration when execution resumes inside optimized instruction sequences.
  - Improved consistency after garbage-collection points, reducing the risk of incorrect execution state.

- **Performance**
  - Reduced repeated overhead when accessing per-thread garbage-collection state during JIT operations.

### DIFF
--- a/majit/majit-gc/src/shadow_stack.rs
+++ b/majit/majit-gc/src/shadow_stack.rs
@@ -212,7 +212,7 @@ thread_local! {
     static MAX_SHADOW_STACK_DEPTH: Cell<usize> = const { Cell::new(DEFAULT_SHADOW_STACK_DEPTH) };
 
     /// Thread-local shadow stack for individual GcRef roots.
-    static SHADOW_STACK: RefCell<ShadowStack> = RefCell::new(ShadowStack::new());
+    static SHADOW_STACK: RefCell<ShadowStack> = const { RefCell::new(ShadowStack::new()) };
 
     /// Thread-local flat jitframe root stack. Rust tests run multiple
     /// JIT/GC tests in parallel, so using one process-global root stack lets
@@ -372,9 +372,9 @@ struct ShadowStack {
 }
 
 impl ShadowStack {
-    fn new() -> Self {
+    const fn new() -> Self {
         ShadowStack {
-            entries: Vec::with_capacity(64),
+            entries: Vec::new(),
         }
     }
 }
@@ -428,6 +428,37 @@ pub fn get(index: usize) -> GcRef {
         let ss = ss.borrow();
         ss.entries[index]
     })
+}
+
+/// An opaque handle to this thread's `SHADOW_STACK` cell.
+///
+/// The pointer is stable for the life of the owning thread (the same
+/// invariant `MutatorEntry` relies on for STW walks): the thread-local cell is
+/// never reallocated once created. Resolving it once and reusing it lets a hot
+/// caller re-read roots without paying the thread-local resolution
+/// (`_tlv_get_addr` on macOS) on every access — the root-stack base is held in
+/// a fixed location and re-read cheaply, matching the C backend's
+/// once-per-function `gc_enter_roots_frame` resolution and the x86 JIT's
+/// register-cached root-stack top.
+#[derive(Clone, Copy)]
+pub struct ShadowStackSlot(*const RefCell<ShadowStack>);
+
+/// Resolve this thread's `SHADOW_STACK` cell once, for reuse by [`slot_get`].
+pub fn shadow_stack_slot() -> ShadowStackSlot {
+    ShadowStackSlot(SHADOW_STACK.with(|ss| ss as *const _))
+}
+
+/// Get a GcRef at `index` through a previously resolved [`ShadowStackSlot`].
+///
+/// # Safety
+///
+/// `slot` must have been produced by [`shadow_stack_slot`] on the current
+/// thread and the thread must still be alive (its `SHADOW_STACK` not yet torn
+/// down). No `&mut` borrow of the cell may be held across this call.
+pub unsafe fn slot_get(slot: ShadowStackSlot, index: usize) -> GcRef {
+    // SAFETY: per the contract, `slot.0` points at the live owning-thread cell.
+    let ss = unsafe { &*slot.0 }.borrow();
+    ss.entries[index]
 }
 
 /// Walk all entries on the GcRef shadow stack.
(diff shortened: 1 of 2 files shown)


youknowone:
Added commit `370e908` — the second gh#394 warmup follow-on (correct `valuestackdepth` on the super-instruction walk-end handback, removing the per-step `stack_effect` reconciliation). Body updated to cover both commits.

A/B for this commit (cand-1 baseline vs this commit, interleaved best-of-N, 21 rounds): **+1.39% best-of-N / +0.17% median, bands overlap** — perf-neutral-to-slightly-positive. It lands as a correctness/orthodoxy cleanup that removes machinery; the A/B confirms no regression. Gates green: check.py 241/241 both backends byte-identical, gc_stress 12/12, pyre-jit 320+12, golden `3004320000`.

— *commented by Claude*

youknowone:
**CI status:** all green except one pre-existing failure.

- `pyre/check.py (ubuntu-24.04)`: `FAIL wasm synth/delete_negative_open_slice_hot wrong output` (+ its cranelift twin), **Linux-x86_64 only**. `dynasm 241/241` passes; macOS and Windows check.py both pass; all cargo test / wasm build / CPython gate / Codex parity / fmt jobs pass.
- This failure is **pre-existing and not caused by this PR**: it reproduces on `main` (e.g. commit `ad785680f`, same `check.py (ubuntu-24.04)` failure), and the negative-open-slice deletion path on the Linux wasm/cranelift backends is untouched by these commits — the diff is confined to `pyre/pyre-jit/src/eval.rs` (a dynasm/interp `valuestackdepth` seam) and the majit-gc shadow-stack slot cache. Local dynasm+cranelift on macOS are byte-identical 241/241.

— *commented by Claude*

PR #706, opened 2026-07-21
jit: fix two exception-path compile defects (last-exc lookahead, except-tuple pin)

youknowone:
Two defects made ordinary `try`/`except` shapes fall off the compiled path, plus two follow-up fixes from review.

## 1. The last-exc lookahead was control-flow-blind

`reads_last_exc_before_next_catch` decides whether a residual call may clear the walker's recorded exception. It advanced by `op.next_pc` only, but jitcode is one flat byte stream whose blocks are stitched by `goto/L` — the physically next op is frequently not the successor.

Instrumented on a `raise`/`except E as e:`/`raise e` frame: the scan from the residual call at jitcode pc=2040 ran past `pc=2077 goto/L → 1017` into the unrelated block at 2080, found *that* block's `catch_exception/L`, and answered "no read". The exception was cleared; the real successor `1017 → 1020` is `last_exc_value/>r`, so every trace died with `LastExcValueWithoutActiveException`. `permanent=false`, so the frame retraced and re-aborted for the whole run — `loops_compiled=0`.

Following labels: **3.3x @80k, 8.8x @200k, 18.0x @400k**.

## 2. `except (A, B):` pinned a rebuilt tuple by identity

The clause builds its match target with `BUILD_TUPLE` on every visit, so the tuple is a fresh object each time. `try_walker_fold_check_exc_match` pinned it with `GuardValue(match_op, const_ref(<the traced tuple's address>))` — unsatisfiable. Guard failures came in at roughly N/2 (the nursery bump sometimes hands back the same address, which is what made it look intermittent).

Discriminator: hoisting `EXC = (TypeError, ValueError)` out of the loop took guard failures 49900 → 1; `except (TypeError,)` with a single element was equally slow, so it was never about tuple length or multi-class matching.

Fix pins the **elements** instead. Two object elements land in `W_SpecialisedTupleObject_oo`, whose inline `value0`/`value1` are immutable → `GetfieldGcPureR` → folded away entirely; other arities go through `wrappeditems` behind an `arraylen` guard. Non-tuple targets keep the identity pin, which does hold for a loaded target.

dynasm **0.627 → 0.067s (9.4x)**, cranelift 0.484 → 0.072s, guard failures 29901 → 2, bridges 149 → 0 (pypy 0.059s). 46/79 corpus programs improved, none regressed. The inline tuple-except form appears at **510 sites across 200 files** of the bundled `lib-python`.

Bench added: `pyre/bench/synth/except_tuple_clause_hot.py`.

## 3–4. Review follow-ups

- The lookahead followed only `goto/L` and `goto_if_not/iL`; the other 23 label-bearing ops fell through to the physically next op, `switch/id` successors were never considered, and `reraise/` was classified as not reading the exception although `opimpl_reraise` re-raises `last_exc_value`. The label offset is now derived from the key's operand signature, and an un-decodable successor set answers "read" — over-reporting leaves a recorded exception standing, under-reporting drops one a later read needs.
- `walker_guard_exc_match_tuple_items` emitted the layout `GuardClass` before reading each element, so an element bail-out returned false with the class already pinned; the caller gates its own identity pin on `is_class_known`, which dropped every value guard from the fold. Elements are now read before anything is recorded.

## Not included, deliberately

Removing the `for_iter_frame_has_raising_named_handler` gate was tried on top of these fixes. It measured 2.5–4.9x faster with zero aborts and passed 1088 differential checks on both backends — and it is still **wrong**. The gate blocks *every* abort, not only the one fixed here, and any abort in a FOR_ITER frame drops the in-flight item. A handler body that mutates a container hits `InplaceContainerMutationUnsupported` and the loop returns 3999 items instead of 4000 on both backends; a second shape aborts the process. The differential corpus missed it because no generated handler body mutated a container. The gate stays.

## Verification

check.py **245/245 dynasm + 245/245 cranelift**; pyre-jit-trace 321, pyre-jit 241, majit-metainterp 1401 (dynasm) / 1399 (cranelift); 688 differential checks across 348 generated exception shapes on both backends, 0 wrong.

## Known, pre-existing, out of scope

A JIT-compiled same-frame raise/except never attaches `__traceback__`, so `e.__traceback__` reads `None` inside the handler once the loop compiles (4000 → 201 on both backends; `e.args` and `type(e)` are unaffected). Confirmed pre-existing by rebuilding with `origin/main`'s `jitcode_dispatch/mod.rs` — same result. Filed as follow-up work, untouched here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved exception handling across branching code paths, preventing incorrect exception state tracking.
  - Improved support for matching exceptions against tuples of exception classes.

- **Performance**
  - Refined exception-match optimization to reduce unnecessary runtime checks while preserving correct behavior.

- **Benchmarks**
  - Added coverage to compare tuple-based and single-class exception handling performance.

### DIFF
--- a/majit/majit-metainterp/src/history.rs
+++ b/majit/majit-metainterp/src/history.rs
@@ -2407,6 +2407,10 @@ impl TraceCtx {
             frames: recorder_frames,
             vable_boxes: vable_boxes.to_vec(),
             vref_boxes: vref_boxes.to_vec(),
+            // The nested-list append fold resumes through the single-frame
+            // collapse, never this multi-frame path, so no extra virtual roots
+            // flow here.
+            extra_virtual_roots: Vec::new(),
         });
         self.set_last_guard_op_resume_position(snapshot_id);
     }
--- a/pyre/bench/synth/except_tuple_clause_hot.py
+++ b/pyre/bench/synth/except_tuple_clause_hot.py
@@ -1,0 +1,42 @@
+# An `except (A, B):` clause builds its match target with `BUILD_TUPLE` on
+# every visit, so the tuple is a fresh object each time the handler runs. The
+# `CHECK_EXC_MATCH` fold pins the match target it saw while tracing; pinning
+# the container rather than its elements makes that guard unsatisfiable on a
+# re-allocated tuple, and the handler then side-exits once per visit. The
+# elements are `LOAD_GLOBAL` constants already covered by the quasi-immutable
+# invalidation guard, so the clause must stay on the compiled path.
+#
+# The single-class `except A:` loop below is the control: same raise rate, a
+# match target that is loaded rather than built. Both loops must agree with the
+# interpreter and neither may accumulate side exits. Output verified against
+# CPython/PyPy.
+N = 60000
+
+
+def tuple_clause(n):
+    acc = 0
+    i = 0
+    while i < n:
+        try:
+            if i > 2:
+                raise TypeError("x")
+        except (TypeError, ValueError):
+            acc += 9
+        i += 1
+    return acc
+
+
+def single_clause(n):
+    acc = 0
+    i = 0
+    while i < n:
+        try:
+            if i > 2:
+                raise TypeError("x")
+        except TypeError:
+            acc += 9
+        i += 1
+    return acc
+
+
+print(tuple_clause(N), single_clause(N))
--- a/pyre/bench/synth/except_tuple_clause_hot.py
+++ b/pyre/bench/synth/except_tuple_clause_hot.py
@@ -8,8 +8,7 @@
 #
 # The single-class `except A:` loop below is the control: same raise rate, a
 # match target that is loaded rather than built. Both loops must agree with the
-# interpreter and neither may accumulate side exits. Output verified against
-# CPython/PyPy.
+# interpreter and neither may accumulate side exits.
 N = 60000
 
 
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch/mod.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch/mod.rs
@@ -2134,18 +2134,77 @@ pub(crate) fn exc_handler_rejoins_loop(code: &[u8], catch_target: usize) -> bool
     false
 }
 
+/// True when a path reachable from `position` reads the walker's active
+/// exception (`last_exception/>i` / `last_exc_value/>r`) before reaching the
+/// `catch_exception/L` that would replace it.
+///
+/// The jitcode is one flat byte stream whose blocks are stitched together by
+/// `goto/L` labels, so the block that physically follows an op is frequently
+/// not its successor. Advancing by `op.next_pc` alone therefore walks out of
+/// the current block and answers from whatever block the assembler happened to
+/// place next — reporting that block's `catch_exception/L` while the real
+/// successor reads the exception. The residual-call sites act on that answer by
+/// clearing `last_exc_value`, so the read then finds no active exception and
+/// the walk aborts (`LastExcValueWithoutActiveException`) for the whole run.
+///
+/// Labels are followed instead: `goto/L` continues at its target,
+/// `goto_if_not/iL` queues its taken arm and continues on the fall-through, and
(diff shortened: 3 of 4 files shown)


youknowone:
Went through every review finding against the source and, where a shape was named, against a running reproducer. Dispositions below.

## Fixed

**`goto_if_exception_mismatch/iL` was misclassified** (Codex parity §3). Confirmed against `rpython/jit/metainterp/pyjitpl.py:1677-1681` — `opimpl_goto_if_exception_mismatch` does `last_exc_value = metainterp.last_exc_value; assert last_exc_value` before testing the class, i.e. it reads the exception exactly like `opimpl_last_exc_value` (`:1716-1719`, same assert). The scan treated it as an ordinary conditional branch and could report no later read, letting the exception be cleared before it. Now answers "read".

**`decode_op_at` returning `None` is not only end-of-stream** (CodeRabbit). Confirmed at `jitcode_runtime.rs:634-637`: it also returns `None` for an opcode byte outside `INSNS_BYTE_TO_OPNAME` and for a payload the stream is too short to hold. Answering `false` there under-approximated the reads the rest of the scan deliberately over-approximates. Now answers "read".

**`w_class` check missing on the specialised arity-2 branch** (CodeRabbit). The helper's own doc says a tuple subclass whose `w_class` diverges must return `false` so the caller pins identity, but only the array-backed branch enforced it. Hoisted the check ahead of the layout split so it covers both. Not reachable today — all three constructors in `specialisedtupleobject.rs` (`:109`, `:141`, `:184`) hardcode `w_class: get_instantiate(&TUPLE_TYPE)` — so this closes a contract gap rather than a live bug.

## Not reproducible — kept as is

**Over-preserving on the untaken sibling branch** (Codex inline, P2). The concern is that OR-ing both successors of a conditional keeps `last_exc_value` on a trace whose taken path would have cleared it, reaching a later `catch_exception/L` with a stale exception. Built the exact shape described — a matched `except TypeError:` whose sibling `except ValueError as e:` is the jump target — plus four variants (reversed clause order, three clauses, a re-raising sibling, a tuple-clause sibling), all at N=60000. Every one is correct against the interpreter with `loops_compiled=1`, `loops_aborted=0` on both backends; the 348-program differential corpus is likewise abort-free.

The suggested fix — use the concrete branch outcome instead of OR-ing — is not available at the decision point: the clear/preserve choice is made *at the residual call*, before the branch condition (frequently derived from that call's own result) has been evaluated. A design that defers the clear to the next `catch_exception/L` would be the real path-specific answer; that is a larger change than this PR, and nothing measured so far motivates it.

The preservation is also narrower than it may read: it is gated on `saved_last_exc_value.is_some()`, so it only applies when an exception is already recorded.

## Deferred, with the blocker named

**`RvmprofCode` collapses to `None`, dropping `cintf.jit_rvmprof_code(arg1, arg2)`** (Codex parity §3, pre-existing). Real divergence from `pyjitpl.py:2547-2555`. Calling the hook needs the frame's `registers_i`, and `try_catch_exception_at` / `finishframe_lookahead_at` take only `&[u8]` — wiring a `WalkContext` through the finishframe lookahead touches call sites unrelated to this PR's subject. `majit_translate::rlib::rvmprof::cintf::jit_rvmprof_code` is a no-op unless an embedder installs a hook, and pyre installs none, so nothing observable is lost today.

## Also in this branch

`main` at `57b01d0e8bd` does not compile: `capture_snapshot_for_last_guard_op_multi_frame_with_vable_vref` (`history.rs:2406`) builds a `recorder::Snapshot` without the `extra_virtual_roots` field that the same commit (`0364c4679f5`) added to the struct. Carried a one-line fix here — an empty vector, matching the sibling multi-frame path at `:2368` — because nothing builds without it. Happy to drop it if it is being fixed on main.

## Verification after these changes

check.py 245/245 dynasm + 245/245 cranelift; pyre-jit-trace 321, pyre-jit 241, majit-metainterp 1401 / 1399; 688 differential checks over 348 generated exception shapes on both backends, 0 wrong; the two iteration-drop / crash reproducers that reverted the gate removal still match the interpreter.

— *commented by Claude*

PR #707, opened 2026-07-21
jit: resolve struct layouts for the target being built, not the extraction host

youknowone:
The build-time descr pool that `pyre-jit-trace/build.rs` serialises is read back
by the target binary, but every offset in it came from a single host extraction
of Charon LLBC plus a hardcoded 8-byte pointer in the layout model. On a 32-bit
target both are wrong — `PyObject.w_class` sits at 4, not 8; `PyType.instantiate`
at 24, not 32 — so the orthodox `w_list_append` sub-walk, the one production
consumer that resolves `d` operands through the pool, read the wrong bytes on
wasm32. #697 papered over this by gating the fold on a runtime layout check.

Three layers, all fixed here:

1. **Extraction was host-only.** A layout sidecar is now produced per configured
   cross target: charon re-runs with `--targets <triple>` and the result is
   reduced to its `type_decls`, written as `<crate>.<target>.layouts.ullbc`
   (1.1 MB / 6.9 MB vs the 46 MB / 383 MB host artefacts). `majit-translate`
   prepends them when `TARGET != HOST`, so their `exact_layouts` win the
   first-writer-wins merge. Reader code is untouched. Passing `--targets` with
   several triples in one pass was tried first and rejected: it aggregates
   *bodies* too, which breaks portal resolution.
2. **The layout model assumed a 64-bit pointer.** `layout::target_word_size()`
   reads `CARGO_CFG_TARGET_POINTER_WIDTH` and replaces `size_of::<usize>()`
   throughout `codewriter::call`; `usize`/`isize` split off `u64`/`i64`. On a
   64-bit target the word is still 8, so native codegen is bit-identical.
3. **A latent 32-bit-only optimizer bug.** `intbounds::propagate_postprocess`
   grouped `Getfield*{I,R,F}` into one arm gated on `field_size < 8`. Only the
   `_I` opcodes have a postprocess, and a 4-byte pointer field satisfies the
   sub-word test — so with correct layouts wasm hit
   `setintbound: expected 'i'-typed operand, got Ref`. Narrowed to `_I`.

`build_descr_layout_matches_target()` is replaced by
`build_time_field_offset(owner, name)` plus a test pinning the pool's offsets to
`offset_of!` for the structs the sub-walk reads.

`pyre-jit` gets no sidecar: with `--target` set, cargo stops applying RUSTFLAGS
to host units, and `pyre-jit-trace`'s build script drags in `rustpython-host_env`,
which uses `cfg_select!` — still unstable on Charon's pinned nightly, and the
`-Zhost-config` that would reach host units panics cargo. The 588 types this
leaves out are all compiler-internal; every runtime type reached through a descr
comes from `pyre-object` or `pyre-interpreter`, which do get sidecars.

## Verification

- `check.py`: dynasm 244/244, cranelift 244/244, wasm 243/243
- `cargo test -p majit-metainterp --lib`: 1401 passed
- `cargo fmt --all -- --check` clean

## Summary by CodeRabbit

- **New Features**
  - Added support for target-specific layout metadata during cross-compilation.
  - Improved build preparation for cross-target and WebAssembly environments.
  - Added validation to detect missing layout artifacts before compilation.

- **Bug Fixes**
  - Improved structure, array, and field layout calculations across different target word sizes.
  - Corrected integer-bound propagation for applicable field operations.
  - Removed an unnecessary WebAssembly restriction affecting list-append optimizations.

- **Tests**
  - Added checks confirming build-time and runtime field layouts remain consistent.

### DIFF
--- a/.github/workflows/pyre-ci.yml
+++ b/.github/workflows/pyre-ci.yml
@@ -124,6 +124,13 @@ jobs:
     # unchanged crates restore while only the changed crate re-extracts; the
     # Extract step below refills any missed crate, so the downstream artefact
     # always carries all three ullbc files.
+    #
+    # Each crate's entry also carries its `*.layouts.ullbc` sidecars — the
+    # same crate re-extracted for a cross target and reduced to its type
+    # declarations, so a wasm32 build reads wasm32 field offsets rather than
+    # the extraction host's. They must restore alongside the artefact they
+    # accompany: a cache hit skips extraction, and `pyre-jit-trace`'s build
+    # script refuses a cross-target build whose sidecars are absent.
     - name: Cache LLBC (pyre-object)
       id: llbc-cache-object
       uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
@@ -131,6 +138,7 @@ jobs:
         path: |
           build/llbc/pyre-object.ullbc
           build/llbc/pyre-object.ullbc.fingerprint
+          build/llbc/pyre-object.*.layouts.ullbc
         key: llbc-pyre-object-${{ runner.os }}-${{ runner.arch }}-${{ env.CHARON_VERSION }}-${{ steps.llbc-fingerprint.outputs.pyre_object }}
     - name: Cache LLBC (pyre-interpreter)
       id: llbc-cache-interpreter
@@ -139,6 +147,7 @@ jobs:
         path: |
           build/llbc/pyre-interpreter.ullbc
           build/llbc/pyre-interpreter.ullbc.fingerprint
+          build/llbc/pyre-interpreter.*.layouts.ullbc
         key: llbc-pyre-interpreter-${{ runner.os }}-${{ runner.arch }}-${{ env.CHARON_VERSION }}-${{ steps.llbc-fingerprint.outputs.pyre_interpreter }}
     - name: Cache LLBC (pyre-jit)
       id: llbc-cache-jit
--- a/majit/majit-metainterp/src/optimizeopt/intbounds.rs
+++ b/majit/majit-metainterp/src/optimizeopt/intbounds.rs
@@ -1636,12 +1636,6 @@ impl Optimization for OptIntBounds {
                 | OpCode::GetfieldRawI
                 | OpCode::GetfieldGcI
                 | OpCode::GetinteriorfieldGcI
-                | OpCode::GetfieldRawR
-                | OpCode::GetfieldGcR
-                | OpCode::GetinteriorfieldGcR
-                | OpCode::GetfieldRawF
-                | OpCode::GetfieldGcF
-                | OpCode::GetinteriorfieldGcF
                 | OpCode::GetarrayitemRawI
                 | OpCode::GetarrayitemGcI
                 | OpCode::CallPureI
--- a/majit/majit-metainterp/src/optimizeopt/intbounds.rs
+++ b/majit/majit-metainterp/src/optimizeopt/intbounds.rs
@@ -1715,15 +1715,14 @@ impl Optimization for OptIntBounds {
             }
 
             // ── Field accesses ──
-            OpCode::GetfieldRawI
-            | OpCode::GetfieldGcI
-            | OpCode::GetinteriorfieldGcI
-            | OpCode::GetfieldRawR
-            | OpCode::GetfieldGcR
-            | OpCode::GetinteriorfieldGcR
-            | OpCode::GetfieldRawF
-            | OpCode::GetfieldGcF
-            | OpCode::GetinteriorfieldGcF => {
+            // `postprocess_GETFIELD_GC_I` and friends: only the int-typed
+            // reads narrow, because the bound this installs is an integer
+            // range and `setintbound` asserts an `'i'` operand.  The `_R` /
+            // `_F` variants have no postprocess of their own.  Their absence
+            // matters on a 32-bit target: a pointer field is 4 bytes wide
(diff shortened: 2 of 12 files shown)


youknowone:
Two more commits: the 32-bit list-append fold now actually folds.

**The remaining 32-bit defect was an array base, not a resume.** `#661` gated
the Integer/Float `w_list_append` fold off on wasm32
(`wasm_unboxed_append_fold_declined`) and attributed the wrong output to a
wasm-backend deopt/resume defect. It is an addressing bug in the array descr:
`arraydescrof_concrete` placed the first item at `len_offset +
array_header_size`, immediately after the length word. The blocks are
`#[repr(C)] { capacity: usize, items: [T; 0] }`, whose items begin at the
length word **rounded up to `T`'s alignment**. An 8-byte word hides this — the
two always coincide. A 4-byte one does not: an `i64` item sits at 8, the descr
said 4, and the whole Integer array was strided a half-word early.

The measurement that pinned it — `[j & 3 for j in range(100000)]` on wasm32,
elements dumped by index:

```
(600, 4294967296), (601, 8589934592), (602, 12884901888), (604, 4294967296)
```

`4294967296 == 1 << 32`, and `601 & 3 == 1`. Every element read back as
`v[i+1] << 32`, exactly what a base 4 bytes below the real one produces for
8-byte items.

- `call.rs`: `array_items_base` / `element_align` at both descr sites. Scalars
  and pointers align to their own width capped at 8; a struct element aligns to
  its widest registered field, falling back to the word. Unit test pinned.
- `jitcode_dispatch`: `wasm_unboxed_append_fold_declined` and its three uses
  are deleted — the fold now folds on wasm32, unboxed arms included.
- `history.rs`: `capture_snapshot_for_last_guard_op_multi_frame_with_vable_vref`
  was missing `Snapshot.extra_virtual_roots`, so `majit-metainterp` did not
  compile at this base. The function and the field landed in separate merges.

check.py after: dynasm 246/246, cranelift 246/246, wasm 245/245 — and 245/245
again with the gate removed rather than merely forced on. `cargo test -p
majit-translate --lib` 2997 passed, `-p majit-metainterp --lib` 1401 passed,
`-p pyre-jit-trace --lib` 242 passed, `cargo fmt --all -- --check` clean.

youknowone:
Worked through every review comment. Commit `91e898a2dc`.

### Applied

**Sidecars drop `type_decls` metadata → merge only `exact_layouts`** (codex, `llbc_extract.py:424`; coderabbit `lib.rs:316`)
Confirmed real: `collect_dedup_bodies` harvests `HashConsedValue` defs from the whole artefact; the sidecar empties `fun_decls`, so a type first hash-consed in a dropped body is unresolvable. The object sidecar has 1379 `Deduplicated` refs vs 456 defs in `type_decls` — ~900 field types would render unresolved, and since sidecars were prepended (first-writer) they'd win over the host's correct types. Field types are target-independent, so the fix is to take them from the host and only offsets from the sidecar: sidecars now append **after** the host artefacts and `exact_layouts` merges **last-writer-wins**. The sidecar's per-type-string tables lose to the host's; its target offsets still win.

**Struct-array base must match `GcTypedArray`** (codex, `call.rs:2157`)
Confirmed real regression in my own change. Interior-field struct arrays address `GcTypedArray` — a flat `{ len, items:[u8;0] }` block with items at `GC_TYPED_ARRAY_ITEMS_OFFSET = WORD` regardless of element alignment (both the live block and `allocate_array_struct` resume path). Element-aligning that base diverged from the block on wasm32. Reverted `2157` to the bare `array_header_size`; the list int/float path (`TypedItemsBlock`, genuinely element-aligned) keeps its aligned base.

**`have_postprocess_op` still lists `_R`/`_F`** (coderabbit, `intbounds.rs:1725`)
Removed — `propagate_postprocess` handles only `_I` now, so those dispatched to a no-op.

**Cross-target recovery hints can't produce/advertise sidecars** (codex `build.rs:151` & `:154`)
The bootstrap command now prints `LLBC_LAYOUT_TARGETS=<target> scripts/extract-llbc.py`, and the `PYRE_MIR_FRONTEND_LLBC` override example appends the sidecar paths.

**Sidecars not in the codegen cache key** (codex `build.rs:592`)
`hash_llbc_inputs` now hashes the sidecars, so a regenerated sidecar rekeys instead of restoring stale descrs.

**Missing sidecar silently skipped** (coderabbit `lib.rs:316`) — `auto_discover_workspace_llbc_paths` now warns.

**Duplicated sidecar naming** (coderabbit `build.rs:614`) — `majit_translate::layout::{is_cross_target, layout_sidecar_filename}` centralise it; both `build.rs` and `auto_discover_workspace_llbc_paths` call them.

**`write_layout_sidecar` field scan hardening** (coderabbit `llbc_extract.py:427`) — each field validates the decoded type and turns a `JSONDecodeError` into the file's clean `SystemExit`.

### Not applied

**"Fix the fold instead of gating it"** (codex `specialize.rs:3182`) — already done in a prior commit: the target-width gate is gone and the root cause (array items base off by a half-word on 32-bit) is fixed, so the fold folds on wasm32.

**Fingerprint target-only inputs before skipping sidecars** (codex `llbc_extract.py:603`) — acknowledged but deferred. The extraction stamp already fingerprints `Cargo.lock` + the full crate source closure, so a struct-layout change that could move offsets rekeys. The residual gap (a path/git dependency's layout changing with no `Cargo.lock` or pyre-source change) is narrow, and with the merge fix above a stale sidecar can only affect offsets, not field types. A full target-metadata walk is disproportionate to that surface; leaving it as a follow-up.

check.py: dynasm 248/248, cranelift 248/248, wasm 247/247.

PR #719, opened 2026-07-22
jit: gate the tracing-frame force on the escaping frame being the traced one

youknowone:
## Problem

A residual (may-force) callee that inspects **its own frame** via `sys._getframe()` in a hot loop makes the JIT produce a wrong answer: a non-journaled `STORE_ATTR` in the callee body is applied one extra time per compiled iteration.

Discriminator (JIT-only; `PYRE_NO_JIT=1`, CPython and PyPy all print `199990000 20000`):

```python
import sys

class Counter:
    pass

c = Counter()
c.n = 0

def bump(x):
    c.n += 1                  # STORE_ATTR: non-journaled body effect
    frame = sys._getframe(0)  # forcing residual inspecting the callee's own frame
    return x if frame is not None else -1

def main():
    total = 0
    for i in range(20000):
        total += bump(i)
    print(total, c.n)         # JIT (for-loop): 199990000 20010 — c.n over-counted

main()
```

This is the P1 left unanswered by the #699 review ("Avoid forcing unrelated tracing frames").

## Cause

`force_pyframe` fires from `gettopframe` / `getnextframe_nohidden` when Python asks for a frame during a residual call. The `sys._getframe(0)` above hands Python the **callee's own** frame — not the caller's traced virtualizable — yet the force path unconditionally forced the traced vable whenever its token was `TOKEN_TRACING_RESCALL`. Forcing a frame that did not actually escape clears the caller's tracing token, which `tracing_after_residual_call` then reads as a real escape and raises `VableEscapedDuringResidualCall`. The abort replays `ContinueRunningNormally` from loop entry, re-running the already-executed non-journaled `STORE_ATTR` → double-apply.

## Fix

Force the traced virtualizable only when the frame handed to Python **is** the traced one (i.e. it was recorded as escaping):

- `flush_active_frame_escape` now returns a `bool` — `true` only when the active escape's `expected` frame matches the frame being forced. The committed-pc write is **decoupled** from flush success (flushing legitimately declines when there is no merge-point depth, and that must not suppress the escape signal).
- `force_pyframe` gates the traced-frame force on that returned flag; the caller's `live_frame_armed` force path is unchanged.

`for i in range(20000): total += bump(i)` moves `c.n` 20010 → **20000**.

## Verification

- `pyre/bench/synth/getframe_residual_callee_own_frame.py` (added) — for-loop regression guard, oracle-matched (CPython == PyPy == pyre == `199990000 20000`).
- `check.py`: **dynasm 286/286, cranelift 286/286, wasm 283/283**, 3/3 backend runs green on the rebased HEAD.
- `inline_helper` and the other inline benches unchanged (no inline-path regression).
- An adversarial static review confirmed the force-target change is a **non-issue** (post-fix force is a strict subset of pre-fix force — it only ever forces *less*).

## Known limitation (tracked, not fixed here)

The **while-loop** twin of the discriminator (`while i < N: total += bump(i)`) still over-counts (`c.n = 20005`). This is **independent** of the fix above (A/B-proven: the count is identical before and after — the fix is monotone and only ever forces less) and is a separate, pre-existing bug.

It is an instance of the known **resume-past-abort** limitation: a side-effecting user function in a hot loop that the JIT keeps aborting re-applies its non-journaled effect `+K` times. In the while shape `bump` is taken down the multiframe-inline seeding path, which declines (`LoopBearingCalleeInlineUnsupported`) after the body already ran concretely; the gh#467 abort-flush then sees an unjournaled effect and keeps the legacy replay from loop entry instead of resuming past the abort.

A perf-safe static fix was **attempted and refuted**: the only available static signal (`extraeffect >= ForcesVirtualOrVirtualizable`) also covers generic Python-to-Python calls, so declining inline on it regresses `inline_helper`/`nbody` to timeouts; and declining inline does not close the class anyway (the residual path double-applies too, for a callee escaping the *traced* frame via `sys._getframe(1)`/`f_back`). The real fix is the resume-past-abort piece in gh#467/#343 territory, tracked separately. No while-loop bench is added here because it would be red.

— *authored by Claude*

### DIFF
--- a/pyre/bench/synth/getframe_residual_callee_own_frame.py
+++ b/pyre/bench/synth/getframe_residual_callee_own_frame.py
@@ -1,0 +1,30 @@
+# Regression guard: a residual (may-force) callee that inspects its OWN frame
+# via sys._getframe() must not clear the traced CALLER's virtualizable tracing
+# token. Clearing it raised a spurious frame-escape with no committed resume pc,
+# replaying the loop body from entry and double-applying the callee's
+# non-journaled STORE_ATTR side effect -- a JIT-only wrong answer (c.n > loops).
+import sys
+
+
+class Counter:
+    pass
+
+
+c = Counter()
+c.n = 0
+
+
+def bump(x):
+    c.n += 1                  # STORE_ATTR: non-journaled body effect
+    frame = sys._getframe(0)  # may-force residual inspecting the callee's own frame
+    return x if frame is not None else -1
+
+
+def main():
+    total = 0
+    for i in range(20000):
+        total += bump(i)
+    print(total, c.n)
+
+
+main()
--- a/pyre/pyre-jit-trace/src/jitcode_dispatch/residual_call.rs
+++ b/pyre/pyre-jit-trace/src/jitcode_dispatch/residual_call.rs
@@ -40,7 +40,14 @@ impl Drop for ActiveFrameEscapeGuard {
     }
 }
 
-pub fn flush_active_frame_escape(ctx: &TraceCtx, frame: *mut pyre_interpreter::PyFrame) {
+/// Returns whether `frame` is the one recorded as escaping this residual call
+/// (`expected == frame`) — i.e. the traced virtualizable itself was handed to
+/// Python, so its tracing token must be forced. A residual callee inspecting
+/// its own frame passes a different `frame` and returns false. Independently,
+/// the walk-end resume pc is committed only when the state flush succeeds (a
+/// merge point with cached depth); a matched frame that cannot flush still
+/// escaped and must be forced, so the two signals are decoupled.
+pub fn flush_active_frame_escape(ctx: &TraceCtx, frame: *mut pyre_interpreter::PyFrame) -> bool {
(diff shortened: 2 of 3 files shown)


youknowone:
Closing the loop on the P2 finding (eval.rs "Preserve f_back escapes from returned callee frames") with empirical results:

- **The stale-locals mechanism is refuted.** Same-epoch A/B on every stored-frame `f_back.f_locals` repro shape gives `stale_hits=0` on both sides: when `.f_back.f_locals` is read, that read is itself a may-force residual whose escape guard matches the caller (`expected == frame`), so the caller IS force-flushed before the read observes it.
- **But the probe surfaced a real regression in a different symptom**: the gate delays the escape from inside-the-callee to the later `f_back` read, whose mid-expression escape flush always declined — legacy replay + FOR_ITER deliver-refuse then **dropped the consumed iteration** (total under-count, not stale reads). The pre-gate unconditional force had incidentally masked this latent resume-past-abort defect.
- Fixed in #740 by committing the escape flush with the walk's operand-stack mirror and resuming forward past the abort; `getframe_stored_fback_walk` guards the shape.

— *commented by Claude*

PR #740, opened 2026-07-23
jit: commit the frame-escape flush with the operand-stack mirror and resume past the abort

youknowone:
## Problem

The orthodox force-gate (#719) unmasked a latent resume-past-abort defect: when a residual forces the traced frame mid-expression (a loop reading a stored frame's `f_back.f_locals`), the escape-pc walk-end flush always declined on "NULL operand-stack shadow slot" — the vable shadow's stack region is only valid at merge points. With no committed resume pc, the abort fell back to legacy replay, the FOR_ITER inflight deliver was refused, and the consumed iteration was **dropped**:

| repro | oracle | before | after |
|---|---|---|---|
| stored `getframe(0)` → `f_back.f_locals` walk in a hot for-loop | 199990000 | 199986990 (5 iterations dropped) | **199990000** |
| + non-journaled side effect in the callee | 199990000 / 20000 | 199986990 / 20005 | **199990000** / 20005¹ |

¹ the remaining +5 is the pre-existing own-frame entry-trace replay double (see Follow-up).

## Mechanism

Frame-level convergence of `run_blackhole_interp_to_cancel_tracing` (pyjitpl.py:2949) / `convert_and_run_from_pyjitpl` (blackhole.py:1799): resume forward from the abort point instead of replaying from loop entry.

- **Latch**: `ActiveFrameEscapeGuard` snapshots the walk's operand-stack OpRef mirror (`vstack_boxes`) per may-force residual — the same authoritative source `reconstructed_all_ref_call_stack` uses for the inline-abort Entry carrier.
- **Flush**: at force time the oprefs resolve through `concrete_of_opref` (GC-forwarded channel) and the walk-end flush commits with the mid-expression stack via a new full-stack variant, height-checked against the forward depth analysis.
- **Resume**: the committed pc adopts through the existing escape leg; the walk's applied effects stand exactly once (journals commit), the escaping opcode re-executes in the interpreter, and the in-flight FOR_ITER item is owned by the flush (no drop, no double).

## Safety gates (re-execution must be idempotent)

- `!writes_live_heap` — every Python-visible frame **mutator** is a Void store or CALL-shaped helper (an object-model invariant: the `f_lineno` setter, `sys.settrace`, `_warnings.warn` all force but are excluded by shape); what survives is frame-family attribute/item reads, which re-execute idempotently against the flushed values.
- **Opcode-scoped effect window** — frame-entry + executed-effect odometers sampled at the first residual of each opcode, so an earlier residual of a multi-residual opcode (rich-compare fallback chains) blocks the latch.
- **Post-call commit withdrawal** — if the forcing callee itself entered a user frame, the commit is cancelled and a **pre-flush undo snapshot** (rooted for its whole armed life through the store-journal root area) restores the frame, so legacy replay re-enters pristine pre-walk state. Discriminator: a journaled `seen.append(i)` before a forcing-then-mutating `@property` — without the restore, 5 appends were lost to the journal rollback.
- **Double-force skip** — after the first commit the frame is heap-authoritative; a second force in the same residual does not re-flush, so callee `f_locals` write-throughs survive adoption.
- Bridge walks and inline sub-walks never latch (their abort paths bypass the adoption epilogue); `PYRE_FBW_ABORT_FLUSH=0` restores byte-identical legacy behavior (verified: the old under-count reappears exactly).

## Verification

- check.py: **dynasm 293/293, cranelift 293/293, wasm 290/290**
- 47/47 foriter57 + getframe/inline_subwalk/mutate guard synth benches, JIT == NO_JIT
- 3 new synth guards: `getframe_stored_fback_walk`, `getframe_force_cancel_journal`, `getframe_force_write_between`
- Perf A/B vs main (same-epoch alternating): inline_helper / nbody / nested_loop / raise_catch — no regression
- Design cross-checked against rpython sources (pyjitpl.py `vable_after_residual_call` → `SwitchToBlackhole(ABORT_ESCAPE)` chain; virtualizable.py force semantics) including an adversarial review cycle that surfaced and closed two unsound holes (cancel-after-flush frame corruption, residual-scoped effect window) and three undo-lifecycle gaps (rooting window, double-force re-flush, cross-attempt staleness).

## Follow-up (not in this PR)

The orthodox endgame is jitcode-granularity blackhole conversion (`convert_and_run_from_walk`): build blackhole frames from the walk's register state and resume AFTER the residual with its result, feeding the existing production `BlackholeInterpreter`. That closes the two remaining bounded defects this PR does not cover — the own-frame `getframe(0)` entry-trace replay double (+5 above) and the while-loop inline-decline twin — and retires the FOR_ITER deliver/refuse machinery wholesale.

— *authored by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved execution reliability when frame inspection occurs during optimized code.
  * Preserved loop iterations, local-variable updates, and list mutations during frame transitions and resumed execution.
  * Improved handling of operand stacks and frame state during optimized execution fallbacks.

* **Tests**
  * Added regression checks covering frame access, caller-local updates, stored frame references, and repeated frame-forcing operations.
  * Integrated the new validation into the pre-merge benchmark checks.

### DIFF
--- a/pyre/bench/getframe_escape_flush_writethrough_regression.py
+++ b/pyre/bench/getframe_escape_flush_writethrough_regression.py
@@ -1,0 +1,43 @@
+# Self-checking regression guard for the frame-escape flush that resumes past
+# the abort (registered via check.py run_selfcheck, NOT the synthetic suite).
+# A residual (may-force) callee stores its own frame; the loop body then reads
+# the caller's f_back.f_locals and mutates it through the 3.14 FrameLocalsProxy
+# write-through. Reading the redirected caller frame forces it mid-expression:
+# the escape flush must commit with the operand-stack mirror (the vable shadow's
+# stack region is NULL there) and resume forward AT the escaping opcode. The
+# legacy replay-from-loop-entry fallback (PYRE_FBW_ABORT_FLUSH=0) drops the
+# in-flight FOR_ITER iteration instead, so `total` comes up short -- the JIT-only
+# regression this guards.
+#
+# The write-through is a 3.14 FrameLocalsProxy behaviour PyPy 3.11 lacks (its
+# f_locals is a snapshot), so cpython and pypy disagree on the mutated value and
+# this cannot be a synthetic bench; the invariant is pyre-internal instead (a
+# correct JIT reproduces the no-JIT result) and is asserted here.
+import sys
+
+LOOPS = 20000
+box = [None]
+
+
+def bump(x):
+    box[0] = sys._getframe(0)
+    return x
+
+
+def main():
+    total = 0
+    marks = 0
+    for i in range(LOOPS):
+        total += bump(i)
+        f = box[0].f_back
+        f.f_locals['marks'] = f.f_locals['marks'] + (1 if f.f_lineno > 0 else 0)
+    if total != sum(range(LOOPS)):
+        print(f"FAIL dropped iteration: total={total} expected {sum(range(LOOPS))}")
+        raise SystemExit(1)
+    if marks != LOOPS:
+        print(f"FAIL write-through dropped: marks={marks} expected {LOOPS}")
+        raise SystemExit(1)
+    print(f"PASS total={total} marks={marks}")
+
+
+main()
--- a/pyre/bench/synth/getframe_force_cancel_journal.py
+++ b/pyre/bench/synth/getframe_force_cancel_journal.py
@@ -1,0 +1,32 @@
+# Regression guard: a frame-forcing user @property that forces FIRST (the
+# escape flush commits mid-property) and mutates SECOND (the commit is then
+# withdrawn because the callee entered a user frame). The withdrawal must
+# RESTORE the pre-flush frame before the legacy replay runs: replaying on the
+# flushed (mid-iteration) frame skipped the journal-rolled-back list append,
+# losing elements (len(seen) < loops, JIT-only).
+import sys
+
+
+class Obj:
+    hits = 0
+
+    @property
+    def peek(self):
+        _ = sys._getframe(1).f_locals  # forces the traced caller -> flush commits
+        Obj.hits += 1                  # user-frame effect AFTER the commit
+        return 1
+
+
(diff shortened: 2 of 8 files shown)


youknowone:
### CI fix (force-pushed `65c501790ae`)

**Root cause of the `pyre/check.py` failure** (Windows job, but every platform's check.py would fail the same way): the synth bench `getframe_force_write_between` printed a value that depends on `f_locals` write-through. That is the 3.14 `FrameLocalsProxy` behaviour (PEP 667): CPython 3.14 writes through and prints `20000`, PyPy 3.11's `f_locals` is a snapshot and prints `0`. check.py's synthetic oracle requires `cpython == pypy`, so the bench was a `BASEFAIL` — the two reference interpreters legitimately disagree on a 3.14-only language feature.

pyre itself was **correct** here (it implements the 3.14 `FrameLocalsProxy` write-through; JIT == no-JIT == `20000`). The bench, not the code, was unsuitable for the PyPy-gated synth suite.

**Fix:** dropped the synth bench and re-landed the coverage as a self-checking guard, `getframe_escape_flush_writethrough_regression.py`, registered via `run_selfcheck` (a pyre-internal oracle — no cpython/pypy comparison), so a 3.14-only feature can be asserted without the `cpython == pypy` gate.

**Honest note on what it guards:** the old commit text framed this bench as "write-through survives a second force" (the double-force skip). I A/B-tested that claim by reverting the `COMMITTED_FRAME_ESCAPE_PC` early-return — the bench still passed, so that framing was not the discriminator. The real discriminator is the escape-flush resume-past-abort: under `PYRE_FBW_ABORT_FLUSH=0` (the legacy replay) it drops the in-flight FOR_ITER iteration and the loop total comes up short (`199986990 != 199990000`); with the fix it passes. The renamed guard now asserts that (the total), on all three backends.

Local `pyre/check.py`: dynasm 296/296, cranelift 296/296, wasm 293/293 (the new guard: dynasm/cranelift/wasm all PASS).

— *commented by Claude*

PR #751, opened 2026-07-23
jit: resolve the jd1 novable unpackiterable-drain live blackhole resume and enable it by default

youknowone:
Resolves the jd1 (`unpackiterable_driver`) novable-drain **live** blackhole
resume, then makes the fixed driver the default: entering the compiled drain
live, blackhole-resuming it on a guard failure against the build-time
jitcode/liveness/descr stores, and draining into a `W_List` — end-to-end
`OK builtin 3000`.

The final commit flips jd1 from opt-in (`PYRE_JD1=1` / `PYRE_JD1_ENTER`) to
**on by default**, so the default runtime now traces, compiles, and
live-enters the `unpackiterable_driver` drain on hot non-tuple unpack sites.
The previous inert behavior is reachable with `PYRE_NO_JD1`; jd1 also follows
the master JIT off-switches (`PYRE_NO_JIT` / `PYRE_JIT=0`).

### Commits
1. **`setarrayitem_gc` trace heap cache** — after the trace-walk store, call
   `ctx.heapcache_setarrayitem` keyed on `descr.index()` so the sibling
   getarrayitem read cannot return a value cached before the store (Codex P2 on
   #741).
2. **Novable vinfo through blackhole resume** — thread the per-jitdriver novable
   `vinfo` so the drain's blackhole frame reconstructs its reds
   (`[w_iterator, items]`, both `Type::Ref`) from the right jitdriver.
3. **Resolve novable drain resume against build-time stores** — the novable
   drain has no Python `CodeObject`; resolve its resume `rd_numb` / liveness /
   descr against the build-time jitcode/liveness/descr stores instead of a
   frame-chain walk.
4. **Null-guard raw_code lookup** (+ orphan-fix, converged with upstream) — the
   novable portal's degenerate `PyJitCode` carries a null `code_ptr`, so
   `raw_code_for_jitcode_index` returns `None` for a null pointer rather than
   handing it to the instruction-decoding consumers (bare-reraise probe,
   traceback lineno). This commit also originally deleted the orphaned
   `build_multi_frame_miframe` `last_caught_exception_value` reads (`#763`
   referenced a field `#756` had removed → E0609); after rebasing onto current
   `main` that deletion **converged with upstream `#765`** ("drop #763's
   residual last_caught_exception_value propagation"), which made the identical
   removal, so the rebase absorbed it and this commit now carries only the
   `state.rs` null-guard.
5. **Collapse the drain append into one registered `dont_look_inside` seam** —
   inlining `w_list_append` into the drain jitcode surfaced each of append's
   strategy/grow helpers (`object_push`, `switch_to_correct_strategy`,
   typed-array grow, …) as a separate unresolved residual funcptr → symbolic
   hash → SIGBUS. Route the drain's append through a `dont_look_inside`
   `drain_list_append` wrapper and register it (with the already-residual
   `w_list_new_empty` prologue and `drain_collect_items` epilogue). The global
   `list.append` keeps calling `w_list_append` directly and **stays traced**, so
   the append fold and the escape-flush replay are unaffected. Also adds
   `int_is_true/i>i` to the curated inline-call blackhole builder — the drain's
   back-edge guard emits it and it was missing, panicking a blackhole-executed
   drain at the first back-edge test.
6. **Firing regression guard for the drain fusion** — `test_result_exc_lowering.rs`
   lowers the real `_unpackiterable_unknown_length` from the production LLBC and
   asserts the Facet A (#703) `try_fuse_drain_match` fusion actually **fires**:
   the synthesized `exc_kind_discriminant` kind-test is present, no
   `StopIteration` ctor survives, and the `next()` site is a `LastException`
   edge. The fusion is fail-safe (silent decline → `catch_and_rewrap`), the
   default non-jd1 run never executed it, and the `unpack_drain_exact_kind`
   parity test only guards the default path — so this drain rework (commit 5)
   or any recognizer regression that silently stopped the fusion was previously
   invisible while reopening the jd1 SIGBUS.
7. **Enable jd1 and the live enter by default** — `jd1_experiment_enabled` flips
   from opt-in (`PYRE_JD1=1`) to on by default, returning false only for
   `PYRE_NO_JD1` / `PYRE_JD1=0` / `PYRE_NO_JIT` / `PYRE_JIT=0`. A new
   `jd1_enter_enabled` gates the `RunCompiled` live enter, on by default with a
   `PYRE_JD1_NO_ENTER` opt-out, replacing the former `PYRE_JD1_ENTER` opt-in.

### Verification
- `check.py --backend dynasm`, **default env** (jd1 now on by default):
  **303/303 ALL PASSED** — including the perf gates.
- Default run of the drain driver (no env): jd1 fires (counter → CloseLoop →
  Compiled) and **live-enters** the compiled drain (4 enters, 200→3000 items
  per enter) → `OK 24000`. `PYRE_NO_JD1=1` makes it fully inert (0 counter/
  compile/enter) and still `OK 24000`.
- `cargo test -p majit-translate --test test_result_exc_lowering`: **6/6**
  (incl. the drain-fusion firing guard: `exc_kind_discriminant=1`,
  `stopiteration_ctors=0`, `lastexc_blocks=2` on the real drain).
- jd0 bit-identical (`899880005 30 73411`).
- `synth/getframe_force_cancel_journal` → `20000 20000 989403` (this test
  briefly regressed to `19995` under an earlier append-`dont_look_inside`
  approach that residualized the **global** append; commit 5 keeps the global
  append folded, restoring it).

### Follow-up (not in this PR)
Making the **global** append a residual exposes a separate pre-existing latent
bug: a journaled-append residual under a FOR_ITER item + a `sys._getframe`
frame-force makes the escape-flush withdraw and the resume disagree — the resume
adopts a forward continuation past `STORE_FAST` instead of the full legacy
replay, dropping the store. That path (post-withdraw resume/journal, touched by
#756/#749/#763) is out of jd1 scope; this PR sidesteps it by keeping the global
append traced/folded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when unpacking iterables of unknown length, including stable iterator-exception handling.
  * Fixed stale heap cache entries after JIT GC array element updates.
  * Hardened JIT inline and residual calls to reject unresolved/symbolic targets and avoid unsafe indirect calls.
  * Improved exception propagation/GC safety across optimized execution, and corrected JIT resume behavior for additional paths.
* **New Features**
  * JD1 tracing is now enabled by default (can be disabled via environment settings).
* **Tests**
  * Added/updated regression and unit coverage for the affected tracing, exception, and call-dispatch behavior.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -7690,6 +7690,9 @@ pub struct CraneliftBackend {
     trace_counter: u64,
     next_trace_id: Option<u64>,
     next_header_pc: Option<u64>,
+    /// `Backend::set_next_frame_value_count_fn` — the compiling driver's
+    /// `-live-` decoder for the `rd_numb` reads below.
+    next_frame_value_count_fn: Option<fn(i32, i32) -> usize>,
     registered_call_assembler_tokens: IndexSet<u64>,
     registered_call_assembler_bridge_traces: IndexSet<u64>,
     /// llmodel.py: self.vtable_offset — byte offset for vtable in objects.
@@ -7917,6 +7920,7 @@ impl CraneliftBackend {
             trace_counter: 1,
             next_trace_id: None,
             next_header_pc: None,
+            next_frame_value_count_fn: None,
             registered_call_assembler_tokens: IndexSet::new(),
             registered_call_assembler_bridge_traces: IndexSet::new(),
             // llmodel.py:64-69: vtable_offset is None when gcremovetypeptr is
@@ -8487,6 +8491,7 @@ impl CraneliftBackend {
             caller_layout,
             &constants_i64,
             attached_descrs,
+            self.next_frame_value_count_fn,
         )?;
         // RPython jitframe layout parity: ref_root slots start AFTER all
         // output slots. max_output_slots must be >= inputs.len() so that
@@ -14404,6 +14409,10 @@ fn collect_guards(
     caller_layout: Option<&ExitRecoveryLayout>,
     _constants: &indexmap::IndexMap<u32, i64>,
     attached_descrs: majit_backend::AttachedDescrPtrs,
+    // `Backend::set_next_frame_value_count_fn` — the compiling driver's
+    // `-live-` decoder for the `rd_numb` reads below.  `None` falls back to the
+    // process-global callback.
+    frame_value_count_fn: Option<fn(i32, i32) -> usize>,
 ) -> Result<(), BackendError> {
     let type_index = OpTypeIndex::new(inputargs, ops);
     let (type_overrides, op_def_positions) = build_type_overrides(ops, &type_index);
@@ -14541,7 +14550,7 @@ fn collect_guards(
                 (op.resolved_rd_numb(), op.resolved_rd_consts())
             {
                 use majit_ir::resumedata::{get_frame_value_count_fn, rebuild_from_numbering};
-                let fvc = get_frame_value_count_fn();
+                let fvc = frame_value_count_fn.or_else(get_frame_value_count_fn);
                 let fvc_ref: Option<&dyn Fn(i32, i32) -> usize> =
                     fvc.as_ref().map(|f| f as &dyn Fn(i32, i32) -> usize);
                 let num_virtuals = op.resolved_rd_virtuals().map_or(0, |v| v.len());
@@ -14588,7 +14597,7 @@ fn collect_guards(
             let rd_vi = op.resolved_rd_virtuals();
             use majit_ir::resumedata::{self, RebuiltValue, rebuild_from_numbering};
             let rd_consts_ref: &[majit_ir::Const] = &rd_consts_data;
-            let fvc = majit_ir::resumedata::get_frame_value_count_fn();
+            let fvc = frame_value_count_fn.or_else(majit_ir::resumedata::get_frame_value_count_fn);
             let fvc_ref: Option<&dyn Fn(i32, i32) -> usize> =
                 fvc.as_ref().map(|f| f as &dyn Fn(i32, i32) -> usize);
             let num_virtuals = rd_vi.as_ref().map_or(0, |v| v.len());
@@ -15384,6 +15393,10 @@ impl majit_backend::Backend for CraneliftBackend {
 
     fn set_next_header_pc(&mut self, header_pc: u64) {
         self.next_header_pc = Some(header_pc);
+    }
+
+    fn set_next_frame_value_count_fn(&mut self, fvc: Option<fn(i32, i32) -> usize>) {
+        self.next_frame_value_count_fn = fvc;
     }
 
     fn set_done_with_this_frame_descr_void(&mut self, descr: majit_ir::DescrRef) {
--- a/majit/majit-backend/src/call_stub.rs
+++ b/majit/majit-backend/src/call_stub.rs
@@ -550,10 +550,87 @@ pub fn collect_call_args_positional(
     out
 }
 
+/// Dispatch a residual call described by `arg_classes`, picking the signature
+/// strategy the active backend supports.
+///
+/// `bh_call_*_dispatch` transmutes the funcptr to an `extern "C" fn` guessed
+/// from the *bucketed* `(int, float)` arity. That is sound only on a C ABI that
+/// tolerates a signature mismatch: SysV/AAPCS pass the surplus in registers the
+/// callee ignores, and a `usize` parameter is register-width either way. wasm32
+/// has neither property — `call_indirect` type-checks the callee's declared
+/// type on every call, and a pointer parameter is `i32` where the transmute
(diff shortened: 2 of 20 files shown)


youknowone:
Added commit `9faa576` — flips jd1 (`unpackiterable_driver`) and its live compiled-drain enter from opt-in to **on by default**:

- `jd1_experiment_enabled` returns false only for `PYRE_NO_JD1` / `PYRE_JD1=0` / `PYRE_NO_JIT` / `PYRE_JIT=0` (the last two match jd0's own kill-switch).
- new `jd1_enter_enabled` gates the `RunCompiled` live enter, on by default, opt-out `PYRE_JD1_NO_ENTER` (replaces the former `PYRE_JD1_ENTER` opt-in).

Verification (default env, no `PYRE_JD1*` forcing): `check.py --backend dynasm` **303/303 ALL PASSED**, perf gates included. The drain driver now traces → compiles → live-enters (4 enters, 200→3000 items each); `PYRE_NO_JD1=1` makes it fully inert (0 counter/compile/enter); jd0 stays bit-identical. The PR body's earlier "default-off, default path unaffected" note is dropped accordingly.

— *commented by Claude*

youknowone:
Rebased on `origin/main` (`447a8c6`) and force-pushed as `2906d4c`. Three commits added since the last push, all found by chasing a jd1 default-on crash the previous head shipped with.

### 1. `try_fuse_drain_match` was silently declining (`dfc7f45`)

The recognizer looked for `PyErrorKind::StopIteration` only as a niladic `SyntheticTransparentCtor`. Once the fieldless enum lowers to its discriminant it arrives as `ConstInt(9)`, so the lookup missed and the fail-safe fell back to `catch_and_rewrap`, leaving `StopIteration` ctor/eq residuals whose `BhDescr::JitCode.fnaddr` are `symbolic_fnaddr_for_path` hashes. That was the SIGBUS and the first `dispatch_step: unwired opcode=0xbc` panic — one root cause, not the three separate defects I had recorded. A stale `.ullbc` had masked it; `LLBC_FORCE_REEXTRACT=1 cargo build` is a no-op, the re-extract is `python3 scripts/extract-llbc.py pyre-interpreter`.

### 2. i64env cargo-test regression (`f757d54`)

`resolve_active_jitdriver_sd_with_vinfo` refused the linear scan whenever *any* slot carried a `virtualizable_info` — including the placeholder `ensure_default_driver_sd` pushes before registration, which `set_virtualizable_info` broadcasts to while no driver is elected. Gated on `jd.index.is_some()` (stamped only by `register_jitdriver_sd`, call.py:46-47). This was the branch's cargo-test red on ubuntu/windows, not the default flip.

### 3. jd1 dropped drain-time exceptions (`5a4e5f5`) — Codex section 1, and it was reachable

The live-enter path discarded every `ExitFrameWithExceptionRef` and cleared the pending slots, relying on the interpreter drain to re-derive the error by calling `next()` again. That only holds for an exhaustion-stable iterator. Repro that aborts the process on the previous head:

```python
class It:                      # raises once, then reports exhausted
    def __next__(self):
        if self.i == 3000 and not self.raised:
            self.raised = True
            raise ValueError("boom")
        if self.raised or self.i >= self.n: raise StopIteration
        self.i += 1; return self.i

f(*It(5000))   # -> panicked at blackhole.rs: dispatch_step: unwired opcode=0xbc pos=157
```

The exception is now taken off the guard exit (and off `ExitFrameWithExceptionRef`) and parked in `TL_JIT_PENDING_EXCEPTION` after the trailing clears; `drain_jit_pending_exception` re-raises it at the caller loop's next call dispatch, before `__next__` re-runs, so no side effect repeats. `warmspot.py:998-1005` propagates the same exception by re-raising out of `ll_portal_runner`; jd1 is entered from a merge-point hook with no return value, so the slot carries it. `walk_jit_pending_exception` roots the parked object — that slot had no GC root walker before.

That also keeps the blackhole out of the drain's re-raise arm, whose `inline_call_r_r` byte the production builder does not register. **Still open, and not fixed here:** `build_inline_call_only_bh_builder` registers 152 of 203 canonical keys, and the canonical `inline_call_*` family is unregistered. Wiring it is not a two-line change — `read_descr` hands back a `BhDescr`, while the global build-time pool holds `RuntimeBhDescr::JitCode(Arc<JitCode>)`, and 812 of 814 build-time jitcodes carry symbolic fnaddrs the value-keyed runtime patcher cannot map. Worth its own issue.

### Test coverage

No test anywhere reached `_unpackiterable_unknown_length` on purpose — `unpack_drain_exact_kind.py` uses `a, b, c = it`, which routes to `unpack_sequence_exact`'s own loop. Added `pyre/bench/synth/unpack_drain_star_raise.py`: `f(*it)` over plain / re-raising / raise-once iterators, output-compared against CPython.

### Gates

- `python3 pyre/check.py --backend dynasm` — 309/309
- `cargo test --all --no-default-features --features dynasm` — clean (40 suites, 0 failures)
- CI's cranelift test set — clean
- `cargo fmt --all -- --check` — clean

— *commented by Claude*

PR #753, opened 2026-07-23
jit(gh#346): retire the complex-division prepass walls via float-`/`→truediv lowering

youknowone:
## Summary

Retires the last foreign-frontier cluster in the two-phase-rtyper `[PREPASS phaseA fail]` census — the complex-division graphs in `descroperation.rs` — plus a determinism fix for the census itself.

Three commits:

1. **`rtyper: sort two-phase Pass-2 pending by path key`** — Pass-2 `populate_call_registry_from_call_graphs` lifted `pending` in `GraphStore::iter()` (`HashMap` order), so last-write-wins error recording + the fail-closed transitive gate made the census head count flicker run-to-run. Sorting `pending` by path key before the lift loop makes it deterministic. Bit-exact (no behavior change).

2. **`front/rtyper: register the f64 complex-division foreign leaves`** — adds `core::f64::<Impl>::abs` / `is_nan` and the `NAN` / `INFINITY` associated-constant accessors to `FOREIGN_STDLIB_EXTERNALS`. These were the remaining unresolved foreign leaves in `_Py_c_quot` / `_Py_c_prod` / `_Py_rc_quot`; `is_infinite` was already present and `is_finite` is front-recognized (`(x-x)==0.0`). Net-neutral alone — prerequisite for #3.

3. **`front, codewriter: lower a float-result \`/\` to the truediv opname`** — the root cause. `canonical_binop_label` maps every Rust MIR `Div` to the `floordiv` opname, correct for integer truncating division but wrong for `f64 /` (true division). The prepass annotator then dispatches `floordiv(SomeFloat, SomeFloat)`, which `pairtype(SomeFloat, SomeFloat)` does not register (add/sub/mul/div/truediv, no floordiv — `binaryop.py:428-447`; RPython's flowspace emits the `truediv` opname for `/` at graph construction so it never sees float floordiv). The fix re-labels a float-result `Div` to `truediv` at the `Rvalue::BinaryOp` lowering. Bit-exact downstream: `jtransform` funnels both `floordiv` and `truediv` over floats to `float_truediv`, and the rtyper maps `(FloatRepr, "truediv")` / `(FloatRepr, "div")` to the same `rtype_template("truediv")`. `canonical_float_arith_binop` is extended to accept `truediv` / `inplace_truediv` on the legacy-walker path too.

## Effect

Retires `complex_quot` / `complex_truediv` / `real_complex_quot` / `float_truediv` prepass-fail heads: **census 309 → 305, no staircase** (the added-head set is empty). Remaining floordiv/mod walls are unrelated: `float_divmod_w`/`float_floordiv` wall on `mod(SomeFloat, SomeFloat)` (`%` on floats → `math.fmod` residual, a separate pre-existing concern), and `int_floordiv` on `checked_div` (integer path).

## Verification

- `cargo test -p majit-translate` — 3004+ tests green.
- `check.py --backend dynasm,cranelift` — **dynasm 297/297, cranelift 297/297**.
- Complex/float division bit-exact vs python3.14, including JIT-traced hot loops (`100.0/x` and `(3+4j)/z` over 100k iterations).
- No LLBC re-extract needed (all edits in `majit-translate`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved floating-point division handling to consistently distinguish true division from integer-style division.
  * Added support for floating-point operations involving absolute values, `NaN`, and infinity.
  * Made translation error reporting more consistent and repeatable across runs.

* **Compatibility**
  * Recognizes additional division operation spellings during translation, improving support for supported Rust and Flowspace code.

### DIFF
--- a/majit/majit-translate/src/codewriter/jtransform.rs
+++ b/majit/majit-translate/src/codewriter/jtransform.rs
@@ -5262,9 +5262,13 @@ fn canonical_float_arith_binop(op: &str) -> Option<&'static str> {
         // `front::mir::canonical_binop_label` collapses Rust `/`
         // (MIR `Div`) to "floordiv"; over floats that operator is
         // true division, so both labels land on `float_truediv`.
+        // The front end re-labels a float-result `/` to "truediv"
+        // (mirroring RPython's flowspace opname) before the prepass
+        // annotator, so this arm also accepts the already-canonical
+        // "truediv" / "inplace_truediv" spellings.
         // Integer "floordiv" never reaches this arm — the guard
         // requires a float operand or a Float result.
-        "div" | "div_assign" | "floordiv" => Some("div"),
+        "div" | "div_assign" | "floordiv" | "truediv" | "inplace_truediv" => Some("div"),
         _ => None,
     }
 }
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -3515,7 +3515,7 @@ impl<'a> Lowering<'a> {
             Rvalue::BinaryOp(op_json, lhs, rhs) => {
                 let lhs_v = self.resolve_operand(mir_bb, lhs)?;
                 let rhs_v = self.resolve_operand(mir_bb, rhs)?;
-                let op_label = binop_label(&op_json)?;
+                let mut op_label = binop_label(&op_json)?;
                 let res = self
                     .graph
                     .alloc_value_var_with_type(crate::model::ConcreteType::Unknown);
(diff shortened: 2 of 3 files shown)


youknowone:
CI note: the single red job — `pyre/check.py (ubuntu-24.04)` — is inherited base-debt, not a regression from this PR:

- `FAIL wasm synth/exception_metadata_hot wrong output` — pre-existing wasm base-debt. The parent commit on `main` (`b54b8665679`) fails the *identical* job with the *identical* signature (run 30003488822).
- `FAIL cranelift int_loop 0.34s > pypy 0.14s x2` — chronic slow-runner perf flake (tiny 0.14s timer-floor baseline).

The same code passes `pyre/check.py` on **macOS and Windows**, and **dynasm is 297/297** on ubuntu; a real regression would fail deterministically on all three OSes. All correctness gates (CPython suite, cargo test ×3 OS, Codex parity review, CodeRabbit) are green.

— *commented by Claude*

Issue #343, opened 2026-07-03
JIT: recover deep-unroll — depth≥2 middle-frame virtual PyFrame rematerialization (walker-arch rework)


youknowone:
> **STATUS UPDATE (2026-07-22).** The depth≥2 SEGV **soundness blocker is resolved** — it was a GC-rooting gap, fixed by rooting the pending blackhole-interp chain across `run()`; raising the bound is now a performance question, not soundness. Recursion-unroll infra landed: `max_unroll_recursion=7` (gated, PR #572), depth-2 recursive inline (PR #674), middle-frame ec fix (PR #643). **Remaining:** the default multiframe depth is still clamped to 1 (`fbw_max_multiframe_depth`); the acceptance is to raise it toward 7 by default and re-measure fib cold toward the PyPy target.

---

Follow-up to #215 item 2 (*virtualize the callee frame and unbox the result int across the `CALL_ASSEMBLER` boundary to remove the per-call heap-frame/box allocation*).

## Status (2026-07-06, HEAD `b6de0492bd`)

**deep-unroll is upstream-orthodox and this issue is valid.** The `50x`/SEGV symptoms are our port being incomplete, **not** a defect in the mechanism. Do not close as superseded/wontfix — the remaining work (depth≥2 middle-frame virtual `PyFrame` rematerialization on resume) is real and load-bearing.

## Upstream basis for deep-unroll (verified)

Recursive unrolling is a first-class RPython optimization, not a pyre invention:

- `_opimpl_recursive_call` (`rpython/jit/metainterp/pyjitpl.py:1375-1416`): walks `metainterp.framestack`, counts how many frames already share the callee's greenkey (`same_constant`), and **inlines the recursive call** (`perform_call(portal_code, allboxes, greenkey=greenboxes)`) while `count < max_unroll_recursion`; only once the count reaches the cap does it stop inlining (`dont_trace_here`) and fall to the `CALL_ASSEMBLER` tail.
- The inline comment is explicit (pyjitpl.py:1386-1387): inlining the recursion *"would be equivalent to unrolling a while loop"* — i.e. bounded recursion unroll is intended.
- `max_unroll_recursion = 7` is the shipped default (`rpython/rlib/jit.py:580,601`; `rpython/jit/metainterp/warmspot.py:96`), settable per-driver (`warmstate.py:317`), and exercised in the test suite (`test_recursive.py:349,1130` set it to `10`).

[...]

PR #759, opened 2026-07-23
jit: retire two default-on FBW inline gates; const-class carrier-raise handler fix

youknowone:
Reduced to the unique delta over `main` after #723 (exc-edge try-block cluster) and #748 (gh#343 branchy plain-branch int-binop inline, landed unconditional) absorbed this branch's earlier work, and rebased onto current `main` (which includes #765 fixing the #763 build break).

## What this PR does

**1. Retire two default-on FBW inline gates** (`fd1a528`)

`PYRE_FBW_TRYBLOCK_INLINE` and `PYRE_FBW_CARRIER_RAISE` were verified default-on with no remaining reason to toggle off, so their behavior is made unconditional and the gates dropped:

- a try-block CALL whose except handler rejoins a loop always multi-frame-inlines; `decline_inline_caller_frame_for_catch_marker` now runs the any-loop rejoin test directly instead of branching between it and a specific-loop fallback.
- the depth-2 carrier-boundary raise delivery in `drive_bridge_carrier_walk` is always active.
- deletes the now-dead `exc_handler_rejoins_specific_loop` / `enclosing_loop_header_jit_pc` and the now-unused `call_jit_pc` parameter of `decline_inline_caller_frame_for_catch_marker`.
- `PYRE_CARRIER_EXC_RESUME` stays (default-off lever).

**2. gate-triage note** (`12ae744`)

The branchy inlined-callee continuation gap's plain-branch subcase is closed unconditionally by #748; its exception-unwind subcase is handled by the now-unconditional try-block callee inline + carrier-boundary raise delivery. `PYRE_CARRIER_EXC_RESUME` recorded as kept (default-off; its `bridge_guard_exc` GC-rooting and unconditional `execute_ll_raised` assign are the parity gaps to close before enabling).

**3. const-class carrier-raise handler fix** (`7b219bb`)

Set `class_of_last_exc_is_const` on the carrier-raise handler entry. #723 shipped the `carrier_raise_seed` branch without setting it.

## Verification

`python3 ./pyre/check.py`: dynasm 303/303, cranelift 303/303, wasm 300/300.

— *commented by Claude*


youknowone:
## Review dispositions

Addressed the automated reviews (Codex + CodeRabbit + CI parity review). Two new commits: `2eb3ea3b3d2` (retire the three verified default-on gates) and `d153e5126e3` (the confirmed handler-flag fix + caveats).

**Fixed**
- **`bridge_subwalk.rs` — `class_of_last_exc_is_const` stale on the carrier-raise handler entry** (CodeRabbit, Major): confirmed real. The carrier-boundary raise branch seeded `last_exc_value` but left the flag at its pre-bridge value, unlike the sibling `exc_edge_catch_target` branch and the walk-level SubRaise routing (`opimpl_raise`). A reraise in the handler read stale exception state. Set it true to match. A reraise-in-handler repro is bit-exact (875000) on both backends and its carrier bridge compiles (`loops_aborted=0`).
- **`resume_snapshot.rs:1112` — `.zip()` required a resolvable loop header even in the ANY-loop branch** (CodeRabbit, Minor): already resolved by `2eb3ea3b3d2`, which retires the gate and rewrites the decline to `try_catch_exception_at(...).is_some_and(exc_handler_rejoins_loop)` — no `enclosing_loop_header_jit_pc` dependency.
- **`gate-triage.md` — document `PYRE_CARRIER_EXC_RESUME`** (CodeRabbit, Minor): added to the WIP-keep inventory with its pre-flip-on parity gaps.

**Not a bug (verified)**
- **`trace.rs` — "probe the carrier catch from the live marker" / backward scan misses the handler** (Codex, P2): empirically refuted. `PYRE_P2_DIAG` shows `carrier_root_catch_target` (backward `find_catch_before_resume_live`) *finds* the root handler for every carrier try-block callee — `catch_before=Some(2892) rejoins=true` on `exceptions`, `Some(1077)` on `break_except`, `Some(2820)` on the reraise repro — and the bridges compile with `loops_aborted=0`. The forward `try_catch_exception_at` returns `None` at the carrier `root_pc` (the CALL's post-call `-live-`, catch sits behind); the forward probe is for the inline-call `op.next_pc` coordinate, a different site.

**Deferred (inert / staged)**
- **CI parity §2 — `PYRE_CARRIER_EXC_RESUME` (`bridge_guard_exc` GCREF-as-i64 not a traced root; `execute_ll_raised` seed guarded by `is_null` rather than unconditional; default-off vs PyPy always-on)**: this gate is default-off and inert (the `bridge_guard_exc` integer is stored but only dereferenced under the gate). These are the parity items to close before it is enabled for the #343/#126 depth-2 exception-resume slice; documented in `trace_ctx.rs` / `gate-triage.md` and tracked as a follow-up.

**Structural adaptations (intended)**
- CI parity §2 `carrier.recipes.len() == 1` — the delivery targets the depth-2 single-recipe carrier; deeper/return-out handlers stay on the residual path (correct, just not bridged) — incremental scope.
- CI parity §2 TLS `FBW_CARRIER_RAISE_SEED` — set in `drive_bridge_carrier_walk`, consumed once in `dispatch_via_miframe` across the carrier boundary, with a defensive post-walk clear; the walk-context threading is a follow-up.
- CI parity §4 sentinel-`w_code` skip and static `exc_handler_rejoins_*` analysis — deliberate adaptations to Pyre's synthetic virtualizable frames and CPython-compatible JitCode coordinates (commented in-code).

`check.py` dynasm 301/301, cranelift 301/301, wasm 297/297.

— *commented by Claude*

PR #762, opened 2026-07-24
jit(gh#346): model a fieldless Rust enum as its discriminant integer (Layer 1+2)

youknowone:
## Summary

Retires the fieldless-enum (C-like enum) `SomeInstance` model in the JIT front-end and replaces it with the orthodox RPython representation: **a fieldless enum is a named integer (`SomeInteger`)**, modeled by-value as its discriminant tag at every site. RPython has no enum type; upstream models a fieldless enum as plain module-level `int` constants with variant names in a side table.

This removes the `RootClassRepr.getclsfield: MissingRTypeAttribute("__discriminant")` rtype (phaseB) failure the compiler-emitted opcode-dispatch `match` hit — the enum value *is* the tag, so its discriminant read has no aggregate base to resolve a class-field against.

## Commits

1. **collapse the fieldless-enum Debug `format!` chain to a discriminant switch** (Layer 1) — lower `format!("… {:?}", enum)` over a fieldless enum to a per-variant constant-message switch, retiring the graph-less `fmt::rt::Argument::new_debug` extern.
2. **drop residual `last_caught_exception_value` stores after field removal** — prerequisite build-fix: an origin/main merge-order break (a later PR referenced a field a prior PR removed from `MIFrame`/`InlineFrame`) that broke the census build config. Dead stores, removed.
3. **model a fieldless enum as its discriminant integer** (Layer 2 A+B) — `tyref_to_value_type` colors a fieldless enum `Int`; `Rvalue::Discriminant` is the identity on the value; `Rvalue::Aggregate` of a fieldless variant folds to `ConstInt(discriminant)`. The Debug-fmt collapse recovers the enum identity from the value's origin `Input` op `class_root` (a side table keyed by the enum type) and switches on the value directly.
4. **retire the fieldless-enum base/variant ClassDef** (Layer 2 C) — `discriminant_value(&enum)` aliases its argument for a fieldless pointee; `derive_program_metadata` stops minting the `__discriminant`-only base + per-variant subclasses for fieldless enums (payload enums keep the sum-type base). `enum_variant_by_discriminant` is kept.

## Verification

- **census (two-phase rtyper)**: phaseB heads 47 → 46 (`OpcodeStepExecutor::call_intrinsic_1` `getclsfield` wall retired); raw phaseA 303 → 302 (`pyre_object::module::w_module_get_w_dict` also retired); `:1919` direct-miss 35 → 35 (no regression); no new walls, no inline-field (`ListStrategy`/`ExcKind`) staircase.
- **check.py**: dynasm 303/303, cranelift 303/303 bit-exact.
- **cargo test -p majit-translate**: 3092 passed.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -1182,27 +1182,34 @@ fn derive_program_metadata(
                 // discriminants, so unsigned is exact. A negative-discriminant
                 // enum would need a signed width here AND a sign-extending field
                 // read; revisit if one is introduced.
-                let disc_ty = if variants.iter().all(|v| v.fields.is_empty()) {
-                    match td.layout_for_target(&target).and_then(|l| l.size) {
-                        Some(1) => "u8",
-                        Some(2) => "u16",
-                        Some(4) => "u32",
-                        _ => "i64",
-                    }
-                } else {
-                    "i64"
-                };
-                let rows: Vec<(String, String)> =
-                    vec![("__discriminant".to_string(), disc_ty.to_string())];
-                struct_fields.fields.insert(name.clone(), rows.clone());
-                struct_fields.fields.insert(leaf.clone(), rows.clone());
-                struct_fields.fields.insert(canon_base.clone(), rows);
-                // Object identity for the enum base (the discriminant
-                // carrier), minted from the crate-stripped `module::Enum`.
+                // A FIELDLESS (C-like) enum is modeled by-value as its
+                // discriminant integer (`tyref_to_value_type` colors it
+                // `Int`; `Rvalue::Discriminant` / `discriminant_value` alias
+                // that int; the constructor folds to `ConstInt`), so it has
+                // no `SomeInstance` base and needs no `__discriminant`-only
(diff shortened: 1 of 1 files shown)


youknowone:
Follow-up census detail (phaseB): this change advances `call_intrinsic_1`/`_2` from a **phaseA (annotate)** wall to a **phaseB (rtype)** wall rather than fully lifting them. Pre-change they failed phaseA on the graph-less `fmt::rt::Argument::new_debug` extern (→ residualize). Post-change they annotate cleanly (the phaseA direct-miss retirement 37→35 is real), reach phaseB, and now fail rtype with `RootClassRepr.getclsfield: MissingRTypeAttribute("__discriminant")` — the synthesized `getattr("__discriminant")` has no registered clsfield on the fieldless enum's `RootClassRepr` (a fieldless enum is modeled as `SomeInstance` with no real discriminant field to rtype against). Net runtime behavior is unchanged (still residualizes; check.py 300/300 bit-exact), so this is not a correctness regression — it is the next wall the same graph hits, and it is exactly what the North Star (model a fieldless enum as an integer discriminant) removes. Tracking that as the follow-up.

— *commented by Claude*

PR #764, opened 2026-07-24
import: __main__ identity, circular-import diagnostics, module getattr precedence, imp completeness

youknowone:
Continues the import-compatibility work. Each change was verified against
PyPy3 (the compat oracle) and CPython 3.14; every case below now matches
PyPy3 exactly.

## Changes

- **`__main__` identity attributes** — seed `__spec__` / `__package__` /
  `__doc__` = `None` on the `__main__` namespace before user code runs, so a
  bare reference no longer falls through to `builtins`. Set `__file__`
  (absolutized) / `__cached__` = `None` for a script run by path. Fixes
  `if __spec__ is None` main-detection (multiprocessing spawn, runpy, pytest)
  and `NameError: __doc__`.

- **Circular-import diagnostics** — `from a import x` and `a.x` on a module
  still executing now report *"partially initialized module 'a' ... (most
  likely due to a circular import)"*, and an unset submodule reports *"cannot
  access submodule ..."*, driven by `__spec__._initializing` /
  `_uninitialized_submodules`.

- **Module `__getattr__` precedence** — a module consults its own dict's PEP
  562 `__getattr__` before the receiver type's class-level `__getattr__`, and
  a class-level `__getattr__`'s own `AttributeError` reaches the caller
  instead of being overwritten by the generic module-miss message.

- **`_imp.is_builtin`** — returns `-1` for a builtin already in `sys.modules`
  (not re-initializable), `1` for a not-yet-imported builtin, `0` otherwise.

- **`_imp.init_frozen`** — added as the legacy no-op returning `None`.

- **`sys.dont_write_bytecode`** — defaults to `False`, consistent with
  `sys.flags.dont_write_bytecode` (no bytecode cache is written regardless).

Also included (earlier on this branch): module subclass `__repr__`/`__str__`
override dispatch in `repr`/`str`, and `chmod` mode validation as `c_int`.

## Verification

- Each case matches PyPy3 (and CPython where they agree; circular-import
  wording follows PyPy3 per the faithful-error-message convention).
- `check.py` dynasm 300/300 ×2 and cranelift 300/300 ×2, all green.

— *commented by Claude*

### DIFF
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -5454,10 +5454,12 @@ unsafe fn module_getattr_hook_or_err(
         Some(w) if !w.is_null() && pyre_object::is_str(w) => {
             pyre_object::w_str_get_wtf8(w).to_string()
         }
-        _ => return Err(PyError::new(
-            PyErrorKind::AttributeError,
-            format!("module has no attribute '{name}'"),
-        )),
+        _ => {
+            return Err(PyError::new(
+                PyErrorKind::AttributeError,
+                format!("module has no attribute '{name}'"),
+            ));
+        }
     };
     // module.py:148-159 — a module still executing (`__spec__._initializing`)
     // or naming an as-yet-unset submodule reports the circular-import cause.
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -5450,10 +5450,8 @@ unsafe fn module_getattr_hook_or_err(
     // when it is a string, falling back to the bare form otherwise
     // (module.py:143-162).  Own the name up front so the `__spec__` lookups
     // below, which allocate, cannot dangle the dict-borrowed slice.
-    let nm = match finditem_str(w_dict, "__name__")? {
-        Some(w) if !w.is_null() && pyre_object::is_str(w) => {
-            pyre_object::w_str_get_wtf8(w).to_string()
-        }
+    let w_name = match finditem_str(w_dict, "__name__")? {
+        Some(w) if !w.is_null() && pyre_object::is_str(w) => w,
         _ => {
             return Err(PyError::new(
                 PyErrorKind::AttributeError,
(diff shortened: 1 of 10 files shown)


youknowone:
Follow-up: implemented the previously-deferred items too, each verified against CPython 3.14 and PyPy3 (respecting PyPy source structure, matching CPython 3.14 behavior).

- **`-B` / `-O` / `-OO`** — the launcher now parses these plus `PYTHONDONTWRITEBYTECODE` / `PYTHONOPTIMIZE` (`_Py_get_env_flag` semantics, `-E`/`-I` gated). `sys.flags.optimize` / `sys.flags.dont_write_bytecode` / `sys.dont_write_bytecode` reflect them; the RustPython compiler's existing `optimize` support now strips `assert` / makes `__debug__` False (`-O`) and discards docstrings (`-OO`); `compile(optimize=-1)` resolves to `sys.flags.optimize`. All flag/env cases match CPython & PyPy3 exactly.

- **Import lock API** — `acquire_lock` / `release_lock` / `lock_held` implement the reentrant depth counter (`ImportRLock`); `release_lock` on a zero count raises `RuntimeError("not holding the import lock")`; `lock_held()` is True during an import.

- **`_imp._fix_co_filename`** — replaces `co_filename` in place, recursing through nested code constants guarded by the root's original name; argument-type validation follows CPython 3.14 (rejects a bytes path; `argument N must be code/str` messages).

- **`_imp.create_dynamic`** — added, raising `ImportError("Not implemented")` (the no-C-extension branch) so the meta-path `hasattr` probe matches while `except ImportError` still falls back.

- **Bootstrap `__module__`** — `importlib._bootstrap{,_external}` now execute under their frozen `__name__` (`_frozen_importlib{,_external}`), so their classes' `__module__` matches CPython/PyPy while the module `__name__` stays `importlib._bootstrap{,_external}`.

`check.py` dynasm 300/300 ×2 and cranelift 300/300 ×2, all green. `pyre -OO` runs stdlib imports without regression.

— *commented by Claude*

youknowone:
Addressed the automated review (Codex inline + Codex parity), each verified against CPython 3.14 and PyPy3. Rebased onto the fixed `origin/main` (the `prepare Charon/LLBC` failure was a base-side compile error from #763 referencing the field #756 removed — fixed on main by #765, not caused by this PR).

**Fixed (3):**
- **Module `__spec__` helpers no longer swallow exceptions** — `get_spec` / `is_spec_initializing` / `is_spec_uninitialized_submodule` now return `Result` and suppress only a missing attribute (`AttributeError`), propagating any other lookup error, the truth test, and the containment test (both the `import_from` message path and the module-getattr miss). This follows `_PyModuleSpec_IsInitializing` / `_PyModuleSpec_IsUninitializedSubmodule` (`PyObject_GetOptionalAttr`) semantics; verified `pyre == CPython 3.14` across a matrix of `__spec__` shapes (raising `_initializing` property, missing `_uninitialized_submodules`, raising `__contains__`, etc.).
- **WTF-8 module `__repr__`/`__str__`** — a `types.ModuleType` subclass whose override returns a lone-surrogate string previously panicked in `w_str_get_value`; `py_repr_wtf8` / `py_str_wtf8` now read it through `w_str_get_wtf8` (and `str()` without a `__str__` override routes back through `py_repr_wtf8`). Matches both oracles.
- **Sandbox `__file__`** — a relative script path is now resolved against `sys_path_cwd()` (the seam-provided virtual cwd) before normalizing, so `std::path::absolute` never reads or leaks the trusted process cwd. Off sandbox the result is unchanged.

**Declined (2), with reasoning:**
- **Module dunder descriptor binding** — every dunder dispatch in `display.rs` (int/float/str subclass, metaclass) uses the same raw `call_and_check(method, &[obj])` convention, and the common plain-function case is already equivalent to `get_and_call_function`'s fast path. Binding only the module branch would make it inconsistent with its siblings; a proper fix is a separate file-wide change.
- **`sys.dont_write_bytecode == False`** — both CPython 3.14 and PyPy3 default this to `False` (with `sys.flags.dont_write_bytecode == 0`). Reporting `True` would diverge from both oracles; the absence of a `.pyc` writer is an orthogonal limitation, not a reason to misreport the flag.

Codex parity §3/§4 (import-lock atomic counter, immutable-constant rebuild in `_fix_co_filename`, CPython env-flag semantics) are intended structural adaptations, already documented, with no parity regression.

`check.py` dynasm 303/303 ×2 and cranelift 303/303 ×2, all green.

— *commented by Claude*

PR #768, opened 2026-07-24
jit(fbw): exception-object virtualization perf + polymorphic-raise resume fix

youknowone:
Extends the FBW walker's exception virtualization and closes a pre-existing
blackhole resume-liveness crash on polymorphic `raise <expr>` loops.

## Perf: virtualize more exception raises + fold more attributes

- **Zero-argument construction** (`raise ValueError()`): the args-list
  `ListStrategy::Empty` build is emitted inline (`NewWithVtable` +
  `SetfieldGc(strategy=Empty)`), mirroring `w_list_new_with_strategy(vec![],
  Empty)`, so the constructor no longer falls to a residual heap alloc.
  ~4900ns → ~1ns.
- **Float-argument construction** (`raise ValueError(1.5)`): the Integer arm's
  mirror via `walker_unbox_float` + `emit_typed_list_inline(Float)`. Float
  subclasses stay Object-strategy (identity preserved). ~4750ns → ~1ns.
- **Bare-class raise** (`raise ValueError`): `try_walker_trace_raise_bare_class`
  synthesizes the zero-arg instance inline and marks it FBW-built so the raise
  records its traceback through the virtual `record_fresh_application_traceback`
  path (staying virtual and DCE-able) instead of the escaping runtime hook.
  ~2644ns → ~1ns.
- **Widen the exception attribute slot fold** to eight more `W_BaseException`
  attributes (`__traceback__`, `name`, `obj`, and the Unicode `object`/`start`/
  `end`/`reason`/`encoding`), load-fold only. Keeps a `__traceback__`-reading
  handler on an otherwise-virtualized exception from escaping.

## Correctness: keep the raise operand in the filtered resume liveness

Fixes a pre-existing crash: a polymorphic `raise <expr>` loop (where the
exception is a per-iteration residual-call result) panics on the type-change
deopt with `bhimpl_raise: excvalue must be non-null`.

`filter_liveness_in_place` narrows the codewriter `-live-` records to the Ref
colors that map to a Python local/stack slot, dropping operand-stack
temporaries. A `raise/r <reg>` reads its exception operand from such a
temporary — the residual `normalize_raise_varargs` result the tracer stored via
`make_result_of_lastop` — so the filter dropped it from the `-live-` preceding
the raise. On the deopt the guard (`GUARD_CLASS`, or the after-residual
`GUARD_NO_EXCEPTION` / `GUARD_NOT_FORCED`) resumed at that marker and the
blackhole seeded `registers_r` without the exception → assert.

The fix re-adds the `raise` op's Ref operand register to the retained set,
extending the existing precedent that re-adds LOAD_ATTR's tracer-populated
residual operand. It mirrors `get_list_of_active_boxes` keeping every register
read after the marker, closing the deferred `num_live`-aware
`capture_resumedata(after_residual_call=True)` gap for the raise operand.

## Verification

- `check.py`: dynasm 303/303, cranelift 303/303.
- Exception attribute reads byte-identical to pypy3 across all folded attributes.
- A 17-shape polymorphic-`raise` corpus (both backends) is crash-free and
  byte-identical to pypy3.

## Summary by CodeRabbit

* **Performance**
  * Improved JIT handling of exception attributes and traceback-related fields, including more precise slot-to-field mapping.
  * Optimized canonical bare-class `raise Type` paths to reduce unnecessary residual exception work.
  * Enhanced exception construction to better support float and empty argument list strategies.
* **Bug Fixes**
  * Improved preservation of exception state during optimized raise/resume flows.
  * Improved Unicode error and attribute-error details for exception attribute access.
  * Unified unbound-local error construction across load and delete cases.

### DIFF
--- a/pyre/bench/synth/exception_traceback_loop_forms.py
+++ b/pyre/bench/synth/exception_traceback_loop_forms.py
@@ -1,0 +1,169 @@
+# Traceback nodes recorded by a compiled loop, driven by BOTH loop forms.
+#
+# The sibling exception-traceback fixtures all drive their workload with
+# `for i in range(N)`. The two loop forms compile through different paths and
+# have failed independently: the same callee raising into a `while` loop lost
+# every node while the `for` twin lost exactly one, so a `for`-only fixture
+# reports the milder symptom or none at all.
+#
+# Each shape reports the SET of chain depths it observed rather than one
+# sample, so a compiled iteration that disagrees with the interpreted ones
+# shows up as a second element instead of being averaged away. A node built on
+# both sides of a frame boundary shows up the same way.
+#
+# The traceback walk is repeated inline in each shape on purpose: moving it
+# into a helper adds a call to the loop body and changes what the body traces,
+# which is enough to hide the divergence.
+
+N = 20000
+
+
+def out(key, value):
+    print(f"{key} = {value}")
+
+
+def thrower(i):
+    raise KeyError(i)
+
+
+def while_callee(n):
+    depths, missing, i = set(), 0, 0
+    while i < n:
+        try:
+            thrower(i)
+        except KeyError as e:
+            tb = e.__traceback__
+            if tb is None:
+                missing += 1
+            else:
+                d = 0
+                while tb is not None:
+                    d += 1
+                    tb = tb.tb_next
+                depths.add(d)
+        i += 1
+    return sorted(depths), missing
+
+
+def for_callee(n):
+    depths, missing = set(), 0
+    for i in range(n):
+        try:
+            thrower(i)
+        except KeyError as e:
+            tb = e.__traceback__
+            if tb is None:
+                missing += 1
+            else:
+                d = 0
+                while tb is not None:
+                    d += 1
+                    tb = tb.tb_next
+                depths.add(d)
+    return sorted(depths), missing
+
+
+def while_same_frame(n):
+    depths, missing, i = set(), 0, 0
+    while i < n:
+        try:
+            raise ValueError(i)
+        except ValueError as e:
+            tb = e.__traceback__
+            if tb is None:
+                missing += 1
+            else:
+                d = 0
+                while tb is not None:
+                    d += 1
+                    tb = tb.tb_next
+                depths.add(d)
+        i += 1
+    return sorted(depths), missing
+
+
+def for_same_frame(n):
+    depths, missing = set(), 0
+    for i in range(n):
+        try:
+            raise ValueError(i)
+        except ValueError as e:
+            tb = e.__traceback__
+            if tb is None:
+                missing += 1
+            else:
+                d = 0
+                while tb is not None:
+                    d += 1
+                    tb = tb.tb_next
+                depths.add(d)
+    return sorted(depths), missing
+
+
+def while_residual_raise(n):
+    """A builtin operation raises for the frame -- no `raise` statement."""
+    depths, missing, i, zero = set(), 0, 0, 0
+    while i < n:
+        try:
+            _ = i // zero
+        except ZeroDivisionError as e:
+            tb = e.__traceback__
+            if tb is None:
+                missing += 1
+            else:
+                d = 0
+                while tb is not None:
+                    d += 1
+                    tb = tb.tb_next
+                depths.add(d)
+        i += 1
+    return sorted(depths), missing
+
+
+def while_bare_reraise(n):
+    depths, missing, i = set(), 0, 0
+    while i < n:
+        try:
+            try:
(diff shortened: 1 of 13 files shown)


youknowone:
## CI triage + review response

### CI: all three red jobs accounted for

| job | verdict |
| --- | --- |
| `pyre/check.py` (macos + ubuntu) — `synth/const_arg_call_resume` | **pre-existing on `main`.** `main@4ec286d5` fails the same test on both runners; its cranelift numbers are `0.38s` (macos) / `0.43s` (ubuntu) against this PR's `0.29s` / `0.43s`. |
| `pyre/check.py` (ubuntu) — `cranelift int_loop` | **boundary flip of a known gap.** pyre's absolute time is `0.41s` on both this PR and `main@4ec286d5`; only the pypy reference run moved (`0.15s` → `0.16s`), which tipped the `x2` gate. |
| `cargo test` (ubuntu) — `eval::tests::test_hot_user_function_loop_survives_compiled_trace` | **flake.** Not reproducible locally: arm64 native and x86_64 under Rosetta, 393/393 on 5 consecutive full-suite runs each. Re-running the job on the same commit passed. The diff's only non-exception hunk is gated on `opname == "raise"`, and that test raises nothing. |

### Review findings — both confirmed and fixed

**1. Float exception args were not guarded by exact class** (Codex P2) — confirmed, and it applied to the Integer arm too.

`list_strategy_for` picks Integer/Float from `is_plain_int1` / `all_floats`, which are strict on `w_class`, but the emitted guard was `walker_unbox_{int,float}`'s `GuardClass` on the payload `ob_type` — which a subclass instance shares, retagging only `w_class`. A later iteration passing a subclass stayed on the trace and was unboxed into the typed storage.

Discriminator (warm the loop with an exact float/int, then re-enter the same compiled loop with a subclass, and read `type(e.args[0]).__name__`):

| | float | int |
| --- | --- | --- |
| pypy3 | `MF` | `MI` |
| pyre `PYRE_NO_JIT=1` | `MF` | `MI` |
| pyre JIT, before | `float` | `int` |
| pyre JIT, after | `MF` | `MI` |

Fixed by pinning the exact `w_class` at both emit sites. A constant operand is the same object on every iteration, so it keeps the bare unbox and the hot path is unchanged (`raise ValueError(1.5)` measures `0.93 ns/iter`).

**2. Decline after emitting bare-raise IR** (Codex P2 / CodeRabbit Major) — confirmed. `walker_ensure_execution_context` was resolved past the commit marker in `try_walker_trace_raise_bare_class`, so a decline there left the class guard, the inline args list and the `NewWithVtable` recorded while the caller fell through to the generic `RaiseVarargs` residual. Resolved before the commit marker instead; that also keeps the recovery getfield from landing after a guard that references it, which is why `ensure_execution_context` recovers eagerly at walk entry.

### Codex parity review (sections 3 and 4)

- **§3 `w_exception_get_traceback` omits `tb.frame.mark_as_escaped()`** — deferred, blocker named. `PyFrame::mark_as_escaped` does exist (the doc comment claiming otherwise was stale and has been corrected), but the flag's only consumer is `ExecutionContext::leave`, whose `frame_vref()` force is still a no-op until `jit.virtual_ref` is ported — and a frame a traceback can reach has already left with `got_exception` set, so `leave` marked its `f_back` regardless. The mark lands with that force, together with the `__traceback__` arm of `exception_attr_slot_fold`, which folds this read to a raw slot load in traced code. Both sites now say so.
- **§4 both items** — kept as documented adaptations; each already cites its upstream decision point (`w_list_new_with_strategy(vec![], Empty)` and `readwrite_attrproperty_w` / `GetSetProperty.descr_property_get/set`).

### Two pre-existing defects found while probing — not part of this PR

- `is` on float/int diverges under the JIT: `y = x + 0.0; y is x` in a hot loop is `True` on pypy3 and on pyre with `PYRE_NO_JIT=1`, `False` under the JIT. No exception is involved. `is` for float/int is value-based on both engines (`W_Root.is_w` / `immutable_unique_id`), so the JIT's pointer compare is the divergence.
- Zero-argument user-function calls are never inlined: `f0()` in a hot loop costs `441 ns/iter` against `1.03 ns` for `f1(1)` and `0.51 ns` on pypy3. `inline_call.rs` declines at `nparams == 0` with the rationale "a residual zero-arg call is cheap", which the measurement contradicts; removing that early return alone only moves it to `548 ns`, so at least one further gate is involved.

### Gates after the fixes

`check.py` dynasm 303/303 and cranelift 303/303; the 17-shape polymorphic-`raise` corpus and the exception-attribute oracle stay byte-identical to pypy3 on both backends.

— *commented by Claude*

PR #771, opened 2026-07-25
jit: fold the LOAD_ATTR bound-builtin-method getattr into a traced Method

youknowone:
`lst.append(x)` in a hot loop kept an opaque `getattr` residual in the
compiled trace, so every iteration ran an MRO walk, allocated a fresh
`Method`, and paid a `GUARD_NOT_FORCED` / `GUARD_NO_EXCEPTION` pair plus a
second `bh_load_method_self_fn` call.

## Why the interpreter gate is not the bug

`load_method_fast_path` declines for `lst.append` because
`flag_method_descriptor` is set only on `Function.typedef`
(`pypy/interpreter/typedef.py:807`). `pypy/objspace/std/callmethod.py:25-80`
restricts its `[w_descr, w_obj]` push the same way, so PyPy also falls
through to `space.getattr` and materialises a `Method`. The port is
faithful and is left alone.

## The gap is at trace level

`PYPYLOG=jit-log-opt` on the same loop shows PyPy's `#18 LOAD_METHOD` merge
point emitting **zero** operations: `space.getattr` is ordinary RPython, so
the meta-tracer traces through it, the MRO lookup constant-folds under
`guard_class` + `guard_not_invalidated`, and the `Method` virtualizes away.
pyre's `getattr` is an opaque residual, so the tracer has to fold it — the
same role `try_walker_specialize_load_attr` / `..._load_method_attr`
already play.

## Change

* `bound_method_attr_fast_path` (beside `load_method_fast_path`) names the
  shape `getattr` reduces to `w_method_new(w_descr, obj, w_type)` purely:
  default `__getattribute__`, cacheable version tag, receiver class
  identical to its `w_class` slot, no instance dict at all, and a
  `FUNCTION_TYPE` descriptor with builtin code.
* `try_walker_specialize_load_bound_method_attr` emits `guard_class` +
  `guard_value(w_class)` + `guard_value(version_tag)` +
  `NewWithVtable(Method)`, so the following CALL reads `w_function` /
  `w_self` back off it and the allocation dies. Restricted to the top
  full-body frame for the reason `try_walker_orthodox_list_append`
  documents (a fold's guards inside an inlined callee collapse resume to
  the caller's CALL and would re-run the callee's earlier side effects).
* `try_walker_fold_load_method_self` answers the companion residual with a
  `PY_NULL` constant under a Method class guard instead of declining —
  `compute_load_method_bound` returns PY_NULL for any `is_method(attr)`.
* `Method`'s descr group gains the inherited `PyObject.w_class` so the
  inline emit's header store is a virtual field of the same size descr.

## Numbers

`synth/const_arg_call_resume`, check.py's own execution-only ratio against
pypy (header gate is x30):

| | before | after |
| --- | --- | --- |
| dynasm | x22.7 | x9.5 |
| cranelift | x28.1 | x12.5 |

The `r.append(i)` loop alone goes x32.2 -> x5.0, and its compiled loop drops
from 50 ops with one `call_may_force` to 45 with none. Against CPython the
whole fixture goes from x4.95 (slower than the interpreter) to x2.4.

The measurement is check.py's metric — user CPU minus an empty-program run
(`_run_timed_unix` + `_exec_time`), not wall clock.

## Verification

`python3 ./pyre/check.py --backend dynasm,cranelift`: 309/309 on both.
`cargo test --all --no-default-features --features dynasm`: green.
`cargo fmt --all --check`: clean.

New fixture `pyre/bench/synth/bound_method_builtin_fold.py` pins the guards:
a `__slots__` subclass whose method is replaced mid-loop (version tag), an
instance-dict entry shadowing the method name (the fold must decline), a
polymorphic receiver at one call site (class guard), and a bound method that
outlives its CALL (materialization). Output matches CPython 3.14 and PyPy.

Not specific to this fixture: it covers `obj.method()` on any dict-less
builtin receiver (list / dict / set / str / ...).

— *commented by Claude*

### DIFF
--- a/pyre/bench/synth/bound_method_builtin_fold.py
+++ b/pyre/bench/synth/bound_method_builtin_fold.py
@@ -8,8 +8,9 @@
 # that fold depends on: a __slots__ subclass whose method is replaced mid-loop
 # (version tag), an instance-dict entry shadowing the method name (the fold
 # must decline for a dict-bearing receiver), alternating receiver types at one
-# call site (class guard), and a bound method that outlives its CALL (the
-# virtual must materialize). Output verified against CPython/PyPy.
+# call site (class guard), `__`-names that bind through the same shape, and a
+# bound method that outlives its CALL (the virtual must materialize). Output
+# verified against CPython/PyPy.
 N = 20000
 
 
@@ -67,6 +68,17 @@ def polymorphic(n):
     return len(a), len(b)
 
 
+def dunder_names(n):
+    # `__`-names bind through the same `function` + builtin-code shape as
+    # `append`, so the fold applies to them too.
+    lst = [0] * 8
+    total = 0
+    for i in range(n):
+        total += lst.__len__() + (i,).__len__()
+        lst.__setitem__(i & 7, i)
+    return total, lst[3]
+
+
 def escaping_bound_method(n):
     r = []
     keep = []
@@ -83,4 +95,5 @@ print(dict_and_set(N))
 print(slots_subclass_override(N))
 print(shadowed(N))
 print(polymorphic(N))
+print(dunder_names(N))
 print(escaping_bound_method(N))
--- a/pyre/bench/synth/bound_method_builtin_fold.py
+++ b/pyre/bench/synth/bound_method_builtin_fold.py
@@ -1,0 +1,86 @@
+# `lst.append(x)` on a builtin receiver: the name resolves to a builtin-code
+# function on the type, so LOAD_ATTR's method fast path declines (it only
+# pushes `[descr, self]` for `flag_method_descriptor` types) and `getattr`
+# materialises a bound method. The tracer folds that construction to
+# guard_class + guard_value(w_class) + guard_value(version_tag) +
+# new_with_vtable(Method) so it virtualizes into the following CALL, instead of
+# leaving an opaque getattr residual in the loop. These cases pin the guards
+# that fold depends on: a __slots__ subclass whose method is replaced mid-loop
+# (version tag), an instance-dict entry shadowing the method name (the fold
+# must decline for a dict-bearing receiver), alternating receiver types at one
+# call site (class guard), and a bound method that outlives its CALL (the
+# virtual must materialize). Output verified against CPython/PyPy.
+N = 20000
+
+
+def build(n):
+    r = []
+    for i in range(n):
+        r.append(i)
+    return r
+
+
+def dict_and_set(n):
+    d = {}
+    s = set()
+    for i in range(n):
+        d.setdefault(i & 7, i)
+        s.add(i & 15)
+    return len(d), len(s)
+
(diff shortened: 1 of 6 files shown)


youknowone:
## Codex parity review dispositions

**§1 (parity regressions):** None reported.

**§2 (other mismatches introduced by this patch):** 2 findings — 1 fixed, 1 won't-fix (documented).

### §2-b — `name.contains("__")` has no upstream equivalent → **fixed** (e385336d544)

Confirmed the finding and removed the carve-out. It had been copied from
`try_walker_specialize_load_method_attr` without checking it applied here. Empirically on this
build, `list.append` and `list.__len__` are *both* `<class 'function'>` and both bind to `method`,
so the fold's `get()` arm treats them identically; `getattr_str` has no dunder shortcut either.
Receivers that resolve `__`-names through slots (weakref proxy, `GenericAlias`) are already
rejected by the existing `has_object_getattribute` gate, so the carve-out was buying nothing.

Coverage check — same loop shape, dunder vs non-dunder, optimized trace:

| loop body | final ops | `call_may_force` |
|---|---|---|
| `lst.__len__()` | 30 | 1 |
| `lst.count(0)` | 32 | 1 |

The remaining `call_may_force` in both is the method body itself; the `bh_load_attr` getattr
residual is gone in both. Before this commit the dunder form kept it.

Added `dunder_names()` (`__len__` on a list, `__len__` on a tuple, `__setitem__`) to
`bench/synth/bound_method_builtin_fold.py`.

### §2-a — `getdict_backing().is_null()` is stricter than `callmethod.py:60-64` → **won't-fix (documented)**

Upstream admits a dict-bearing receiver and merely checks that `name` is absent from the dict.
This fold cannot: it emits **no dict-shape guard**, so a later store of `name` into the instance
dict would not side-exit and the trace would keep calling the type-level function. The dictless
requirement *is* the soundness condition for the guard set actually emitted, which is why Codex
also classifies it as safe. Admitting those receivers means emitting that guard first — recorded
as the convergence path in a comment on `bound_method_attr_fast_path`.

### Verification (base `c38f0f2dc83`)

- `check.py --backend dynasm,cranelift` — **309/309 both**
- Fixture vs CPython 3.14 oracle (identical to PyPy): dynasm/cranelift × JIT/`PYRE_NO_JIT=1` — **4/4 match**
- `cargo fmt --all --check` — clean

One caveat, stated plainly: `cargo test -p majit-translate --test test_rbigint_mir` fails 4/5
locally. That is a stale local Charon/LLBC extraction (`build/llbc/*.ullbc` dated Jul 23; #761's
rbigint port landed Jul 25), not this patch — `majit-translate` has no dependency on any `pyre`
crate. Confirmed: after `scripts/extract-llbc.py pyre-object` + `pyre-interpreter`, the test is
**5 passed / 0 failed**.

— *commented by Claude*

PR #778, opened 2026-07-25
jit: model maybe-null type pointers as niche Option<NonNull<PyObject>>

youknowone:
## What

Retires a class of two-phase-rtyper `[PREPASS phaseB fail]` census heads (gh#346) by remodelling `gettypefor` and `r#type` — the two producers of a maybe-null type object — from a two-word `Option<PyObjectRef>` to a one-word niche `Option<NonNull<PyObject>>`.

A niche `Option<NonNull<T>>` is a single pointer word: `None` is the null pointer, `Some(p)` is the non-null pointer. This matches RPython exactly — RPython has no `Option` type; a maybe-null `W_Root` **is** a nullable pointer, and `w is not None` is a pointer null-test. The front-end folds the niche `Option`'s discriminant read to `ptr_ne(p, null)` and its `Some(x)` payload read to the pointer identity, so the rtyper prepass no longer reads a `__pos_0` payload that rtypes to `Void`.

## How

Six commits, bottom-to-top:

- **interp: fold null-valued Option<PyObjectRef> producers to None** — the front-end recognizes a producer whose `Some` arm is a proven-null pointer and collapses it to `None`.
- **jit: fold niche Option<NonNull<T>> discriminant/payload to pointer null-test** — `Rvalue::Discriminant` becomes `ptr_ne(base, null)`; the `Some(x)` `__pos_0` read becomes the identity on the pointer.
- **jit: fold niche Option<NonNull> in the Option combinator post-passes** — `unwrap_or` / `map_or` / `unwrap` / `?` / `map`+`and_then`+`unwrap_or_else` residual-combinator lifts skip the synthetic `__discriminant`+`__pos_0` switch for a niche pointer Option.
- **jit: fold niche Option<NonNull> construction and route its null through null_mut()** — `Rvalue::Aggregate` construction folds (`None` → `null_mut()` call, `Some(p)` → identity); every niche null is a repr-adaptive `core::ptr::null_mut()` call rather than a fixed-GCREF `ConstRefNull`, so it unions/converts against the pointer receiver's own `InstanceRepr`.
- **interp: return Option<NonNull<PyObject>> from gettypefor** — producer swap; callers adapt at the leaf.
- **interp: return Option<NonNull<PyObject>> from r#type** — producer swap across the interpreter and the two `pyre-jit-trace` walker call sites; `if-let` / `match` / `let-else` control structures are left untouched (the niche fold handles them natively), `.as_ptr()` is added only at leaf sites that thread a `NonNull` into a `*mut`-typed helper, and `.map_or(NULL, |p| p.as_ptr())` at raw-consuming producers.

## Verification

- `cargo test -p majit-translate` — full suite green (3020 lib + all integration + 5 real-LLBC rbigint anchors), 0 failed.
- `check.py --backend dynasm,cranelift` — bit-exact **308/308** on both backends.

Rebased onto `main` (`#761` rbigint port); one conflict in `model.rs` resolved by keeping upstream's `tuple_owner_for_var` deletion plus this branch's `push_null_mut_ptr` addition.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -4226,6 +4226,35 @@ impl<'a> Lowering<'a> {
                     let base = self.resolve_place(mir_bb, place)?;
                     return Ok((None, base));
                 }
+                // A niche `Option<NonNull<T>>` is one pointer word: `None`
+                // = null, `Some(p)` = the non-null pointer.  Its
+                // discriminant is the pointer-null test `base != null`
+                // (`None` = 0, `Some` = 1), not a `__discriminant` field off
+                // a two-word aggregate base (there is none).  Emit the null
+                // constant, then compare — a `ne` on two `Ref` operands
+                // lowers to `ptr_ne` (`jtransform`), and its `Int` result
+                // feeds the match switch exactly as a `__discriminant` read
+                // would.  Modelling the maybe-null pointer this way is the
+                // orthodox representation (RPython has no `Option`; a
+                // maybe-null `Ptr` tests `ptr_nonzero`).
+                if self.tyref_is_niche_option_ptr(&place.ty) {
+                    let base = self.resolve_place(mir_bb, place)?;
+                    let bb_id = self.block_id[mir_bb];
+                    let nullc = self.graph.alloc_value_var();
+                    self.graph.block_mut(bb_id).operations.push(SpaceOperation {
+                        result: Some(nullc.clone()),
+                        kind: OpKind::ConstRefNull,
+                    });
+                    return Ok((
+                        Some(OpKind::BinOp {
+                            op: "ne".to_string(),
+                            lhs: base,
+                            rhs: nullc,
+                            result_ty: ValueType::Int,
+                        }),
+                        res,
+                    ));
+                }
                 let base = self.resolve_place(mir_bb, place)?;
                 Ok((
                     Some(OpKind::FieldRead {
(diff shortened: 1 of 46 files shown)


youknowone:
Re: the parity review's Section 1 (niche status not recorded for `Option::is_some`/`is_none` at `mir.rs` `recognize_is_none_site`, claimed to regress the `int()` fast path at `builtins.rs:6490`).

Verified against the two-phase-rtyper census (phaseA/phaseB heads) built with `--no-default-features --features dynasm` on this branch:

- **phaseB after this patch: 36 unique fns; the `isinstance_w` cluster retired** (only `isinstance_str_w` remains, a distinct fn). No `Option::is_none`/`is_some` head appears.
- The `int()` fast path at `builtins.rs:6490` uses `w_type.is_some()` + `w_type == w_int` on the niche `Option<NonNull>`; these `is_some`/`is_none` calls are **inlined by rustc** before residualizing, so they never reach `recognize_is_none_site` / `option_is_none::rewire_*`. No new phaseB head, no rtype `getclsfield` on `__discriminant`. check.py is bit-exact 308/308 on dynasm + cranelift, so there is no runtime regression either.

The structural point is nonetheless valid: `recognize_is_none_site` (`mir.rs`) omits the `niche` flag its siblings `recognize_unwrap_or_site`/`recognize_unwrap_site` carry, and `option_is_none.rs` always emits a `__discriminant` `FieldRead` with no pointer-null-test arm. That path is currently unreached for niche Options (they inline), so it is latent hardening rather than a live regression — tracked as a follow-up to mirror the `option_unwrap_or` niche arm into `option_is_none` should such a residual ever appear.

Separately: this PR's merged HEAD left one JIT-only leaf (`bound_method_attr_fast_path`, `baseobjspace.rs:7905`) calling `is_null()` on the now-`NonNull` result, which breaks the `--no-default-features --features dynasm` build; fixed on main in `beee634acb7`.

— *commented by Claude*

PR #787, opened 2026-07-25
majit: evict per-loop side tables on retirement and harden MAJIT_BRIDGE_ONLY parsing

youknowone:
Follow-up to #773, which was merged before its review comments were addressed.
Four findings from CodeRabbit and Codex; each is described with what it does and
does not affect.

## Changes

* **`pyjitpl.rs` — per-loop side tables outlived the loops they describe.**
  `loop_header_pcs` (pre-existing) and `loop_header_greens` (added by #773) were
  inserted on compile and never removed, so both kept growing as loops were
  retired. `remove_compiled_loop` and the memmgr eviction path in
  `try_to_free_some_loops` now drop the retired key from both, through a shared
  `forget_loop_side_tables`.

  This is a growth fix, not a targeting fix: `compiled_key_for_greens` already
  gates its match on `has_compiled_targets`, so a stale entry could never have
  resolved a bridge onto a retired loop.

* **`jitdriver.rs` — `MAJIT_BRIDGE_ONLY` silently swallowed unparsable entries.**
  The list was parsed with `filter_map(|t| t.parse().ok())`, so
  `MAJIT_BRIDGE_ONLY=oops` produced an *empty* allowlist — which suppresses every
  bridge, the exact inverse of the documented "unset means all" default, with no
  diagnostic. Since this knob exists to bisect a miscompiling bridge out of a
  run, a typo would have produced a confidently wrong result. An unparsable entry
  now panics.

* **`resume_box_reader.rs` — the two `fieldnums` arity checks are unconditional
  `assert_eq!`.** As `debug_assert_eq!` they compiled out of release builds,
  which then indexed past the end of the slice and panicked without the message
  the check exists to print.

* **`jitdriver.rs`** — dropped a redundant `clone()` of the close greens; the
  value is not used again.

## Not addressed here, with reasons

* **Multi-frame bridge giveup** (`jitdriver.rs`, Codex P1). #773 gave up on
  bridges whose resume data spans more than one frame, in the same commit that
  fixed the liveness splitter those bridges decode through. The giveup may
  therefore now be unnecessary conservatism costing bridge coverage. That is a
  behaviour change needing its own measurement, not a review fixup, so it is
  left for a separate change together with CodeRabbit's request for
  `building_bridge` regression tests in `optimizeopt/optimizer.rs`.

* **Shared `FRAME_VALUE_COUNT_FN`** (`majit-ir/src/resumedata.rs:20`, Codex P1).
  Reported as thread-local state; it is a process-global `AtomicUsize` that
  predates #773 and is already consumed by the cranelift backend. It is only
  hazardous with two `#[jit_interp]` drivers registering different splitters in
  one process, which no current configuration does. Making it per-driver is a
  cross-crate signature change through `rebuild_from_numbering`.

## Verification

`cargo test -p majit-metainterp --features dynasm` — 1501 passed, 0 failed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup and invalidation for retired compiled loops, preventing stale per-loop side data from lingering.
  - Added stricter release-mode validation for virtual string/unicode rematerialization inputs.
  - Fixed resume/decoding liveness handling to consistently use the runtime snapshot.
  - Updated headerless allocation behavior to support a non-collecting path for headerless nursery allocations.
  - Improved descriptor field-key stability/compatibility to avoid mismatched descriptor resolutions.
- **New Features**
  - Added support for reporting field-descriptor identity census on demand via an environment setting.
- **Tests**
  - Added unit tests for compiled-loop removal and compiled-trace invalidation side-table cleanup.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -11683,36 +11683,164 @@ impl CraneliftBackend {
                         &known_values,
                         op.arg(0).to_opref(),
                     );
-                    let result = emit_collecting_gc_call(
-                        &mut builder,
-                        ptr_type,
-                        call_conv,
-                        jf_ptr,
-                        &ref_root_slots,
-                        &defined_ref_vars,
-                        &stale_ref_vars,
-                        &demoted_failarg_slots,
-                        ref_root_base_ofs,
-                        per_call_gcmap,
-                        gc_alloc_nursery_headerless_shim as *const () as usize,
-                        &[size],
-                        Some(cl_types::I64),
-                    )
-                    .expect("headerless nursery allocation helper must return a value");
-                    jf_ptr = emit_reload_frame_if_necessary(&mut builder, ptr_type, call_conv);
-                    builder.ins().set_pinned_reg(jf_ptr);
-                    builder.def_var(var(vi), result);
-                    // malloc_cond parity: the headerless slow path
-                    // (gc_alloc_nursery_headerless_shim) returns NULL on
-                    // host/bounded out-of-memory; propagate before the
-                    // following stores dereference it.
-                    emit_memory_error_check(
-                        &mut builder,
-                        ptr_type,
-                        result,
-                        propagate_exception_descr_ptr,
-                        preamble_phase,
-                    );
+                    // x86/assembler.py:2556-2565 malloc_cond parity, headerless
+                    // variant — the same inline bump the `CallMallocNursery`
+                    // arm below emits and the dynasm backends already emit for
+                    // this opcode (`genop_call_malloc_nursery_headerless`).
+                    // Three differences from the headered shape: the bump is by
+                    // `size` alone (no `GcHeader::SIZE` reservation), no header
+                    // word is zeroed, and the result is the old nursery base
+                    // rather than `base + GcHeader::SIZE`.
+                    //
+                    // The raw bump is correct only while the active GC is
+                    // headerless-aware, which is the same invariant
+                    // `alloc_nursery_headerless`'s panicking default enforces on
+                    // the overflow path — whoever declares `headerless_structs`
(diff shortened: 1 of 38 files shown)


youknowone:
Rebased onto `origin/main` (`948340d19e`) and added one commit.

**`jit: skip the finish_setup republish when neither writer buffer grew`** — `ensure_finish_setup` runs on every `jitcode_for` and built its arguments by cloning `Assembler.insns` and `Assembler.all_liveness` whole. Since `jit: seed the runtime assembler's insns from the build-time opcode table` and `jit: resume the build-time liveness buffer instead of forking the pool` seed both buffers from the build-time tables, that copy became proportional to the whole opcode and liveness universe, once per blackhole resume. `assembler.py:29-31` only appends to them, so equal lengths mean equal contents and the snapshot is now skipped when neither grew.

Measured on `bench/synth/depth{2,3,7}_inline_chain_typeflip`, which run 119873 guard failures with identical trace structure on both sides:

| | depth2 | depth3 | depth7 |
|---|---|---|---|
| `origin/main` | 1.78s | 2.47s | 4.50s |
| this branch, before the fix | — | — | timed out |
| this branch, after | 1.65s | 1.85s | 3.61s |

(user+sys CPU, min of 3. Before the fix, wall clock was 5.56s / 7.38s on depth2 / depth3 against 1.95s / 2.54s for `origin/main`; depth3 and depth7 timed out under `check.py`.)

`python3 ./pyre/check.py --backend dynasm`: 2 failed / 329 passed. Both failures are independent of this branch — `synth/getframe_force_cancel_journal` (pre-existing wrong output) and `synth/const_arg_call_resume`, whose ratio gate divides by a 0.01s pypy denominator; its CPU time is at parity with `origin/main` (0.27–0.29s here vs 0.26–0.34s there).

— *commented by Claude*

PR #791, opened 2026-07-25
builtins, str, import: surrogate-safe entry points, builtin arity checks, settrace under the JIT

youknowone:
Seventeen commits, rebased onto `origin/main`. The first nine are described below under **sys.flags / `_imp` / surrogates / binascii, zlib**; the last eight close the items the first round reported as "found, not fixed".

## sys.flags is a structseq

`sys.flags` was a class-namespace stand-in, so `isinstance(sys.flags, tuple)` was false. It is now a structseq: it indexes, slices, compares, hashes and unpacks as a tuple, carries `n_sequence_fields` (18) / `n_fields` (21), and its fields are read-only. `gil`, `thread_inherit_context` and `context_aware_warnings` sit past the sequence as named-only extras.

## `_imp`

- `find_frozen` takes the keyword-only `withdata`; when true it returns a read-only `memoryview` over the marshalled code object, so `marshal.loads(bytes(data))` reconstructs what `get_frozen_object` returns. Adds `marshal::dumps_bytes` for that stream.
- `create_builtin` encodes the spec name to ascii the way the extension-module loader does, so a non-ascii name raises `UnicodeEncodeError`, and a correctly spelled name that names no builtin returns `None` instead of raising.

A 26-row `_imp` matrix is byte-identical to CPython 3.14.

## Lone surrogates no longer abort the process

pyre strings are backed by WTF-8, so a lone surrogate has no `&str` view and `w_str_get_value` panicked — killing the whole process, not raising. A fuzz sweep over ~3300 module entry points (each probe run with a lone surrogate, and again with a plain ASCII string as a control) found **152** such aborts.

All 152 are gone. `str_utf8_w` reports the `UnicodeEncodeError` the strict utf-8 encoder raises (unicodehelper.py:245); `text_w` and the `#[pyre_function]` `&str` parameter route through it, which covers the whole `pyre_function` surface at once.

Where the value is not required to be utf-8 it is now **kept**, not rejected: format specs, regex subjects and replacements, pickle strings, warning text, strftime formats, keyword names, `__qualname__`, `super().__getattribute__`, memoryview cast formats, array typecodes, `compare_digest` and the time clock names read the WTF-8 buffer. The format-spec parser carries its fill as a `CodePoint`, so `format(x, '\ud800<6')` pads with the surrogate it was given; `_sre` drives `str` subjects over `&Wtf8`, so a pattern matches a lone surrogate like any other character and `'\ud800'.encode('idna')` no longer dies inside the codec's regex.

## binascii, zlib

`b2a_hex`/`hexlify`, `b2a_base64`, `b2a_qp`, `b2a_uu`, `crc32` and `crc_hqx` accepted a `str` through the `ascii_buffer_converter` that belongs to the `a2b_*` decoders. They declare a `Py_buffer`, so a `str` is now rejected by its type, matching CPython and PyPy. zlib's converter did the same and now rejects a `str` too.

---

# Second round — the reported items

## Wrong argument counts no longer abort the process

A builtin body indexes its slots directly, so a call whose count does not fit read past the end of `args` and aborted. The control arm of the fuzz stopped the process on 81 such labels (`operator.and_('x')`, `cmath.rect(1)`, `itertools.combinations('ab')`, `_signal.signal(1)`, `builtins.__build_class__(1, 2)`, …).

The count is now checked where it is known exactly:

- `#[pyre_function]` / `#[pyre_methods]` wrappers reject a count outside their required..total range, derived from the typed parameter list.
- `py_module!`'s `inline_functions:` arm derives `fast_natural_arity` from that same list, so a function with an optional or varargs parameter registers `HOPELESS` instead of the raw parameter count. This replaces `pyre_count_typed_args!`, which counted `#[default(...)]` parameters as required and so declared a *maximum* as if it were exact.
- `py_module!`'s `functions:` / `module_functions:` arms wrap each declared arity in `gateway::check_declared_arity`. `_opcode.stack_effect`, which takes 1-2 positional arguments plus a keyword-only `jump`, moves to `*` and keeps its own check.
- `_signal.signal`, `itertools.combinations` and `posix.urandom`, registered outside `py_module!`, call `check_declared_arity` themselves.

Both fuzz arms now run to completion with **zero** process aborts. The messages match CPython's: `and_ expected 2 arguments, got 1`, `neg() takes exactly one argument (0 given)`, `list.append() takes exactly one argument (2 given)`, `center expected at least 1 argument, got 0`.

`__build_class__` additionally rejects a body that is not a Python function — `__build_class__(1, 2)` crashed reading the code object, and now raises `__build_class__: func must be a function` exactly as both oracles do.

## `sys.settrace` emits events again

`call_trace` / `return_trace` / `bytecode_trace` are driven from the plain eval path, so a frame the JIT took over reported nothing. `eval_with_jit_inner` now declines a frame whose execution context carries a trace or profile hook, and `demo()` under `settrace` yields `[('demo', 'call'), ('demo', 'return')]`.

`line` events are still missing — that reproduces with the JIT off too and is a separate gap.

## `_hashlib.new` validates at construction

`HASH.__init__` stored the digest name unchecked, so `_hashlib.new('nope')` succeeded and only failed once a digest was taken. A new interp-level `_check_digest_name` rejects a non-str and an unknown algorithm where the digest is resolved.

## `_socket.gethostbyname` idna-encodes

The converter demanded a utf-8 view before its ASCII test, and replaced a failed `.encode('idna')` with a generic TypeError. The ASCII test now reads the raw buffer and the codec's own error propagates: `_socket.gethostbyname('\ud800')` raises `UnicodeEncodeError: 'idna' codec can't encode character '\ud800' in position 0: Invalid character '\ud800'`, byte-identical to CPython. pyre's `idna` codec itself was already correct (`'예'.encode('idna')` → `b'xn--2j5b'`).

## `operator._compare_digest`

The earlier report was wrong about the cause: `_compare_digest` *does* exist upstream. The real divergence is that pyre installed the interp-level table under both `operator` and `_operator`, while `moduledef.py:5` gives it `applevel_name = '_operator'`. `import operator` now resolves to `operator.py`, whose `from _operator import *` drops the underscore names.

## `array.array` typecode message

The typecode length was measured in bytes, so a single non-ASCII character was reported as a string of its UTF-8 length. It now counts code points, and both rejections name what was passed: `must be a unicode character, not <type>` and `not a string of length N`.

## `posix.urandom`

`urandom` read the argument's integer field raw, so `posix.urandom('x')` asked for an arbitrary number of bytes and hung. The size goes through `int_w` (`__index__`), and a negative size raises `ValueError`.

## Gate

`check.py` 315/321 on dynasm and 316/321 on cranelift, `cargo fmt --all --check` clean.

The failures are **pre-existing on this branch's base** — a binary built from the branch tip before this round's changes produces the identical set:

- `synth/except_star`, `synth/exception_group_type`, `synth/exception_subclass_attrs` — crash (exit -10) after JIT warmup; passes under `PYRE_JIT=0`
- `synth/getframe_force_cancel_journal`, `synth/unpack_drain_star_raise` — wrong output
- `synth/const_arg_call_resume` — the known boundary perf gate (34-39x against a 30x gate; cranelift passed it on the final run)

## Verification

- 83-row cross-interpreter matrix against CPython 3.14 and PyPy3: every row matches CPython except four, and all four (`compile` filename, `type()` name label, memoryview cast, `super().__getattribute__` message) are **byte-identical to PyPy** — left alone per the project's rule of not diverging from PyPy where pyre already agrees with it.
- Both fuzz arms (~3300 probes each, lone surrogate and ASCII control) complete with no process aborts.

## Summary by CodeRabbit

* **New Features**
  * Added support for executing Python code from standard input, with improved interactive vs piped behavior.
  * Enhanced `PYTHONSAFEPATH` handling, including correct `-P` / environment precedence and `-E` interactions.
  * Updated `sys.flags` to better match expected runtime behavior.
* **Bug Fixes**
  * Preserved lone-surrogate strings across formatting, regex, imports, warnings, pickling, and related Unicode paths.
  * Tightened bytes-only APIs to reject `str` with clearer `TypeError`s (notably `binascii` and `zlib`).
* **Other Changes**
  * Standardized builtin argument validation and arity-related errors (and improved tracing behavior when using JIT).

### DIFF
--- a/lib-python/3/test/test_importlib/util.py
+++ b/lib-python/3/test/test_importlib/util.py
@@ -15,9 +15,6 @@ import sys
 import tempfile
 import types
 
-import_helper.import_module("_testmultiphase")
-
-
 BUILTINS = types.SimpleNamespace()
 BUILTINS.good_name = None
 BUILTINS.bad_name = None
--- a/majit/majit-translate/src/codewriter/assembler.rs
+++ b/majit/majit-translate/src/codewriter/assembler.rs
@@ -17,6 +17,90 @@ use vecset::VecSet;
 use crate::call::CallControl;
 use crate::flatten::{FlatOp, IntOvfOp, Label, RegKind, SSARepr};
 
+fn reg_label(reg: crate::flatten::Register) -> String {
+    reg.repr()
+}
+
+fn reg_or_const_label(value: &crate::flatten::RegOrConst) -> String {
+    match value {
+        crate::flatten::RegOrConst::Reg(reg) => reg_label(*reg),
+        crate::flatten::RegOrConst::Const(c) => {
+            let kind = crate::flatten::constant_kind(c);
+            format!("Const({kind})")
+        }
+    }
+}
+
+fn flatop_debug_label(op: &FlatOp) -> String {
+    match op {
+        FlatOp::Label(label) => format!("Label({})", label.0),
+        FlatOp::Op(op) => {
+            let has_result = op.result.is_some();
+            format!("Op(has_result={has_result})")
+        }
+        FlatOp::Jump(label) => format!("Jump({})", label.0),
+        FlatOp::GotoIfNot { cond, target } => {
+            format!("GotoIfNot(cond={}, target={})", reg_label(*cond), target.0)
+        }
+        FlatOp::GotoIfNotOp {
+            opname,
+            args,
+            target,
+        } => {
+            let args = args
+                .iter()
+                .map(|reg| reg_label(*reg))
+                .collect::<Vec<_>>()
+                .join(",");
+            format!("GotoIfNotOp({opname}, args=[{args}], target={})", target.0)
+        }
+        FlatOp::Switch { value, targets } => {
+            format!(
+                "Switch(value={}, targets={})",
+                reg_label(*value),
+                targets.len()
+            )
+        }
+        FlatOp::IntBinOpJumpIfOvf {
+            op,
+            target,
+            lhs,
+            rhs,
+            dst,
+        } => format!(
+            "IntBinOpJumpIfOvf({op:?}, lhs={}, rhs={}, dst={}, target={})",
+            reg_label(*lhs),
+            reg_label(*rhs),
+            reg_label(*dst),
+            target.0
+        ),
+        FlatOp::CatchException { target } => format!("CatchException(target={})", target.0),
+        FlatOp::GotoIfExceptionMismatch { target, .. } => {
+            format!("GotoIfExceptionMismatch(target={})", target.0)
+        }
+        FlatOp::Move { dst, src } => {
+            format!(
+                "Move(dst={}, src={})",
+                reg_label(*dst),
+                reg_or_const_label(src)
+            )
+        }
+        FlatOp::Push(reg) => format!("Push({})", reg_label(*reg)),
+        FlatOp::Pop(reg) => format!("Pop({})", reg_label(*reg)),
+        FlatOp::LastException { dst } => format!("LastException(dst={})", reg_label(*dst)),
+        FlatOp::LastExcValue { dst } => format!("LastExcValue(dst={})", reg_label(*dst)),
+        FlatOp::Live { live_values } => format!("Live(count={})", live_values.len()),
+        FlatOp::Reraise => "Reraise".to_string(),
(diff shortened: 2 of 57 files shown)


youknowone:
@copilot resolve the merge conflicts in this pull request. blend the pypy-parity way

PR #792, opened 2026-07-25
majit: Vec-ctor newlist + reserved __pyre_range, union-pair diagnostic, mixed-signedness const fold

youknowone:
## Summary

Three `majit-translate` fixes found while walking the rtyper prepass census (`#73`), plus one comment fix in the method-cache front door.

**`majit`: name the colliding operands in the union fallthrough error**
`union()`'s unhandled-pair arm reported only `no upstream pair(s1, s2).union() handler in current subset`, and both operands usually render as the bare `<other>` `KnownType`, so the skip-classified panic could not be acted on. It now appends the classdef name for `Instance` / the pointer target for `Ptr`, the way the `SomeInstance` arm already names its two classdefs. The existing phrase stays first and verbatim — `cutover.rs:1291` (`is_known_unported`) and `:2718` (bucket `UNION-PAIR-PORT`) match it with `contains`.

**`majit`: lower `Vec::{new,with_capacity}` to `newlist`, reserve a `__pyre_range` path**
- `vec::Vec::{with_capacity,new}` fell through to `foreign_container_ctor`, so the result annotated as a classdef-less opaque `SomeInstance` and the already-wired `push`→`append` / `reverse` recognizers aborted their `getattr` against it. The adapter now emits `newlist` with no element args (`rtype_newlist`, `rlist.py:338-344`). The capacity operand is dropped — upstream keeps a size hint only through `objectmodel.newlist_hint` (`objectmodel.py:443-447`), a path `rtype_newlist` here does not take. `op_canraise` classifies the call non-raising: `operation.py:551-553` gives `NewList` `canraise = []`.
- `front::range_iter`'s synthesized range call moves from a bare `["range"]` to a reserved `["__pyre_range"]` spelling with its own `translate_op` arm binding the `HOST_ENV` singleton. A single-segment path resolves against the `PyreCallRegistry` before `HOST_ENV`, and that registry is one flat leaf-keyed namespace, so `pyre_interpreter::module::_ast::convert::range` registered as `["range"]` and captured the call; `findbltintyper` is keyed by `Arc` identity (`rbuiltin.rs:98-101`) and then failed with *"don't know about built-in function `<host range>`"*. This also un-poisons the `_ast::convert::range` / `_ast::convert::<Impl>::parts` annotations that #767's bare spelling was capturing on main.

Census at `2360d766bfa`: phaseA 318 → **311**, phaseB 22 → **19**, classdef-less roots 38 → **6**, no new failures.

**`majit`: fold mixed-signedness shifts and comparisons in `const_eval_binop`**
`const_eval_binop` had arms for `(Int,Int)`, `(UInt,UInt)` and `(Bool,Bool)` only, so every mixed-signedness pair hit the catch-all decline. `pub const MEMBER_DIRECT_FLAG: u32 = 1 << 31` records one — an unsuffixed shift amount defaults to `i32` while the shifted value is `u32` — and the shift-overflow assert Charon plants ahead of it (`31i32 as u32 < 32u32`) is mixed too, so the const stayed a residual accessor call no registry can bind. The new arm fires only when the signed side is non-negative (both carriers then agree on the mathematical value) and only for shifts and comparisons; **arithmetic stays declined**, since a mixed pair is exactly what leaves the result *type* undetermined.

It also closes a latent wrong-constant hole: `const_narrow_to_target` re-narrows every `Assign` result to the destination width, because the `ConstLit` carrier is 64-bit while rustc computes at the destination width and Rust does not count shifted-out bits as an overflow (`200u8 << 1` is a legal const equal to **144**, not 400).

Census at `beee634acb7`: phaseA 295 → **292**, skips 338 → **335**, no new failures; `MEMBER_DIRECT_FLAG` mentions 15 → 0.

**`typeobject`: note the pair-coherence consequence of the split elidable lookup**
`lookup_where_with_method_cache` reads `(w_class, w_value)` through two single-register elidable residuals because the residual-call ABI carries one raw register, where `typeobject.py:510` returns one tuple. Each residual takes the cache mutex separately, so a concurrent `mutated()` can land between them — worth stating, and no weaker than the `version_tag` promotion immediately above it. The same comment still said "thread-local `MethodCache`", which #784 made stale.

### Dropped during the rebase onto `0b706504dde`
This branch previously carried a fourth line of work — converting `baseobjspace::METHOD_CACHE` from `thread_local!` to a process-global prebuilt singleton, on the grounds that `space.fromcache` is a `specialize:memo` and a thread-local space cache is a deviation. **`#784` (free-threaded runtime port) landed the same conversion on main and did it properly**, as `LazyLock<parking_lot::Mutex<MethodCache>>` with every probe, fill, clear and GC walk under the lock, plus the same treatment for `MAP_ATTR_CACHE`. The local version's premise — one mutator, so an unlocked publication protocol suffices — is false as of #784, so re-applying it would have replaced a correct mutex with a racy protocol. Those three commits were dropped rather than rebased; only the comment fix above survives.

### Verification
- `cargo test --all --no-default-features --features dynasm` — **7084 passed, 0 failed**.
- `python3 pyre/check.py --backend dynasm,cranelift`.
- Capstone gate for the `newlist` adapter arm: `strings target/release/build/pyre-jit-trace-*/out/insns.bin | rg newlist` is empty (a raw `newlist/…>r` reaching the assembler default arm would be an opname with no blackhole handler), and `default_bh_builder_unwired_set_matches_task_85_snapshot` is green.
- The LLBC corpus was re-extracted (`scripts/extract-llbc.py --force`) after the rebase; a stale corpus fails `test_rbigint_mir` on the `jit_bigint_div` marker, which is not a code defect.

## Self-review

Reviewed against the upstream sources cited above (`typeobject.py`, `rlist.py`, `objectmodel.py`, `operation.py`, `rbuiltin.rs`). One thing worth reviewer attention:

- `const_narrow_to_target` changes the behaviour of the pre-existing `(UInt,UInt)` `Shl` arm, not only the new mixed arm. That is deliberate — the old arm could bake a constant wider than its destination type.

### DIFF
--- a/majit/majit-translate/src/annotator/model.rs
+++ b/majit/majit-translate/src/annotator/model.rs
@@ -2958,6 +2958,37 @@ impl std::error::Error for AnnotatorError {}
 // union() dispatch (A4.6) — model.py:750-784 + binaryop.py pair().union().
 // ---------------------------------------------------------------------------
 
+/// Short identity of a `union` operand for the unhandled-pair error text.
+///
+/// The bare [`SomeValueTag`] is often not enough to act on: `Instance`
+/// covers every classdef, and the failing pair is usually diagnosed by
+/// *which* class collided, not by the variant. This renders the one
+/// discriminating field for the two payload-bearing variants that show up
+/// in unhandled pairs and falls back to the tag for the rest, so the error
+/// stays one short line. Same motivation as the `SomeInstance ∪
+/// SomeInstance` arm naming its two classdefs.
+fn union_operand_id(s: &SomeValue) -> String {
+    match s {
+        SomeValue::Instance(inst) => match &inst.classdef {
+            Some(cd) => format!("Instance({})", cd.borrow().name),
+            None => "Instance(classdef-less)".to_string(),
+        },
+        SomeValue::Ptr(ptr) => {
+            use crate::translator::rtyper::lltypesystem::lltype::PtrTarget;
+            let target = match &ptr.ll_ptrtype.TO {
+                PtrTarget::Struct(st) => st._name.clone(),
+                PtrTarget::Opaque(op) => op.tag.clone(),
+                PtrTarget::Func(_) => "Func".to_string(),
+                PtrTarget::Array(_) => "Array".to_string(),
+                PtrTarget::FixedSizeArray(_) => "FixedSizeArray".to_string(),
+                PtrTarget::ForwardReference(_) => "ForwardReference".to_string(),
+            };
+            format!("Ptr({target})")
+        }
+        other => format!("{:?}", other.tag()),
+    }
+}
+
 /// RPython `union(s1, s2)` (model.py:750-769).
 ///
 /// The join operation in the lattice of annotations. Returns the most
@@ -3424,7 +3455,18 @@ pub fn union(s1: &SomeValue, s2: &SomeValue) -> Result<SomeValue, UnionError> {
         _ => Err(UnionError {
             lhs: s1.clone(),
             rhs: s2.clone(),
-            msg: "no upstream pair(s1, s2).union() handler in current subset".into(),
+            // Name the two colliding operands, for the same reason the
+            // `SomeInstance` arm above names its classdefs: both sides
+            // often render as the bare `<other>` `KnownType`, which makes
+            // the skip-classified panic undiagnosable.  The classifier
+            // phrase is kept verbatim and first — `cutover.rs:1291` /
+            // `:2718` match it with `contains`.
+            msg: format!(
+                "no upstream pair(s1, s2).union() handler in current subset: \
+                 {} ∪ {}",
+                union_operand_id(s1),
+                union_operand_id(s2)
+            ),
         }),
     }
 }
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -14356,7 +14356,13 @@ fn const_eval_init_body_lit(llbc: &Llbc, u: &Unstructured, depth: usize) -> Opti
                         )?,
                         _ => return None,
                     };
-                    locals.insert(dst, value);
+                    // rustc computes each assignment at the destination's
+                    // width; re-narrow so a `Shl` that shifted bits out does
+                    // not bake a too-wide constant.
+                    locals.insert(
+                        dst,
+                        const_narrow_to_target(const_literal_ty(llbc, &place.ty), value),
+                    );
                 }
                 _ => return None,
             }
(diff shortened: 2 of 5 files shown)


youknowone:

The review ran against `b857962`, which carried a `METHOD_CACHE` conversion this branch no longer has. Rebasing onto `0b706504dde` showed that **#784 (free-threaded runtime port) had already made both space caches process-global**, as `LazyLock<Mutex<…>>` with every probe, fill, clear and GC walk under the lock. The three `typeobject:` commits were therefore dropped rather than rebased — their premise (one mutator, so an unlocked publication protocol suffices) is false as of #784, and re-applying them would have replaced a correct mutex with a racy protocol. That resolves most of the review by deletion; the dispositions below are recorded anyway, since the findings were right.

**1 (regressions): none reported.**

**2 (new mismatches): 1 finding — correct, and now moot.**

The finding had two halves, and both were real:

- The ABA half is a genuine defect, not just a documentation problem. `method_hash` mixes `(version_tag, name)` into 2048 slots, so two different names for the **same** type can collide, and both fills then write the *same* `version_tag`. A post-payload re-check that compares only the version cannot witness that: a reader could validate `(version_tag, name A)`, read name B's payload, re-read an identical version, and accept it. It was fixed on the branch by re-checking **both** key halves — sufficient rather than merely narrower, since the only other way both match again is a refill for the identical `(version_tag, w_name)`, whose payload is the same MRO-walk result by construction — and the fix then went away with the commit it fixed.
- "The fences do not make those accesses atomic" is likewise correct. #784's mutex is the right answer to it, and a better one than the unlocked protocol plus honest documentation that the branch had settled on.

**3 (pre-existing): 1 finding — correct, and already fixed on main.**

`MAP_ATTR_CACHE` is `static LazyLock<Mutex<MapAttrCache>>` as of #784 (`mapdict.rs:852`), so the thread-local deviation the review names is gone. `MAPDICT_METHOD_CACHE_CODES` is still a `thread_local!`, but it is not a member of this family: it is a pyre-only GC-root registry with no upstream counterpart (upstream reaches `CacheEntry.w_method` through its GC-managed holder), and it retires when code objects become GC-managed rather than being converted.

**4 (structural adaptations): 4 findings, all correctly classified; 1 documentation fix kept.**

- `range_iter.rs:349` (reserved `__pyre_range`) and `flowspace_adapter.rs:1645` (`Vec` ctor → empty `newlist`, capacity dropped) — agreed, and both already carry the in-code citation for the upstream decision point they encode.
- `mapdict.rs:820` (deterministic FNV instead of the configured `compute_hash`) — agreed, and already documented as a `PRE-EXISTING-ADAPTATION` at `mapdict.rs:808-819`, including why it is benign: `find_map_attr` rechecks name and attrkind, so a divergent hash costs at most a miss.
- `baseobjspace.rs:7841` (the tuple-returning elidable split into two single-register residuals) — agreed, and this was the one adaptation with no in-code note about the consequence, so the note is kept in this PR, restated for the post-#784 model: upstream reads the pair atomically because it returns one tuple, whereas each residual here takes the cache mutex separately, so a concurrent `mutated()` can land between them — no weaker than the `version_tag` promotion immediately above, where both halves are then read under a tag that is no longer current, exactly as a lookup completing one instruction earlier would have been.

— *commented by Claude*

PR #793, opened 2026-07-25
none: expose the CPython 3.14 slot-wrapper surface

kyokuping:
Assisted-by: Codex:gpt-5.6 Sol

## Summary

- Preserve PyPy’s identity-based equality semantics for `None` and its use of `NotImplemented` for ordering comparisons, while exposing the comparison and hash slots as distinct `wrapper_descriptor` objects to match CPython 3.14’s public surface.
- Reuse the immutable, `BuiltinCode`-backed Function layout without exposing Python-function-specific metadata, and extend descriptor binding, call dispatch, `LOAD_METHOD`, GC vtable registration, and JIT static-address resolution to recognize the new carrier.
- Use CPython’s reproducible `None` hash value, `0xFCA86420`, and document both its origin in [gh-99541](https://github.com/python/cpython/issues/99541) and the intentional divergence from PyPy’s identity-based hashing.

## Self-review

## Summary by CodeRabbit

* **New Features**
  * Added CPython-compatible slot wrapper descriptors for built-in operations.
  * Improved support for calling and binding slot wrappers as methods.
  * Added support for fixed-arity slot wrapper creation.

* **Bug Fixes**
  * Updated `hash(None)` to use a consistent CPython-compatible hash value.
  * Improved callable detection across function, method, and descriptor execution paths.

### DIFF
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -8460,6 +8460,13 @@ pub(crate) unsafe fn get(
     // PyPy splits BuiltinFunction from FunctionWithFixedCode at the typedef
     // layer: BuiltinFunction omits __get__, while FunctionWithFixedCode keeps
     // Function.__get__ and binds like a normal method descriptor.
+    if crate::is_slot_wrapper(descr) {
+        if obj.is_null() {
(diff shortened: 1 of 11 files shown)


youknowone (approved):
great, thank you so much!

PR #794, opened 2026-07-25
jit: make JIT-created interpreter frames GC-owned (retire FrameArena and the portal frame rebuild)

youknowone:
Makes every JIT-created interpreter frame an ordinary GC object. `FrameArena` had no upstream analogue: `space.createframe` allocates a fresh `PyFrame(W_Root)` per call (`baseobjspace.py:799-801`, `pyframe.py:52`) and lifetime is the GC's — `ec.leave` (`executioncontext.py:91-107`) frees nothing.

## Commits

1. **`jit: enter/leave the inlined callee frame around its concrete residual calls`** — names the frame a sub-walk executes concretely so each residual it runs can `enter`/`leave` it on the interpreter frame chain.
2. **`jit: allocate JIT callee frames as GC objects instead of recycling arena slots`** — deletes `FrameArena`.
3. **`jit: run the CALL_ASSEMBLER force path on the callee frame it is handed`** — deletes the portal's frame rebuild.
4. **`interp: remove the unused raw-value PyFrame constructors`**
5. **`jit: drop the unread concrete_frame from PendingInlineFrame`**

## The recycle bug (commit 2)

`scratchpad/s4_arena_tb_recycle.py` — a self-recursive `fib(n, k)` whose base case raises `ZeroDivisionError`, an outer function that walks `e.__traceback__` and retains every `tb_frame`, then more `fib` churn, then re-reads `f_locals` on the held frames:

| | |
|---|---|
| pypy3 | `held 38 mutated 0` |
| pyre-dynasm | `held 38 mutated 15` — silent data corruption |
| pyre-dynasm `PYRE_NO_JIT=1` | `held 38 mutated 0` |
| pyre-cranelift | non-unwinding panic in `gc_alloc_nursery_shim`, abort |

`jit_drop_callee_frame` unconditionally returned the slot; `put` disarmed it (removing it from the root walk) and the next call at that depth overwrote it in place, including `drop_in_place` of the old locals array — which is what made cranelift crash in `PyFrame::fast2locals`.

`escaped()` cannot protect it even in principle: the bit is set lazily when Python *reads* `tb.tb_frame`, long after the drop.

The four create sites now call one `alloc_callee_frame` = `FrameBox::new(PyFrame::new_for_call_with_closure_and_globals_obj(..., OldGenGc))`, and `jit_drop_callee_frame` only unroots. Between create and drop the frame's only root is `LIVE_CALLEE_FRAMES`, so that walker must **mark the frame object itself**, not just walk its slots — slot-walking alone (what the arena needed) would let a major collection sweep the block compiled code is running on.

`pytraceback_object_custom_trace` already forwards `tb.frame` guarded by `try_gc_owns_object`, so making the frame GC-owned lights that edge up on its own; no `PYTRACEBACK_GC_PTR_OFFSETS` change was needed.

★The reason my first probes found nothing: the arena is only entered when `recursive_force_cache_safe` accepts the callee, and it rejects any `LoadGlobal`/`LoadName` that does not name the function itself. `sys._getframe()` needs `LOAD_GLOBAL sys`, so such a callee **never uses the arena** — it escapes its frame via a **traceback**, not `sys._getframe`.

## The portal rebuild (commit 3)

`jit_force_callee_frame` read `pycode`/`w_globals`/`execution_context` off the incoming pointer by raw offset and built a fresh `PyFrame` with an empty argument slice. Two defects on one line: the rebuilt frame was a Rust **stack local**, so `portal_runner` → `install_current_frame` published a C-stack address as `CURRENT_FRAME` and `ec.topframeref`; and every parameter slot was `PY_NULL` with `last_instr` reset to -1. The repo already recorded the second failure once — the cranelift comment about `w_list_append`'s sibling path notes *"left `locals_w[0]` uninitialized … the stale `n` value drove the self-recursive CA into an unbounded loop"* — and fixed it by deleting one call site, leaving the body.

The premise justifying the rebuild — *"the callee frame may be a nursery-allocated JitFrame-like block"* — is false. `emit_new_pyframe_inline_with_params` emits `NewWithVtable` + `SetfieldGc` with the real `PyFrame` layout, which is why the raw offsets worked at all; a dynasm regression test is named *"force_fn must NOT be called with jitframe pointer … expected a PyFrame pointer"*. `#777` independently reached the same conclusion and stamped `assembler_call_helper resumes the callee frame the rewritten CALL_ASSEMBLER passed as arg 0`.

Upstream never rebuilds: `ll_portal_runner(*args)` forwards greens+reds verbatim (`warmspot.py:941-959`), the `frame` red is a live heap `PyFrame(W_Root)` (`interp_jit.py:66-92`), and forcing a virtualizable writes back **into the existing object** (`resume.py:1399-1408`, `virtualizable.py:126-137`). `rg "flavor\s*=\s*'stack'"` over `rpython/`+`pypy/` returns 0 hits — RPython has no stack-allocated frame flavor. `ll_portal_runner_shim` already ran the frame directly; both now share `run_frame_through_portal`.

**No reachable repro for this one.** An `lldb` breakpoint on the mangled symbol (it resolves, so 0 is meaningful) counted 0 hits across 8 hand-written probes × 2 backends, all 19 `bench/*.py` + 31 `bench/synth/*.py`, and 112 corpus scripts on cranelift (`[ca-shim] entering` = 0 in every one). On dynasm the leg is unreachable by construction — the GC rewriter `expect()`s the CA token to be registered, and the same `compile_loop` body that registers it stamps `set_ll_function_addr` a few dozen lines later with no early return between. So this commit restores upstream structure on a path that would be wrong whenever it does fire, rather than fixing an observed crash.

## Cleanups (commits 4-5)

An `ast-grep -p 'PyFrame { $$$ }'` census found exactly five `PyFrame` struct literals repo-wide, and 20 of 22 `new_for_call*` callers already wrap in `FrameBox::new`. The two that did not were commit 3's rebuild and `PendingInlineFrame.concrete_frame` — a `StdAlloc` by-value frame whose locals seeding and `last_instr` stamp were dead weight (**no code reads the field**; the callee's concrete state is carried by `sym.concrete_locals` / `sym.concrete_stack` and the frame vable bound to `sym.frame`). `PyFrame::new_minimal`, `new_with_namespace` and `alloc_fixed_array_from_vec` had no callers at all.

## Verification

- `s4_arena_tb_recycle`: `held 38 mutated 0` on **both** backends (a control binary built without commit 2 still prints `mutated 15`).
- 9 `sys._getframe` / `f_back` discriminators: 9/9 × 2 backends × {default, `PYRE_FBW_MULTIFRAME=1`}.
- Recursion battery (self-recursive, 2-arg, mutual, 3-arg) and a `RecursionError`-inside-CALL_ASSEMBLER probe: correct on every combination.
- `check.py`: **313/313 × 4** — dynasm and cranelift × default and `PYRE_FBW_MULTIFRAME=1`, run serially.
- `cargo test -p pyre-jit-trace -p pyre-jit --features dynasm`: 644 passed.
- No pinning leak: `fib` RSS 60 MB with the JIT vs 593 MB under `PYRE_NO_JIT=1`.
- Perf A/B (dynasm, arms **interleaved**, 9 reps, min): fib_recursive 0.985, raise_catch 0.955, int_loop 0.961, inline_helper 1.002 — consistent with the measurement that only 0.4% of calls ever reached the arena. (A first A/B that ran arm A's reps then arm B's showed a fake +20% on `fib_recursive` *and* +15% on `int_loop`, a pure loop that allocates no callee frame at all; that impossible second number is what exposed it as load drift.)

Re-verified end to end on the current base (`cb4f6bd409`, after #768 / #784 / #788 / #795):

- `check.py`: **322/322 x 4** — dynasm and cranelift x default and `PYRE_FBW_MULTIFRAME=1`, run serially.
- 9 discriminators + `s4_arena_tb_recycle` (`held 38 mutated 0`): 9/9 x both backends.
- `cargo test -p pyre-jit-trace -p pyre-jit --features dynasm`: 645 passed.

## Review follow-ups

- **`InlineConcreteFrameGuard::enter` inherited the enclosing callee's frame on a null seed** — the slot is now set unconditionally. A nested sub-walk whose seed block bailed has no frame of its own, and publishing the outer callee's frame for the inner callee's residuals resolves one level too shallow, which is the error the guard exists to stop; a null slot makes `ResidualFrameChainGuard::enter` publish nothing.
- **`unroot_callee_frame` did not re-arm the remembered set** — it does now, on the frame and its old-gen locals array. The live-frame list is what makes a minor collection scan those slots; once the frame leaves it, a young ref stored since the last minor (argument boxing, the CALL_ASSEMBLER writeback in `jit_frame_set_slot_*`) is reachable only through the array's own items, which a minor does not walk. Before this PR the frame was recycled at that point, so the window is new. `FrameBox::new` arms the same barrier at creation and the first minor consumes it.
- **`run_frame_through_portal` dereferenced its argument unchecked** — now a `debug_assert!`. The suggested "return NULL for a tagged/absent slot" guard was not taken: the portal's result is a Ref whose NULL spelling means *exception stored*, so returning NULL for a non-frame manufactures an exception-less error instead of a clean deopt. `jit_drop_callee_frame` tolerates a tagged word only because it does nothing but unroot.
- **`mark_as_escaped` inside the `escaped_published` predicate** — left as is. `escaped()` is set lazily, when Python reads `tb.tb_frame`; relying on `ResidualFrameChainGuard::drop` alone would drop the mark on a frame whose escape has not been observed yet.

## Unrelated rider

`display: skip the subclass-dunder probe for exact builtin instances` — `builtin_subclass_dunder_obj` ran two uncached MRO walks plus a descriptor call on every `str`/`repr`/`format` of an exact `int`/`float`/`bool`/`str`/`list`/`bytearray`, although only a subclass can redirect the dunder. `long` is excluded because its `__str__`/`__repr__` also enforce `sys.set_int_max_str_digits` (`longobject.py descr_repr`) and that check sits in the descriptor rather than in the conversion — `synth/int_max_str_digits` caught the first, broader version of this change on all four arms. `synth/comprehension_object_append_hot` goes 27.8/32.9/27.5/51.7x to 22.6/25.0/22.1/25.7x; its gate goes 40 to 32.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of JIT-compiled calls involving nested or inlined functions.
  * Corrected frame-chain tracking during residual execution, helping stack information and frame state remain accurate.
  * Improved garbage-collection handling for active JIT callee frames.
  * Streamlined forced JIT execution to preserve the correct interpreter frame context.

* **Documentation**
  * Clarified internal frame-layout, tracing, and execution behavior documentation.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -17295,7 +17295,7 @@ mod tests {
     /// GC_STORE against callee jitframes. Mirrors the dynasm test layout
     /// (runner.rs install_call_assembler_test_layout); the offsets match
     /// the production cranelift mapping in pyre-jit call_jit.rs
-    /// arena_jitframe_descrs. Reads jitframe_gc_type_id() after
+    /// jitframe_layout_descrs. Reads jitframe_gc_type_id() after
     /// set_gc_allocator has lazily registered JITFRAME.
     fn install_call_assembler_test_layout() {
         register_jitframe_layout(JitFrameLayoutInfo {
--- a/pyre/bench/synth/comprehension_object_append_hot.py
+++ b/pyre/bench/synth/comprehension_object_append_hot.py
@@ -1,4 +1,4 @@
-# pyre-check: max-pypy-ratio=40
+# pyre-check: max-pypy-ratio=32
 # An inlined list comprehension whose LIST_APPEND element lands in a list
 # Object-strategy (tuple / None / str / dict / f-string) folds through the #171
 # orthodox append. Its Object arm stores a GC ref and runs list_write_barrier,
--- a/pyre/pyre-interpreter/src/display.rs
+++ b/pyre/pyre-interpreter/src/display.rs
@@ -353,6 +353,24 @@ pub(crate) unsafe fn builtin_subclass_dunder_obj(
         if w_class.is_null() || !pyre_object::is_type(w_class) {
             return Ok(None);
         }
+        // Only a subclass can redirect the dunder: it keeps the builtin
+        // `ob_type` and retags `w_class` (`typedef::subclass_to_tag`), which
+        // is exactly what `is_exact_builtin_instance` tests. An exact
+        // instance resolves the dunder to the builtin the caller is about to
+        // run natively, and the builtin types are immutable, so the two MRO
+        // walks and the descriptor call below can only reproduce it.
+        //
+        // `long` is the one leaf where the descriptor does more than the leaf
+        // formatter: `longobject.py descr_repr` also enforces
+        // `sys.set_int_max_str_digits`, and that check sits in the descriptor
+        // rather than in the conversion, so an exact `long` keeps going
+        // through it. A machine `int` cannot reach any settable limit — 19
+        // digits against a floor of 640.
+        if !std::ptr::eq(tp, &LONG_TYPE as *const PyType)
+            && pyre_object::is_exact_builtin_instance(obj)
+        {
+            return Ok(None);
+        }
         let Some((src, found)) = crate::baseobjspace::lookup_where_pair(w_class, name) else {
             return Ok(None);
         };
--- a/pyre/pyre-interpreter/src/executioncontext.rs
+++ b/pyre/pyre-interpreter/src/executioncontext.rs
@@ -351,6 +351,8 @@ impl ExecutionContext {
         // builtins_module, matching a fresh PyPy ExecutionContext.
         ec.builtin_dict_cache = std::cell::Cell::new(pyre_object::PY_NULL);
         ec.sys_exc_value = pyre_object::PY_NULL;
+        // executioncontext.py:53 — a fresh ExecutionContext starts at 0.
+        ec.coroutine_origin_tracking_depth = 0;
         ec.current_gen_or_coroutine = pyre_object::PY_NULL;
         ec.w_asyncgen_firstiter_fn = pyre_object::PY_NULL;
         ec.w_asyncgen_finalizer_fn = pyre_object::PY_NULL;
--- a/pyre/pyre-interpreter/src/importing.rs
+++ b/pyre/pyre-interpreter/src/importing.rs
@@ -357,8 +357,8 @@ pub fn mount_embedded_stdlib(mount: &Path) {
 // PyPy equivalent: `space.sys.get('modules')`, `space.sys.path`, and
 // `space.builtin_modules` are object-space/process state, shared by every
 // ExecutionContext.  Raw GC references use the established process-global
-// `usize` representation; the GIL serializes semantic access while the mutex
-// also makes foreign STW root walks well-defined.
+// `usize` representation; the mutex serializes semantic access and keeps
+// foreign STW root walks well-defined.
 static SYS_MODULES: LazyLock<Mutex<HashMap<String, usize>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 static SYS_MODULES_DICT: AtomicUsize = AtomicUsize::new(0);
--- a/pyre/pyre-interpreter/src/module/thread/mod.rs
+++ b/pyre/pyre-interpreter/src/module/thread/mod.rs
@@ -712,17 +712,13 @@ mod handle_class {
                     // native abort.  The negated comparison rejects NaN too.
                     let secs = floatobject::w_float_get_value(timeout);
                     if !(secs <= TIMEOUT_MAX) {
-                        return Err(crate::PyError::overflow_error(
-                            "timeout value is too large",
-                        ));
+                        return Err(crate::PyError::overflow_error("timeout value is too large"));
                     }
                     Some(Duration::from_secs_f64(secs.max(0.0)))
                 } else if is_int(timeout) {
                     let secs = w_int_get_value(timeout);
                     if secs as f64 > TIMEOUT_MAX {
-                        return Err(crate::PyError::overflow_error(
-                            "timeout value is too large",
-                        ));
+                        return Err(crate::PyError::overflow_error("timeout value is too large"));
                     }
                     Some(Duration::from_secs(secs.max(0) as u64))
                 } else {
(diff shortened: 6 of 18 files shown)


youknowone:
Applied the review findings on `5c6acce6bd` (rebased onto `cb4f6bd409`).

**`InlineConcreteFrameGuard::enter` — null callee frame inherited the parent** (codex P2 / coderabbit): fixed. The slot is set unconditionally; a null slot makes `ResidualFrameChainGuard::enter` publish nothing rather than the enclosing callee's frame.

**`unroot_callee_frame` — remembered set not re-armed** (codex P1): fixed. The live-frame list is what makes a minor collection scan the frame's slots, so leaving it is exactly the moment the old-gen locals array has to join the remembered set. `FrameBox::new` arms the same barrier at creation (`try_gc_alloc_stable_raw` + `try_gc_write_barrier`) and the first minor consumes it; before this PR the frame was recycled at unroot, so the window only exists now.

**`run_frame_through_portal` unchecked deref** (coderabbit major): guarded with a `debug_assert!` rather than the proposed early `return PY_NULL`. The portal's result is a Ref whose NULL spelling means *exception stored*, so returning NULL for a tagged/absent slot manufactures an exception-less error instead of a clean deopt. `jit_drop_callee_frame` tolerates a tagged word only because it does nothing but unroot — there is nothing sensible to *run* for one.

**`mark_as_escaped` inside the `escaped_published` predicate** (coderabbit nitpick): not changed. `escaped()` is set lazily, when Python reads `tb.tb_frame`; moving the propagation into the taken branch or relying solely on `ResidualFrameChainGuard::drop` would drop the mark for a frame whose escape has not been observed yet.

**Frames relocating under the collector** (coderabbit, outside-diff): `FrameBox::new` allocates through `try_gc_alloc_stable_raw` — a non-moving old-gen block — so the raw `*mut PyFrame` copies in compiled code, `callee_concrete_r` and the sub-walk local stay valid. That is what the walker's comment asserts; no change made.

Verification on the current base: `check.py` 322/322 x 4 (dynasm/cranelift x default and `PYRE_FBW_MULTIFRAME=1`, serial), 9/9 getframe discriminators x both backends with `s4_arena_tb_recycle` at `held 38 mutated 0`, `cargo test -p pyre-jit-trace -p pyre-jit --features dynasm` 645 passed, `cargo fmt --all --check` clean.

— *commented by Claude*

PR #795, opened 2026-07-25
jit(fbw): drop the unwired PendingInlineFrame vref residue; record _MULTIFRAME's measured corpus coverage

youknowone:
One commit: delete FBW walker residue that a call-graph search shows is
unreachable, correct three comments that name mechanisms which do not exist,
and record a measurement of the multi-frame blackhole path in the gate ledger.

## Dead state

`PendingInlineFrame::drop_frame_opref` is constructed at exactly one site,
always `None`, and never read. Its comment said `None` "skips the
opimpl_virtual_ref emission in push_inline_frame". `push_inline_frame`
(`majit-metainterp/src/compile.rs`) does exist, but it is a depth guard — it
pushes a `(code_ptr, pc)` greenkey onto `inline_frames` and returns bool. It
takes no `PendingInlineFrame`, mutates no framestack, and contains no vref
code. The emission the comment names was never built.

`InlineTraceStepAction` — the enum that would have carried a
`PendingInlineFrame` into such a push — is declared and referenced nowhere in
the tree. Both go.

## Comments corrected against the call graph

- `MetaInterpreter::push_inline_frame`: the identifier `MetaInterpreter` occurs
  exactly once in the whole tree, inside that comment. There is no such type.
- `opimpl_virtual_ref`'s "Called from metainterp push_inline_frame": its only
  callers, in both the `majit-metainterp` and the `pyre-jit-trace` port, are
  `#[test]`. `virtualref_boxes` is therefore always empty and no live trace
  records a `VIRTUAL_REF`. The replacement says that, and names upstream's real
  caller (`executioncontext.py:89 enter`, which the tracer reaches by tracing
  the interpreter's own frame-entry code — something the pyre walker, which
  builds its inline levels itself, never does).
- "`push_inline_frame` stamps `parent_frames.first().pending_result_*`":
  `pending_result_stack_idx` is never assigned a `Some` value anywhere in the
  tree.

## Measurement recorded in gate-triage §1

The multi-frame blackhole path has no corpus coverage. The vable-escape latch
site was instrumented and all **318** benchmarks (`pyre/bench/*.py` +
`pyre/bench/synth/*.py`) were run under `PYRE_FBW_MULTIFRAME=1`. The site is
reached in **3** benches — `getframe_escape_flush_writethrough_regression`,
`synth/getframe_inlined_callee_own_frame`, `synth/getframe_stored_fback_walk` —
5 events each, and **all 15 have `inline_subwalk=false`**, so every one takes
the single-frame arm and adopts. `build_multi_frame_miframe` is never called,
the image is never latched, and `try_adopt_multi_frame_blackhole` never sees a
candidate.

Consequences recorded in the ledger: flipping `_MULTIFRAME` ON is a no-op
across the corpus; the three items §1 lists as its remaining work are all
unexercised, so porting any of them would be unvalidatable until a benchmark
reaches `inline_subwalk=true` at a vable escape. §4's `_BLACKHOLE_RESUME` row is
amended — its retirement was conditioned on `_MULTIFRAME` landing, which that
measurement makes unevaluable. (The multi-frame latch is also nested inside
`single_frame_blackhole_resume_enabled()`, so it depends on `_BLACKHOLE_RESUME`
staying ON regardless.)

The instrumentation was temporary and is not in this diff; a pre-existing
`[s2-gate]` eprintln under `PYRE_FBW_DEBUG_ABORT` already reports
`inline_subwalk` at that site.

## Gates

`cargo check -p pyre-jit-trace --no-default-features --features dynasm` is clean
(one pre-existing `unreachable_patterns` warning, unrelated).

An earlier full gate on this diff — `pyre-jit-trace` + `majit-metainterp` unit
tests green (1399 / 294 / …) and `check.py --backend dynasm` 313/313 — was run
against the pre-rebase base `7b70e1ebff`, so it does not certify the current
base. The re-gate on `0b706504dd` is in flight at the time of writing; CI on
this PR is the authoritative x86_64 check. A first attempt at that re-gate
produced a spurious `E0609: no field frame_value_count_fn` — the compile raced a
mid-session pull of #751, which adds that field to `jitdriver.rs` and its
assignment in `unpack_state.rs` in the same commit. `cargo check` on the settled
tree passes, so that error was a torn-tree artifact, not a base break.

The diff is dead-code deletion plus comments plus markdown; nothing in it
changes codegen.

## Summary by CodeRabbit

* **Documentation**
  * Updated gate-triage records with the latest audit outcomes and measured multi-frame coverage.
  * Clarified prerequisites, instrumentation findings, and retirement criteria for blackhole-related feature gates.

* **Refactor**
  * Simplified inline-call resume and guard-state handling.
  * Removed obsolete inline trace-step routing and clarified virtual-reference tracing behavior.

### DIFF
--- a/pyre/gate-triage.md
+++ b/pyre/gate-triage.md
@@ -118,7 +118,7 @@ upstream lines:
 | gate | orthodox side | outcome |
 |---|---|---|
 | PYRE_FBW_VABLE_SCALAR_CA | **OFF** | **RETIRED** — the ON design contradicts upstream |
-| PYRE_FBW_MULTIFRAME | **ON** | keep; the ON path is the port, and it is unfinished |
+| PYRE_FBW_MULTIFRAME | **ON** | keep; the ON path is the port, it is unfinished, and §1 measures it as never reached by the corpus |
 | PYRE_FBW_CALLEE_VSTACK | NEITHER | keep OFF; see §5 |
 
 The walker's default-ON `PYRE_FBW_*` cluster was retired separately in #757.
@@ -187,7 +187,29 @@ keeps inlined callee frames unmaterialized (`fbw_strict_fold_frame_reg`,
 upstream does not have — a consequence of pyre's virtual-callee-frame inlining —
 plus per-frame vable binding, since `PyjitplBlackholeFrameConfig` stamps one
 shared `virtualizable_ptr` onto every frame in the chain and the adopt writes
-only `last_instr`.
+only `last_instr`.  A third item sits below both: `try_adopt_multi_frame_blackhole`
+(`pyre-jit-trace/src/trace.rs`) declines outright when the recovered chain is not
+rooted at the walked frame, and names the `jit.virtual_ref` emit at the inline
+push as the prerequisite.  That emit does not exist — `opimpl_virtual_ref` /
+`_finish` are ported in both `majit-metainterp/src/pyjitpl.rs` and
+`pyre-jit-trace/src/state.rs`, and **neither has a caller outside a `#[test]`**,
+so `virtualref_boxes` is empty and no live trace records a `VIRTUAL_REF`.
+
+**Measured 2026-07-25: the multi-frame path has no corpus coverage.**  The
+vable-escape latch site was instrumented and all **318** benchmarks
+(`pyre/bench` + `pyre/bench/synth`) run under `PYRE_FBW_MULTIFRAME=1`.  The site
+is reached in **3 benches** (`getframe_escape_flush_writethrough_regression`,
+`synth/getframe_inlined_callee_own_frame`, `synth/getframe_stored_fback_walk`),
+5 events each, and **all 15 have `inline_subwalk=false`** — every one takes the
+single-frame arm and adopts.  `build_multi_frame_miframe` is therefore never
+called, the image is never latched, and the adopt never sees a candidate.  So
+flipping `_MULTIFRAME` ON is a no-op across the corpus, none of the three items
+above is exercised, and any port of them would be unvalidatable until a
+benchmark that reaches `inline_subwalk=true` at a vable escape exists.  Building
+that benchmark is the prerequisite for the rest.  Note the multi-frame latch is
+nested inside `single_frame_blackhole_resume_enabled()`, so it also requires
+`_BLACKHOLE_RESUME` to stay ON.  The pre-existing `[s2-gate]` eprintln (under
+`PYRE_FBW_DEBUG_ABORT`) already reports `inline_subwalk` at that site.
 
 ## §2 — Not gates (11): Rust identifiers, not env vars
 
@@ -229,7 +251,7 @@ OFF path is a needed safety net. Retire at the listed trigger (A7).
 
 | var | subsystem | retire when |
 |---|---|---|
-| PYRE_FBW_BLACKHOLE_RESUME | single-frame resume-past-escape (#754) | flipped default-ON 2026-07-25; retire once the multi-frame twin (`_MULTIFRAME`) lands and the escape/replay fallback is no longer the only other leg |
+| PYRE_FBW_BLACKHOLE_RESUME | single-frame resume-past-escape (#754) | flipped default-ON 2026-07-25; retirement was conditioned on the multi-frame twin (`_MULTIFRAME`) landing, but §1 now measures that twin as having zero corpus coverage, so the condition is unevaluable — keep the gate and re-open the question only once a benchmark reaches `inline_subwalk=true` at a vable escape |
 | PYRE_TWO_PHASE_RTYPE, PYRE_TUPLE_PER_SHAPE_CLASSDEF | rtyper prepass / per-shape tuple classdef | WS2 / #346 rtyper epic |
 | PYRE_ORIGINAL_BOXES | greens++reds original_boxes index shape | box-identity #202 / resume F1 |
 | PYRE_MIR_FRAMESTATE | framestate-threaded MIR lowering | MIR front-end #176/#181/#346 |
--- a/pyre/pyre-jit-trace/src/state.rs
+++ b/pyre/pyre-jit-trace/src/state.rs
@@ -5491,7 +5491,12 @@ impl PyreSym {
 /// records VIRTUAL_REF(box, cindex), and pushes
 /// [virtualbox, vrefbox] onto virtualref_boxes.
 ///
-/// Called from metainterp push_inline_frame (executioncontext.enter parity).
+/// Upstream's caller is `executioncontext.py:89 enter`, which the tracer
+/// reaches by tracing through the interpreter's own frame-entry code.  The
+/// pyre walker builds its inline levels itself and never traces `enter`, so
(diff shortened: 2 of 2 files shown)


youknowone:
Gate update on the current base (`ade259b614` on `0b706504dd`), and a follow-up
finding that supersedes part of this PR's ledger text.

## Gates — green

- `pyre-jit-trace` + `majit-metainterp` unit tests: **all pass** (exit 0).
- `check.py --backend dynasm`: **320 passed**, one perf bench over gate —
  `synth/comprehension_object_append_hot` at 43.6x vs its 40x gate. Discriminated
  as machine load, not a regression: that run had **load average 28.76 with 52
  concurrent build/cargo processes** from sibling worktrees; re-run in isolation
  three times it is **35.5x / 33.0x / 31.8x**, all under the gate.

A first attempt at this re-gate failed differently (two `list_append` unit tests
panicking, `synth/getframe_force_cancel_journal` wrong output, and
`synth/const_arg_call_resume` 34.0x). Root cause was a stale
`build/llbc/pyre-object.ullbc` — `w_list_append_inner` lives in `pyre-object`,
and only `pyre-interpreter` had been re-extracted. All three cleared after
extracting `pyre-object` and `pyre-jit`. Recorded here because the three
symptoms look unrelated but share one cause.

An `E0609: no field frame_value_count_fn` seen earlier was a torn-tree artifact:
the compile raced a mid-session pull of #751, which adds that field to
`jitdriver.rs` and its assignment in `unpack_state.rs` in the same commit.
`cargo check` on the settled tree is clean.

## Superseding finding: the multi-frame path IS reachable

This PR's `gate-triage.md` §1 addition says building a benchmark that reaches
`inline_subwalk=true` at a vable escape is the prerequisite for the rest. That
benchmark now exists, and the result goes further than expected — **the
multi-frame image is not just latched, it is built.**

The §1 text also says there are declines to clear before the adopt can fire, and
names the `jit.virtual_ref` emit as the prerequisite for the root-mismatch
decline. That framing is wrong for this path and will be corrected in a
follow-up commit; the measurement below is what should be recorded instead.

Five candidate programs, all sharing one shape — a `while` loop calling a
straight-line inlined callee that calls `sys._getframe` — produce, per run:

```
[s2-gate] inline_subwalk fs=1 flag=true writes_live=true odo_unchanged=true
          committed_none=false not_bridge=true bh_result_some=true sym_nonnull=true
[s2-build-decline] BUILT multi-frame depth=2
[fbw-blackhole] adopted single-frame terminal ...
```

Minimal form:

```python
import sys
_gf = sys._getframe

def leaf(x):
    _gf()
    return x + 1

def main():
    total = 0
    i = 0
    while i < 30000:
        total = leaf(total)
        i = i + 1
    return total

print(main())
```

Three things this establishes:

1. `[s2-gate]` has **never** fired for any of the 318 corpus benchmarks. It fires
   here. (Note `[s2-gate]` is itself gated on `inline_subwalk`, so it can only
   ever print `true` — its presence alone is the signal.)
2. `build_multi_frame_miframe` succeeds: **depth=2**, so none of the build-side
   conditions is the blocker.
3. The adopt still declines — `try_adopt_force_blackhole` is
   `try_adopt_multi_frame_blackhole(..) || try_adopt_single_frame_blackhole(..)`,
   and the single-frame arm is what adopts. So the remaining blocker is entirely
   inside `try_adopt_multi_frame_blackhole`, downstream of a successfully built
   image.

Why every existing bench misses: all five `getframe_*` benches drive with `for`.
With a FOR_ITER item in flight the callee's nested residual is declined by
`fbw_abort_nested_unjournaled_residual` before `execute_residual_call` runs, so
the force never happens inside the sub-walk. The driving loop must be `while`.

A depth-3 variant (`outer` → `inner`, both inlined) additionally reports
`[s2-build-decline] frame 1: parent.blackhole None (capture missing)`, so the
nested-parent capture is a separate, deeper gap.

Locating the adopt-side decline is in progress. The follow-up commit will carry
the corrected §1 text plus a benchmark for the corpus, so the path stops being
uncovered.

— *commented by Claude*

youknowone:
Follow-up to the previous comment: the adopt-side decline is located, and it is
one line whose own comment misdescribes it.

## The decline

`try_adopt_multi_frame_blackhole` (`pyre/pyre-jit-trace/src/trace.rs`):

```rust
if per_frame.first().map(|&(frame_ptr, _)| frame_ptr) != Some(cf_addr as i64) {
    return false;
}
```

The comment above it attributes the decline to "a chain rooted at an
intermediate frame … the walk descended into a residual call and inlined inside
it", and names the `jit.virtual_ref` emit at the inline push as what is needed
to close it. Instrumenting the repro shows that is not what happens. All five
events print:

```
first = 0xa4be2db40    live (ec.topframeref) = 0xa4be2db40    cf_addr = 0xa4c0515e8
all   = [0xa4be2db40, 0xa4be89e70]        # [main, leaf] — depth 2
```

`per_frame[0]` equals the **live** frame exactly. The chain *is* rooted at the
walked frame. The two sides are different representations of that same frame,
and the source says so:

```rust
let cf_addr = &*concrete_frame as *const PyFrame as usize;
// "The snapshot stands in for concrete stepping only"
sym.set_live_vable_frame_addr(live_frame_addr);
// "gap 10 slice 2b: set this BEFORE `init_symbolic` so the root vable identity
//  (seed_virtualizable_boxes) is baked against the live frame address, not the
//  discarded snapshot's."
```

The root vable identity that lands in the blackhole's frame register is the
**live** pointer by construction; the check compares it against the **snapshot**
pointer. The two can never be equal, so for this shape the decline is
unconditional — and no vref emit is involved.

## What this means for the ledger text in this PR

Two things in the §1 paragraph this PR adds are wrong and I will correct them in
a follow-up rather than silently leave them:

- It carries forward the decline comment's claim that the `jit.virtual_ref` emit
  gates the root-mismatch. It does not, on this path.
- It implies the §1 items (per-frame vable binding, outer-locals materialization)
  are what stand between here and a working multi-frame adopt. The build side
  already succeeds — `build_multi_frame_miframe` returns an image at **depth=2** —
  so those items are downstream of a gate that never opens.

What the paragraph got right and should keep: the multi-frame path has no corpus
coverage, and a benchmark reaching `inline_subwalk=true` at a vable escape was
the prerequisite. That benchmark now exists.

## Not fixing it in this PR

Swapping the comparison to the live address is a one-line change but not
obviously correct: `cf_addr` is used downstream in the same function for
`drive_multi_frame_blackhole(..., cf_addr, stack_base, ..)`,
`stack_base = concrete_nlocals(cf_addr)`, `apply_blackhole_crn(cf_addr, ..)`, and
the root `f_backref` link. Each needs deciding on its own, and the result wants
its own gating and A/B. This PR stays a dead-code-and-comments change.

A depth-3 variant (`outer` → `inner`, both inlined) additionally reports
`[s2-build-decline] frame 1: parent.blackhole None (capture missing)`, a
separate gap in the nested-parent capture.

## Gate status unchanged from the previous comment

Unit tests pass; `check.py --backend dynasm` 320 passed with the one perf bench
discriminated as machine load (isolated 35.5x / 33.0x / 31.8x vs its 40x gate,
against 43.6x under load average 28.76).

— *commented by Claude*

PR #815, opened 2026-07-26
jit: fold the immutable-type STORE_ATTR/DELETE_ATTR raise to a traced inline exception

fregataa:
## Problem

`synth/type_immutable_reject` is ~133x slower than CPython (pyre 0.74s vs cpython 0.17s vs pypy 0.02s). Profiling (`sample`, 5s window) attributes ~65% of the loop time to the `bh_store_attr_fn` / `bh_delete_attr_fn` residuals, almost all of it inside `PyError::to_exc_object`: every compiled iteration re-materialises the TypeError instance, its message string, and its args list through the runtime GC hooks (`dynasm_gc_owns_object` → `OldGen::contains`, shadow-stack pins, write barriers).

The trace records the raising store as `CallMayForceN + GuardNotForced + GuardException`, so the exception is born *inside* the opaque residual and comes out as an unvirtualizable guard-result box — even though everything downstream (inline PyTraceback construction, `sys_exc_value` save/restore, handler matching) is already traced. By contrast, an explicit `raise TypeError("...")` loop already runs at PyPy-class speed (~0.05s loop time) because the B3 construct fold (`try_walker_trace_exception_new`) exposes the allocation to the optimizer, which virtualizes it away when the exception never escapes.

## Change

Extend the B3 treatment to the one attribute-store raise whose outcome is provably iteration-invariant: a STORE_ATTR / DELETE_ATTR on a non-heap (immutable) builtin type.

- **`pyre-interpreter`**: new stability predicate `type_immutable_attr_raise_is_stable(obj, name, is_delete)` — accepts only a non-heaptype type receiver whose metaclass is the canonical `type` (MRO `(type, object)`, both frozen) with no metaclass descriptor for `name`. Under these conditions every dict the `setattr_str`/`delattr_str` raise-path decision consults is frozen, so the raise and its message are a pure function of trace-time constants.
- **`pyre-jit-trace`**: new walker fold `try_walker_trace_immutable_type_attr_raise` — pins the receiver with `GuardValue` (when not already constant), runs the authentic `setattr_str`/`delattr_str` concretely for the authoritative walk's exception, and emits the inline construction (`emit_object_list_inline` + `emit_exception_new_inline`, message as a rooted trace constant shared across iterations like a co_consts string) in place of the residual + guards, then seeds the residual-executor Err-arm state (`last_exc_value`, `BH_LAST_EXC_VALUE`, `fbw_built_exc`) and routes through the ordinary `SubRaise` path.
- **`majit-ir` / `pyre-jit`**: `PyreHelperKind::DeleteAttr` calldescr tag (appended at the enum tail, existing discriminants preserved) + tagging in `lower_delete_attr_hlop_to_insn`, so the walker can recognise the DELETE_ATTR residual — mirroring the existing `StoreAttr` tag.

Python-visible semantics are unchanged: the same TypeError (same message, fresh instance per iteration, traceback attached) is raised; only how the compiled trace produces it changes. Any shape the predicate cannot prove (heap types, custom metaclasses, metaclass descriptors like `type.__name__`) declines to the unchanged residual path.

## Results (Apple Silicon)

- `synth/type_immutable_reject` (dynasm): **0.74s → 0.11s** — now faster than CPython (0.15s); 10x from PyPy (was ~37x). The optimized trace contains no `CallMayForceN` residuals; both the store and delete exceptions virtualize. cranelift: 0.16s.
- `check.py --synthetic-only`: dynasm **314/314**, cranelift **314/314** — ALL PASSED.
- Edge-case parity (JIT output == `PYRE_NO_JIT=1` interpreter output, 7 scenarios): escaping exception message/args/traceback, delete-variant message, `int.__name__ = x` (metaclass-descriptor decline), heap-type store success in the same loop shape, custom-metaclass `__setattr__` override, per-iteration fresh exception identity.

## Status

- [x] dynasm: full synth suite green (314/314), target fixture 0.11s vs cpython 0.15s.
- [x] cranelift: full synth suite green (314/314), target fixture 0.16s.
- [ ] wasm backend untested (no local wasm32 target run).

## Summary by CodeRabbit

* **New Features**
  * Improved tracing for attribute assignments and deletions on immutable types.
  * Deterministic `TypeError` outcomes are now captured and replayed directly in optimized execution paths.

* **Bug Fixes**
  * Attribute deletion operations are now correctly recognized during optimization.
  * Unsupported or unstable attribute operations continue using the existing fallback behavior.

### DIFF
--- a/majit/majit-ir/src/effectinfo.rs
+++ b/majit/majit-ir/src/effectinfo.rs
@@ -712,6 +712,14 @@ pub enum PyreHelperKind {
     /// inserting a helper in the middle changes existing discriminants consumed
     /// by serialized/stable helper metadata.
     LoadMethodSelf,
+    /// `bh_delete_attr_fn(obj, code, name_idx)` — the plain DELETE_ATTR
+    /// residual (`lower_delete_attr_hlop_to_insn` → `space.delattr`), the
+    /// deletion counterpart of [`StoreAttr`](PyreHelperKind::StoreAttr).
+    /// The full-body walker recognises this tag to fold the deterministic
+    /// immutable-type raise (`del int.x` → TypeError before any dict
+    /// access) to a traced inline exception construction the optimizer
+    /// can virtualize; every other shape keeps the generic residual.
+    DeleteAttr,
 }
 
 impl EffectInfo {
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -9105,6 +9105,69 @@ pub fn setattr_str(obj: PyObjectRef, name: &str, value: PyObjectRef) -> PyResult
     object_setattr(obj, name, value)
 }
 
+/// Trace-time stability predicate for the walker's immutable-type
+/// STORE_ATTR / DELETE_ATTR raise fold: whether `setattr_str` /
+/// `delattr_str` on `obj` with `name` is *guaranteed* to take the
+/// terminal `object_setattr` / `object_delattr` non-heaptype raise —
+/// not just this call, but on every later call with the same `(obj,
+/// name)` pair.
+///
+/// The raise-path decision consults, in order: the weakref-proxy
+/// `force` (identity for a type receiver), the metaclass
+/// `__setattr__` / `__delattr__` resolution, the metaclass-MRO data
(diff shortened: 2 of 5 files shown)


youknowone (approved):
awesome! thank you so much!

PR #816, opened 2026-07-26
fix: collect all positional-only-as-keyword violations before raising

leehanjeong:
## Summary

`resolve_kwargs`, `bind_kwargs_to_signature`, and `call_with_kwargs` each raised on the first keyword naming a positional-only parameter instead of scanning the rest first, unlike PyPy's `_match_keywords` (`argument.py:464-501`) and CPython's `positional_only_passed_as_keyword` (`Python/ceval.c`), both of which collect every violator and report them together.

`f(a=1, b=2)` against `def f(a, b, /)` reported only `'a'` instead of `'a, b'`. Fixes `test_positional_only_arg`.

`ArgErr::PosonlyAsKwds::getmsg` also had a singular-vs-plural branch that used "got a positional-only argument passed as keyword argument" for a single name. Neither PyPy (`argument.py:635-637`) https://github.com/youknowone/pyre/blob/a527e14dcb51f446744430c251bad01e303702b7/pypy/interpreter/argument.py#L635-L637 nor CPython special-case a single violation — both always use the plural wording. Dropped the singular branch.

## Self-review

- [x] I fully resolved all reasonable code review comments from Codex and CodeRabbit.
  - If this is not checked, commits must include `Assisted-by`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting when positional-only parameters are provided as keyword arguments.
  * Error messages now list all affected parameter names together instead of stopping at the first one.
  * Standardized wording for single and multiple invalid arguments.

### DIFF
--- a/pyre/pyre-interpreter/src/argument.rs
+++ b/pyre/pyre-interpreter/src/argument.rs
@@ -1671,17 +1671,13 @@ impl ArgErr {
                 }
             }
             ArgErr::PosonlyAsKwds { posonly_kwds } => {
-                if posonly_kwds.len() == 1 {
-                    format!(
-                        "got a positional-only argument passed as keyword argument: '{}'",
-                        posonly_kwds[0],
-                    )
-                } else {
-                    format!(
-                        "got some positional-only arguments passed as keyword arguments: '{}'",
-                        posonly_kwds.join(", "),
-                    )
-                }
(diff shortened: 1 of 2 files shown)


youknowone (approved):
looks good, thank you!

PR #837, opened 2026-07-27
jit(fbw): type the walk-end commit contract and make the callee rebuild the preferred abort leg

youknowone:
17 commits: 10 code commits on the walk-end commit path in `pyre/pyre-jit-trace`, one majit parity fix, five plan-document updates and one rustfmt pass.

`run_perfn_walk`'s epilogue can keep the walk's eagerly-executed store journal instead of falling back to the legacy replay-from-entry. Whether a path's resume pc sits *behind* effects the walk already applied — a rewind, which re-executes them on top of committed ones — was decided ad hoc per leg across three files, and one path had no gate at all. This series types that obligation, then converges the abort-inside-an-inlined-callee case onto the leg the plan classes as orthodox.

Correctness/convergence only; **no performance claim is made**.

`walk-end-commit.plan.md` (new, tracked in-tree) is the authoritative record: R1, R2, R4 and R5 landed; **R3 deliberately landed no deletion**; R6 closed as refuted; R7 rejected.

## Walk-end commit contract (R1–R4)

- New `WalkEndResume { Terminal, Rewind{effects_at_resume_point}, RewindUnproven, RewindProvenAtLatch, AfterApplied }` in `trace.rs`. `commit_walk_end` takes it and is `#[must_use]`. Every call site is retagged (9 `commit_walk_end` sites across 7 pre-existing legs — `VableEscape` covers three of them).
- **This is a runtime change, not only a retag.** The escape flush's merge-point fallback — the path that had no gate — is now tagged `RewindUnproven`, always declines, and restores the live frame via `restore_escape_flush_undo()`. Legs 3 and 6 also re-check their odometer sample a second time at the commit point, so a path that previously committed after only the latch-time gate can now decline. The plan records the merge-point fallback as measured 0/334 on both backends, so this decline has no corpus witness either way.
- `WalkEndCommitLeg::TerminateNoReplay = 8` names the portal/CALL_ASSEMBLER no-replay shortcut. Of the eight tagged legs, seven hand the portal a resume pc; **this eighth hands none** — it keeps the journal under a different caller protocol. `record_walk_end_leg` (tag only) is split from `commit_walk_end` (tag + flush flag) so it stops reporting as `leg=0` in the census without setting `WALK_END_FLUSH_COMMITTED`.
- `ESCAPE_OPCODE_WINDOW` is re-keyed from a runtime `(frame_entry_count, executed_effect_count)` sample to the residual's declared `EffectInfo` class, matching the axis upstream licenses `resumepc=orgpc` on (`jtransform.py:620-630`). Note the new `reentrant_residual` is deliberately **stricter** than the `provably_side_effect_free` predicate it was split out of: it drops the `PyreHelperKind::ForIterNext` exemption, so a FOR_ITER's own next-residual now dirties the opcode window where the runtime-count gate left it clean.
- Two comments the sources contradict are corrected: upstream *does* rewind (`opimpl_str_guard_value`, `pyjitpl.py:1498-1511`), and the `vstack_cur_pypc` "one past the executing op" note was wrong.
- majit (`6de46f5ae4`): `finalize_standard_virtualizable_may_force` reloads the virtualizable fields from the heap before `TraceAction::Abort`, matching `vable_after_residual_call`. Without it the blackhole rebuilt from pre-call boxes.

## Leg 4 / R5

Leg 4 (`CalleeRebuild`) is the leg the plan classes as orthodox: `convert_and_run_from_pyjitpl` builds one blackhole per framestack frame at its own pc and splices the callee result into the caller *past* its call (`blackhole.py:1799-1821`, `:1653-1662`), and upstream's form is unconditional (`pyjitpl.py:2956`). Leg 3 (`EntryCarrierCall`) rewinds the caller to its CALL and has no upstream counterpart. The MidBody latch previously required the executed-effect odometer to have moved, which routed every zero-effect callee abort to leg 3.

This series drops that filter, then removes or widens each narrowing the resulting debug census named as a cause of leg-4 refusal:

- Drop the odometer conjunct; `fbw_set_abort_call_resume` no longer overwrites an already-latched MidBody carrier. Payload building becomes `Result<_, &'static str>` so every refusal **inside the builder** names itself under `PYRE_FBW_DEBUG_ABORT`. The outer `is_top_inline && !fbw_has_unjournaled_effect()` gate remains silent — see the residue table.
- **Widen** the abort anchor (`portal_vable_bookkeeping_anchor`) to admit a prefix of `setarrayitem_vable_r/rirdd` and `setfield_vable_i/rid` on any `VableField` aimed at this frame's `portal_frame_reg`. `getarrayitem_vable_r` stays refused, and 3 such refusals remain.
- Route a declining rebuild to the entry carrier rather than to the legacy replay — declining is **not** a safe fallback, since the store journal covers only five folded classes.
- Refuse to inline generator/coroutine/async-generator callees in `resolve_inlinable_callee`. That is the single shared resolve site, so the refusal also reaches the LOAD_METHOD specialization gate, the callee-candidate scan and the self-recursive-callee check.
- Deliver a rebuild into an expression-position caller. `outer_call_operands_below` computes the operand residue below the CALL; the return path's `depths[post_call_py_pc] == 1` gate becomes `== below + 1`, and the raise path's `handler_depth == 0` becomes `handler_depth <= below.len()`, restoring the operands the CALL sat on instead of resetting to the bare stack base.
- Delete the `x_arg` seed **and the refusal it carried** ("first callee argument is not a `Ref`") — the third refusal class the census named.

## R6 — refuted, not fixed

R6 claimed pyre's virtualizable shadow is lazy where upstream's is write-through, and that this is why the mid-expression operand-stack latch exists. If true it would retire `flush_escape_state_with_latched_stack`, `ESCAPE_OPCODE_WINDOW`, `EscapeFlushUndo`, `EscapeResumeKind`, `RewindProvenAtLatch` and `RewindUnproven`.

**The outcome is a refutation recorded in the plan, not a code fix.** A debug-gated `[r6-latch]` probe compares each latched slot against the vable shadow. Over `pyre/bench/synth`, every slot that disagreed (10 disagreements in 2 files) was the in-progress opcode's TOS, and in every one it held a compile-time NULL *constant* written by `popvalue_maybe_none` (`pyframe.py:411-417` → `setarrayitem_vable_r(_, depth, ConstPtr.NULL)` via `jtransform.py:1898`) — the in-flight opcode had already **popped** it. RPython's `popvalue` NULLs the slot the same way, so on this evidence the array is correct and matches upstream, and the latch is the orthodox register-level mirror of what upstream keeps in `MIFrame.registers_r`. It does not retire.

An intermediate commit's stated root cause ("the missing write is the `pushvalue` array store") is refuted by the next one: `LOAD_ATTR` — the escaping opcode in both witness files — already emits the push mirror via `emit_pushvalue_ref!` and its slot still reads NULL, because the pop follows the push.

The only executable change in this slice is the probe block, wholly inside `if fbw_debug_abort_enabled()` (`PYRE_FBW_DEBUG_ABORT`); it reads pure accessors and prints. The rest of R6 is documentation. (The rustfmt commit is separate and reformats R5 code that landed earlier on the branch.)

## Not done / residue

Recorded in the plan's `## Residue` table so it is not re-derived. Nothing here is blocking:

| row | why not worked |
|---|---|
| 3 remaining `getarrayitem_vable_r` anchors | needs a color-liveness argument, not a measurement |
| depth-1 / N-frame (the assumed R5 blocker) | **unobserved, not refuted** — its gate sits outside the named-refusal path, so the census cannot see it either way |
| `push_and_bump!` omits the `jtransform.py:1898` array store | real gap, no corpus witness; prototyped under a `Kind::Ref` gate and reverted — it took one bench from 140 to 175 emitted `setarrayitem_vable_r` and changed none of the 10 disagreements |
| legs 3/6 and the store journals | keep — deleting them re-opens the gh#467 double-apply |

## Verification

The series was verified **as a whole at HEAD**, after the branch was rebased onto current `main`:

- `cargo test --all --features dynasm` — 7083 passed, 0 failed
- `python3 ./pyre/check.py` — dynasm 334/334, cranelift 334/334, wasm 331/331
- `cargo fmt --all -- --check` — clean

Most individual commits also carry their own gate line, but five do not (`6de46f5ae4`, `9faf14afd9`, `f7ecad700b`, `0d6e137717`, `3928e1606e`); the per-commit numbers in those messages predate the rebase. Census figures quoted in commit messages come from debug-gated runs and are not re-derivable from the tree alone.

## Reviewer notes

- `try_commit_entry_carrier_call` has three decline points; only the first — `walk_end_resume_provable(Rewind{..})` on `FBW_EXECUTED_EFFECT_COUNT` — is a soundness re-check. That odometer is walker-side and plain interpretation does not bump it.
- Only `try_execute_residual_call_via_executor` calls `escape_opcode_window_note`, so residuals reaching the opcode by another path do not constrain the window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JIT state recovery after aborted or interrupted calls.
  * Ensured virtualizable values reflect current heap state instead of stale cached values.
  * Improved operand-stack and exception-state restoration during complex tracing scenarios.
  * Strengthened validation for inline calls, resume points, and virtualizable bookkeeping.
  * Generators, coroutines, and async generators are now handled more safely during optimization.

* **Documentation**
  * Added a detailed walk-end commit contract and roadmap.

* **Tests**
  * Expanded coverage for virtualizable state refresh, stack restoration, and tracing safety.

### DIFF
--- a/majit/majit-metainterp/src/pyjitpl.rs
+++ b/majit/majit-metainterp/src/pyjitpl.rs
@@ -4457,68 +4457,10 @@ impl<M: Clone> MetaInterp<M> {
             None => return,
         };
         let vable_ptr = self.vable_ptr;
-        if vable_ptr.is_null() {
-            return;
-        }
-        let (vable_box, array_lengths) = match self.tracing.as_ref() {
-            Some(ctx) => {
-                let Some(vable_box) = ctx.standard_virtualizable_box() else {
-                    return;
-                };
-                let array_lengths = ctx
-                    .virtualizable_array_lengths()
-                    .map(|lengths| lengths.to_vec())
-                    .unwrap_or_default();
-                (vable_box, array_lengths)
-            }
-            None => return,
-        };
-        let (static_boxes, array_boxes) = unsafe { info.read_all_boxes(vable_ptr, &array_lengths) };
         let Some(ctx) = self.tracing.as_mut() else {
             return;
         };
-        let cap = static_boxes.len() + array_boxes.iter().map(Vec::len).sum::<usize>() + 1;
-        let mut boxes = Vec::with_capacity(cap);
-        let mut values = Vec::with_capacity(cap);
-        for (index, value) in static_boxes.into_iter().enumerate() {
-            let (opref, concrete) = match info.static_fields[index].field_type {
-                majit_ir::Type::Int => (ctx.const_int(value), Value::Int(value)),
-                majit_ir::Type::Ref => (
-                    ctx.const_ref(value),
-                    Value::Ref(majit_ir::GcRef(value as usize)),
-                ),
-                majit_ir::Type::Float => (
-                    ctx.const_float(value),
-                    Value::Float(f64::from_bits(value as u64)),
-                ),
-                majit_ir::Type::Void => continue,
-            };
-            boxes.push(opref);
-            values.push(concrete);
-        }
-        for (array_index, items) in array_boxes.into_iter().enumerate() {
-            let item_type = info.array_fields[array_index].item_type;
-            for value in items {
-                let (opref, concrete) = match item_type {
-                    majit_ir::Type::Int => (ctx.const_int(value), Value::Int(value)),
-                    majit_ir::Type::Ref => (
-                        ctx.const_ref(value),
-                        Value::Ref(majit_ir::GcRef(value as usize)),
-                    ),
-                    majit_ir::Type::Float => (
-                        ctx.const_float(value),
-                        Value::Float(f64::from_bits(value as u64)),
-                    ),
-                    majit_ir::Type::Void => continue,
-                };
-                boxes.push(opref);
-                values.push(concrete);
-            }
-        }
-        boxes.push(vable_box);
-        // The vable identity's concrete value is the heap pointer itself.
-        values.push(Value::Ref(majit_ir::GcRef(vable_ptr as usize)));
-        ctx.set_virtualizable_boxes_with_info(boxes, values, &info, &array_lengths);
+        ctx.load_fields_from_virtualizable(&info, vable_ptr);
     }
 
     /// pyjitpl.py:1167-1172 `opimpl_getfield_vable_i(box, fielddescr, pc)`.
--- a/majit/majit-metainterp/src/pyjitpl/dispatch.rs
+++ b/majit/majit-metainterp/src/pyjitpl/dispatch.rs
@@ -1030,13 +1030,15 @@ where
         Some(active)
     }
 
-    fn finish_standard_virtualizable_after_residual_call(
+    /// The escaping half of `vable_after_residual_call` (pyjitpl.py:3372):
+    /// `Some` when `tracing_after_residual_call` reports the virtualizable
+    /// escaped during this CALL_MAY_FORCE.  Returns the still-rooted
+    /// [`ActiveStandardVirtualizable`] so the caller can read the heap object
+    /// before dropping the shadow-stack root.
+    fn escaped_standard_virtualizable(
         active: Option<ActiveStandardVirtualizable>,
-    ) -> bool {
-        let Some(active) = active else {
-            return false;
-        };
-        unsafe { active.info.tracing_after_residual_call(active.obj_ptr()) }
+    ) -> Option<ActiveStandardVirtualizable> {
+        active.filter(|active| unsafe { active.info.tracing_after_residual_call(active.obj_ptr()) })
     }
 
     fn finalize_standard_virtualizable_may_force(
@@ -1045,7 +1047,18 @@ where
         sym: &mut S,
         active: Option<ActiveStandardVirtualizable>,
     ) -> TraceAction {
-        if Self::finish_standard_virtualizable_after_residual_call(active) {
+        if let Some(active) = Self::escaped_standard_virtualizable(active) {
+            // pyjitpl.py:3377 `self.load_fields_from_virtualizable()` runs
+            // BEFORE the abort: the residual call forced the virtualizable and
+            // wrote its fields through the heap object, so the tracing-time
+            // box cache is stale.  Without this reload the blackhole rebuilds
+            // the resumed interpreter state from the pre-call boxes and the
+            // forced writes are lost.  The `MetaInterp` side already does this
+            // (`vable_after_residual_call`); this dispatcher holds no
+            // `MetaInterp`, so it calls the shared `TraceCtx` body directly.
+            // `active` stays alive across the read to keep the shadow-stack
+            // root for `obj_ptr` — its `Drop` pops that root.
+            ctx.load_fields_from_virtualizable(&active.info, active.obj_ptr());
             // pyjitpl.py:3373-3375 `raise SwitchToBlackhole(ABORT_ESCAPE,
             // raising_exception=True)` — stash the upstream-orthodox
             // abort reason + `raising_exception` flag on TraceCtx so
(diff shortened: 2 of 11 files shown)


youknowone:

Section 1 was empty. Acting on the rest.

### §2.1 — post-execution rewind fallback — **CONFIRMED, fixed in `539aa04`**

This one was real and I introduced it in this series. Verified against the source before acting:

`try_commit_midbody_abort` returned a `bool`, so the call site (`trace.rs:3126` pre-fix) took the `EntryCarrierCall` fallback for **every** decline — including the four that occur *after* `frame.execute_frame` has already run the callee body. That leg rewinds the outer frame to its CALL, so the body runs a second time.

The re-check does not catch it, and the reason is one I had already written into this PR's own reviewer notes without connecting it here: `try_commit_entry_carrier_call`'s `walk_end_resume_provable` samples `FBW_EXECUTED_EFFECT_COUNT`, which is walker-side, and the plain interpretation inside `execute_frame` never bumps it.

Fix: declines now carry which side of `execute_frame` they came from (`MidBodyDecline::{BeforeRun, AfterRun}`), and the rewinding leg is taken only on `BeforeRun`.

⚠️ **This narrows the hazard; it does not close it.** `AfterRun` now takes no leg, which drops through to the legacy replay — and that re-enters the outer frame at its entry and re-runs the CALL too. The callee's effects are user code, which the store journal does not cover, so neither branch can undo them. The real close is to make the post-run path infallible: every `AfterRun` decline already has a pre-run counterpart (`can_flush_walk_end_state_after_outer_call`, the propagate licence), so reaching one means a pre-check is too weak, not that another fallback is needed — which matches the observation that upstream has no post-execution decline at all. Tracked as **R8** in `walk-end-commit.plan.md`.

No corpus witness: the census row "rebuild latched but declined at the flush" measured 0.

### §2.2 — incomplete frame reconstruction vs. copying every `MIFrame` — **accurate, deferred (tracked)**

Agreed as a description. It is the plan's existing R5 residue row (the N-frame generalization), not a new divergence: `is_top_inline` and the closure/cellvar/freevar refusals pre-date this patch, which removed several *other* narrowings and named the rest. What this patch did change is that the path is now the preferred leg, which makes the incompleteness more visible.

One correction from this review's own line of attack: the plan had recorded "row 1 (`is_top_inline`) never fired" as a measurement. It is **not measurable** — that gate sits outside the closure whose `Err(&'static str)` arms print under `PYRE_FBW_DEBUG_ABORT`, so a depth≥2 sub-walk abort emits no line at all. Corrected to "unobserved, not refuted" in `d69ebbd`.

### §3.1 — single-latch escape recovery — **pre-existing, and investigated this cycle**

This is exactly what R6 examined. Outcome recorded in the plan: the latch is **orthodox**, not a workaround. Over `pyre/bench/synth` the only slot that ever disagreed with the vable shadow was the in-progress opcode's TOS, holding a compile-time NULL *constant* written by `popvalue_maybe_none` — the opcode had already popped it, and RPython's `popvalue` NULLs the slot identically. The latch holds what upstream keeps in `MIFrame.registers_r`.

### §3.2 — the escaping residual is not in its own purity window — **accurate, open, tracked as R9**

Correct and worth stating precisely: `escape_opcode_window_note` runs *after* `execute_residual_call` returns, so a residual constrains only those that follow it inside the opcode, never itself — while the escape resume pc is that same opcode. Upstream's licence for `resumepc=orgpc` is `EF_ELIDABLE_CANNOT_RAISE` on the residual being rewound over (`jtransform.py:620-630`), i.e. the residual itself.

Deliberately not changed here: folding the current residual into its own window converts latches into declines on a path with no measured decline today, and R6 established the latch is orthodox. Needs a witness first — a forcing residual that is neither elidable nor loop-invariant and whose re-execution is observable.

### §4 — structural adaptations — **accepted as-is**

All three match decisions already carrying in-code citations.

---

Verification after the fix, on the rebased branch: `cargo test --all --features dynasm` 7083 passed; `check.py` dynasm 334/334, cranelift 334/334, wasm 331/331.

— *commented by Claude*

PR #841, opened 2026-07-27
jit: stop the frame-chain walkers from forcing; keep last_instr current across residual calls

youknowone:
## What

`ExecutionContext::gettopframe_nohidden` and `getnextframe_nohidden` called `force_frame` on every frame they walked. `executioncontext.py` has no force in either walk — upstream emits `jit_force_virtualizable` per redirected **field access** (`rvirtualizable.hook_access_field`) and `jtransform` deletes it again in every graph the codewriter looks inside, so only graphs the JIT cannot see keep the call.

Forcing inside the walk escaped the traced virtualizable for every frame-walking helper, so any residual call that reached one raised `VableEscapedDuringResidualCall`. Concretely: `~bool` emits a DeprecationWarning, `_warnings::setup_context` walks the frame chain, and that walk aborted the trace — so a loop containing `~bool` never compiled at all.

Both walks are now force-free and the **consumers** force instead: `sys._getframe`, `sys._current_frames`, `PyFrame::fget_f_back`, and the coroutine-origin walk.

## Follow-on: `last_instr`

Nothing else kept the live frame's `last_instr` current once the walk stopped forcing, and that field has two incompatible required values at a residual-call site:

| meaning | value | written by |
|---|---|---|
| executing opcode (line reporting) | `vstack_cur_pypc` | `eval.rs frame.last_instr = pc`; the existing `mirror_vable_static_to_boxes` box half |
| resume coordinate (replay) | `resume_py_pc - 1` | `flush_walk_end_state_to_frame_inner`, with `ActiveFrameEscapeGuard` passing that same pc |

Exactly off by one, always. The old code never hit the conflict because forcing during the walk always ended in `ABORT_ESCAPE`, so the resume coordinate never had to survive the residual.

- **Compiled code**: a `SetfieldGc` of `last_instr` recorded from the *same constant* as the box mirror, so shadow and heap cannot disagree.
- **Recording walk**: `LiveLastInstrGuard` publishes the executing pc for the residual's duration and restores it afterwards. It targets `INLINE_CONCRETE_FRAME` when one is set — inside an inline sub-walk `vstack_cur_pypc` is in the *callee's* code, and writing it onto the walk's virtualizable strands that frame's replay on a stack depth it never had.

Without this, `~bool` in a traced loop reported **two** DeprecationWarnings (one at the `~` statement, one at the loop-entry line) where CPython reports one.

## Lift fixes

Four interpreter functions are marked `dont_look_inside` with their addresses registered in `jit_fnaddr`: `builtins::lookup_exc_class` (reads the `EXC_CLASS_REGISTRY` static), `host_seam::emit_stdout` / `emit_stderr` (host stdio handles), and a new `descroperation::bool_invert_deprecation_text` accessor wrapping a `PrebuiltText` static. Each static failed the front-end lift and, transitively, the lift of every caller — which is how `warn::warn_category_w` and `descroperation::invert` lost their jitcode.

Also ports `vable_after_residual_call`'s `debug_print` under `PYRE_FBW_DEBUG_ABORT`, naming the callee that forced the virtualizable.

## Review follow-ups

Two fixes from the review of this branch:

- `LiveLastInstrGuard` now **saves and restores** the previous publication instead of clearing it on drop. The guards in `residual_call.rs` nest — a residual runs user code that records a nested walk whose own residual enters the guard again — so clearing on the inner drop hid a still-live outer publication. Every sibling (`InlineConcreteFrameGuard.previous`, `ResidualFrameChainGuard.previous_published`, `ActiveFrameEscapeGuard.prev`) already saved-and-restored; being the odd one out was the tell. The corpus does not cover the nesting, so this was latent rather than reproducible.
- The executing-pc publish is now **skipped inside an inline sub-walk**. `mirror_vable_static_to_boxes` and `vable_setfield_descr` both named `standard_virtualizable_box()` — the walk's virtualizable — while using `vstack_cur_pypc`, which inside a sub-walk indexes the *callee's* code. Compiled code therefore stamped a foreign pc onto the caller's frame and `offset2lineno` resolved it against the caller's code object. `LiveLastInstrGuard` already retargeted the *concrete* store to `INLINE_CONCRETE_FRAME`; the IR emission did not. Both halves are now gated on the same predicate.

Verifying this area needs care — two natural-looking instruments are unsound. The warning dedup registry keys on lineno, so 99 999 compiled iterations that all report the same wrong line show up as **one** extra histogram entry, indistinguishable from a single stray recording event; and `warnings.simplefilter("always")` makes `show_warning` do stderr I/O every iteration, which stops the loop compiling (`Total # of loops: 0`) so the measurement is of the interpreter. The instrument that works is a counting `showwarning` with no I/O, plus confirming `Total # of loops > 0`.

## Numbers

`synth/arith_int_bool`: **11.6s → 2.65s** dynasm, 2.77s cranelift (cpython 1.00s, pypy 0.03s — PyPy has no `~bool` deprecation at all, so this bench's pypy ratio is structurally unwinnable; CPython is the honest comparison).

Warning attribution matches CPython exactly in the **flat** shape on both backends (`[(19, 100000)]`). It does **not** match when the warning is raised from a JIT-inlined callee: the frame chain has no callee frame, so the warning is attributed to the caller. This divergence is known and deliberately left here.

Closing it is not a matter of forcing in the consumer. That was built and measured: correct (`[(14,5),(15,100005)]` vs CPython `[(15,100000)]`) but 15.97/17.17/15.24 s against 4.61/6.81/7.05 s without — i.e. straight back to the regression this PR removes, because the force lands as a live may-force residual inside the (still lifted) `setup_context`, escapes the virtualizable and aborts every trace. It is "correct" only because nothing compiles. Upstream's `setup_context` jitcode carries **zero** force ops; forcing there is a property of the call boundary, not of the consumer.

The orthodox fix is to wire the vref producer — emit `enter`/`leave` (`f_backref` + `topframeref = virtual_ref(callee)`) at the inline push/pop, as upstream gets for free by tracing `executioncontext.enter`. Forcing a **vref** is a graceful `VIRTUAL_REF_FINISH`, not the `ABORT_ESCAPE` that forcing the **virtualizable** causes. That work is an epic with its own prerequisites (walker-path `_do_jit_force_virtual`, the vref residual-call bracket, `ResidualFrameChainGuard` vref-awareness, `EC_DESCR_GROUP` extension, scoped callee `last_instr`) and a blast radius covering the whole inline-call corpus, so it is deliberately out of scope for this PR.

check.py: **328/328 on both backends** (dynasm and cranelift), measured at `769da5c8` on base `2580479628` with the box quiet (load 13.1). The branch has since been rebased onto `512e5cb32b`; the files this PR touches are byte-identical across that rebase, but the base moved, so CI is the authority for the current head.

— *commented by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection object detection and forwarding behavior across supported backends.
  * Fixed frame inspection and traceback behavior when execution is optimized or inlined.
  * Corrected function-call and exception-result handling in translated code.
  * Improved warning attribution and initialization reliability.
  * Fixed hashing behavior, including cached string hashes and empty-value compatibility.

* **Performance**
  * Reduced temporary allocations during object creation, hashing, and warning processing.
  * Improved optimized handling of tuples, exception arguments, and residual calls.

* **Compatibility**
  * Strengthened inline handling for positional-only and star-argument calls.
  * Improved specialized tuple construction for exception arguments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,6 @@ aheui/
 CLAUDE.md
 
 # Personal planning docs (not committed)
+
+# Scratch probes, profiles and baseline binaries (not committed)
+/scratchpad/
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -352,6 +352,9 @@ fn ensure_jitframe_type_registered(gc: &mut dyn GcAllocator) -> Option<u32> {
 /// Install a GC box into TLS and register all `set_active_*` hooks.
 /// Shared by `install_gc_standalone` (production) and `set_gc_allocator` (tests).
 fn install_gc_box(mut gc: Box<dyn GcAllocator>) {
+    // Per-thread allocator: its nursery is not the singleton's, so the
+    // process-wide published range can no longer answer `is_nursery_object`.
+    majit_gc::disarm_published_nursery();
     let jitframe_type_id = ensure_jitframe_type_registered(gc.as_mut());
     gc.freeze_types();
     let supports_guard_gc_type = gc.supports_guard_gc_type();
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -211,6 +211,10 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
 /// [`install_gc_standalone`], which registers the same hooks WITHOUT a
 /// box so the trampolines fall through to `gc_sync`.
 fn install_gc_box(gc: Box<dyn majit_gc::GcAllocator>) {
+    // This thread now answers heap queries from its own allocator, whose
+    // nursery is not the singleton's, so the process-wide published range can
+    // no longer stand in for `is_nursery_object`.
+    majit_gc::disarm_published_nursery();
     let supports_guard_gc_type = gc.supports_guard_gc_type();
     DYNASM_ACTIVE_GC.with(|cell| {
         let mut guard = cell.borrow_mut();
--- a/majit/majit-backend-wasm/src/lib.rs
+++ b/majit/majit-backend-wasm/src/lib.rs
@@ -331,6 +331,9 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
 /// registers the same hooks WITHOUT a box so the trampolines fall through
 /// to `gc_sync`.
 fn install_gc_box(gc: Box<dyn majit_gc::GcAllocator>) {
+    // Per-thread allocator: its nursery is not the singleton's, so the
+    // process-wide published range can no longer answer `is_nursery_object`.
+    majit_gc::disarm_published_nursery();
     let supports_guard_gc_type = gc.supports_guard_gc_type();
     WASM_ACTIVE_GC.with(|cell| {
         let mut guard = cell.borrow_mut();
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -3763,6 +3763,15 @@ impl GcAllocator for MiniMarkGC {
         self.is_nursery_object_start(addr)
     }
 
+    fn nursery_bounds(&self) -> Option<(usize, usize)> {
+        let start = self.nursery.start_ptr() as usize;
+        Some((start, start + self.nursery.size()))
+    }
+
+    fn taggedpointers(&self) -> bool {
+        self.config.taggedpointers
+    }
+
     fn write_barrier(&mut self, obj: GcRef) {
         self.do_write_barrier(obj);
     }
--- a/majit/majit-gc/src/gc_sync.rs
+++ b/majit/majit-gc/src/gc_sync.rs
@@ -93,6 +93,7 @@ pub fn store_singleton(gc: Box<dyn GcAllocator>) {
     // SAFETY: gc_mutex held, no concurrent access.
     unsafe {
         *GC_STORE.0.get() = Some(gc);
+        crate::publish_singleton_nursery(&**(*GC_STORE.0.get()).as_ref().unwrap());
     }
     GC_INITIALIZED.store(true, Ordering::Release);
 }
@@ -115,6 +116,7 @@ pub fn replace_singleton_leaking_old(gc: Box<dyn GcAllocator>) {
             std::mem::forget(old);
         }
         *GC_STORE.0.get() = Some(gc);
+        crate::publish_singleton_nursery(&**(*GC_STORE.0.get()).as_ref().unwrap());
     }
     GC_INITIALIZED.store(true, Ordering::Release);
 }
--- a/majit/majit-gc/src/lib.rs
+++ b/majit/majit-gc/src/lib.rs
@@ -398,6 +398,19 @@ pub trait GcAllocator: Send {
         false
     }
 
+    /// The nursery's `[start, end)`, when it is a fixed range for the life of
+    /// this allocator.  `None` (the default) keeps callers on the full
+    /// `is_nursery_object` query.
+    fn nursery_bounds(&self) -> Option<(usize, usize)> {
+        None
+    }
+
+    /// `gc/base.py:380-383 is_valid_gc_object`'s tagged-immediate setting
+    /// (`translationoption.py:185 taggedpointers`, default off).
+    fn taggedpointers(&self) -> bool {
+        false
+    }
+
     /// Current nursery free pointer.
     fn nursery_free(&self) -> *mut u8;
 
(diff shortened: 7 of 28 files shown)


youknowone:
Re-verified against the current base (`70a68163c6`, on top of #835) after re-extracting LLBC and rebuilding both backends:

```
ALL PASSED: dynasm 326/326
ALL PASSED: cranelift 326/326
```

Correctness spot-checks on both backends: `~bool` in a hot loop emits exactly one DeprecationWarning at the `~` statement (matching CPython 3.14 in both count and line), `warnings.simplefilter("always")` reports all 30000 occurrences at the correct line, results match, and `PYRE_FBW_DEBUG_ABORT=1` shows zero `VableEscapedDuringResidualCall`.

Two earlier local runs failed a boundary perf bench (`raise_catch` 1.6x/1.5x gate, then `nested_loop` 2.2x/2x gate) while the box was at load 40–150. Those were machine noise, not regressions — the *failing bench migrated* between runs, a same-epoch binary A/B showed no difference (0.21–0.24s baseline vs 0.21–0.28s patched), and at load ~13 the patched build scored `raise_catch` **1.4x** where the parent commit scored 1.5x. This full-green run is the confirmation.

`synth/arith_int_bool` reads 2.65s dynasm / 2.77s cranelift on a quiet box; the number rises to ~4.2s under load, so compare it only against a same-epoch baseline.

— *commented by Claude*

youknowone:
Rebased onto the current `main` (`daca722c01`) and force-pushed; the branch head is now `8ca19b19da`.

**Added a commit from the Codex parity review.** The review flagged `builtins.rs` hashing; one of its two findings held up and the root cause was bigger than reported:

`_hash_str` short-circuited empty input to 0 without running siphash. `ll_strhash`'s `return 0` arm is a *null-pointer* check (`if s:`), not an empty-string case — upstream digests `b""` like any other value. Measured against `PYTHONHASHSEED=0` PyPy:

| | before | after | PyPy (seed 0) |
|---|---|---|---|
| `hash("")` | 29742 | 2202906307356721367 | 2202906307356721367 |
| `hash(b"")` | 0 | 2202906307356721367 | 2202906307356721367 |
| `hash(memoryview(b""))` | 0 | 2202906307356721367 | 2202906307356721367 |

That value is SipHash-2-4 under the all-zero key over empty input (`0x1e924b9d737700d7`). Every non-empty digest already matched PyPy byte-for-byte and is unchanged — verified by diffing a table of str and bytes hashes against seed-0 PyPy, identical.

The commit also moves the zero-digest substitute from the str caller into `_hash_str` (where `_ll_strhash` applies it) and corrects its value from `29742` to upstream's `29872897`, so bytes and memoryview agree with the value a str caches for the same bytes.

For the record on the review's other two points: finding 1 ("empty strings should hash as 0, per `ll_strhash`") is **not** correct — PyPy's `hash("")` is nonzero, as measured above. Finding 3 (permanently zero SipHash key vs a random `PYTHONHASHSEED`) is real and pre-existing, and is deliberate for reproducibility; not touched here.

`cargo fmt --check` was failing on a line in `LiveLastInstrGuard::drop`; fixed.

Verification on `8ca19b19da`, with LLBC re-extracted and both backends rebuilt (per-crate fingerprints confirmed current):

```
ALL PASSED: dynasm 326/326
ALL PASSED: cranelift 326/326
```

— *commented by Claude*

youknowone:
Fixed the one finding from the latest Codex parity review that lands on new code in this PR, and it was a real latent defect.

`PUBLISHED_LAST_INSTR` is a single TLS slot but `LiveLastInstrGuard` is entered per residual call, and a residual can run user code that records a nested walk whose own residual enters a second guard. The inner drop *cleared* the slot, so a still-live outer publication went invisible and `capture_escape_flush_undo` would snapshot the executing pc as the outer frame's resume coordinate — precisely the failure the TLS exists to prevent. It now saves the displaced entry and restores it, the same discipline `InlineConcreteFrameGuard` and `ResidualFrameChainGuard` already use. (My guard was the only thread-local holder in that file that cleared instead of restoring, which is what made the report convincing.)

Not exercised by the corpus — check.py was green before and after — so this is a latent fix, not a repro.

The review's other two points are noted but not addressed here:
- **`STATE_NS` process-global vs `space.fromcache(State)`** — real, pre-existing, and only observable with multiple object spaces, which pyre does not create today. Worth noting that profiling this warn path earlier showed `fromcache(State)` accounting for ~0% of it, so there is no perf motive either; it belongs in a `_warnings`-ownership change, not here.
- **`W_UnicodeObject.hash` plain `i64` reads/writes under free-threading** — also pre-existing (the field arrives with the str-hash-memoization commit that predates this work) and correctly identified as needing an atomic cache field. Out of scope for this PR.

Final verification on `10bf87dbc8`, both backends rebuilt, on a quiet box (load ~5):

```
ALL PASSED: dynasm 326/326
ALL PASSED: cranelift 326/326
```

Low-load numbers from that run: `synth/arith_int_bool` **2.61s** dynasm / 2.80s cranelift (cpython 0.97s), `raise_catch` 1.4x, `nested_loop` 1.9x. `hash("")`/`hash(b"")`/`hash(memoryview(b""))` identical to `PYTHONHASHSEED=0` PyPy, and `~bool` in a hot loop still emits exactly one DeprecationWarning at the right line.

— *commented by Claude*

PR #848, opened 2026-07-28
interp: complete module __getattr__ routing (descriptor __get__ errors + surrogate-name misses)

youknowone:
Two related fixes closing gaps where a module attribute lookup failed to reach the module's `__getattr__` hooks. Both mirror `module.py Module.descr_getattribute` / `objspace.py:694-699` / `descroperation.py:242-245`, and both are oracle-agreed (python3.14 == pypy3).

## 1. Descriptor `__get__` AttributeError (flagged by Codex on #844)

`getattr_str_impl` inlines the default module `__getattribute__` for the hot path but propagated an `AttributeError` raised by a module-type data / non-data descriptor's `__get__` instead of routing it — unlike the `module_getattribute` slot's own `try`/`except`. So `getattr(module, name)` where a `ModuleType`(-subclass) descriptor's `__get__` raises `AttributeError` leaked the error rather than consulting the module's PEP 562 `__getattr__`.

- On the `space.getattr` path (`call_getattr`), catch that `AttributeError` and route it through the module dict's PEP 562 `__getattr__`, then the class-level `__getattr__`. The bare `object.__getattribute__` slot (`!call_getattr`) still propagates, so `module_getattribute`'s own `try`/`except` performs the routing.
- This makes the module fast-path consistent with the instance and type paths, which already route descriptor `__get__` `AttributeError`s.
- Extract the shared tail as `module_getattr_fallback` and reuse it in the module-miss tail.
- Test: `synth/module_getattr_descr_error`.

## 2. Surrogate-name module miss

`getattr_surrogate` consulted the module dict's PEP 562 `__getattr__` on a lone-surrogate attribute miss but returned the `AttributeError` instead of falling through to the class-level `__getattr__` on the module's type, unlike the non-surrogate `module_getattr_fallback`. `getattr(module, "\udc80")` on a `ModuleType` subclass with a class-level `__getattr__` raised instead of calling it.

- Drop the early return so a surrogate-name miss reaches the class-level `__getattr__` too.
- Test: `synth/module_getattr_surrogate_cls`.

## Verification

- Both fixes: `getattr` result identical across python3.14, pypy3, pyre-dynasm JIT, pyre-dynasm NOJIT.
- `check.py --backend dynasm`: ALL PASSED 328/328.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module attribute lookup so `AttributeError` from descriptor `__get__` is routed consistently through module-level fallback behavior, with reliable class-level fallback when needed.
  * Updated lone-surrogate attribute name handling to consult module-level `__getattr__` first before falling back.

* **Tests**
  * Added benchmark coverage for descriptor failures and module attribute fallback behavior, including repeated surrogate-name lookups.

### DIFF
--- a/pyre/bench/synth/module_getattr_descr_error.py
+++ b/pyre/bench/synth/module_getattr_descr_error.py
@@ -1,0 +1,63 @@
+# module.py Module.descr_getattribute (PEP 562): an AttributeError raised by
+# the normal module lookup — including a module-type data / non-data
+# descriptor's __get__ (objspace.py:694-699) — is caught and routed to the
+# module-level __getattr__, the same as a plain attribute miss.  Without a
+# module __getattr__ the error propagates.  The descriptors live on a
+# ModuleType subclass so the base module slot stays intact.
+
+import types
+
+
+class DataDesc:
+    def __get__(self, obj, objtype=None):
+        raise AttributeError("data descr")
+
+    def __set__(self, obj, value):
+        pass  # __set__ makes it a *data* descriptor
+
+
+class NonDataDesc:
+    def __get__(self, obj, objtype=None):
+        raise AttributeError("non-data descr")
+
+
+class M(types.ModuleType):
+    ddesc = DataDesc()
+    ndesc = NonDataDesc()
+
+
+def with_hook():
+    m = M("m")
+
+    def modgetattr(attr):
+        return "hook:" + attr
+
+    m.__getattr__ = modgetattr
+    return m
+
+
+def main():
+    m = with_hook()
+    # a data descriptor whose __get__ raises AttributeError routes to the hook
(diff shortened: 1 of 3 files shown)


youknowone:
Addressed the P2 in 6777990be9: `getattr_surrogate` now catches an `AttributeError` from the module-dict `__getattr__` hook and falls through to the class-level `__getattr__`, carrying the hook's error as the terminal one — matching the identifier path's `module_getattr_fallback`. Verified against python3.14 and pypy3 (both call the class hook), identical on pyre jit/nojit; `synth/module_getattr_surrogate_cls` extended with the raising-hook cases (`ident3`/`surr3` with a class fallback, `ident5`/`surr5` without).

— *commented by Claude*

PR #850, opened 2026-07-28
gc: deliver __del__ for builtin subclasses; walk referents; hash a weakref as its referent

youknowone:
Three-way differential work against `python3.14` (3.14.5) and `pypy3` (7.3.22).
Classification used throughout: `pyre==pypy!=cpython` is left alone,
`pyre!=pypy==cpython` is fixed, `pyre!=pypy!=cpython` is a pyre-only invention.

## `__del__` was never called for a subclass of most builtins

`objspace.py:485-487 allocate_instance` ends with `user_setup` and then
`if w_subtype.hasuserdel: register_finalizer(instance)`. pyre has that tail
scattered across per-allocation-site code, and most sites tagged `w_class`
without registering. `class D(int)` with a `__del__` never ran it.

Two commits complete the pattern:

- int, float, complex, bytes, bytearray, enumerate, map, filter, zip.
  int/float/complex needed a collectable allocator (`w_int_subclass_new` and
  friends) — the existing `w_int_new_unique` serves four JIT hot-path call
  sites and stays immortal.
- the exception constructors (`exc_new_wrapper!`, the OSError family,
  ExceptionGroup), type objects and their metaclass, staticmethod,
  classmethod, property, ModuleType, the itertools allocator, `array`,
  `_random.Random`, and the `_io` types.

`_io` registers through its own helper rather than the shared one.
`iobase_del` is a builtin method on the interp-level type, so `hasuserdel` —
which `type.__new__` sets only for a class whose own dict carries `__del__` —
is false and the gated hook skips the object; `interp_iobase.py:63-64`
registers explicitly for the same reason. Using both paths would run a
subclass's `__del__` twice. One consequence is that a file dropped without
`close()` now flushes: `open(p,'w')`, `write`, `del`, collect left the file
empty and now leaves the bytes on disk.

`BufferedWriter.__new__` and its three siblings ignored their `cls` argument
entirely; they now tag with it.

`array` instances stay unreclaimable: `W_Array::allocate` is `malloc_typed`,
so their registration is inert until that moves to the managed heap.

## weakref hashed its own address

`interp__weakref.py:218` is `self.w_hash = self.space.hash(w_obj)`. `descr_hash`
returned the referent's address instead, so `hash(ref) != hash(obj)` and a
weakref could not be used to look up its referent in a dict. It now delegates
to `hash_w_strict`, which also supplies the unhashable-referent TypeError, the
-1 to -2 fold and propagation of a raising `__hash__`. The lazy cache that
keeps the hash stable after the referent dies is unchanged. The referent is
rooted across the call because a user `__hash__` can allocate.

## gc module

`gc.get_referents` and `gc.get_referrers` returned an empty list. PyPy
implements both, so `referents.py:53-78 _list_w_obj_referents` is ported as
`MiniMarkGC::do_get_referents`, reached through a `get_referents` hook next to
the existing `get_objects` one with a trampoline in each of the three
backends. The walk stops at the first app-level object on a branch and expands
anything else, so a list reports its items rather than the array holding them;
GCFLAG_EXTRA is restored on every visited header before returning.
`get_referrers` composes the two the way `referents.py:147-169` does.

- `gc.is_tracked` answered False for everything, including lists and
  instances. It now reports whether the collector owns the object.
- `gc.set_threshold` took no arguments, so `gc.set_threshold(*gc.get_threshold())`
  raised TypeError. It now accepts one to three, parses them all before
  writing any, and writes only the positions given. `gc.get_threshold` reads
  the stored values back and starts at 3.14's `(2000, 10, 10)` rather than
  `(700, 10, 10)`.
- `gc.unfreeze` and `gc.get_freeze_count` join the `gc.freeze` no-op that was
  already there. The collector has no permanent generation, so nothing is ever
  frozen and the count is zero.

### Deliberately still absent

`get_stats` (PyPy's `W_GcStats` has 13 fields; `active_heap_stats()` supplies
two), `set_debug`/`get_debug` (storing the flag without honoring DEBUG_SAVEALL
would silently drop uncollectables the caller asked to retain),
`gc.callbacks` invocation (firing only on explicit `collect()` reads as "no
collections happened"), and `is_finalized` (`GCFLAG_IGNORE_FINALIZER` is a
*don't-run* marker, `incminimark.py:3043`, not a *has-run* record). A fake
would be worse than the AttributeError.

### Known divergences, verified as correct

- `gc.get_referents([1,2,3])` is `[]`, matching PyPy exactly (its list
  strategies unbox small ints); CPython reports 3.
- `gc.get_referents({'a': 1})` is `[]` where PyPy gives 2 and CPython 1.
  pyre's exact str and small int are immortal and the walk stops at the
  managed-heap boundary — the same boundary `gc.get_objects` already drew.
  Widening it has no safe address test: an immortal object carries a GC
  header, but a slot such as `ob_type` can hold a static that has none.
- `gc.get_referrers(a, a)` reports the referrer twice, as PyPy does (the
  append sits inside the per-argument loop); CPython reports it once.
- `gc.is_tracked((1, 2))` is True; CPython untracks tuples of untracked items.

## Verification

- `check.py --backend dynasm,cranelift,wasm`: dynasm 326/326, cranelift
  326/326, wasm 323/323, after re-extracting LLBC. Re-run after the rebase
  onto current main.
- `cargo test`: majit-gc 205, pyre-object 275, pyre-interpreter 401.
- Every builtin base in a 16-base subclass matrix and all seven weakref-hash
  cases now agree with CPython 3.14.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -388,6 +388,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_alloc_oldgen_typed(Some(alloc_oldgen_typed_via_active_runtime));
     majit_gc::set_active_collect_full(Some(collect_full_via_active_runtime));
     majit_gc::set_active_get_objects(Some(get_objects_via_active_runtime));
+    majit_gc::set_active_get_referents(Some(get_referents_via_active_runtime));
     majit_gc::set_active_collect_oldgen(Some(collect_oldgen_nonmoving_via_active_runtime));
     majit_gc::set_active_heap_stats(Some(heap_stats_via_active_runtime));
     majit_gc::set_active_root_hooks(
@@ -1610,6 +1611,11 @@ fn collect_full_via_active_runtime() {
 fn get_objects_via_active_runtime(generation: i8, visitor: majit_gc::GetObjectsVisitorFn) {
     let mut visit = visitor;
     with_cranelift_gc(|gc| gc.get_objects(generation, &mut visit));
+}
+
+fn get_referents_via_active_runtime(obj: GcRef, visitor: majit_gc::GetObjectsVisitorFn) {
+    let mut visit = visitor;
+    with_cranelift_gc(|gc| gc.get_referents(obj, &mut visit));
 }
 
 /// Non-moving old-gen-only major trampoline — sweeps dead old-gen objects
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -389,6 +389,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_collect_full(Some(collect_full_via_active_runtime));
     majit_gc::set_active_get_objects(Some(get_objects_via_active_runtime));
     majit_gc::set_active_get_referents(Some(get_referents_via_active_runtime));
+    majit_gc::set_active_is_tracked(Some(is_tracked_via_active_runtime));
     majit_gc::set_active_collect_oldgen(Some(collect_oldgen_nonmoving_via_active_runtime));
     majit_gc::set_active_heap_stats(Some(heap_stats_via_active_runtime));
     majit_gc::set_active_root_hooks(
@@ -1616,6 +1617,10 @@ fn get_objects_via_active_runtime(generation: i8, visitor: majit_gc::GetObjectsV
 fn get_referents_via_active_runtime(obj: GcRef, visitor: majit_gc::GetObjectsVisitorFn) {
     let mut visit = visitor;
     with_cranelift_gc(|gc| gc.get_referents(obj, &mut visit));
+}
+
+fn is_tracked_via_active_runtime(obj: GcRef) -> bool {
+    with_cranelift_gc(|gc| gc.is_tracked(obj)).unwrap_or(false)
 }
 
 /// Non-moving old-gen-only major trampoline — sweeps dead old-gen objects
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -192,6 +192,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_alloc_oldgen_typed(Some(dynasm_alloc_oldgen_typed));
     majit_gc::set_active_collect_full(Some(dynasm_collect_full));
     majit_gc::set_active_get_objects(Some(dynasm_get_objects));
+    majit_gc::set_active_get_referents(Some(dynasm_get_referents));
     majit_gc::set_active_collect_oldgen(Some(dynasm_collect_oldgen_nonmoving));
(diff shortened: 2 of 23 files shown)


youknowone:
Codex review dispositions, both verified against `python3.14` and `pypy3` before acting.

### ① `is_tracked` — confirmed, fixed in `9a36daa3c9`

Reproduced: `gc_owns_object` made the answer depend on where the instance happened to land.

| | `int` | `float` |
|---|---|---|
| pyre default | False | False |
| `PYRE_GC_INTERP=1` | **True** | **True** |
| CPython 3.14 | False | False |

`gc.is_tracked` now asks the registered type whether the collector traverses references out of it (`MiniMarkGC::object_is_tracked` → `has_gc_ptrs || items_have_gc_ptrs`), reached through an `is_tracked` hook with a trampoline in each backend, plus a unit test that asserts the answer survives promotion out of the nursery. The two configurations now print identical tables, and `(1, 2)` reports False — what CPython converges to after one `gc.collect()`.

`bytes` and `int` still answer True where CPython answers False. That is not the predicate being wrong: `W_BytesObject` carries `w_dict` and `w_weakreflifeline` slots that the collector really does follow, and the CPython objects have no equivalent. Documented at the call site.

### ② `_io` / weakref duplicate registration — half confirmed, not fixed here

**Confirmed in code.** `register_weakref_finalizer` (`executioncontext.rs:203`) excludes `hasuserdel` objects and the two native-buffer types but does not know about `register_iobase_finalizer`, and the collector's `register_finalizer` (`collector.rs:3875`) does not deduplicate. A weakref'd instance of one of the five wired `_io` types is therefore queued twice.

**The stated harm does not reproduce.** `__del__` runs exactly once, with and without a weakref, for `BufferedWriter`, `TextIOWrapper` and `BufferedReader` (counted through an overridden `close`), for an inherited `__del__`, and for a suspended generator (the other undeduplicated path, pre-existing). The finalization-ordering pass absorbs the duplicate — consistent with the review's own "the duplicate survives into another finalization cycle". The residual cost is a lingering deque entry.

**One premise is wrong.** "subclasses that merely inherit `__del__`, whose own `hasuserdel` remains false" — `hasuserdel` *is* set for an inherited `__del__` (`call.rs:4257-4258`), and `class B(A)` with `A.__del__` measures 1 call, not 2.

**A real adjacent gap surfaced while verifying**, pre-existing rather than introduced here: registration is attached to the five `_io` types that own a `__new__`, so the ones built through the generic instance path — `io.FileIO`, and user subclasses of `io.IOBase` — never register, and `close()` is not called when they are collected (CPython calls it). Upstream attaches registration to `W_IOBase.__init__` (`interp_iobase.py:63-64`), not to `__new__`, so the correct fix moves it to the `__init__` level and then lets `register_weakref_finalizer` skip that whole set. Doing that safely needs a per-object "already registered" marker that does not exist yet, so it is left as a follow-up rather than landed half-done.

### Gate

`check.py --backend dynasm,cranelift,wasm` after re-extracting LLBC on the current base: dynasm 328/328, cranelift 328/328, wasm 325/325. `cargo test`: majit-gc 206, pyre-interpreter 401.

The earlier `pyre/check.py (ubuntu-24.04)` red was `cranelift int_loop ratio 2.1x > gate 2x` — the known perf-ratio flake on that runner; every correctness case passed and macOS/Windows were green.

— *commented by Claude*

PR #852, opened 2026-07-28
jit(gh#346): pin the symmetric-difference result set across the merge loop

youknowone:
## What

Follow-up to #846. That slice hoisted the symmetric-difference merge loop into `set_symmetric_difference_storage` and, in doing so, dropped the `push_roots()` + `pin_root(d_new)` that the removed `pyre_object::setobject` helper carried.

`d_new` (the fresh result set from `w_set_new()`) is reachable only from the merge's Rust frame. The GC sees roots via the shadow stack only — the two operands are rooted through the interpreter frame, but `d_new` is not. A probe `eq_w` that triggers a collection therefore sweeps the in-progress result body; the next insert or the return then touches freed storage.

- interpreter path: `GC BUG: invalid type_id` crash
- JIT path: `AttributeError` type-confusion (freed storage reused)

Both reproduced on the merged slice, then fixed here.

## Fix

Wrap the merge in `push_roots()` and `pin_root(d_new)`. The set body is old-gen (non-moving) so it keeps its address across a collection — the pin prevents the sweep and needs no reload. `+7/-2` in `typedef.rs`; the traced-loop shape (matching `set_intersect_update`) is unchanged, so the census head stays retired.

## Why not re-residualize the helper

Restoring the deleted `#[dont_look_inside]` + `jit_fnaddr` helper (the #846-slice-1 pattern) **revives** the census head: the helper returns `Result<PyObjectRef, SetUpdateError>`, and the annotator cannot decode the custom enum across the residual boundary (`map_set_update_error() takes exactly 1 argument (0 given)`), regressing phaseA 321→322. Void residuals and null-niche `Option<PyObjectRef>` residuals cross the boundary; custom-enum-`Result` residuals do not. So the pin-only path is the census-clean fix.

## Test

Extends `set_symdiff_raising_eq.py` with:
- a `HashBoom` raising-`__hash__` case (the error must surface before any probe),
- return-type annotations on the `__hash__` methods,
- a `Collect` class whose `__eq__` runs `gc.collect()` — this is the pin guard; without the pin it crashes as above. The collect loop is kept small because a full collection per single-bucket probe is O(heap).

## Verification

- census A/B: head-neutral (phaseA 321 before ≡ 321 after; `comm` diff empty both ways)
- `check.py --backend dynasm,cranelift`: dynasm 326/326, cranelift 326/326, bit-exact
- `cargo test` pyre-object / majit-translate / pyre-interpreter (dynasm,host_env): green
- snippet passes under JIT, interpreter, and CPython oracle

Note: `set_intersect_update` shares the same live-read shape (a pre-existing class deviation, PyPy `_intersect_unwrapped` / `_symmetric_difference_unwrapped` both unerase once); restoring capture-once for both jointly is a separate follow-up, out of scope here.

### DIFF
--- a/pyre/extra_tests/snippets/set_symdiff_raising_eq.py
+++ b/pyre/extra_tests/snippets/set_symdiff_raising_eq.py
@@ -12,13 +12,23 @@ class Boom:
         self.key = key
         self.explode = explode
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return 7  # single bucket -> __eq__ runs on every probe
 
     def __eq__(self, other):
         if self.explode or getattr(other, "explode", False):
             raise ValueError("boom in __eq__")
         return self.key == other.key
+
+
+class HashBoom:
+    """__hash__ raises: symmetric_difference must surface it before any probe."""
+
+    def __hash__(self) -> int:
+        raise ValueError("boom in __hash__")
+
+    def __eq__(self, other):
+        return self is other
 
 
 def run_symdiff(a, b):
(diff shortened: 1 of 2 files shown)


youknowone:
### CI note: `pyre/check.py (ubuntu-24.04)` red is the known `cranelift int_loop` perf-gate flake, not a regression

The only failing check is:

```
FAIL cranelift int_loop  exec 0.56s > pypy 0.27s  ratio 2.0x > gate 2x
ALL PASSED: dynasm 326/326
FAILED: cranelift 1 failed, 325 passed
ALL PASSED: wasm 323/323
```

Evidence it is a flake, not caused by this PR:

- **`main` at `4a0f926140` (#843) — the exact commit this PR is rebased onto — fails the identical bench** (`pyre CI` run `30321615653`, job `90161394033`): `FAIL cranelift int_loop exec 0.63s > pypy 0.31s ratio 2.0x > gate 2x`, dynasm 326/326 + wasm 323/323 green. So the red predates and is independent of this change.
- The failing bench is `int_loop` — a pure integer loop with **no set / GC involvement**, orthogonal to this PR's `pin_root(d_new)` in `set_symmetric_difference_storage`.
- It is the documented `cranelift int_loop` 2.0×-on-gate-2× coin-flip: cranelift-x64 measures 2.0–2.3× across many unrelated commits while dynasm-x64 and cranelift-aarch64 stay ~1.1×. Absolute times swing 3× between runs on the same image; only the ratio is stable, and it sits exactly on the gate.
- Local `check.py --backend dynasm,cranelift`: dynasm 326/326, cranelift 326/326, bit-exact.

Re-running the failed job for a green tick; no code change warranted.

— *commented by Claude*

PR #859, opened 2026-07-28
jit(aarch64): close the loop gap vs PyPy in the backend; fix NaN float compares on both dynasm arches

youknowone:
Two independent lines of work sit on this branch: the **aarch64 backend's steady-loop code quality** (which is where the loop-bench gap against PyPy actually was), a **float-comparison miscompile on both dynasm architectures**, and the earlier **#841 review follow-ups + vref-producer hardening**.

---

# 1. The loop gap vs PyPy was the guard veneer

Wall-clock A/B on the dev box is not usable for this: running the **same binary under two labels** put `nested_loop` at 2.11x and 1.96x — opposite sides of the 2x gate — and the absolute level drifts ~40% between runs minutes apart. So the work was driven by **disassembling pyre's steady loop and PyPy's steady loop for the same program and diffing them**, which is deterministic and load-immune.

For `while j < 30000: s = s + i*j; j = j + 1`: **pyre 24 instructions, PyPy 20.**

**Root cause.** `emit_bcond_to_label` emitted *every* guard as

```
b.<inv_cc> skip ; b =>label ; skip:
```

so the branch to the failure stub got the unconditional `b`'s 26-bit range. That was a blanket workaround for a single case — a trace larger than `b.cond`'s +1MB forward reach (logo's 70000-op trace), which cannot use one conditional branch at all. PyPy emits a single patched `b.cond` (it appears as `brk #0` in `jit-backend-dump` before patching).

Every guard therefore cost **two instructions instead of one**, plus an extra branch in the fetch stream — and traces are guard-dense, so it scaled with everything. It is now decided per trace in `_assemble` (`ops.len() * 1024 >= 1<<20` keeps the long form). The estimate is not taken on faith: `_assemble` asserts the emitted body plus a stub allowance fits the reach whenever the short form was used, and all 337 `pyre/bench` + `pyre/bench/synth` fixtures run clean against a release build with `-C debug-assertions=yes`.

**Three smaller parity gaps in the same sweep**, each a real divergence from `opassembler.py`:

| | before | upstream | 
|---|---|---|
| `int_mul_ovf` | `asr x14, dst, 63` + `cmp x15, x14` | `CMP_rr_shifted(ip0, res, 63)` = `cmp Xn, Xm, asr #63` |
| compare vs immediate | materialize into x17, then `cmp` reg,reg | `emit_int_comp_op` takes `CMP_ri` for a 12-bit operand |
| array base offset | `mov x16, index` + `add x16, x16, #ofs` | a single `ADD_ri(ip0, index, ofs)` |

The last two are encoded as raw words, because dynasm's `cmp`/`add Xd|SP, Xn|SP, #uimm` forms reject a **dynamic** register operand — it cannot prove the register isn't SP. Both encodings are `codebuilder.py`'s own (`CMP_ri`, `ADD_ri`) and were checked against `llvm-mc` at the range endpoints.

Also on this branch: the compare→guard condition-code fold, which is two commits. The first records an adjacent comparison's cc so the guard branches on the live NZCV instead of re-testing a materialized boolean. The second completes the **frame-register sentinel** that x86 already had — `force_allocate_reg_or_cc` hands a comparison the frame register when the next op consumes the flags, and the emit side publishes `guard_success_cc` and emits nothing. It was `#[cfg(target_arch = "x86_64")]`-gated precisely because aarch64 had no `flush_cc` (`cset x29, cc` would destroy the frame pointer). aarch64 now has one, the gate is lifted, `consider_int_is_true_j2` routes through it (`regalloc.py:469 prepare_comp_unary` allocates no destination for the folded form), and `genop_discard_cond_call` consumes a published condition the way `_emit_op_cond_call` skips its CMP when `arglocs[0] is None`.

### Result

Steady-loop instruction counts, pyre vs PyPy — pyre is now shorter on five of six:

| | pyre | pypy |
|---|---|---|
| nested_loop | **19** | 20 |
| inv | **14** | 15 |
| int_loop | **15** | 17 |
| raise_catch | **31** | 33 |
| fannkuch | **50** | 102 |
| lst (list-heavy) | 32 | **21** |

Wall clock, taken on a quiet box with an interleaved, order-alternated min-of-11 harness, and reproduced on a reversed-order repeat (the after-column landed within ±0.05x both times):

| bench | before | after | gate |
|---|---|---|---|
| int_loop | 1.02x | **0.79x** | 2.0 |
| nested_loop | 1.54x | **1.02x** | 2.0 |
| raise_catch | 1.18x | **0.91x** | 1.5 |
| spectral_norm | 1.82x | **1.48x** | 5.0 |
| fannkuch | 3.20x | **2.92x** | 5.0 |
| float_loop | 0.64x | **0.57x** | 1.5 |
| fib_loop | 1.30x | 1.29x | 3.0 |
| inline_helper | 1.08x | 1.09x | 1.5 |
| nbody | 1.90x | 1.95x | 5.0 |

`nbody` is float-heavy and flat within the noise floor; nothing here touches float codegen paths. `lst` at 32 vs 21 is the remaining per-loop gap — list indexing carries two bounds checks, a class guard *and* a storage-kind guard, plus three instructions to materialize a constant class-pointer address each iteration. Upstream also has `prepare_comp_op_float_*` / `emit_comp_op_float_*` (float compare folded into the following guard), which pyre still lacks. Both are follow-ups, not in this PR.

### Where the residual gap now lives (measured after these commits)

Worth recording because it **inverts** the reading this branch started from. With the veneer gone, wall clock tracks instruction count again:

| loop | pyre insns | pypy insns | wall vs pypy | cost per instruction |
|---|---|---|---|---|
| nested_loop | 19 | 20 | 1.02x | 1.07x |
| list read (`acc + q[i] + q[j]`) | 50 | 35 | 1.67x | 1.17x |
| list read+write (fannkuch swap) | 56 | 38 | 2.0x | 1.36x |

Per-instruction cost is 1.1–1.4x, not the ~2.3x measured before the veneer fix, so **instruction count is the dominant term** and "the gap is per-op cost, not IR" no longer holds. The 15 extra instructions in the list loops are exactly the per-iteration class guard (6), storage-kind guard (3), `length`/`items` reload (2) and bounds checks (5) that PyPy hoists into its preamble and pyre re-executes; PyPy's steady body carries none of them.

That hoisting is the short-preamble heap import, which is separately known to be dead in pyre and to miscompile bridges when revived — an optimizer epic, not something reachable from the backend. The aarch64 backend itself is now at parity for these loops: the constant-first `mov`+`cmp` and the four-instruction scaled array access that remain are byte-for-byte what PyPy emits (`prepare_int_cmp` only tests `arg1`; `load_supported_factors = (1,)` forces the explicit `int_lshift`).

Ratios below `~0.15s` of PyPy time are startup noise and must not be quoted — the same list loop reads as 4x at the fixture's own N and 1.67x once scaled.

---

# 2. Float comparisons miscompiled NaN on both dynasm architectures

Found while reviewing the above. The **live** float-comparison arm (`regalloc_perform`) emitted a bare `SETcc`/`cset` from a condition table with no unordered handling. One x86 table had been copied to aarch64, so each architecture got wrong exactly the half the other handles correctly:

| | FLOAT_LT | FLOAT_LE | FLOAT_EQ | FLOAT_NE | FLOAT_GT | FLOAT_GE |
|---|---|---|---|---|---|---|
| aarch64 — `fcmp` sets NZCV=0b0011 | ok `lo` | ok `ls` | ok | ok | **`hi` → True** | **`hs` → True** |
| x86 — `ucomisd` sets ZF=PF=CF=1 | **`setb` → True** | **`setbe` → True** | **`sete` → True** | **`setne` → False** | ok `seta` | ok `setae` |

In a compiled loop with `x = float('nan')`, 30000 iterations: aarch64 evaluated `nan > 1.0` as True 28558 times; x86 evaluated `nan < 1.0` True 27358 times, `nan == 1.0` True 26958 times, and `nan != 1.0` True only 3242 times (the interpreter prefix before the trace compiled).

- **aarch64**: `gt`/`ge` test N == V and are false when unordered, which is what `opassembler.py:310-315` uses. `hi`/`hs` test C, which `fcmp` *sets* for NaN.
- **x86**: ported `assembler.py:1322 _cmpop_float` — FLOAT_LT/LE compare in the reverse operand order so they can use the `A`/`AE` forms, and FLOAT_EQ/NE get `_if_parity_clear_zero_and_carry` (`jnp skip; cmp rbp, 0; skip:`).

Both verified against CPython over 15 operand pairs covering ±0.0, ±inf and NaN in either position; x86 verified under Rosetta on `x86_64-apple-darwin`.

**cranelift and wasm were never affected** — `FloatCC::LessThan` and `f64.lt` are IEEE-ordered by construction.

Worth flagging for future backend audits: the **dead** legacy `genop_float_cmp` in both files gets this right (it swaps for lt/le and uses `sete`+`setnp`). It has no callers. A reading-only review that lands on the dead spelling will conclude the backend is fine.

---

# 3. #841 review follow-ups and vref-producer hardening

Unchanged from the previous revision of this PR; restated in brief.

**`force_all_frames` forced nothing.** Flagged by both reviewers; the finding is real, the attribution was not. Upstream gets the force out of the *walk* — `pyframe.py` declares `f_backref = jit.vref_None`, so `getnextframe_nohidden`'s `frame.f_backref()` is a `jit_force_virtual`. Making pyre's walkers force-free was correct; what is missing is the vref *producer*, so `force_vref` is the identity. Until it exists, this consumer states the force directly. Still unobservable — a regression bench was built, did not discriminate (same count under `PYRE_NO_JIT=1`), and was deleted rather than committed; `sys.settrace` still fires **zero** events under pyre, before and after #791.

**The other seven #841 findings were refuted**, each verified against current code and re-attacked by an independent skeptic: `_warnings` `STATE_NS` is rooted via `walk_process_import_roots`; `hash("")` cannot disagree between entry points; the wasm nursery-hook disarm is reachable only from a `#[test]`; the `_warnings` stale-`category` local is always a `W_TypeObject` and heap classes allocate old-gen; specialised tuples cannot reach the `wrappeditems` path; the `FnDef` constant is a synthetic 0-arg `Call` whose `result_ty` is the return value; the string-hash memo belongs to the threading epic as one change.

**Three vref hardening changes**, each independently correct and a no-op today, each a crash the moment a producer exists: the frame-chain re-entry test resolved through `vref_referent` rather than by forcing (forcing would clear `TOKEN_TRACING_RESCALL` and make a reader report its own read as a callee escape); the displaced `topframeref` rooted through `majit_gc::shadow_stack` at the two sites that parked it raw across collectable code; and the vref given its own materializer header, since `materialize_virtual_object` was dereferencing the `JIT_VIRTUAL_REF_VTABLE` magic as a type object.

**The producer itself is deliberately not here** — `alloc_virtual_ref` is `Box::into_raw` (host heap, not GC-registered), the normal non-escaping `leave` leaves `forced == NULL` which is exactly what `eval.rs`'s `expect` aborts on, and the walker never calls the vref bracket. Those are design changes that cannot be exercised without the producer and vice versa. There is also a fork to decide first: `stop_tracking_virtualref` must record `VIRTUAL_REF_FINISH` **before** the `CALL`, but pyre's walker records the call op first and concrete-executes after, in all four dispatchers.

**Five comments corrected** across `trace.rs`, `jitcode_dispatch/mod.rs` and `mir.rs`, including one that framed an orthodox `ValueType::Int` stamp as a deviation and invited a future "fix" to `Ref` that would be the actual deviation.

---

## Verification

- `python3 pyre/check.py --backend dynasm,cranelift` re-run after **every** commit in section 1 and 2: **331 passed on both backends** each time. The single failure is `synth/ast_compile_roundtrip`, which check.py itself marks `BASEFAIL` and which fails identically on cranelift — a backend none of these changes touch.
- `cargo test -p majit-backend-dynasm` and `-p majit-metainterp` pass.
- All 337 `pyre/bench` + `pyre/bench/synth` fixtures swept against a release build with `-C debug-assertions=yes`: neither the guard-branch range assumption nor `flush_cc`'s precondition ever fires.
- Hand-encoded `CMP_ri` / `ADD_ri` words checked against `llvm-mc --disassemble --triple=aarch64` at both range endpoints.
- Output compared to CPython for float/NaN (15 pairs), `int_mul_ovf` boundary values, and every loop bench.

A note for whoever reads CI: **a perf-gate failure on this suite is not by itself evidence of a regression.** The same tree failed `dynasm nested_loop ratio 2.3x > gate 2x` and then passed minutes later with no code change; the noise floor on a shared box exceeds the gate margin. Re-run before believing it.

— *commented by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JIT guard branching and comparison handling on AArch64 after moves/spills, with safer NZCV-based guard evaluation.
  * Fixed x86 floating-point compare handling for unordered/NaN cases to prevent misclassification.
  * Improved tracing/residual-call and frame/virtual-reference behavior, including more reliable frame rooting during multi-frame “blackhole” handling.
* **Performance**
  * Reduced emitted instruction sequences for AArch64 indexed addressing and selected compare/arithmetic paths.
* **Documentation**
  * Updated comments covering funcptr constant materialization, residual-call parity behavior, and Unicode/list layout details.
* **Chores**
  * Removed an internal list helper used for in-place append GC-block querying.


youknowone:
## CI triage — `pyre/check.py (ubuntu-24.04)`

Two failures, neither attributable to this branch. Evidence for both, so this does not have to be re-derived at merge time.

### 1. `dynasm synth/exception_traceback_frame_lineno` — `GC BUG: invalid type_id ... site=copy_nursery_object`

Pre-existing and intermittent. It has a deterministic amplifier:

```
PYPY_GC_NURSERY=131072 ./target/release/pyre-dynasm \
    pyre/bench/synth/exception_traceback_frame_lineno.py
```

A/B under that amplifier, 8 runs each:

| binary | panics |
|---|---|
| this branch (`6cc8dbdd12`) | **8/8** |
| pre-change baseline (this branch's merge-base build) | **8/8** |

Identical, so the commits here neither introduce nor aggravate it. The nursery-size band is narrow — `32768` and `1048576` are both 0/8 — so a single small value is not enough to call it unreproducible.

Being worked separately on `fix-dynasm-gc-nested-listcomp` (#872); deliberately not touched here.

### 2. `cranelift int_loop  exec 0.73s > pypy 0.36s  ratio 2.0x > gate 2x`

This branch touches **no cranelift file** (`git diff --name-only <merge-base>..HEAD` has zero matches for `cranelift`), and `check.py`'s own bench table names `int_loop` as one of the "known slow-runner flakes" left deliberately at low headroom.

It is also already red on `main` itself for the same job/bench/backend — run 30378431710 at `b8a00beb9`. It is x86_64-only: pyre's absolute time on that runner is unchanged across recent main runs (0.71 / 0.80 / 0.72 / 0.79 s, all of which passed); the ratio crossed because *pypy* got faster on the runner. Local arm64 measurement on this tree, interleaved against pypy over 5 rounds, is 0.33s vs 0.30s — under 1.2x raw and below 1.0x once startup is subtracted the way check.py does.

The underlying cause is cranelift x86_64 register pressure (`preserve_frame_pointers` + `enable_pinned_reg` cost two of sixteen GPRs), not this branch's IR.

### Local verification of this tree

`python3 pyre/check.py --backend dynasm,cranelift` was re-run after every commit here: **331 passed on both backends** each time, the only failure being `synth/ast_compile_roundtrip`, which check.py marks `BASEFAIL` and which fails identically on cranelift.

— *commented by Claude*

youknowone:
Rebased onto current `origin/main` and force-pushed (`--force-with-lease`). One commit added beyond the previous push.

## Added: cherry-pick of `b301a3c21b` from #872

`jit(fbw): always record list_write_barrier on the Object in-place append arm`, recorded with `-x`.

A 337-fixture x 4-config panic census on this branch found that shrinking `PYPY_GC_NURSERY` exposes four fixtures that abort at `GC BUG: invalid type_id` — across **two** sites, not one:

| fixture | site | before | after |
|---|---|---|---|
| `nested_list_comprehension_hot` | `object_total_size` | 5/5 abort | 0/56 |
| `comprehension_object_append_hot` | `object_total_size` | 5/5 abort | 0/56 |
| `const_arg_call_resume` | `object_total_size` | 5/5 abort | 0/56 |
| `exception_traceback_frame_lineno` | `copy_nursery_object` | 5/5 abort | 0/56 |

(after = 7 nursery sizes x 8 reps each). #872's commit clears **both** sites, including `const_arg_call_resume` and the `copy_nursery_object` fixture, which its title does not name — the two sites shared one producer, a missing remembered-set entry for an appended old->young element.

The census also confirms the config CI actually runs (`PYRE_STDLIB` + `MAJIT_STRICT=1` + `MAJIT_STATS=1`, default nursery) is **0 panics / 337 fixtures** on arm64 and on x86_64 under Rosetta, both before and after.

## Verification

`python3 pyre/check.py --backend dynasm,cranelift`: **331 passed on both backends**. The sole failure is `synth/ast_compile_roundtrip`, which check.py itself marks `BASEFAIL` and which fails identically on cranelift.

## Not addressed here

`synth/attr_instance_shadows_class` is 148x slower than pypy at its shipped `N=30000` (14490 ms vs 98 ms) and scales superlinearly — doubling N roughly triples pyre's time while pypy stays flat. It is not a panic and not JIT-related (`PYRE_NO_JIT=1` is *slower*). Profile: `mapdict_root_walker_area` 31.1%, `OldGen::contains` 25.3%. `walk_mapdict_roots_area` clones the entire `INSTANCE_DICT` side table on every minor collection with a per-entry mutex re-lock, so minor GC costs O(live instances) rather than O(nursery); there is no remembered set for that table. That is mapdict + collector work, a different subsystem from this branch's dynasm backend changes, so it is left for a separate change.

— *commented by Claude*

PR #861, opened 2026-07-28
gc: drive the interpreter safepoint from the collector, and root the __main__ frame across the bootstrap

youknowone:
Four commits on the interpreter GC safepoint. The first two replace the safepoint's hand-rolled heap model and its jitframe gate with the collector's own answers; in between sits the frame-lifetime bug that removing the gate exposes, and the diagnostic that found it.

## `gc: ask the collector when to run the interpreter safepoint's major`

The safepoint kept its own model of heap growth: bytes charged by `note_alloc` at eight allocation sites, compared against a threshold re-derived after each collection from `get_total_memory_used()` times a growth delta. Both halves described a different heap from the one being collected — the tally saw only the sites that remembered to report (dicts, lists, tuples, instances and every JIT promotion grow the old generation without charging it), the charges were payload-only while the collector's total counts each `GcHeader`, and the re-derived threshold dropped the `growth_rate_max` / `max_delta` caps and the `max_heap_size` bound that `set_major_threshold_from` (incminimark.py:575-594) applies.

All of that is already ported in `majit-gc`. `threshold_reached` (incminimark.py:1288-1290) weighs `get_total_memory_used()` against the threshold set from the last major's survivors, and `do_collect_oldgen_nonmoving` already runs `major_collection_step` through `finish_incremental_cycle`, which sets the next one. Exposed as `GcAllocator::major_threshold_reached`. What remains on the pyre side is a poll interval carrying no heap semantics.

Interleaved min-of-7 against the previous policy: every workload within 1% with `PYRE_GC_INTERP` off and on, completed majors per run unchanged (14/13, 34/33, 12/12), RSS over a `s.lower()` loop flat at 154 MB either way.

## `pyframe: name the code object and pc when the value stack underflows`

`pop`'s bare depth assertion said an opcode popped from an empty stack, but not which opcode of which code object — the whole question when the underflow means the frame is no longer the one the loop started with. Replaced with a `#[cold]` reporter carrying depth/base/nlocals/ncells, pc, `last_instr`, the code object's name and file, the frame/pycode/f_backref pointers, and an opcode window.

## `interp: root the __main__ frame across the bootstrap imports`

`run_source` built the `__main__` frame and then ran the `sys` importhook, the importlib bootstrap, `seed_main_loader` and `import_site` before handing it to `eval_with_jit`. The frame is a GC-managed `FrameBox`, and the only root that reaches one is the `CURRENT_FRAME` / `f_backref` chain `walk_pyframe_roots` follows — which it joins when `eval_with_jit` installs it. A collection driven by any bootstrap step therefore reclaimed a live frame and handed its block to the next stable frame allocation.

With min_heap_size near 1 MB the first major lands inside `import_site` and the block goes to importlib's `_get_module_lock` weakref callback, which leaves `last_instr` at its `ReturnValue` and a value stack at base; `eval_with_jit` then enters that block and `pop` underflows. Confirmed under lldb: the `FrameBox` pointer returned by `new_with_context` is the frame the panic names.

Pin the frame for the span, as `FrameBox::new`, `call_function_impl_result` and `finalize_weakrefs` already do for the same reason — an RPython local holding a GC object is a shadow-stack root, a Rust local is not. `pyre-wasm`'s `run_python_impl` has the same shape over a shorter window and gets the same pin.

## `gc: drop the jitframe-empty gate from the interpreter safepoint`

The safepoint only ran a major when no compiled trace was suspended on this thread. A trace suspends by making a residual call into the interpreter, and a call site is where the backend emits a gcmap, so `enumerate_root_walker_values` already reaches a suspended frame's live refs by expanding each jitframe on the JF shadow stack through `trace_libc_jitframe`. The gate has no upstream counterpart and suppressed most of the safepoint's collections.

Removed root and branch, along with its hook layer, trampoline and fn-addr aliases. On `p_calleeloop.py` this takes the safepoint from 12 majors to 23, against a JIT-off baseline of 26.

---

Verification is running locally (repro matrix, `check.py` on both backends × default and gated, `cargo test`, perf A/B); this PR will be updated if anything comes back red.

— *opened by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved garbage-collection scheduling by using the collector’s major-threshold status directly.
  * Reduced unnecessary allocation bookkeeping during object creation.
  * Improved frame lifetime handling during Python execution to prevent premature collection.
  * Added detailed diagnostics for interpreter stack underflow errors.
  * Improved compatibility across native, WebAssembly, and JIT execution backends.
* **Refactor**
  * Simplified garbage-collection safepoint and polling behavior for more consistent collection decisions.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -398,6 +398,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_is_tracked(Some(is_tracked_via_active_runtime));
     majit_gc::set_active_collect_oldgen(Some(collect_oldgen_nonmoving_via_active_runtime));
     majit_gc::set_active_heap_stats(Some(heap_stats_via_active_runtime));
+    majit_gc::set_active_major_threshold_reached(Some(major_threshold_reached_via_active_runtime));
     majit_gc::set_active_root_hooks(
         Some(gc_add_root_via_active_runtime),
         Some(gc_remove_root_via_active_runtime),
@@ -1662,6 +1663,12 @@ fn finalizer_next_dead_via_active_runtime(fq_index: usize) -> Option<GcRef> {
 /// Report `(oldgen_total, nursery_used)` for the interpreter GC safepoint.
 fn heap_stats_via_active_runtime() -> (usize, usize) {
     with_cranelift_gc(|gc| gc.heap_byte_stats()).unwrap_or((0, 0))
+}
+
+/// Report whether the GC wants a major collection, for the interpreter GC
+/// safepoint (incminimark.py:1288-1290 `threshold_reached`).
+fn major_threshold_reached_via_active_runtime() -> bool {
+    with_cranelift_gc(|gc| gc.major_threshold_reached()).unwrap_or(false)
 }
 
 /// Host-side root-register trampoline. Bridges
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -199,6 +199,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_is_tracked(Some(dynasm_is_tracked));
     majit_gc::set_active_collect_oldgen(Some(dynasm_collect_oldgen_nonmoving));
     majit_gc::set_active_heap_stats(Some(dynasm_heap_stats));
+    majit_gc::set_active_major_threshold_reached(Some(dynasm_major_threshold_reached));
     majit_gc::set_active_root_hooks(Some(dynasm_gc_add_root), Some(dynasm_gc_remove_root));
     majit_gc::set_active_gc_owns_object(Some(dynasm_gc_owns_object));
     majit_gc::set_active_gc_is_nursery_object(Some(dynasm_gc_is_nursery_object));
@@ -627,6 +628,19 @@ fn dynasm_heap_stats() -> (usize, usize) {
         return r;
     }
     majit_gc::gc_sync::gc_op(|g| g.heap_byte_stats())
+}
+
+/// Report whether the GC wants a major collection, for the interpreter GC
+/// safepoint (incminimark.py:1288-1290 `threshold_reached`).
+fn dynasm_major_threshold_reached() -> bool {
+    if let Some(r) = DYNASM_ACTIVE_GC.with(|c| {
+        c.borrow_mut()
+            .as_deref_mut()
+            .map(|g| g.major_threshold_reached())
+    }) {
+        return r;
+    }
+    majit_gc::gc_sync::gc_op(|g| g.major_threshold_reached())
 }
 
 /// Host-side root-register trampoline. Bridges
--- a/majit/majit-backend-wasm/src/lib.rs
+++ b/majit/majit-backend-wasm/src/lib.rs
@@ -321,6 +321,7 @@ fn register_active_hooks(supports_guard_gc_type: bool) {
     majit_gc::set_active_is_tracked(Some(wasm_is_tracked));
     majit_gc::set_active_collect_oldgen(Some(wasm_collect_oldgen_nonmoving));
     majit_gc::set_active_heap_stats(Some(active_gc_heap_stats));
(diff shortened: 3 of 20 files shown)


youknowone:
## Local verification

Built on the rebased base (`b8a00beb94`), both backends, release.

**Defect-3 repro matrix** — the eight configurations that previously died with a value-stack underflow in importlib's `cb`:

| binary | result |
|---|---|
| `pyre-g9-dyn` (this branch) | **8/8 pass** |
| `pyre-g9-cl` (this branch) | **8/8 pass** |
| pre-fix control (gate removed, frame not rooted) | **6 fail / 2 pass** — the 2 passes are the two controls that never armed the safepoint |

Every control failure is the same frame: `code="cb" pc=77 last_instr=76 depth=2 base=2 back=0x0`, window `75:ReturnValue 76:PushExcInfo`. Under the previously-breaking config (`PYPY_GC_NURSERY=262144 PYPY_GC_MIN=1048576`) majors now proceed to 7 on both backends instead of dying on the first.

**check.py** — serially, both backends × default and `PYRE_GC_INTERP=1`: **331 passed, 1 failed** in all four runs. The failure is `synth/ast_compile_roundtrip`, reported as `cpython/pypy output mismatch`; it is an oracle version skew in my local environment, not a pyre defect:

- pyre output is byte-identical to CPython 3.14.2 (zero diff over 87 lines).
- The local pypy3 is 3.11.13, and `ast.NamedExpr(target=Constant(1))` validation changed between 3.11 and 3.14 (`TypeError` vs `ValueError`). check.py requires its two oracles to agree and fails the backend without running pyre at all.
- The fixture was added by #856 in `origin/main`; this branch touches no AST file.

CI's ubuntu run does not hit it (331 passed there too).

**cargo test** (`-p pyre-object -p pyre-interpreter -p pyre-jit -p majit-gc --features dynasm`): **1231 passed, 0 failed**, 21 ignored.

**Perf A/B** — `PYRE_GC_INTERP=1` vs `=0` on the same binary, interleaved arms, 9 rounds, p25 (min in parentheses):

| bench | ratio p25 | ratio min |
|---|---|---|
| fib_loop | 0.970 | 0.997 |
| int_loop | 0.963 | 0.859 |
| nested_loop | 1.011 | 1.016 |
| raise_catch_loop | 0.943 | 0.857 |
| inline_helper | 1.034 | 0.968 |
| float_loop | 0.983 | 1.224 |

No systematic cost from removing the gate. `float_loop`'s min outlier is noise — its p25 ratio is 0.983.

## On the CI `pyre/check.py (ubuntu-24.04)` failure

`FAIL cranelift int_loop exec 0.63s > pypy 0.31s ratio 2.0x > gate 2x`, 331 passed.

PR #858 — which *is* this branch's base commit `b8a00beb94` — failed the identical gate at `ratio 2.1x` and was merged. The bench sits exactly on the 2x boundary on that runner; this run is marginally under the base's ratio, so it is not a regression from these commits.

## Follow-up that was investigated and dropped

The planned next step was an `EB_GC` eval-breaker bit armed by the collector, so hot compiled loops would deopt at their back edge into the dispatch-loop safepoint. **The premise does not reproduce.** Measured with `MAJIT_LOG=1` on three probe designs — `s.lower()` (N=400k), `{"a": i, "b": i+1}` (N=1.2M), and `d = {}` (N=2M, the stable-allocation path with no nursery storage) — old-gen is flat in every configuration: 341,536 B with `PYRE_GC_INTERP=1`, 293,608 B with it off, unchanged across a 4× longer run. `PYRE_GC_INTERP=1` also *adds* majors (20 vs 15), so compiled loops demonstrably do return to the dispatch loop.

The reason is that any real Python loop allocates something in the nursery, nursery fill drives a minor, and `run_major_progress_after_minor` (incminimark.py:832, 849-860) already runs `major_collection_step` after every minor. Adding `EB_GC` would introduce a pyre-only mechanism plus a deopt-storm risk (an armed bit the safepoint declines to service re-fails every back edge) for a problem with no measurement behind it. Not implemented.

What is real is the structural gap the parity review flagged independently: `alloc_oldgen_typed` has no counterpart to `external_malloc`'s `threshold_reached` check (incminimark.py:987-994). Porting it needs the callers to be root-safe, which is exactly what the `__main__` frame bug in this PR shows pyre is not yet. Documented at the allocator in `32ef2eeb9c` rather than worked around.

— *commented by Claude*

PR #864, opened 2026-07-28
jit(fbw): resume a blackhole ContinueRunningNormally on the frame the drive wrote

youknowone:
## What this fixes

`try_adopt_single_frame_blackhole` drives a blackhole image and then decides whether to adopt the result. Every decline taken **after** `drive_single_frame_blackhole` hands an already-executed region back to a replay that runs it again.

The frame undo that guards this was recently widened to cover the resume coordinate, not just the locals. That closes the frame half completely — the only vable fields with a production emit site are `last_instr` (index 0) and `valuestackdepth` (2), plus the `locals_cells_stack_w` array; `setfield_vable_r`/`_f` have dispatch arms and no emitter. **It is still not sufficient**, because the drive also ran the region's residual calls, and those are heap effects rather than frame state.

The corpus fixture hid this. Its driven region re-adds an already-present element to a set, which is idempotent, so a replayed region leaves no trace in the output. With a list append instead, and `apply_blackhole_crn` forced to decline every time:

| driven-region effect | result |
|---|---|
| `set.add` of a present string (idempotent) | `199990000` — looks correct |
| `list.append` (non-idempotent) | `total` correct, **20005 appends for 20000 iterations** |

One extra per declined drive. So the widened undo made the failure *quieter*, not smaller — before it, the accumulator was visibly wrong too.

## The fix

Delete the decline rather than gate around it.

`apply_blackhole_crn` rebuilt the frame from the terminal register banks and validated pcdep coverage, register-bank bounds and NULL Refs — all post-drive, all fallible. Upstream does none of that:

- `bhimpl_setarrayitem_vable_r` / `bhimpl_setfield_vable_i` (`blackhole.py:1390-1490`) are **write-through**: they fetch the array out of the virtualizable and store into it. So when `bhimpl_jit_merge_point` raises (`blackhole.py:1068-1069`) the frame already holds the driven region's result.
- `handle_jitexception` (`warmspot.py:970-982`) re-enters the portal with the greens and reds the exception carries. The frame is among the reds; nothing reconstructs its contents.

`adopt_blackhole_crn` does the same thing. pyre's walked frame has two representations, so the mirror is what bridges them: the drive's vable ops address the frame in the jitcode's frame register (the live root), while the portal epilogue publishes the snapshot — the same move the escape leg already makes. `last_instr` is normalized to `resume_py_pc - 1` because the drive stops having recorded the pc it is *at*; skipping that normalization makes both loop fixtures produce no output at all.

With the arm unable to decline, the two pre-drive gates that existed only because it could — the loop-header refusal and the pcdep-trivia precondition — have nothing left to stand on and are removed, along with `code_has_loop_header` and both `pcdep_trivia_populated` bodies (`pub`, so no dead-code warning would have caught them).

## Scope

Both adopt arms take the new handoff. A CRN is raised only at the bottommost blackhole level, which `convert_and_run_from_pyjitpl` makes the portal frame, so the multi-frame chain is the same case — and that arm already folded its live root into the snapshot before the match, leaving only the coordinate to set. With no caller left, `apply_blackhole_crn` is deleted.

Five post-drive declines go: the register-image rejection in both arms, the absent terminal image, the out-of-range terminal jitcode index, and — with the arms no longer able to reject — the two pre-drive gates that existed only because they could. What survives is an empty green list, a codewriter contract violation with no upstream counterpart.

The multi-frame path has no `pyre/bench/synth` coverage (censused: 45 `s2-gate` lines, zero `s2-drive-outcome`), so its conversion rests on the argument above rather than on a measurement of that path. The corpus, test and `check.py` runs confirm nothing else regressed.

## Write barrier

`restore_resume_state_from` copies Ref slots between two frames with raw array writes and no write barrier. `apply_blackhole_crn` armed `remember_frame_locals_array` per store, so routing the handoff through that helper dropped the barrier — an old destination array taking a young reference could miss the remembered set. The barrier is armed in the helper, covering this handoff and its three pre-existing callers. One arm suffices per call: `to_vec` has already allocated by the time the copy loop runs, so nothing in it can collect.

This one was found by the Codex parity review, not by the verification below — it is a GC-timing defect, so the corpus, the test suite and `check.py` all passed with it present.

## Verification

- `pyre/bench/synth` (320 files): no output differs from the interpreter.
- Census: 15 CRN adoptions and 50 frame-terminal adoptions, **zero declines**. Only the three `getframe_root_loop_force_*` fixtures reach the CRN arm.
- `cargo test --all --features dynasm`: 7119 passed, 0 failed.
- `check.py`: dynasm 335/335, cranelift 335/335, wasm 332/332.
- CPU time on a 5000/20000/80000 trip ladder, min of 9 interleaved rounds, one binary with the old gate behind an env arm so both sides share a build: `0.060/0.090/0.200s` on both arms.

Loop-bearing bodies now drive where they previously declined before the drive, so the perf question is real rather than rhetorical; it measures neutral, and adoption fires once per trace attempt rather than per iteration, which is why the delta does not grow along the ladder.

## Note on the history

The branch introduces the loop-header gate (`0ca67ddefd`) and the pcdep precondition (`c7020f552e`), records what lifting the gate was measured to do (`3db91cc100`), and then removes both in the final commit. That is the actual order the investigation went in: the gates were the correct call while the CRN arm could still reject an image, and only became removable once it could not.

— *authored by Claude*

### DIFF
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -2280,6 +2280,12 @@ impl BlackholeInterpBuilder {
         interp.virtualizable_ptr = 0;
         interp.virtualizable_info = std::ptr::null();
         interp.virtualizable_stack_base = 0;
+        // Same footing: the layout describes the register file of the machine
+        // this run resumed, so it must not outlive it.  A stale one does not
+        // fail loudly — `StateFieldLayout::default()` is all-zero and the
+        // `scalar_slot`/`array_elem_slot` range checks are `debug_assert!`, so
+        // a release build reads whatever slot the arithmetic lands on.
+        interp.state_field_layout = StateFieldLayout::default();
         self.pool.push(interp);
     }
 
@@ -3895,6 +3901,30 @@ mod tests {
             assert!(
                 builder.acquire_interp().jitdrivers_sd.is_empty(),
                 "a pooled interp must not carry the previous run's table",
+            );
+        }
+
+        /// `state_field_layout` describes the register file of the machine one
+        /// resume belongs to, so it must not survive into the pool.  The
+        /// range checks in `scalar_slot`/`ref_scalar_slot`/`array_elem_slot`
+        /// are `debug_assert!`, so a release build given the all-zero default
+        /// silently returns `0 + field_idx` — aliasing the dispatch JitCode's
+        /// own argument registers rather than the identity slots
+        /// `ref_scalar_base` exists to keep clear of them.
+        #[test]
+        fn released_interp_does_not_leak_its_state_field_layout_to_the_next_user() {
+            let mut builder = super::build_inline_call_only_bh_builder();
+            let mut bh = builder.acquire_interp();
+            bh.state_field_layout =
+                super::StateFieldLayout::with_ref_scalars(1, vec![4], 0, 2, 8, 0);
+            assert_eq!(bh.state_field_layout.ref_scalar_slot(0), 8);
+
+            builder.release_interp(bh);
+            let reused = builder.acquire_interp();
+            assert_eq!(
+                reused.state_field_layout,
+                super::StateFieldLayout::default(),
+                "a pooled interp must not carry the previous resume's layout",
             );
         }
 
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -716,6 +716,17 @@ impl BlackholeInterpreter {
                 self.registers_r[i] = 0;
             }
         }
+        // `tmpreg_r` is a reference slot on the same footing as `registers_r`:
+        // `run()` hands both to `push_bh_regs`, and `walk_bh_regs` visits the
+        // tmpreg word as a `GcRef` root.  Its live window is one opcode — the
+        // gap between a call helper returning and the `registers_r[dst]` store
+        // — but nothing resets it, so a released interp carries the last ref
+        // return value into the pool.  Nothing roots it there, so the object
+        // can be collected; the next `run()` on that pooled interp then pushes
+        // the dead address as a root and the collector reads a header off it.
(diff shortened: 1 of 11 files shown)


youknowone:
Updated after the Codex parity review above.

**Section 1** — confirmed real. `restore_resume_state_from` copies Ref slots with raw array writes and no write barrier, and the handoff this PR introduces routes through it, replacing `apply_blackhole_crn`'s per-store `remember_frame_locals_array`. Barrier armed in the helper, which covers this call and its three pre-existing callers.

The second item in section 1 (the multi-frame fold at `trace.rs:2620`) is **pre-existing**, not introduced here — that fold is already on `main` (`f43cee9650`, #830). Same one-line fix covers it, so it was fixed rather than reclassified-and-deferred.

**Section 2** — the point stands, so the multi-frame arm was converted too rather than left documented. It already folds its live root into the snapshot before the match, so only the coordinate was left to set. `apply_blackhole_crn` now has no callers and is deleted, taking the NULL-Ref decline with it.

**Section 3.1** (multi-frame reconstruct/decline protocol unlike PyPy) — resolved by the section 2 fix.

**Section 3.2** (float registers seeded only for the liveness-selected set) — deferred. The blocker is the one the report itself names: `WalkContext` has no concrete-float mirror, so a remaining-float-bank pass has nothing to read. It is also not reachable from this patch — the CRN arm no longer reads the register image at all, so incomplete float seeding now affects only the drive's input.

**Section 4** — both won't-fix, documented. The snapshot/live copy and the `MIFrame`-from-concrete-banks construction are consequences of pyre keeping a tracing snapshot separate from the live frame, where PyPy carries the same virtualizable as a portal red argument. Both call sites already cite that.

Re-verified: corpus 320 files no output differing from the interpreter; 15 CRN and 50 frame-terminal adoptions, zero declines; `cargo test --all --features dynasm` 7119 passed 0 failed; `check.py` dynasm 335/335, cranelift 335/335, wasm 332/332.

— *commented by Claude*

PR #865, opened 2026-07-28
descr: gate struct identity on the serialized raw-set universe, and fix the two ordering defects it exposes

youknowone:
## The gate

Struct-identity work needs a gate that cannot be greened without actually
joining two spellings. The field-descr identity census is not one: it counts
only the descrs a particular run happens to mint, so deleting a call site
greens it while nothing converges.

The serialized raw-set universe is. `descr_from_set_member` looks every member
of `descrs.bin`'s `DescrSetKeys` back up in the runtime gccache, and its three
outcomes are exactly the three states of the identity question:

| outcome | meaning |
|---|---|
| `Resolved` | the analyzer's key and the runtime's publish agree |
| `Ambiguous` | container published, **not** under this member's key — the two-spellings defect by construction |
| `AbsentContainer` | nothing in this process ever named the container |

`Ambiguous` degrades the whole `EffectInfo` to `EF_RANDOM_EFFECTS`, so each one
is a residual call that stopped invalidating precisely.

**`ambiguous == 0` alone is not the gate.** A container the runtime publishes
under a *different* key is missing from both `_cache_field` and `_cache_size`
at the analyzer's key, so it reads exactly like a container this process never
traced — the defect hides in `AbsentContainer`. So the ledger reports all
three, `absent` by name, and the gate is: `ambiguous == 0` **and** no `absent`
container key is a spelling of a group the runtime does publish.

`PYRE_DESCR_SPELLING_GATE=1` prints it.

## Reading on this base

```
[descr-spelling-gate] resolved=88 absent=58 ambiguous=0
```

Identical on `int_loop`, `fannkuch`, `nbody`.

Intersecting the 58 `absent` container keys against `path_hash` of all three
spellings (simple / crate-relative / crate-qualified) of the 18 published
groups: **empty**. The same intersection against the `resolved` container keys
is non-empty — `intobject::W_IntObject`, `listobject::W_ListObject`,
`pyframe::PyFrame` — so the join is real, not vacuous, and the analyzer and
the runtime already agree on the crate-relative spelling for the containers
that carry raw-set members. The 58 absent are containers the runtime has no
group for at all: `W_TypeObject` (`mro_w`, `version_tag`), `PyCode`
(`arg_count`, `varnames`, `cellvars`), the dict/map storages, the rclass
`OBJECT_VTABLE` (`instantiate`, `subclassrange_min/max`), thread state.

**The gate is green.** The 9 field-descr identity misses on `W_ListObject`
that opened this task are therefore latent: they cost no `EffectInfo`
precision today.

## What landed

**`majit-ir`: authority in `register_keyed_size`.** The cached `SizeDescr` was
upgraded whenever the incoming frozen field list was longer. Measuring the
colliding keys shows the analyzer-side list is longer by exactly the header
words at offsets 0 and 8 — and `heaptracker.py:60-71 all_fielddescrs`, the list
`descr.py:125-126` assigns onto the cached descr, skips `typeptr` and `c__pad*`
outright while the GC header is not part of the lltype STRUCT at all. Upstream's
positional list is the header-free one, which is what the runtime groups build,
so "more fields" is not "closer to upstream". The vtable is the reliable
discriminator — only the runtime publish has one, and losing it is separately
fatal, since `new_with_vtable` skips its typeptr store when `vtable == 0` and
the fresh object reads back with a null `ob_type` over zeroed nursery memory.
Now a `vtable == 0` descr never displaces a vtable-bearing one; among descrs
that both carry one, the longer list still wins. Two tests, built at the
measured offsets.

**`pyre-jit-trace`: publish the runtime groups before resolving raw-set
members.** The 18 module-scope `PyreObjectDescrGroup` statics published lazily,
on first trace touch — after the `Once` that resolves the raw sets. Anything
not yet published answered `AbsentContainer` and was dropped from its set for
the life of the process; that is the KNOWN GAP already documented on the
variant. Forcing all 18 first restores the order upstream gets for free
(`cpu.*descrof` during codewriting, universe complete before
`compute_bitstrings`). On the current `descrs.bin` it recovers exactly one
member: `intobject::W_IntObject.intval` in `readonly_descrs_fields`. This is
also why the vtable guard has to land with it — publishing runtime-first puts
the `vtable: 0` analyzer specs on the displacing side.

**`pyre-jit-trace`: measure the `AbsentContainer` staleness gap.** The variant's
doc said a container registered after the `BhCallDescr` is materialised leaves
the frozen `EffectInfo` claiming the callee does not touch it, "no corpus
fixture reproduces it today" — an assertion, never a measurement. The ledger now
retains the dropped members and re-asks the lookup against the grown universe at
the post-`real_main` hook. Over `pyre/bench` + `pyre/bench/synth`, 337 scripts,
dynasm: 330 reach the hook and every one reports `stale_absent=0`.

## What did not land

The identity rewrite behind task #12 (`assembler.rs:3208`
`type_id: path_hash(owner)` → a canonicalized struct key). Both one-site
variants segfault identically — `EXC_BAD_ACCESS address=0x0` in
`py_str_wtf8`, all 5 JIT benches — and with the gate green there is no live
defect motivating it. It stays open as a latent-identity cleanup, now with a
gate that will show when it starts costing something.

## Verification

- `check.py --backend dynasm,cranelift`: **6 failed / 326 passed** per backend.
  Same 6 test names as the pre-change baseline at this base (`except_star`,
  `exception_group_type`, `exception_subclass_attrs`,
  `getframe_force_cancel_journal`, `slots_class_var_conflict`,
  `unpack_drain_star_raise`) — all pre-existing, `jd1 drain body must contain a
  jit_merge_point` / wrong output, unrelated to descr registration.
  One cranelift run additionally reported `synth/const_arg_call_resume  exec
  0.16s > pypy 0.01s  ratio 31.3x > gate 30x`; the same binary passes it on
  re-run and the script's own user CPU is 0.19s across 5 rounds, so that is the
  known perf-gate boundary flake.
- `cargo test -p majit-ir register_keyed_size_authority`: 2 passed.
- `cargo fmt --check`: clean.
- Codex parity review §1 verified then reclassified, §3 measured rather than
  deferred — see the adjudication comment.

### DIFF
--- a/majit/majit-ir/src/descr.rs
+++ b/majit/majit-ir/src/descr.rs
@@ -1240,7 +1240,10 @@ impl GcCache {
         //
         // The vtable is the reliable discriminator between the two: only the
         // runtime publish has one, because a build-time spec has no runtime
-        // type object to point at. Losing it is separately fatal —
+        // type object to point at. It decides in both directions, so the
+        // surviving descr is the same whichever producer registers first —
+        // the closest this can get to upstream's single owner per STRUCT.
+        // Losing the vtable is separately fatal —
         // `new_with_vtable` skips its typeptr store (`majit-gc/src/rewrite.rs`
         // gates `gen_initialize_vtable` on `vtable != 0`) over zeroed nursery
         // memory, so the fresh object reads back with a null `ob_type`.
@@ -1267,10 +1270,17 @@ impl GcCache {
                     .unwrap_or(0);
                 let existing_vtable = existing.as_size_descr().map_or(0, |sd| sd.vtable());
                 let new_vtable = descr.as_size_descr().map_or(0, |sd| sd.vtable());
-                if existing_vtable != 0 && new_vtable == 0 {
-                    false
-                } else {
-                    new_count > existing_count
+                // Both directions, so the surviving descr does not depend on
+                // which producer happened to register first. The 18
+                // module-scope groups are published before the build-time
+                // specs, but the per-exception-class groups behind
+                // `W_BASE_EXCEPTION_DESCR_CACHE` are not — they arrive after,
+                // and a one-sided rule would reject their vtable exactly the
+                // way the other order loses it.
+                match (existing_vtable != 0, new_vtable != 0) {
+                    (true, false) => false,
+                    (false, true) => true,
+                    _ => new_count > existing_count,
                 }
             }
         };
@@ -5443,6 +5453,46 @@ mod register_keyed_size_authority_tests {
                 .collect::<Vec<_>>(),
             vec![16, 24, 32],
             "the header words must not enter the positional list",
+        );
+    }
+
+    /// The mirror order. The 18 module-scope groups publish before the
+    /// build-time specs, but the per-exception-class groups do not — they
+    /// arrive after a `vtable: 0` spec is already cached, and a one-sided rule
+    /// would reject their vtable, losing it exactly the way the other order
+    /// does.
+    #[test]
+    fn a_vtable_bearing_publish_still_wins_when_it_arrives_second() {
+        let mut gc = GcCache::new();
+        let key = LLType::Struct(0x8463ec159694d229);
+        gc.register_keyed_size(
+            key.clone(),
+            size_descr_at(0x9694d229, 0, &[0, 8, 16, 24, 32]),
+        );
+        gc.register_keyed_size(key.clone(), size_descr_at(7, 0x1000, &[16, 24, 32]));
+        let cached = gc._cache_size.get(&key).unwrap().as_size_descr().unwrap();
+        assert_eq!(cached.type_id(), 7);
+        assert_ne!(cached.vtable(), 0, "the vtable must win from either side");
+    }
+
+    /// The two orders above must agree; upstream has one owner per `STRUCT`
+    /// and never arbitrates, so an order-dependent answer is a defect by
+    /// itself.
+    #[test]
+    fn the_surviving_descr_does_not_depend_on_registration_order() {
+        let runtime = || size_descr_at(7, 0x1000, &[16, 24, 32]);
+        let analyzer = || size_descr_at(0x9694d229, 0, &[0, 8, 16, 24, 32]);
+        let survivor = |first: DescrRef, second: DescrRef| {
+            let mut gc = GcCache::new();
+            let key = LLType::Struct(3);
+            gc.register_keyed_size(key.clone(), first);
+            gc.register_keyed_size(key.clone(), second);
+            let sd = gc._cache_size.get(&key).unwrap().as_size_descr().unwrap();
+            (sd.type_id(), sd.vtable(), sd.all_fielddescrs().len())
+        };
+        assert_eq!(
+            survivor(runtime(), analyzer()),
+            survivor(analyzer(), runtime())
         );
     }
 
--- a/majit/majit-ir/src/descr.rs
+++ b/majit/majit-ir/src/descr.rs
@@ -1227,20 +1227,33 @@ impl GcCache {
         // report partial layouts, so the cached owner is upgraded when the
         // incoming frozen list has more fields.
         //
-        // Field count alone is not authority. The analyzer-side producer
-        // (`majit-translate` `bh_size_spec_from_callcontrol`) reports
-        // `vtable: 0` because a build-time spec has no runtime type object to
-        // point at, and it reports the inherited header fields the runtime
-        // groups deliberately keep out of their positional list — so it
-        // outranks the runtime publish on count while carrying strictly less
-        // information. Letting it displace a vtable-bearing descr loses the
-        // only copy of that vtable, and `new_with_vtable` then skips its
-        // typeptr store (`majit-gc/src/rewrite.rs` gates
-        // `gen_initialize_vtable` on `vtable != 0`) over zeroed nursery
+        // Field count alone is not authority, and "more fields" is not
+        // "closer to upstream". Where the two producers collide today, the
+        // analyzer-side one (`majit-translate`
+        // `bh_size_spec_from_callcontrol`) is longer by exactly the header
+        // words at offsets 0 and 8 — and `heaptracker.py:60-71
+        // all_fielddescrs`, the list `descr.py:125-126` assigns onto the
+        // cached descr, skips `typeptr` and `c__pad*` outright while the GC
+        // header is not part of the lltype STRUCT at all. Upstream's
+        // positional list is therefore the header-free one, which is what the
+        // runtime groups build.
+        //
+        // The vtable is the reliable discriminator between the two: only the
+        // runtime publish has one, because a build-time spec has no runtime
+        // type object to point at. Losing it is separately fatal —
+        // `new_with_vtable` skips its typeptr store (`majit-gc/src/rewrite.rs`
+        // gates `gen_initialize_vtable` on `vtable != 0`) over zeroed nursery
         // memory, so the fresh object reads back with a null `ob_type`.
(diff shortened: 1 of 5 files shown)


youknowone:

### §1 `register_keyed_size` — verified, then **reclassified**; comment and test rewritten (`25886f59`)

The finding is that the `existing_vtable != 0 && new_vtable == 0` arm regresses
the fuller-layout upgrade, because the retained runtime groups omit inherited
header fields from the positional list.

I first checked the premise I had asserted in the original commit message —
that the two producers sit on different keys, so the arm is a no-op today.
**That premise was wrong.** Instrumenting `register_keyed_size` over
int_loop / fannkuch / nbody, the arm fires on 7+ keys per run and materially
changes which descr survives (e.g. `n=12 vtable≠0` vs `n=15 vtable=0`,
`n=3` vs `n=6`). So this is a live path, exactly as the review says.

Dumping the field lists on both sides of each collision decides it:

```
exist(vt=0x…,n=3) [off=16, off=24, off=32]
new (vt=0x0 ,n=5) [off=0, off=8, off=16, off=24, off=32]
```

The analyzer's list is longer by exactly the **header words at offsets 0 and
8**. `heaptracker.py:60-71 all_fielddescrs` — the list `descr.py:125-126`
assigns onto the cached descr — skips `typeptr` and `c__pad*` outright, and the
GC header is not part of the lltype STRUCT at all. So upstream's positional
list is the header-free one, which is what the runtime groups build. Keeping
the runtime publish moves *toward* `all_fielddescrs`, not away from it; the
count-based upgrade was the deviation.

Behaviour is unchanged. What changed is the justification: the comment now
cites `heaptracker.py:60-71` and `descr.py:105-127` (`get_size_descr` takes the
vtable as a *parameter* and mints once per STRUCT, so upstream has no
arbitration rule for two producers at all), and the test builds the two sides
at the measured offsets — runtime `[16, 24, 32]` vs analyzer
`[0, 8, 16, 24, 32]` — asserting the header words stay out of the surviving
list.

One case the comment now records rather than papers over: some colliding specs
disagree on offsets *and* field sizes (`off=40 size=8 Int` vs
`off=40 size=1 Int`) rather than one being a subset. There neither side is
right — the shared key is itself the defect, i.e. the struct-identity spelling
collision tracked in the linked task.

### §3 `AbsentContainer` drops ×3 — **measured instead of deferred** (`a8561cde`)

All three findings are the same mechanism, and the variant's own doc already
stated the consequence — a container registered *after* the `BhCallDescr` is
materialised leaves the frozen `EffectInfo` claiming the callee does not touch
it — with the note "no corpus fixture reproduces it today". That note was an
assertion, never a measurement, so I made it one.

The ledger now retains the dropped members and re-asks the lookup against the
grown universe at the post-`real_main` hook. Sweeping `pyre/bench` +
`pyre/bench/synth`, **337 scripts, dynasm: 330 reach the hook and every one
reports `stale_absent=0`**; the 267 that reach rehydrate all report the same
`resolved=88 absent=58 ambiguous=0`.

So the gap is real in principle and currently unreached in practice, and there
is now an instrument that says which it is on any given corpus rather than a
comment claiming it. The conservative repair (degrade every such EI to
`EF_RANDOM_EFFECTS`) stays rejected on the previously measured cost — 211 drops
across ~40 containers on one synth script would turn most residual calls into
whole-heap barriers.

This PR already narrowed the struct half by publishing the 18 module-scope
groups eagerly. What stays lazy is named in the doc: the per-exception-class
groups behind `W_BASE_EXCEPTION_DESCR_CACHE`, and `ARRAY_DESCR_REGISTRY`, which
is minted on demand from call-site `array_type_id` strings and has no static
table to publish from — that is why the array/interiorfield halves are not
closed the same way.

### §4 — accepted as stated, no action.

### Verification after these two commits

`check.py --backend dynasm,cranelift`: **6 failed / 326 passed per backend**,
same 6 names as the pre-change baseline. One cranelift run additionally
reported `synth/const_arg_call_resume  exec 0.16s > pypy 0.01s  ratio 31.3x >
gate 30x`; the same binary passes it on re-run and the script's own user CPU is
0.19s across 5 rounds, so that is the known perf-gate boundary flake, not a
regression.

— *commented by Claude*

youknowone:
## Codex parity review, round 2 — §1 and §2 are now `None.`

### §3 order-dependence — **fixed now** (`ae13aef7`)

> descriptor authority is still order-dependent in the opposite direction: an
> existing vtable-less descriptor with a longer field list is retained when a
> shorter vtable-bearing runtime descriptor arrives.

Correct, and it is the same fatal consequence as the case the rule already
covered — the vtable is lost, `new_with_vtable` skips its typeptr store, the
object reads back with a null `ob_type`. My reasoning for leaving the rule
one-sided was that this PR publishes the runtime groups first, so the
vtable-bearing descr is always the incumbent. That holds for the 18
module-scope groups and **not** for the per-exception-class groups behind
`W_BASE_EXCEPTION_DESCR_CACHE`, which are published on first touch and arrive
after the build-time specs. The one-sided rule would reject their vtable.

The rule now decides on the vtable in both directions, which makes the outcome
order-independent — as close as two producers get to `descr.py:105-127`'s
single cache-or-mint owner per `STRUCT`. Two more tests: the mirror order, and
that the two orders agree.

Measured: sweeping `pyre/bench` plus `pyre/bench/synth`, 337 scripts, dynasm,
the new arm fires **0 times**, so this closes the order dependence without
moving anything today. `check.py --backend dynasm,cranelift` stays at 6 failed
/ 326 passed per backend, the same 6 names as baseline.

### §3 `AbsentContainer` drops — **deferred, blocker named**

Now measured rather than asserted (`stale_absent=0` on 330 of 337 scripts, see
the previous comment). The struct half is closed by the eager publication in
this PR. The array/interiorfield half is not closable the same way:
`ARRAY_DESCR_REGISTRY` entries are minted on demand from call-site
`array_type_id` strings and there is no static table to publish from, so the
convergence path remains re-resolving from the retained `ei.descr_set_keys` as
the universe grows — which is what the variant's doc already names, and which
is blocked on `compute_bitstrings`' per-`jitcode_for` cost. Tracked, not worked
here.

### §4 — accepted as stated, no action.

— *commented by Claude*

youknowone:
## CI: the one ubuntu failure is inherited from the base, not from this PR

`pyre/check.py (ubuntu-24.04)` reports `dynasm 1 failed, 331 passed`:

```
FAIL dynasm synth/exception_traceback_frame_lineno
  rust panic: majit/majit-gc/src/collector.rs:1699
  GC BUG: invalid type_id=3546422560 at obj_addr=0x7f036bbfdf48
          (header_addr=0x7f036bbfdf40, nursery_start=0x7f036b7fe010,
           site=copy_nursery_object)
```

`main` fails the same test with the same panic at the same line — run
[30409036844](https://github.com/youknowone/pyre/actions/runs/30409036844)
(`31a4f2037a`), `dynasm 1 failed, 331 passed` plus a cranelift `int_loop`
`ratio 2.0x > gate 2x` perf flake this branch does not have.

It is also a recent arrival on `main`, not longstanding. At `b8a00beb94` — the
base this PR was originally cut from — run
[30378431710](https://github.com/youknowone/pyre/actions/runs/30378431710) shows
`synth/exception_traceback_frame_lineno` **passing** (`0.63s 12.8x`), with the
cranelift `int_loop` perf flake as the only failure. The branch was rebased
externally mid-review onto a base that now carries `#866`, `#861` and `#851`,
and the crash came with it.

So on Linux this branch is strictly no worse than `main`: same single
pre-existing failure, and cranelift is green here where `main` is not. Not
reproducible on macOS (0 GC BUG over 3 runs of the script), consistent with the
other Linux-only entries in this crash family.

Out of scope for this PR — filed separately with the bisect evidence above.

— *commented by Claude*

PR #888, opened 2026-07-29
jit: reload bridge virtualizable fields from the live frame

youknowone:
## Summary

- reload bridge-loop virtualizable fields from the per-trace rebuilt frame
- avoid using the stale MetaInterp portal-frame anchor for an inlined callee
- fix `str_search_index_bounds` input arity panic (`16 != 26`) on dynasm, cranelift, and wasm

## Validation

- `cargo fmt --all -- --check`
- `cargo check --features dynasm`
- `cargo test --features dynasm`
- `python3 pyre/check.py --backend dynasm --no-synthetic target/release/pyre-dynasm` (15/15)
- dynasm `pyre/bench/synth/str_search_index_bounds.py`
- cranelift `pyre/bench/synth/str_search_index_bounds.py`

### DIFF
--- a/majit/majit-metainterp/src/pyjitpl.rs
+++ b/majit/majit-metainterp/src/pyjitpl.rs
@@ -5231,13 +5231,22 @@ impl<M: Clone> MetaInterp<M> {
         }
         // Bridge traces start from rebuilt resume state, not a fresh portal
         // entry, so `initial_inputarg_consts` is not seeded with the
-        // virtualizable inputarg's ConstPtr. The live virtualizable pointer
-        // cached by `set_vable_ptr` during JitState setup is the same heap
-        // object (`orig_inpargs[idx].getref_base()`), so fall back to it.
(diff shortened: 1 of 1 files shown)


youknowone:
Closing as redundant. This branch's single commit (`e9f7be0551`, reorder `orig_vable_ptr_from_trace_ctx` to read `ctx.virtualizable_heap_ptr()` before the ambient `self.vable_ptr`) is byte-identical in code to what landed independently on main via #891 (`fdca72a7fe`, "read the loop's virtualizable out of the merge point it is cut from"). #891 is a strict superset: it adds `orig_vable_ptr_for_cut`, which reads `mp.vable_ptr` from the merge point the loop is cut from, and covers the same expanded-inputargs mismatch (26 boxes vs 18-length locals) this branch targeted. Rebasing onto current main drops this commit as empty.

The two red CI checks were unrelated to this change: the ubuntu `int_loop` cranelift gate was a stale calibration already relaxed on main by #898, and the macos `load_super_attr` dynasm crash did not reproduce — main at the identical tree is green.

— *commented by Claude*

PR #894, opened 2026-07-30
majit: reserve the header word in front of off-GC jitframes

youknowone:
## What

The emitted `COND_CALL_GC_WB` fast path loads the flag byte at `jit_wb_if_flag_byteofs`. `WriteBarrierDescr::extract_flag_byte` measures that offset **from the object pointer**, so it is negative — the flags are the high half of `tid_and_flags`, at `obj-4`.

Frames built outside the GC were handed to compiled code as bare allocation bases, which puts that load in front of the allocation. Usually it reads unrelated bytes and enters the barrier helper for nothing whenever they happen to carry `TRACK_YOUNG_PTRS`; when the block lands at the start of a mapped region, the load faults.

`majit_backend::jitframe::alloc_libc_jitframe` now allocates `GcHeader::SIZE + size` and returns the payload with the header word zeroed — the same thing a nursery-allocated frame presents to that load — and `free_libc_jitframe` releases the matching base.

## Call sites

Six, not the nine `libc::calloc` sites in `runner.rs`. `dynasm_nursery_slowpath`, `..._varsize`, `dynasm_raw_varsize_alloc_typed_and_set_len` and `dynasm_raw_fixedsize_alloc_typed` already allocate `+ GcHeader::SIZE` and return `raw + gc_hdr`; `..._headerless` is headerless by contract. The ones handing out a *jitframe* were the gap:

| site | note |
| --- | --- |
| `DynasmBackend::execute_token` | **production** — every JIT entry frame |
| `DynasmBackend::execute_token_ints_raw` | **production** |
| `dynasm_realloc_frame` | frame-realloc slowpath |
| `dynasm_nursery_slowpath_jitframe` fallback | no-GC path |
| `cranelift_realloc_frame` fallback | host-test path |
| cranelift `run_compiled_code_inner` `vec![0i64; …]` | host-test path |

`free_jitframe_chain` moves to the paired free: the frame pointer is no longer the block base.

Alignment is unaffected — `collector.rs` rounds allocations to 8 (`(total_size + 7) & !7`), so a GC frame's payload is 8-aligned too; `base + 8` matches it rather than degrading it.

## Verification

- 800-seed 8-way `test_strtod` load sweep: **0 hits**. Prior baseline on the same script was ~3 hits per 400 seeds (all `rc=138`, SIGBUS), so `P(0 hits | unfixed) ≈ e⁻⁶`.
- Three macOS `.ips` crash reports for those hits agreed exactly: `fault == fp - 4`, `fp` at the start of a `MALLOC_SMALL` region, `sp` in an unrelated range (not a stack overflow), `pc` outside every `usedImage` (JIT-generated code), `KERN_PROTECTION_FAILURE`.
- `majit-backend` 15, `majit-gc` 213, `majit-backend-dynasm` 59 tests green.
- New regression test `jitframe::tests::libc_jitframe_reserves_the_negative_flag_byte` pins that the flag byte is in-bounds and clear.

The hazard was already half-documented at `collector.rs:3254-3270`, and `write_barrier_ignores_unmanaged_jitframe_with_flag_byte_set` covered the case where the out-of-bounds read *succeeds* and returns garbage. Nothing covered it faulting; that test even reserved a leading word by hand, which production did not. Both comments are updated to match.

## Second commit

`cpython_tests/run.py` `classify()` discarded stderr entirely for `rc < 0 or rc > 128`, so a signal death recorded no evidence at all while the FAIL path kept the last non-empty line. It now shares that line and decodes the signal: `signal/abort SIGBUS rc=-10 <last stderr line>`.

Seeding was deliberately not added: this failure is load-driven (a failing seed replayed alone always passes), and every way to seed a script-mode child perturbs what the suite sees — `sitecustomize` needs the temp cwd on `PYTHONPATH` (`test_site`, `test_cmd_line`), and a driver wrapper changes `__main__` identity, which is exactly why `DOTTED_IDENTITY_MODULES` exists. `--mode regrtest` is the non-perturbing route if seeds are ever wanted.

— *authored by Claude*

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -3779,10 +3779,11 @@ pub unsafe extern "C" fn cranelift_realloc_frame(old_jf: *mut i64, new_depth: us
             gcref.0 as *mut i64
         }
         None => {
-            let layout = std::alloc::Layout::from_size_align(new_bytes, 8)
-                .expect("cranelift_realloc_frame: invalid layout");
-            let p = unsafe { std::alloc::alloc_zeroed(layout) as *mut i64 };
-            assert!(!p.is_null(), "cranelift_realloc_frame: alloc_zeroed failed");
+            // Reserves the header word the emitted write-barrier fast path
+            // reads at a negative offset, so this frame answers that read the
+            // way the nursery-allocated one above does.
+            let p = majit_backend::jitframe::alloc_off_gc_jitframe(new_bytes) as *mut i64;
+            assert!(!p.is_null(), "cranelift_realloc_frame: alloc failed");
             // Host-test fallback only — register with the libc-jitframe
             // tracer so any interior Ref scan still works.
             majit_gc::shadow_stack::register_libc_jitframe(p as usize);
@@ -7260,8 +7261,14 @@ fn run_compiled_code_inner(
         }
         (gcref, None)
     } else {
-        let mut buf = vec![0i64; jf_total];
-        let gcref = GcRef(buf.as_mut_ptr() as usize);
+        // One leading word stands in for the GcHeader the nursery branch
+        // above puts in front of the frame: the emitted write-barrier fast
+        // path loads the flag byte at a negative offset from the frame
+        // pointer, so without it that load reads outside the buffer.
+        const HEADER_WORDS: usize = majit_gc::header::GcHeader::SIZE / 8;
+        const _: () = assert!(HEADER_WORDS * 8 == majit_gc::header::GcHeader::SIZE);
+        let mut buf = vec![0i64; HEADER_WORDS + jf_total];
+        let gcref = GcRef(unsafe { buf.as_mut_ptr().add(HEADER_WORDS) } as usize);
         // jitframe.py:84 parity — `jf_frame.length` is the count of `Signed`
         // payload slots after the length word.  The GC-alloc branch above
         // already does this at line 6282; mirror it here so frame-size
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -862,7 +862,7 @@ pub extern "C" fn dynasm_nursery_slowpath_jitframe(frame_size: u64) -> u64 {
             gc.alloc_nursery(frame_size as usize).0 as u64
         }
     })
-    .unwrap_or_else(|| unsafe { libc::calloc(1, frame_size as usize) as u64 })
+    .unwrap_or_else(|| majit_backend::jitframe::alloc_off_gc_jitframe(frame_size as usize) as u64)
 }
 
 /// `_build_malloc_slowpath(kind='var')` parity: varsize nursery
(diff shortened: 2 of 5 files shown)


youknowone:
Force-pushed: the first revision used `libc::calloc`/`libc::free`, which do not exist on `wasm32-unknown-unknown` — and `majit-backend` is compiled for that target during LLBC extraction, so all three `prepare Charon/LLBC` jobs went red and everything downstream skipped.

Now on `std::alloc::alloc_zeroed`/`dealloc`. That needs the layout back at free time, so the block carries its total size in a word ahead of the reserved header:

```
[ total size ][ header word ][ JitFrame … ]
  base          base+8         base+16 == frame pointer
```

The pair is renamed `alloc_off_gc_jitframe` / `free_off_gc_jitframe`, and the `libc` dependency added to `majit-backend` is reverted.

Re-verified locally:
- `cargo check -p majit-backend --target wasm32-unknown-unknown` — clean
- the exact CI surface, `cargo check -p pyre-interpreter --target wasm32-unknown-unknown --features dynasm` with the repo's `--cfg getrandom_backend="custom"` — clean
- `majit-backend` 15 and `majit-backend-dynasm` 59 tests, `cargo fmt --check`
- 400-seed 8-way `test_strtod` load sweep against a release build of this revision: **0 hits** (the allocator changed, so the earlier 800-seed sweep was re-run rather than assumed)

— *commented by Claude*

youknowone:
## CI status

All green except `pyre/check.py` on ubuntu and macOS, and **that failure is pre-existing on `main`**, not from this PR.

**Green (17):** `cargo test` ubuntu/macOS/windows, **CPython suite (gate)**, `pyre/check.py (windows-latest)`, `prepare Charon/LLBC` ×3, `wasm build` ×2, sandbox build + e2e, `cargo fmt --check`, pre-commit, CodeRabbit, Codex parity review.

**Red (2):** `pyre/check.py` ubuntu + macOS, one benchmark each — `synth/const_arg_call_resume`, `342 passed, 1 failed`.

### Why it is not this PR

`main` at `d873f1396e` (latest CI run 30517880085) fails the same benchmark with the same shape:

```
FAIL cranelift synth/const_arg_call_resume  exec 0.16s > pypy 0.01s  ratio 32.7x > gate 30x
FAILED: cranelift 1 failed, 342 passed
```

The `pypy 0.01s` denominator is at `EXEC_TIME_FLOOR_S`, so the ratio swings hard by runner. Measured values on `main` are 41.2x dynasm / 50.6x cranelift (ubuntu) and 45.7x cranelift (macOS); this PR reports 36.9x / 46.5x / 35.0x — **inside `main`'s range, and lower on every comparable reading**.

Mechanically it is also not a plausible target. The bench is `while k < 400` around three 1000-iteration calls, so JIT entry frames number in the hundreds, and this PR adds 16 bytes to an allocation that already happened per entry (`libc::calloc` → Rust's `System::alloc_zeroed`, the same underlying allocator). On the cranelift side the changed paths are not even reached under `check.py`: a configured GC takes the `use_gc_alloc` branch, leaving the Vec-backed entry frame and the realloc fallback — both host-test-only — inert.

I have not run a dedicated A/B for this benchmark: it fails on `main`, and with the denominator collapsed a single-run comparison would not be evidence either way. Happy to run the size ladder if you want a number.

— *commented by Claude*

PR #900, opened 2026-07-30
importing: demote module-dict functions to non-binding BuiltinFunction

youknowone:
## Summary

Ports PyPy's `MixedModule._load_lazily` demotion (`pypy/interpreter/mixedmodule.py:112-139`)
to pyre: a function directly in a mixed-module is a non-descriptor
`BuiltinFunction` (no `__get__`), so storing it on a user class and reading it
through an instance does not synthesize a bound method.

New `demote_module_function_to_builtin` retags a globals-less `FUNCTION_TYPE`
object in place to `BUILTIN_FUNCTION_TYPE` — only `ob_type` (the binding
dispatch tag) and `w_class` (the Python-visible `type()`) change; code, name,
and every other field are preserved and object identity is kept. Both PyTypes
share the `Function` layout and GC type id, so the retag is sound and
allocation-free (`get_instantiate` is a cached atomic load).

Called at the two mixed-module namespace-load sites:
- `new_builtin_module_dict` (the `builtins` module)
- `load_builtin_module` (every other builtin module)

In `load_builtin_module` the demoted functions now also receive
`builtin_function_set_module_obj` in the second key loop (they are
`BUILTIN_FUNCTION_TYPE` after demotion), matching PyPy's
`bltin.w_moduleobj = self`.

## Parity notes (deliberate structural adaptations)

- **In-place retag vs. distinct wrapper**: PyPy creates a separate
  `BuiltinFunction(func)` cached as `func._builtinversion_`; pyre retags the
  single object to preserve identity/GC-tid. Callers pass module dict values
  only, so no external reference to a pre-demotion function exists.
- **globals-null guard**: pyre walks the whole completed module dict, so it
  excludes app-level functions (which carry globals) from demotion; PyPy's
  loader population is interp-level (globals-less) only, so behavior matches on
  the real population.
- `name`/`qualname` are preserved from the source function rather than set to
  the dict key, which diverges from PyPy only for aliased exports.

## Verification

- `cargo check -p pyre-interpreter` — clean
- `pyre/check.py --backend dynasm` — ALL PASSED: dynasm 343/343
- Codex parity review (scoped to this change): section 1 = none; section 2
  findings are all deliberate, already-documented structural adaptations.

— *commented by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of built-in functions retrieved from built-in modules so they remain non-descriptors when attached to user-defined classes (preventing unexpected instance binding).
  * Preserved standard binding behavior for user-defined methods.
  * Ensured consistent behavior for common built-ins like `len`, `math.sqrt`, and `sys.exc_info` across class and instance attribute access.

* **Tests**
  * Added a benchmark/test script that validates non-binding behavior for multiple built-in entry points and confirms related stdlib behaviors remain intact.

### DIFF
--- a/pyre/bench/synth/module_function_not_descriptor.py
+++ b/pyre/bench/synth/module_function_not_descriptor.py
@@ -1,0 +1,69 @@
+# A function directly in a builtin module is a non-descriptor
+# `builtin_function_or_method` (no `__get__`), so storing it on a user class
+# and reading it through an instance does not synthesize a bound method and
+# does not inject `self`.  Mirrors mixedmodule.py:_load_lazily; guards the
+# demote_module_function_to_builtin pass at both call sites:
+#   - builtins.rs new_builtin_module_dict   (the `builtins` module: len)
+#   - importing.rs load_builtin_module      (every other builtin module:
+#       sys.exc_info from a non-macro register_module; math.sqrt; time.sleep)
+# A plain user `def` on a class is the control: it MUST still bind.
+#
+# A function that carries an instance attribute (populated `w_func_dict`) must
+# NOT be demoted: `builtin_function_or_method` has no instance `__dict__`, so
+# the retag would hide the attribute.  `itertools.chain` is a function-shaped
+# constructor whose `from_iterable` alternate constructor is attached as such
+# an attribute; demoting it dropped `chain.from_iterable`, which broke
+# `import unittest.mock` -> asyncio -> dataclasses.  It stays binding here.
+# Output verified against CPython 3.14.
+import sys
+import math
+import time
+import itertools
+
+
+def check_non_binding(f, expected_name):
+    assert type(f).__name__ == "builtin_function_or_method", type(f).__name__
+    assert not hasattr(type(f), "__get__"), expected_name
+    assert f.__name__ == expected_name, f.__name__
+
+    class C:
+        m = f
+
+    inst = C()
+    assert C.m is f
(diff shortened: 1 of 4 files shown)


youknowone:
The first push's CI was red only on **CPython suite (gate)** — 4 tests (`test__colorize`, `test_apple`, `test_tabnanny`, `test_xml_dom_xmlbuilder`) regressed PASS → IMPORTERROR.

**Root cause:** all four import `unittest.mock` → `asyncio` → `dataclasses`, which calls `itertools.chain.from_iterable`. In pyre `itertools.chain` is a function-shaped constructor with `from_iterable` attached as an attribute in its `w_func_dict`. `demote_module_function_to_builtin` retagged it to `BUILTIN_FUNCTION_TYPE`, which exposes no instance `__dict__`, so `chain.from_iterable` became unreachable.

**Fix** (`f6bebd79e69`): guard the demote on an empty `w_func_dict` — a module function carrying instance attributes is left `FUNCTION_TYPE`. Among module-level functions only `itertools.chain` carries such an attribute, so the change is surgical. Added `pyre/bench/synth/module_function_not_descriptor.py` asserting non-binding at both call sites and that `chain.from_iterable` survives.

Local `pyre/cpython_tests/run.py --backend dynasm` now: **PASS 42 / IMPORTERROR 0 / no regressions**; the new synth test passes on pyre-dynasm and CPython 3.14.5.

— *commented by Claude*

Issue #80, opened 2026-05-20
Retire backend `fail_descr_registry` in favor of JitCellToken-owned descr lists


youknowone:
## Summary

Both backends maintain a process-global `Arc<Mutex<HashMap<usize, DescrRef>>> fail_descr_registry` that maps raw `jf_descr` addresses back to live `DescrRef` Arcs. This is a pyre-introduced indirection with no PyPy counterpart — RPython instead ties descr liveness to specific `CompiledLoopToken` instances via `asmmemmgr_gcreftracers`, and recovers the descr through `AbstractDescr.show(cpu, descr_gcref)` (a cast, not a lookup).

This is a fallout of #36 (descr-identity split) that #36's body did not explicitly call out, but that the same orthodoxy principle covers: pyre should match PyPy's mechanism, not document a new mechanism as a "Rust adaptation."

## Current pyre state

- `majit-backend-dynasm/src/runner.rs:677-678` — `fail_descr_registry: Arc<Mutex<HashMap<usize, DescrRef>>>`.
- `majit-backend-cranelift/src/compiler.rs:6947` — same shape.
- Eviction is per-process scan keyed on `descr_owning_jct` (`runner.rs:2269+`), required because the registry's lifetime is independent of any single token.
- `Arc::as_ptr(descr)` is baked into the JIT-emitted code as a `usize`; the trampoline returns that `usize`, and `fail_descr_arc_from_addr` does the reverse lookup.
- Companion HashMap on dynasm: `bridge_addr_by_descr` (`runner.rs:689`) — same shape, same lifetime issue.

The registry also violates the [No HashMap, ever](https://github.com/youknowone/pyre/issues) policy on pyre/majit.

## PyPy mechanism (line-by-line target)

- `rpython/jit/metainterp/history.py:109-114` — `AbstractDescr.hide(cpu)` / `show(cpu, gcref)`: pure casts between the descr instance and a `GCREF`.
- `rpython/jit/backend/llsupport/llmodel.py:412-420` — `get_latest_descr(deadframe)` reads `deadframe.jf_descr` and calls `AbstractDescr.show(cpu, descr)`. No side table.
[...]

PR #902, opened 2026-07-30
jit: root the forced-virtual caches, store them inside handle_async_forcing, and zero PyFrame.vable_token on JIT-allocated frames

youknowone:
Follow-up to #899, acting on its review. Two findings on `MetaInterp::forced_virtuals` — pyre's stand-in for the `AllVirtuals` upstream hides in the deadframe's `jf_savedata` word — plus the unit test coderabbit asked for.

Sections 4-6 were added later: chasing this PR's failing `pyre/check.py (macos-latest)` job led to a GC-corrupting descr-census gap that was already red on `main`, plus one adjacent hardening and the `main` build fix the branch needs to be gated at all.

## 1. The cache was in no root set

The ptr half holds what `force_all_virtuals` (`resume.py:969-981`) materialized, and it lives from the force until the following `GUARD_NOT_FORCED` consumes it. Nothing rooted those slots.

The exposure is *reclamation*, not relocation: virtuals are allocated non-moving (`dynasm_alloc_oldgen_typed`), so there is nothing to forward — but `OldGen::sweep_arenas_step` frees any old-generation object that lacks `flags::VISITED`, and `finish_alloc_in_oldgen` sets that flag only when a marking cycle is already in progress. A virtual named *only* by a resume frame's ref registers is written nowhere else at force time, so this `Vec` is its only referent. Upstream is immune because `jf_savedata` is traced as a real GCREF field (`majit-backend/src/jitframe.rs:354` — the previous citation, `:278`, is `JitFrame::init`, and is corrected here).

Fix: enroll it as the fifth `register_mutator_extra_area` member, alongside `rd_consts` / `partial_trace` / `active_trace` / `compile_snapshot`. Only the ptr half is walked — the int half is `virtuals_int_cache`, unboxed integer field values, and handing those to the visitor would test integers as heap addresses. `prepare_resume_heap_with_roots` roots the same one half for the same reason.

`push_resume_ref_roots` cannot serve here: it snapshots `(ptr, len)` and releases by depth truncation, and this window spans a return to compiled code — every intervening `Drop` would truncate past the force-time push. `residual_call.rs:589-594` already documents that and solves it with a permanent extra area.

Rooting alone would leak, so an ephemeron pruner is registered too, modelled on `mapdict::prune_dead_owner_entries`: an entry the guard never consumes is dropped when its owner frame is swept, which is the lifetime `jf_savedata` gets for free by living on the deadframe. Without it an unconsumed entry pins its objects for the process lifetime and leaves a stale `PyFrame` key that a later frame at the same address could fish.

The pruner is registered per mutator, not through the global `register_ephemeron_pruner`. That registration hands the classifier no way to name a thread, so it would have forced the pyre side to read `JIT_DRIVER` from caller TLS and see only the collecting thread's driver — contradicting the contract stated on `MutatorExtraWalkFn` ("must derive every thread-specific address from `data`, never from caller TLS") and leaving another mutator's dead-owner entries pinned. So `register_mutator_pruner` / `prune_all_mutator_areas` / `prune_my_mutator_areas` join the extra-area equivalents in the same `MutatorEntry`, and the collector picks between all-mutator and own-mutator on `gc_sync::mutators_quiesced()` — the same predicate `do_collect_nursery` and `enumerate_root_walker_values` use, so a collection's prune reach always equals its own root-walk reach. That branch is load-bearing: the unconditional form tripped the quiescence assertion, because the pre-sweep point does not own STW. The mapdict tables stay on the global registration; they are a process-global `Mutex` map, reachable from any thread.

The entry is keyed by a bare address, which is sound only because the virtualizable is move-stable: it comes from `FrameBox::new` → `try_gc_alloc_stable_raw`, whose registered contract is "stable across minor and major collections (MiniMark mark-sweep does not move old-gen objects)", never from the frames a trace builds virtually (`emit_new_pyframe_inline_with_params` → `NewWithVtable`, which can land in the nursery). `save_forced_virtuals` now asserts that, so the invariant is checked rather than argued.

## 2. `force_pyframe_vref` dropped the cache

`compile.py:996-1000` calls `set_savedata_ref` **inside** `handle_async_forcing`. #899 did it at the `force_pyframe` hook instead, so `force_pyframe_vref` — the materializing arm of `virtualref.py:134 force_virtual_if_necessary`, which calls `force_virtualizable_token` as a statement — discarded what the force produced, and its `GUARD_NOT_FORCED` rebuilt the virtuals from `rd_virtuals`.

Fix: move the store inside, as upstream has it. `force_from_resumedata` now also returns the virtualizable its vable section named (`resume.py:1404`), and `handle_async_forcing` keys the store on that — so the key comes from the guard's own resume data rather than from a caller-side guess, and both force entry points are covered by construction. `force_virtualizable_token` returns nothing, like `force_now`.

## 3. Tests and a diagnostic

Two tests for the `all_virtuals = Some(..)` resume:

- `consume_vref_and_vable` jumps the vable and vref sections (`resume.py:1433-1435`) — asserted by decoding the *same* resume data twice and checking that the ordinary path surfaces the virtualizable while the `GUARD_NOT_FORCED` path surfaces none and the frame section behind them still decodes (which is what proves the jump lengths line up).
- `_prepare_virtuals` zeroes a preloaded cache (`resume.py:990-991`), which is why `blackhole_from_resumedata` must pass `rd_virtuals`/`rd_guard_pendingfields` as `None` on that path.

Plus a `[jit][take_forced_virtuals] owner=.. hit/miss` counterpart to the existing `handle_async_forcing` line. Only a `GUARD_NOT_FORCED` reaches it (`is_guard_forced()` gates the callers), so hit/miss is the force→resume handoff itself.

## Two review findings adjudicated as false positives

- **`jitdriver.rs:3766/6209` pass `all_virtuals = None`.** Those are the majit state-field macro JIT's resume paths. That JIT has no virtualizable force — its virtualizable is a host-stack `&state`, `force_virtualizable_token` is never called on it, and `handle_async_forcing` is unreachable there — so there is never a cache to hand over. `None` is correct, not a dropped edge.
- **The miss fallback should resume with `ragnf = 2` and an empty cache.** It resumes with `None` instead, i.e. it rebuilds. That is strictly the safer of the two: `ragnf = 2` with an empty cache would make any `TAGVIRTUAL` reference index an empty `virtuals_cache`, whereas rebuilding produces a correct (if redundant) object set. This mirrors `compile.py:959-960`, which also tolerates an absent cache.

## 4. `PyFrame.vable_token` was never zero-initialized on a JIT-allocated frame

Found while chasing the `synth/mutate_then_raise_caught` abort that was already
red on `main` (`pyre/check.py` cranelift 344/345, and the `macos-latest` job on
this PR): `GC BUG: invalid type_id=… site=minor_custom_trace_target`, reached
from `gc_alloc_nursery_shim`.

The nursery is deliberately not zero-filled — `Nursery::reset` mirrors
`incminimark.py:211 malloc_zero_filled = False` — and the backend's
`CallMallocNursery` fast path clears only the 8-byte header word. So the whole
zero-initialization of a JIT-inline-allocated object is the pending stores that
`rewrite.py:498-504 clear_gc_fields` derives from `descr.gc_fielddescrs()`.

`rvirtualizable.py:29` appends `('vable_token', llmemory.GCREF)` to the
virtualizable's own fields, so upstream's `gc_fielddescrs` names it and the slot
is cleared on every `new`. pyre types the field `usize` and
`PYFRAME_DESCR_GROUP`'s positional census never listed it, so
`emit_new_pyframe_inline_with_params`' `NewWithVtable(pyframe_size_descr())`
left it holding bytes from the previous nursery cycle.
`emit_force_virtualizable` (`pyjitpl.py:1148-1158`) then read them with
`GETFIELD_GC_R`; the backend spilled that Ref into a jitframe ref-root slot and
marked it live in the call site's gcmap, so the next minor collection
dereferenced them.

Fix: carry the edge through a new `build_object_descr_group_with_extra_gc_edges`
lane — the one the inherited `PyObject.w_class` edge already uses — which adds to
`gc_fielddescrs` without disturbing the positional `all_fielddescrs` that
`field_descr_from_group` indexes. Adding it positionally instead would risk a
descr split.

### How it was pinned down

The panic's neighbourhood dump is what settled it. Reading each word around
`obj_addr` as a header and stepping by that type's size landed **exactly** on the
next candidate header (`0 → 8 + 24 = 32`), which proves `obj_addr` was a *header*
address and the real object was `obj_addr + 8` — so the slot held a bad value,
not a stale pointer. From there: the holder's `jf_gcmap` bits gave the slot
(`bit = max_output_slots + slot`, `holder_offset = 64 + 8*bit`), the backend's
`ref_root_slots` named the producing var, and its defining op was
`GcLoadR(v651, #80)` off a `CallMallocNursery(240)` whose offset 80 had no store.
The `GcStore(p, ofs, #0)` ops already in the trace **are** the `clear_gc_fields`
pending zeros, so "which offsets got a zero" reads the census straight off the
trace.

It looked nondeterministic because it reproduces 5/5 under non-tty stdio
(`capture_output=True, stdin=DEVNULL`) and 0/100 standalone — stdio setup shifts
the allocation history, hence what the recycled bytes happen to hold.

### Audit of the same class

Every hand-written descr group that JIT code can allocate was checked against its
Rust struct's GC pointers. `PyFrame` was the only gap.

| descr | GC pointers in the struct | census |
|---|---|---|
| `PyFrame` | …, **`vable_token`** | missing → fixed here |
| `W_ListObject` | `items`, `int_items.block`, `float_items.block` | all `Type::Ref` |
| `PyTraceback` | `w_class`, `frame`, `w_next`, `w_code` | all declared; the emitter stores all six fields |
| `W_FloatObject` | `w_dict`, `w_slots` (forwarded unconditionally) | both `Type::Ref` |
| `W_LongObject` | `value` | `Type::Ref` |
| `W_IntObject`, `W_TupleObject`, `Method`, `W_SliceObject`, specialised tuples | — | complete |

`W_DictObject` (`dstrategy`, `dstorage`) and `W_ObjectObject` (`storage`) have
incomplete censuses, but neither has a `*_size_descr()` accessor at all, so no
allocation op can carry them — unreachable today, not fixed here.

## 5. The inline empty list assumed a memzero that does not happen

Same false premise, one level down: `emit_empty_list_inline` and
`w_list_size_descr`'s doc both stated that `NewWithVtable` leaves untouched slots
zero-filled. `items` / `int_items.block` / `float_items.block` are GC-pointer
fields of the descr and so *are* covered by `clear_gc_fields`; `length` is a plain
int and was not. It is unreachable today — `set_object_items_from_vec` resets
`length` on the Empty→Object transition and the typed strategies read
`*_items.len` instead — so this is hardening, not a live bug. Stored explicitly,
as `rlist.py ll_newlist` does, and both comments restated in terms of the
mechanism that actually clears the slots.

## 6. `main` did not compile

`#907` added a second `#[cfg(not(feature = "sandbox"))] fn
interp_return_log_enabled` with an identical body next to the existing one, so
`pyre-interpreter` fails E0428. The branch cannot be built or gated without
removing it, so the duplicate is dropped here.

## Verification

Witness for the handoff, on the #899 fixture `getframe_caller_locals_nested_compiled_callee`:

| backend | forces | hits | misses | output |
|---|---|---|---|---|
| dynasm | 5 | 5 | 0 | 105005 ✓ |
| cranelift | 5 | 5 | 0 | 105005 ✓ |

Identical to the pre-change measurement, which is what confirms the new key (the decoded virtualizable) resolves to the same frame the consumer looks up.

The `mutate_then_raise_caught` abort of section 4 reproduces 5/5 on `origin/main` without any commit from this branch, and 0/30 with the fix.

Gate, rebased onto `origin/main` `25b2442c4e`: `check.py` **dynasm 345/345, cranelift 345/345, wasm 341/341**. Before section 4 the same tree was cranelift 344/345 — that one failure is this bug. Earlier in the branch: differential JIT-vs-`PYRE_NO_JIT` corpus 329/329 on both backends · `cargo test -p majit-metainterp --lib` 1418 passed · `cargo fmt --check` clean · type-checks on dynasm, cranelift and wasm32.

Honest scope note: the current corpus produces `forces == hits` — every entry is consumed, so the pruner never fires and no collection lands inside the rooting window. Both are structural fixes; the paths that leave an entry unconsumed are the escaped-virtualizable raise and `handle_fail`'s bridge-compiled arm.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `GUARD_NOT_FORCED` resumption when virtualizable values are present, ensuring decoding stays aligned.
  - Fixed forced-virtual cache handling to avoid keeping stale entries after their owning frames are gone.
  - Added GC tracking so forced-virtual references are properly walked and pruned.

- **Improvements**
  - Enhanced forced-virtual caching behavior, including hit/miss logging.
  - Added/updated metainterp resume tests to cover cache reset and forced-virtual edge cases.

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -2857,7 +2857,7 @@ impl MiniMarkGC {
         // An owner outside old-gen is either immortal (`malloc_typed`, no
         // header to read) or, under a non-moving major, still in the live
         // nursery; neither can be proven dead here, so both are kept.
-        crate::shadow_stack::prune_ephemeron_tables(&mut |owner| {
+        let mut classify_owner = |owner: usize| -> Option<usize> {
             if owner == 0 || !self.oldgen.contains(owner) {
                 return Some(owner);
             }
@@ -2867,7 +2867,19 @@ impl MiniMarkGC {
             } else {
                 None
             }
-        });
+        };
+        crate::shadow_stack::prune_ephemeron_tables(&mut classify_owner);
+        // The same question for tables a single mutator owns in its own TLS.
+        // Those cannot go through the global registration: it names no thread,
+        // so a major driven here would leave every other mutator's dead-owner
+        // entries pinned. Reach exactly as far as this collection's own root
+        // walk did (`enumerate_root_walker_values`) — foreign TLS only while
+        // this thread owns STW.
+        if crate::gc_sync::mutators_quiesced() {
+            crate::shadow_stack::prune_all_mutator_areas(&mut classify_owner);
+        } else {
+            crate::shadow_stack::prune_my_mutator_areas(&mut classify_owner);
+        }
         // incminimark.py:2510-2511 — run destructors of dying old objects
         // before the sweep frees them (VISITED still distinguishes
         // survivors from the dying at this point).
--- a/majit/majit-gc/src/shadow_stack.rs
+++ b/majit/majit-gc/src/shadow_stack.rs
@@ -263,6 +263,7 @@ struct MutatorEntry {
     bh_regs_stack: *const RefCell<Vec<BhRegsEntry>>,
     resume_ref_roots_stack: *const RefCell<Vec<(*mut i64, usize)>>,
     extra_areas: Vec<MutatorExtraArea>,
+    pruners: Vec<MutatorPruner>,
 }
 
 /// Walker for one opaque root area owned by a registered mutator.
@@ -274,6 +275,18 @@ pub type MutatorExtraWalkFn = unsafe fn(*const (), &mut dyn FnMut(&mut GcRef));
 #[derive(Clone, Copy)]
 struct MutatorExtraArea {
     walk: MutatorExtraWalkFn,
+    data: *const (),
+}
+
+/// Pruner for one owner-keyed side table owned by a registered mutator.
+///
+/// Same contract as [`MutatorExtraWalkFn`]: it runs on the collecting thread
+/// and must derive everything from `data`, never from caller TLS.
+pub type MutatorPrunerFn = unsafe fn(*const (), &mut dyn FnMut(usize) -> Option<usize>);
+
+#[derive(Clone, Copy)]
+struct MutatorPruner {
+    prune: MutatorPrunerFn,
     data: *const (),
 }
 
@@ -307,6 +320,7 @@ pub fn register_mutator() {
         bh_regs_stack,
         resume_ref_roots_stack,
         extra_areas: Vec::new(),
+        pruners: Vec::new(),
     });
 }
 
@@ -326,6 +340,72 @@ pub unsafe fn register_mutator_extra_area(walk: MutatorExtraWalkFn, data: *const
         .find(|entry| entry.thread_id == thread_id)
         .expect("register_mutator_extra_area called before register_mutator");
     entry.extra_areas.push(MutatorExtraArea { walk, data });
+}
+
+/// Append an owner-keyed-table pruner to the current registered mutator.
+///
+/// The ephemeron half of [`register_mutator_extra_area`]: a table whose keys are
+/// owner addresses and whose values a walker roots needs its dead-owner entries
+/// dropped, and needs it with the same per-mutator reach the root walk has.
+/// [`register_ephemeron_pruner`] cannot serve TLS-owned state — it hands the
+/// classifier no way to name a thread, so a major driven by one thread would
+/// leave every other thread's dead-owner entries pinned.
+///
+/// # Safety
+///
+/// Same as [`register_mutator_extra_area`]: `data` must stay valid until
+/// [`unregister_mutator`] runs on this thread, and `prune` must derive every
+/// address it dereferences from `data`, never from caller TLS.
+pub unsafe fn register_mutator_pruner(prune: MutatorPrunerFn, data: *const ()) {
+    let thread_id = std::thread::current().id();
+    let mut registry = MUTATOR_REGISTRY.lock().unwrap();
+    let entry = registry
+        .iter_mut()
+        .find(|entry| entry.thread_id == thread_id)
+        .expect("register_mutator_pruner called before register_mutator");
+    entry.pruners.push(MutatorPruner { prune, data });
+}
+
+/// Prune every registered mutator's owner-keyed tables during STW.
+///
+/// Called from the same pre-sweep point as [`prune_ephemeron_tables`] and with
+/// the same classifier; see that function for why only a major prunes. Callers
+/// pick between this and [`prune_my_mutator_areas`] on
+/// `gc_sync::mutators_quiesced()`, exactly as the root walk picks between
+/// [`walk_all_extra_areas`] and [`walk_my_extra_areas`] — so a collection's
+/// prune reach always equals its own root-walk reach.
+pub fn prune_all_mutator_areas(classify: &mut dyn FnMut(usize) -> Option<usize>) {
+    debug_assert!(
+        crate::gc_sync::mutators_quiesced(),
+        "prune_all_mutator_areas reaches foreign mutator TLS; caller must own collector-side STW",
+    );
+    let registry = MUTATOR_REGISTRY.lock().unwrap();
+    for mutator in registry.iter() {
+        for pruner in mutator.pruners.iter() {
+            // SAFETY: gc_sync has quiesced every registered owner, and each
+            // pruner's data remains valid until its MutatorEntry is removed.
+            unsafe { (pruner.prune)(pruner.data, classify) };
+        }
+    }
+}
+
+/// Prune the current mutator's owner-keyed tables.
+///
+/// The single-thread collection path, mirroring [`walk_my_extra_areas`]:
+/// callers without a registered mutator have no per-thread tables and are a
+/// no-op. A collection that only walked its own roots must only prune its own
+/// tables — another mutator's owner was never marked here, so its entries
+/// cannot be classified.
+pub fn prune_my_mutator_areas(classify: &mut dyn FnMut(usize) -> Option<usize>) {
+    let thread_id = std::thread::current().id();
+    let registry = MUTATOR_REGISTRY.lock().unwrap();
+    let Some(mutator) = registry.iter().find(|entry| entry.thread_id == thread_id) else {
+        return;
+    };
+    for pruner in mutator.pruners.iter() {
+        // SAFETY: this is the owning thread's synchronous collection path.
+        unsafe { (pruner.prune)(pruner.data, classify) };
+    }
 }
 
 /// Walk every registered mutator's opaque extra root areas during STW.
--- a/majit/majit-metainterp/src/jitdriver.rs
+++ b/majit/majit-metainterp/src/jitdriver.rs
@@ -4590,14 +4590,14 @@ impl<S: JitState> JitDriver<S> {
     /// virtuals through the same resume allocator used by ordinary guard
     /// failure.  In particular, a `jit.virtual_ref` frame must not be decoded
     /// through `NullAllocator`, or its `forced` writeback remains null.
-    pub fn force_virtualizable_token(&mut self, token: u64) -> Option<(Vec<i64>, Vec<i64>)> {
+    pub fn force_virtualizable_token(&mut self, token: u64) {
         let fallback_alloc = crate::resume::NullAllocator;
         let allocator: &dyn crate::resume::BlackholeAllocator = self
             .blackhole_allocator
             .as_deref()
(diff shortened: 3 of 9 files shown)


youknowone:
Adjudicating the three Codex findings. Two refuted on a shared premise, one correct and fixed in `78e30de`.

### P1 "Re-key the forced-cache owner during minor collections" — refuted

The premise is that the `PyFrame` owner can be forwarded by a minor. It cannot: the owner is the *virtualizable*, and a virtualizable is an interpreter-created frame from `FrameBox::new` → `try_gc_alloc_stable_raw`, whose registered contract is "The backend routes this to an old-gen allocator whose returned pointer is **stable across minor and major collections** (MiniMark mark-sweep does not move old-gen objects)" (`pyre-object/src/gc_hook.rs:140-142`; `pyframe.rs:563-566` says the same at the allocation site).

The frames a *trace* builds do go through a movable path — `emit_new_pyframe_inline_with_params` emits `NewWithVtable(pyframe_size_descr())`, which can land in the nursery. But such a frame is never the vable identity: `consume_vable_info` (`resume.rs`) reads the one frame that entered the compiled loop, which existed before the trace was recorded. An inlined-callee frame is a virtual in the callee's MIFrame, not the virtualizable.

Rather than leave that as an argument, it is now a checked invariant — `save_forced_virtuals` carries `debug_assert!(!majit_gc::gc_is_nursery_object(owner))`. A debug build running the `getframe_caller_locals_nested_compiled_callee` shape completes **11 major and 44 minor collections** with 5 forces and 5 cache hits, and the assertion never fires. (The assertion is not vacuous: the sibling quiescence assertion added in the same commit *did* fire and caught a real defect — see P2.)

The second half of the finding — that an unconsumed young-owner entry can never be classified dead — is the documented contract of this mechanism, not new: the classifier keeps every owner outside old-gen, because "An owner outside old-gen is either immortal (`malloc_typed`, no header to read) or, under a non-moving major, still in the live nursery; neither can be proven dead here" (`collector.rs`). It is shared with the mapdict tables, and a `PyFrame` owner is old-gen regardless.

### P1 "Root the virtualizable key while forcing virtuals" — refuted

Same premise, same refutation: the virtualizable is move-stable, so copying `virtualizable_ptr` before `force_all_virtuals()` cannot desynchronize from anything.

Reading it before the materialization is also deliberate, not incidental. The sibling `blackhole_from_resumedata` can root the reader's slot only because `prepare_resume_heap_with_roots` opened a resume-root scope whose `Drop` releases it; `force_from_resumedata` calls the bare `prepare`, which opens none, so a `push_resume_ref_roots` there would leak past the return to compiled code — the hazard `pyre-jit/src/residual_call.rs:589-594` documents and solves with a permanent extra area instead.

### P2 "Prune caches from every mutator's driver" — correct, fixed

Right, and citing the in-repo contract makes it sharper: `MutatorExtraWalkFn` states the callback "must derive every thread-specific address from `data`, never from caller TLS", and the global `register_ephemeron_pruner` gives the classifier no way to name a thread, so the pyre pruner had to violate that.

Fixed by giving the mechanism the missing half: `register_mutator_pruner` / `prune_all_mutator_areas` / `prune_my_mutator_areas` alongside the extra-area equivalents, in the same `MutatorEntry`, invoked from the pre-sweep point the collector already prunes at with the same classifier. pyre registers it on the same `jit_driver` data as `forced_virtuals_root_walker_area`.

One correction to the suggested shape, found by asserting rather than assuming: pruning cannot unconditionally iterate all mutators. The pre-sweep point does **not** own collector-side STW — the unconditional form tripped `prune_all_mutator_areas`'s quiescence assertion on the very first run. So it branches on `gc_sync::mutators_quiesced()`, exactly as `do_collect_nursery` and `enumerate_root_walker_values` branch between `walk_all_extra_areas` and `walk_my_extra_areas`. That makes the real invariant explicit: a collection's prune reach always equals its own root-walk reach.

The mapdict tables stay on the global registration — they are a process-global `Mutex` map, reachable from any thread, so they need no per-mutator dispatch.

### Noted, not touched

`blackhole_from_resumedata`'s rooting comment justifies itself with "That relocates the young virtualizable frame". Under the allocation contract above that rationale is inaccurate. The push is harmless and I have not removed it or rewritten the comment: deciding whether *any* allocator regime makes a virtualizable movable means enumerating the `install_gc_box` per-thread-allocator and wasm nursery regimes (`install_gc_box` disarms the published nursery, so the new assertion cannot even answer there), which is its own investigation. Flagging it here rather than replacing one inaccurate claim with another.

### Verification after the fix

`check.py` dynasm 345/345, cranelift 345/345, wasm 341/341 · differential JIT-vs-`PYRE_NO_JIT` corpus 330/330 on both backends · `cargo test -p majit-metainterp --lib` 1423 passed · `cargo fmt --all --check` clean.

— *commented by Claude*

PR #908, opened 2026-07-30
jit-trace: anchor an inlined level's walk-time traceback on its own frame

youknowone:
Rebased onto current `main`. Two of the four original commits were dropped: **#910 landed the same blocker on `main` by a better mechanism**, mirroring an inlined MIFrame's standard-vable writes onto that level's own concrete red frame as they are made (`current_inline_concrete_frame`, `store_live_frame_static_int`) instead of publishing register images at the adopt. `#910` also absorbed the per-level ref-bank rooting, so `trace.rs` / `state.rs` here resolve entirely to `main`'s side.

What is left is the third blocker and its guard.

### The walk-time traceback named a fabricated frame

`record_inline_application_traceback` recorded the walk's concrete node through `record_inline_traceback_for_recording`, which `createframe_obj`s a throwaway frame from the promoted code and globals. That hook predates the inline seed. A seeded level owns a real frame — the sub-walk runs the callee on it, and the **emitted** node already names that same object, since `traceback_node_site` resolves its frame operand from the level's portal frame register and the seed stamps the concrete frame onto that operand's box.

So the walk and the compiled run named different frames for the same invocation, and the walk's answer is the one a multi-frame blackhole adopt commits: `tb.tb_frame is sys._getframe()` read `False` for exactly one iteration per adopt.

### The obligation that rides with the anchor

The fabricating hook does **two** things — `frame.last_instr = last_instruction`, then record. Porting only the second reintroduced a different wrong answer: a level's own frame carries the `-1` entry sentinel, because the recording walk does not make `dispatch_bytecode`'s per-opcode `last_instr` store and a frame that leaves by the exception never reaches an exit that would publish one. `f_lineno` then answers the `def` line.

CI caught it and the local run did not: `synth/exception_traceback_frame_lineno` grew a second `('raises_out', 1, 0)` shape beside the correct `('raises_out', 1, 1)`, on all three backends and all three OSes. The anchor now makes the same store the blackhole already makes for its replay in `publish_last_instr_at_live_marker`.

### The runtime half was investigated and declined

The `emit_runtime` arm still emits the fabricating hook. Two measurements say to leave it:

- **It is unreachable.** An lldb breakpoint on that arm's own call-descr construction counts **0 hits across 43 corpus exception fixtures and 15 hand-built probes**, corroborated by a `MAJIT_LOG` scan finding no call with the hook's `[Ref, Ref, Ref, Int, Int]` signature in any dumped trace. `emit_runtime` is the negation of `record_prepend_application_traceback`, which never declines — the `exc.is_constant()` arm is suppressed because every raising residual assigns `class_of_last_exc_is_const = false` immediately before `walker_record_guard_exception` reads it. (This also retracts the "compiled code still calls the fabricating hook" claim from the previous revision of this PR.)
- **The obvious port breaks an allocation contract.** `w_pytraceback_new` roots `w_next` and `w_code` but deliberately not `frame`, on the documented ground that executing frames are non-moving oldgen blocks. The top-level sibling, the walk and the fabricating hook all satisfy that; a compiled trace's inlined callee frame does not — it is the trace's own `NewWithVtable`, lowered to a nursery allocation. Passing it would hold a movable pointer across the parking allocation inside `w_pytraceback_new`. A future port needs the root-and-reload shape on the recorder first, which also covers the same pre-existing exposure on its `w_next` argument.

Both findings are recorded in `gate-triage.md` rather than left in a PR thread.

### Verified

Rebuilt from a fresh LLBC extraction on the new base.

| | |
|---|---|
| `synth/exception_traceback_frame_lineno` (the CI failure) | matches the oracle, dynasm **and** cranelift |
| `synth/blackhole_inlined_callee_local_after_escape` | `[(True, 2)]` = oracle, dynasm and cranelift |
| `getframe_inline_subwalk_multiframe` | 5 builds / 5 adopts / 0 declines, output unchanged |
| `getframe_while_escaping_read_frame_identity` | 10 / 10 / 0, output unchanged |
| `getframe_while_inlined_callee_subwalk` | 5 / 5 / 0, output unchanged |
| full corpus, dynasm | **ALL PASSED 349/349** |
| `cargo fmt --check` | clean |

### DIFF
--- a/pyre/bench/synth/blackhole_inlined_callee_local_after_escape.py
+++ b/pyre/bench/synth/blackhole_inlined_callee_local_after_escape.py
@@ -1,0 +1,47 @@
+# Guard for what an adopted multi-frame blackhole chain owes its inner levels.
+#
+# An inlined callee assigns a local, the frame then escapes through a residual
+# `sys._getframe()`, and an attribute read POSITIONED AFTER that escape reads the
+# local back.  The read is executed by the blackhole, not by the walk, so the
+# shape holds the adopt to two separate obligations and fails differently on
+# each:
+#
+#   * every LOAD_FAST lowers to `getarrayitem_vable_r` on the level's own frame
+#     array, so a level whose locals were left unpublished resumes `tb` as null
+#     and the attribute read faults in `object_getattr_miss` -- a hard SIGSEGV
+#     (rc=139) with no output at all;
+#   * the traceback the callee stored has to name the frame the callee runs on,
+#     so a walk-time node anchored on any other object prints `False` here while
+#     still exiting 0.
+#
+# The second is the quieter one and the reason the assertion is an identity
+# rather than a liveness check.  Both need the escape to happen inside an
+# INLINED callee: the same shape through the single-frame arm was always
+# correct.
+#
+# Deliberately carries no `# pyre-check: max-pypy-ratio=` header: this guards an
+# output, and the forcing read makes it a poor perf subject.
+import sys
+
+N = 20000
+
+
+def catches_here(i):
+    try:
+        raise ValueError(i)
+    except ValueError as e:
+        tb = e.__traceback__
+        f = sys._getframe()
+        return (tb.tb_frame is f, tb.tb_lineno - f.f_code.co_firstlineno)
+
+
+def drive():
+    seen = set()
+    k = 0
+    while k < N:
+        seen.add(catches_here(k))
+        k += 1
+    return sorted(seen)
+
+
+print(drive())
--- a/pyre/bench/synth/getframe_inline_subwalk_multiframe.py
+++ b/pyre/bench/synth/getframe_inline_subwalk_multiframe.py
@@ -9,9 +9,10 @@
 # residual frame rather than the walked frame, and
 # `try_adopt_multi_frame_blackhole`'s chain-root identity gate declined it. What
 # that decline wanted was the `jit.virtual_ref` emit at the inline push
-# (`executioncontext.py:89`); `walker_ec_enter` / `walker_ec_leave` landed it.
-# The gate no longer fires here: the shape now adopts once per build and returns
-# the same result as before. This fixture pins that result.
+# (`executioncontext.py:89`); `walker_ec_enter` / `walker_ec_leave` landed it, so
+# the chain-root gate no longer fires here, and with every level resumable from
+# the concrete frame it owns, the chain is adopted rather than replayed. This fixture pins the
+# result across that adopt.
 #
 # One `sys._getframe(1)` level does NOT reach the build: the chain needs a
 # residual level under the walked frame and an inlined level under that, so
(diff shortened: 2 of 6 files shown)


youknowone:
Pushed two more commits: both remaining blockers behind the multi-frame adopt
are closed, so the arm adopts rather than declining.

## `874730b930` — the walk-time traceback named a frame nothing else could reach

The concrete traceback node for an inlined level went through
`record_inline_traceback_for_recording`, which `createframe_obj`s a
traceback-only frame from the promoted `w_code` / `w_globals`. Its comment says
why — "the recording walker cannot force the optimizer's virtual locals" — but
that predates the inline seed. A seeded level HAS a frame, the sub-walk runs the
callee on it, and the emitted node already names it: `traceback_node_site` takes
its frame operand from the level's portal frame register and the seed stamps the
concrete frame onto that operand's box.

So the walk and the compiled run disagreed about `tb.tb_frame` for the same
invocation, and only a path that COMMITS the walk's concrete effects can see it
— which is why it showed up as exactly one wrong iteration per adopt and never
in steady state.

The node is now recorded against that frame; a level inlined with no frame at
all keeps the fabricating hook, which is then its only frame identity. The
python pc is resolved first, so the record-or-skip decision is byte-identical
and only the frame changes.

## `857f48dccc` — publish every level's locals, drop the depth decline

`semantic_slot_owner` resolves one `locals_cells_stack_w` slot to the single
`(bank, color)` that owns it at the level's resume pc; `write_back_miframe_
locals` boxes and stores every resolved slot behind an all-or-nothing preflight.
Every level's old locals image is captured before the first publish, and the
images, the latched Ref words and the pending exception stay
`push_resume_ref_roots`-rooted for the whole drive: publication boxes, boxing
allocates, and each image is the only remaining reference to the pre-walk
objects its publish overwrote.

## Measured

| fixture | builds / adopts / declines | output |
|---|---|---|
| `getframe_inline_subwalk_multiframe` | 5 / 5 / 0 | matches reference |
| `getframe_while_escaping_read_frame_identity` | 10 / 10 / 0 | matches reference |
| `getframe_while_inlined_callee_subwalk` | 5 / 5 / 0 | matches reference |
| `blackhole_inlined_callee_local_after_escape` | — | `[(True, 2)]`, matches |

`dynasm: 1 failed, 345 passed` — the failure is `synth/ast_compile_roundtrip`,
the cpython/pypy reference mismatch that predates this branch. `cargo fmt
--check` clean.

## Not fixed here

The runtime-emitted hook is unchanged. When `record_prepend_application_
traceback` declines but the level is seeded — mainly `exc.is_constant()`, which
holds for every raise after the first in a walk — compiled code still calls the
fabricating hook, so a traceback built inside a compiled inlined callee names a
frame nothing else can reach. Same defect class, one layer over.

— *commented by Claude*

youknowone:
Force-pushed after a rebase onto current `main`, and the CI failure is fixed.

**Rebase changed the shape of this PR.** `#910` landed the inner-level-locals blocker on `main` by mirroring an inlined MIFrame's standard-vable writes onto that level's own concrete red frame, rather than publishing register images at the adopt. That supersedes two of the four commits here outright, so `trace.rs` and `state.rs` resolved to `main`'s side and the branch is now three commits.

**The `pyre/check.py` failure was mine.** `synth/exception_traceback_frame_lineno` went WRONG on all three backends and all three OSes with a second `('raises_out', 1, 0)` shape. Root cause: the frame-fabricating hook this PR replaces does two things — `frame.last_instr = last_instruction`, then record — and only the second was ported. A level's own frame carries the `-1` entry sentinel, so `f_lineno` answered the `def` line. Fixed by making the same store `publish_last_instr_at_live_marker` already makes for the blackhole's replay.

**The runtime-emitted half is deliberately not ported**, and the previous revision's claim that "compiled code still calls the fabricating hook" is retracted. An lldb breakpoint on that arm counts 0 hits across 43 corpus exception fixtures and 15 probes, because `record_prepend_application_traceback` never declines in practice. And the obvious port would be the first path to hand `record_application_traceback` a nursery-resident frame, which its `w_pytraceback_new` contract explicitly does not allow. Both are written down in `gate-triage.md`.

Verified on the new base after a fresh LLBC extraction: full corpus **ALL PASSED 349/349** on dynasm, the two target fixtures matching the oracle on cranelift as well, latch fixtures 5/5/0, 10/10/0, 5/5/0, `cargo fmt --check` clean.

— *commented by Claude*

PR #909, opened 2026-07-30
gc: root the working values in any/all and set.union/update, and make shadow-stack reads cheap

youknowone:
Three commits: a rooting regression harness, the two fixes it demonstrates, and the shadow-stack read cost that rooting a streaming loop exposes.

## What was wrong

A `PyObjectRef` in a plain Rust local is not a root — the collector walks Python frames, not the Rust stack. Two helpers held working values across calls that run arbitrary Python:

- **`any` / `all`** bound the iterator in a Rust local and called `next()` and `is_true()` — a user `__next__` and `__bool__` — without pinning it and without reloading it between calls. `collect_iterator`, in the same file, states the opposite rule above its own loop and follows it.
- **`set.union` / `set.update`** held the set being built across every operand's drain. `collect_iterable` runs a user `__iter__`/`__next__` and `builtin_set_add_items` a user `__hash__`/`__eq__`, so with two or more non-set operands the set crosses two full drains. The sibling `set_init_from_iterable_impl` already pins for exactly this reason.

## How they were demonstrated first

The first commit adds the tests and lands them **red**, so the defects are reproducible from history rather than asserted. At that commit:

```
any_all_iterators_survive_python_callbacks
  GC BUG: invalid major child type_id=... site=major_varsize_item
set_union_update_values_survive_python_callbacks
  Went past end of probe sequence
```

The reported `type_id` differs between runs.

Two of the seven programs are **baselines** over helpers that already root correctly — `min`/`max` and `sorted` with a collecting `key=`. Both pass, which is what makes the other results attributable to the helper under test rather than to the harness. Operands are passed as temporaries, never bound to a Python local, so the frame cannot keep them reachable incidentally.

Three programs — reflected binop, comparison order, and `sum` — **pass**. Their operands stay reachable in these shapes, so those helpers are untested, not shown sound. They are deliberately left unfixed: there is no discriminator for them yet.

`builtin_set_add_items_impl` is unchanged. It already pins the set and every slice element and reloads them after each hash.

## The third commit

Rooting a streaming loop costs one `shadow_stack_get` per element, which was `SHADOW_STACK.with(|s| s.borrow()[index])` — a TLS access, a `RefCell` borrow flag with its guard drop, and a bounds check. Measured on a 300k-element `any`/`all` walk (user CPU, min of 15, machine quiet): control 2.75s → 3.02s, **9.8%**, with `any` and `all` agreeing to three decimals (1.0981 / 1.0978) and an untouched `sorted` control at 0.9944.

`ShadowStackPool` (`shadowstack.py:277-281`) is a fixed-capacity raw array addressed through `root_stack_base`/`root_stack_top` with raw address arithmetic (`:81-88`) and `ll_assert` (`:319`) as its check — so the `RefCell` was the deviation. Access now goes through an `UnsafeCell`, and the re-entrancy net moves to a signed depth counter under `cfg(debug_assertions)`, where `cargo test` runs. Bounds checks stay; `pin_root`'s publish-then-normalize ordering is untouched.

`walk_shadow_stack_area` names the cell type in a raw cast and moves in lockstep — the two walkers over this storage are independent, and leaving that one behind would still compile while walking a quiesced thread's roots through the wrong layout.

## Verification

| | status |
|---|---|
| `gc_stress` | 21 passed / 2 failed at commit 1 → **23 passed / 0 failed** at commit 2, held at commit 3 |
| `cargo test -p pyre-object` | 282 passed |
| `check.py` both backends | 344 passed, 1 failed — the pre-existing `synth/ast_compile_roundtrip` cpython/pypy oracle mismatch |

**Pending, and I will post them as a comment:** `check.py` has been run through commit 2 but not yet over commit 3, and the A/B has not been re-run to show how much of the 9.8% the third commit recovers. Both need a quiet machine; load has been sitting above 100. The accessor change stands on the orthodoxy argument regardless of the recovery figure, but the figure is owed.

— *commented by Claude*

### DIFF
--- a/pyre/pyre-interpreter/src/builtins.rs
+++ b/pyre/pyre-interpreter/src/builtins.rs
@@ -10173,9 +10173,23 @@ fn builtin_any(args: &[PyObjectRef]) -> Result<PyObjectRef, crate::PyError> {
             positional.len()
         )));
     }
-    let it = crate::baseobjspace::iter(positional[0])?;
+    // `iter` and every `next` run arbitrary Python. Pin the iterable, the
+    // iterator, and a dict view's backing dict for the whole walk, and reload
+    // the iterator from its slot before each call: a moving minor collection
+    // relocates it behind a plain Rust local. The shape `collect_iterator` uses.
+    let _roots = pyre_object::gc_roots::push_roots();
+    let iterable_slot = pyre_object::gc_roots::shadow_stack_len();
+    pyre_object::gc_roots::pin_root(positional[0]);
+    let it = crate::baseobjspace::iter(pyre_object::gc_roots::shadow_stack_get(iterable_slot))?;
+    let iterator_slot = pyre_object::gc_roots::shadow_stack_len();
+    pyre_object::gc_roots::pin_root(it);
+    if unsafe { pyre_object::dictmultiobject::is_dict_view_iterator(it) } {
+        let w_dict = unsafe { pyre_object::dictmultiobject::w_dict_view_iterator_get_dict(it) };
+        pyre_object::gc_roots::pin_root(w_dict);
+    }
     loop {
-        match crate::baseobjspace::next(it) {
+        let it_now = pyre_object::gc_roots::shadow_stack_get(iterator_slot);
+        match crate::baseobjspace::next(it_now) {
             Ok(item) if crate::baseobjspace::is_true(item)? => return Ok(w_bool_from(true)),
             Ok(_) => {}
             Err(e) if e.kind == crate::PyErrorKind::StopIteration => return Ok(w_bool_from(false)),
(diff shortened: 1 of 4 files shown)


youknowone:
Closing out the two measurements this PR's description left owed.

## `check.py` over the final commit

Both backends: **344 passed, 1 failed** — `synth/ast_compile_roundtrip`.

Not attributable to this PR. `check.py` reports it as `cpython/pypy output
mismatch`: the two oracles disagree *before* pyre runs. It was fixed
independently by #906, which merged at 02:05 — after the base this branch was
measured against (`0cda9d43aa`, 01:16). The tested tree simply predates the
fixture fix.

## The A/B — the rooting cost stands at 9.8%; the recovery figure does not

| probe | control | rooted | ratio |
|---|---|---|---|
| `any` over 300k × 800 | 2.75s | 3.02s | **1.0981** |
| `all` over 300k × 800 | 2.76s | 3.03s | **1.0978** |
| `sorted` (untouched path) | 16.11s | 16.02s | 0.9944 |

User CPU, min of 15, at load ~5. Both soundness checks hold: `any` and `all`
are the same loop with opposite polarity and agree to three decimals, and the
untouched path sits at 0.9944.

That 9.8% is an upper bound, not a representative slowdown. The probe is 240M
`next()` calls whose per-element work is essentially nothing — every element is
`0` — so one added `shadow_stack_get` per element is close to the entire cost.
Anything doing real work per element dilutes it.

**The third commit's recovery figure is still unmeasured, and I am not quoting
the attempt I have.** It reported `rooted/base` 1.0975 for `any` but 1.0225 for
`all` — the same loop, disagreeing by 7 points — and the untouched `sorted`
control moved 5% (0.9467). The fault was the harness, not the box: the arms ran
in sequence (15 base, then 15 rooted, then 15 final) while load fell
monotonically from 10.9 to 6.4, so each arm was measured on a different machine.
The re-run interleaves the arms within each rep and rotates their order, but it
has been sitting on a load gate that has not opened — this box has been between
60 and 130 for hours.

## Follow-up on the Codex review

#917. Short version: the mechanism the review describes is real, but it cites a
walker that is dead outside tests, and the walker that actually runs had no
check before this PR either — it read through `RefCell::as_ptr()`, which
bypasses the borrow flag. Details and the reachability analysis are in that PR.

— *commented by Claude*

youknowone:
The re-run finished. The recovery figure I owed above is now measured, and the
answer is that there is no recovery to report.

| | `rooted/base` | `final/base` |
|---|---|---|
| `any` over 300k × 800 | 1.1054 | 1.1090 |
| `all` over 300k × 800 | 1.1217 | 1.1291 |
| `sorted` (untouched) | 1.0120 | 0.9955 |

User CPU, min of 15, arms interleaved within each rep with their order rotated,
load 3–6 throughout.

Read the noise floor first: the untouched `sorted` path moved 1.2% in one arm
and −0.45% in the other, and `any`/`all` — the same loop with opposite polarity
— disagree by 1.6 to 2.0 points. So this run resolves to about ±1.5%, coarser
than the ±0.03 the two-arm measurement managed. Any conclusion has to survive
at that resolution.

At that resolution: **`final` and `rooted` are the same.** The third commit's
cheap-accessor change recovered nothing measurable — if anything `final` is
0.4–0.7 points slower, in both probes, which is inside the floor.

That is consistent with where the cost actually is. `shadow_stack_get` carries
`#[majit_macros::dont_look_inside]`, which forces `#[inline(never)]`, so every
read is a real call plus a TLS resolve. Dropping the `RefCell` borrow flag
removes a few instructions from underneath a much larger fixed cost. The 9.8–11%
is the per-element call itself, not the borrow bookkeeping.

So the accessor commit stands on the orthodoxy argument alone — which is what
the PR description claimed for it, and now that is not merely the main reason
but the only one. It is worth recording that a change justified by a measured
9.8% did not move that 9.8%.

This also sets expectations for the fixed-capacity conversion discussed in #917:
converting the storage to upstream's `root_stack_base`/`root_stack_top` raw form
closes the remaining soundness gap, but on this evidence it should not be
expected to pay for itself in speed unless it also removes the never-inlined
call and the TLS resolve.

— *commented by Claude*

PR #911, opened 2026-07-30
review: act on the Codex and CodeRabbit findings on #887

youknowone:
Follow-up to #887, acting on the automated review findings that landed on it after the merge.

## Removed

- **`CriticalCodeGuard` in the `bound_reached` compile path** (`pyre-jit/src/eval.rs`). It spanned the whole trace-compile-and-dispatch region, so `report_error = 0` covered the walk's residual calls into user code — `stack.c:111` returns that flag in place of the overflow verdict, turning a `RecursionError` into an unreported native overflow. Upstream marks only regions that run no user code (`compile.py:976`, `pyjitpl.py:3305`, `resume.py:1318`). Measured on the fixture it was added for, `synth/wasm_ca_trampoline_decline`: with the guard suppressed the overflow no longer occurs at all, so the region is inert.
- **The `type_id == 9` backtrace capture** in `remember_young_pointer`.

## GC correctness

- **`w_tuple_new_array_backed` published an uninitialized tuple.** `try_gc_alloc_stable_raw` returns uninitialized payload and `tuple_object_custom_trace` reads `ob_header.w_class` and `wrappeditems`, so a collection between `pin_root` and the items-block allocation read garbage. The header now goes in first with a null items block — the state the trace hook returns early on — and the block is stored after.
- **`W_TextIOWrapper` published its reference fields with one barrier at the end of `attach_buffer`**, but every `w_str_new` in between allocates and `set_newline` / `getattr_str` / `set_encoder_decoder` run Python. Barrier after each reference store instead.
- **`OwnerRootGuard` was auto-`Send`.** Its `index` names a slot in the acquiring thread's `OWNER_ROOTS`; a `PhantomData<*mut ()>` pins it to that thread at compile time.

## Error handling

- **`attach_stdio_codec` dropped both the codec lookup and the encoder/decoder construction errors**, leaving a stream that reports itself unreadable however readable its buffer is. `init_stream_codecs` and `init_importlib_bootstrap` now propagate; a bootstrap failure still wins as the more fundamental one.
- **`allocate_stdio`** still cannot propagate — it runs inside `sys` module creation — but no longer discards the result blind: it finishes the two fields the early return skipped instead of forcing `STATE_OK` over a half-applied init.
- **`faulthandler.enable` restores `FAULTHANDLER_FD`** when `enable_fatal_handlers` fails; a failed re-enable was redirecting the handlers already installed.

## faulthandler file ownership

`enable` resolved `file=None` to a hard-coded fd 2 and dropped the file object. `get_fileno_and_file` (`handler.py:35-49`) resolves the *current* `sys.stderr`, asks it for `fileno()`, and flushes it ignoring an ordinary flush error; `enable` then parks it as `fatal_error_w_file` (`:145`) and `disable` clears it (`:150`), because the descriptor the installed handler writes to belongs to that object. pyre has no Handler instance, so the owner is a process-global slot walked as a GC root alongside the other interpreter globals — not a module attribute, which would show up in `dir(faulthandler)` where upstream has none.

## Performance

`acquire_owner_root` scanned `OWNER_ROOTS` from 0 for a free slot, so rooting the frames of a recursion `depth` deep cost O(depth²). Released interior slots now go on a free stack; a released tail still shrinks the walked range.

## Documentation

Recorded why `fbw_inline_recursion_count` matching `w_code` alone *is* the element-wise greenkey compare of `pyjitpl.py:1396-1401`: `MIFrame.setup` (`pyjitpl.py:74-80`) assigns `greenkey` once, so a frame's greenkey stays its entry greens, and every `InlineFrame` is pushed at a callee entry with the `(w_code, 0)` identity `reconstruct_inline_recipe` also stamps. Codex had flagged this as a mismatch; it is not one.

## Verification

```
ALL PASSED: dynasm 345/345
ALL PASSED: wasm   341/341
FAILED:     cranelift 1 failed, 344 passed
```

`cargo test`: pyre-interpreter 432 passed, majit-gc 215 passed.

The one cranelift failure, `synth/mutate_then_raise_caught`, panics in `collector.rs` (`minor_custom_trace_target`, invalid type_id). It reproduces 3/3 on a rebuild of `main` with none of these changes applied, and passes on the Linux CI runner — it is a pre-existing macOS-only failure, not introduced here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved garbage collection reliability when managing interpreter and stream references.
  - Fixed faulthandler file ownership and signal-handler behavior, including safer file descriptor validation.
  - Improved tuple initialization safety during garbage collection.
  - Prevented stream codec setup errors from being silently ignored.

- **Reliability**
  - Interpreter startup now reports failures when standard stream codecs cannot be initialized.
  - Improved reuse and cleanup of internal root slots for more consistent runtime behavior.

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -3402,9 +3402,6 @@ impl MiniMarkGC {
                 unsafe { (*hdr).type_id() },
                 self.gc_state
             );
-            if unsafe { (*hdr).type_id() } == 9 {
-                eprintln!("{}", std::backtrace::Backtrace::force_capture());
-            }
         }
         unsafe { (*hdr).clear_flag(flags::TRACK_YOUNG_PTRS) };
     }
--- a/majit/majit-gc/src/shadow_stack.rs
+++ b/majit/majit-gc/src/shadow_stack.rs
@@ -483,12 +483,19 @@ pub fn release_owner_root(index: usize) {
 /// RAII owner for one fixed translated-livevar root slot.
 pub struct OwnerRootGuard {
     index: usize,
+    /// `index` names a slot in the ACQUIRING thread's `OWNER_ROOTS`, so the
+    /// guard must not outlive that thread's stack frame.  Bare `usize` would
+    /// make it auto-`Send`, and dropping it elsewhere would release a foreign
+    /// thread's slot (or trip the inactive-slot assert).  The raw-pointer
+    /// marker pins it to one thread at compile time.
+    _not_send: std::marker::PhantomData<*mut ()>,
 }
 
 impl OwnerRootGuard {
     pub fn new(root: GcRef) -> Self {
         Self {
             index: acquire_owner_root(root),
+            _not_send: std::marker::PhantomData,
         }
     }
 
--- a/majit/majit-gc/src/shadow_stack.rs
+++ b/majit/majit-gc/src/shadow_stack.rs
@@ -220,6 +220,14 @@ thread_local! {
     /// when Rust RAII scopes nest across a `FrameBox`.
     static OWNER_ROOTS: RefCell<Vec<Option<GcRef>>> = const { RefCell::new(Vec::new()) };
 
+    /// Interior `OWNER_ROOTS` slots that have been released but still sit
+    /// below the live tail.  Popping one is what keeps `acquire_owner_root`
+    /// O(1): searching for the first `None` instead is a linear scan, and a
+    /// recursion `depth` frames deep pays O(depth²) just to root its frames.
+    /// Trailing slots never land here — `release_owner_root` shrinks those
+    /// away so an unwound recursion leaves no walked tail behind.
+    static OWNER_ROOTS_FREE: RefCell<Vec<usize>> = const { RefCell::new(Vec::new()) };
+
     /// Thread-local flat jitframe root stack. Rust tests run multiple
     /// JIT/GC tests in parallel, so using one process-global root stack lets
     /// one GC walk another test's jitframe. RPython's root stack is per
@@ -454,7 +462,9 @@ pub fn get(index: usize) -> GcRef {
 pub fn acquire_owner_root(root: GcRef) -> usize {
     OWNER_ROOTS.with(|roots| {
         let mut roots = roots.borrow_mut();
-        if let Some(index) = roots.iter().position(Option::is_none) {
+        let reused = OWNER_ROOTS_FREE.with(|free| free.borrow_mut().pop());
+        if let Some(index) = reused {
+            debug_assert!(roots[index].is_none(), "reusing a live owner-root slot");
             roots[index] = Some(root);
             index
         } else {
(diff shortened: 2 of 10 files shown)


youknowone:
Pushed `27e7baf70e`, acting on the two review findings above plus a `main` breakage found while verifying.

### `main` does not compile

`interp_return_log_enabled` is defined twice in `pyre/pyre-interpreter/src/eval.rs`, under the same `#[cfg(not(feature = "sandbox"))]` — #874 added one copy and #907 the other, and neither PR's CI saw the other's, so `origin/main` fails with E0428. Deleted the later duplicate here, since this branch cannot build otherwise.

### Retain each registered signal's file owner — fixed

`register` discarded the resolved file while the installed handler kept only its descriptor. `handler.py:125` keeps `user_w_files[signum] = w_file` and `:132` pops it on `unregister`; the same per-signal owners now live in a process-global table walked by the faulthandler root walker, which covers both it and the fatal-error file. Verified: `register(sig, f); del f; gc.collect()` leaves the fd open, and `unregister` returns `True` then `False`.

### Preserve asynchronous flush exceptions — **not** applied, after measuring

This one is PyPy-only. `handler.py:44-48` does re-raise via `e.async(space)`, and PyPy propagates a `SystemExit` raised out of `flush`. But pyre's behaviour target is CPython 3.14, and 3.14 clears the flush error unconditionally:

| | `enable(obj)` where `obj.flush()` raises `SystemExit(7)` |
|---|---|
| PyPy 3.11 | `SystemExit propagated: 7` |
| CPython 3.14.5 | `SystemExit SWALLOWED` |

I had implemented the propagation before checking, then reverted it — applying it would have moved pyre away from its target. The divergence is now recorded in a comment at the call site.

### Also fixed while in there

A `fileno()` returning a negative value reached `FAULTHANDLER_FD` and the handler registration unchecked. PyPy accepts it; 3.14 raises `RuntimeError: file.fileno() is not a valid file descriptor`, distinct from the direct-int arm's `ValueError: file is not a valid file descriptor`. Follows 3.14 now.

Resolving the file argument runs a user `fileno()` and `flush()`, and both `enable` and `register` ran it *before* the support gate — so a build that can only answer `NotImplementedError` executed those side effects first. Both moved inside the gate, which also puts `register`'s integer coercions ahead of the file resolution, as its argument parsing does.

### Verification

```
ALL PASSED: dynasm    345/345
ALL PASSED: cranelift 345/345
ALL PASSED: wasm      341/341
```

All three backends green. `synth/mutate_then_raise_caught`, which panicked in `collector.rs` (`minor_custom_trace_target`, invalid type_id) 3/3 on macOS before — including on a rebuild of the base with none of this branch applied — now passes 5/5. `cargo test`: pyre-interpreter 438.

Every faulthandler probe (file retention across GC for both `enable` and `register`, `unregister`, negative fileno, missing `fileno`, ordinary and async flush errors, `dir()` surface) matches `python3.14` line for line.

— *commented by Claude*

PR #915, opened 2026-07-31
majit: carry _immutable_fields_ through ullbc and fold w_class past forcing (append loop 45 → 31 ops)

youknowone:
Six commits on the append-loop class-identity cluster (`for i in range(n): r.append(i)`), ending with a build-reproducibility fix without which the last two are a coin flip.

Steady loop body over the series: **45 → 43 → 37 → 31 ops**.

## What lands

**`majit: record class_now_known alongside new_object in trace_box_int/float`**
`execute_new_with_vtable` (`pyjitpl.py:2714-2718`) records both `heapcache.new(resbox)` and `heapcache.class_now_known(resbox)`; only plain `execute_new` (`:2720-2723`) stops at `new`. `trace_box_int` / `trace_box_float` emit `NewWithVtable` but recorded only `new_object` and took the type address as an ignored parameter. Measured perf-neutral — the optimized-trace op sequence is byte-identical.

**`majit: make inline_short_preamble's final _map_args strict on unmapped args`**
`unroll.py:364-370 _map_args` is strict for non-Const boxes; the closing `_map_args` at `unroll.py:438-439` used `mapping.get(..).unwrap_or(jump_arg)`, handing the JUMP the preamble's own box — the loop-entry value, re-supplied on every back edge. Treat the miss as the `InvalidLoop` the KeyError amounts to. Measured inert: fires 0 times across the corpus on both backends.

**`majit: make optimize_getfield's field cache lookup get-or-create`**
`heap.py:648 cf = self.field_cache(descr)` is get-or-create (`heap.py:392-397`), so `getfield_from_cache` — and its `PreambleOp` arm — runs for the first touch of a descr too. `optimize_getfield` looked the cache up read-only and skipped the whole consult when absent, which is every field at the head of the peeled loop. Also fixes the `TestDescr` / `ImmutableDescr` fixtures, which left `index_in_parent` at 0 so every fixture descr mapped to ptr_info slot 0 of the same parent — a state `info.py:212-214` cannot represent.

**`majit: fold the w_class header read from a forced allocation's kept class`** (43 → 37)
`info.py:152 force_box` keeps the SAME info object and only clears `_is_virtual`, so `info.py:324 get_known_class` still answers after the allocation escapes. pyre installs `PtrInfo::Instance { descr, known_class }` on force but the consumer arm was missing, so the class became unusable the moment the allocation escaped. The stored-write probe goes through the layout's own `w_class` field descr, never the read descr's index — a parent-bearing `w_class` spelling has `index_in_parent == 0` and would alias the first value field.

**`majit: carry #[jit_immutable_fields] through ullbc and declare PyType's slots`** (37 → 31)
The attribute was a pure pass-through. Like every other `majit_macros` attribute it is consumed at expansion time and does not survive in Charon's `attr_info`, so `SemanticProgram.immutable_fields` stayed empty for every build and **no struct's declaration ever reached** `CallControl::field_immutability`, `StructFieldLayout::is_immutable`, or `BhDescr::Field::is_immutable` — `pyre-object`'s one existing declaration (`RBigInt`, matching `rbigint.py:157`) was inert.

Routed through the carrier the other attributes already use: the macro leaves `#[doc(hidden)] const _immutable_fields_<Struct>: &str` next to the struct, Charon extracts it into `global_decls`, and `harvest_immutable_fields_from_llbcs` reads it back — the analog of the translator reading `cls._immutable_fields_` off the class object.

`PyType`'s four slots are then declared immutable: `rclass.py:173-174` gives OBJECT_VTABLE `hints={'immutable': True, 'static_immutable': True}` and `lltype.py:372-374 _immutable_field` answers True for every field of a struct carrying that hint, so each vtable slot reaches `descr.py:231`. The walker's ConstPtr fast path (`pyjitpl.py:876-880`) then folds the read during the walk instead of recording it.

**`majit: give every struct-name spelling the struct's _immutable_fields_ entries`**
`struct_fields` carries more than one name per struct — the bare leaf (`PyType`) and the crate-qualified path (`pyre_object::pyobject::PyType`) — while the harvester registers only the leaf and the crate-stripped path. The layout loop iterates a `HashMap` and stores one layout per `StructId`, so the undeclared spelling could overwrite the declared spelling's layout and drop `is_immutable` from it, and which one won depended on iteration order.

`PyType.*` and `RBigInt.*` therefore reached `bh_all_field_specs_for_struct_into` as mutable on some builds and immutable on others, while the standalone field descrs for the same fields always carried the flag. **Three forced codegen runs at one unchanged source state produced 37 / 31 / 31 ops.** After the fix, six runs all produce 31.

## Verification

- `check.py --backend dynasm,cranelift`: **dynasm 348/348, cranelift 348/348**
- `cargo test --release`: majit-metainterp `--features dynasm` 1425, pyre-jit-trace `--features dynasm` 313, pyre-object 3048, majit-macros 34, majit-translate 105+ — 0 failed except the pre-existing red below
- Steady loop body re-measured on this base: **31 ops**, and `all_descrs=4038` agrees across the two independently-run codegen pipelines (dynasm and cranelift have different `codegen_cache_key`s, so this is two full runs converging, not one cached result)
- Class-identity regression (run out of band, since a wrongly skipped class check is a silent miscompile): a hot append loop interleaving plain ints with a `MyInt(int)` subclass reads the same sum and preserves `type(x) is MyInt` under JIT and `PYRE_NO_JIT` on both backends; `synth/callee_store_global_read_after_call` stays at 240008 on both backends and under `PYRE_NO_JIT`.

### Pre-existing red, not from this branch

`majit-translate`'s `slow_generated_jitcodes_preserve_complete_dispatcher_graph` fails at
`test_make_jitcodes_produces_graph_keyed_output.rs:255` with `left: 115, right: 111`. That operand
is produced by `syn`-parsing `pyre/pyre-interpreter/src/pyopcode.rs` and comparing against a
hardcoded arm count. Both that file and the test are byte-identical to `origin/main` in this branch
(`git diff origin/main --name-only` lists neither), so it fails the same way on the base. The
constants were last updated 2026-07-16 (#573) while `pyopcode.rs` kept gaining dispatcher arms
through 2026-07-29 (#654, #814) — belongs to whoever added them.

## Not the destination

Upstream never traces a typeptr read at all: `jtransform.py:833-834 rewrite_op_getfield` routes it to `handle_getfield_typeptr` (`:1004-1009`), which yields a codewriter-time `Constant(cls)` for a constant receiver and `guard_class` otherwise, and `rewrite_op_setfield` (`:909-911`) deletes typeptr stores outright. This series is convergence toward that codewriter shape.

The remaining `GetfieldGcR(list, w_class)` + `GuardValue` / `GetfieldGcI(type, _version_tag)` + `GuardValue` pair in the 31-op body is blocked on two separate things: `ensure_ptr_info_arg0` builds a parent-less `PtrInfo::instance(None, None)` for `is_w_class()` and never calls `init_fields`, so `produce_short_preamble_ops` (`info.py:255-271`) has no `_fields` to export; and the version-tag read sits on a `ConstPtr`, whose `ConstPtrInfo` upstream exports nothing to the short preamble either — the orthodox route there is `_version_tag?` quasi-immutable (`typeobject.py:177`), not hoisting.

— *authored by Claude*

### DIFF
--- a/majit/majit-macros/src/lib.rs
+++ b/majit/majit-macros/src/lib.rs
@@ -1523,21 +1523,75 @@ fn parse_release_gil_save_err(attr: proc_macro2::TokenStream) -> syn::Result<i32
 ///
 /// Usage:
 /// ```ignore
-/// #[majit_macros::jit_immutable_fields(pools)]
+/// #[majit_macros::jit_immutable_fields("pools", "size?", "digits[*]")]
 /// pub struct Storage {
 ///     pub pools: [*mut Stack; STORAGE_COUNT],
 ///     ...
 /// }
 /// ```
 ///
-/// The proc-macro is a pass-through: it leaves the struct definition
-/// untouched and exists solely so `rustc` accepts the attribute. The
-/// codewriter front-end reads the struct's field list from the lowered
-/// MIR type metadata (`majit-translate::front::mir` `struct_field_attrs`)
-/// and feeds it into the struct layout / descr pipeline.
+/// Entries carry the `rclass.py:644-678 _parse_field_list` suffixes
+/// (`?` quasi-immutable, `[*]` immutable array); a bare identifier is
+/// accepted as shorthand for a plain immutable field.
+///
+/// The struct definition itself is left untouched.  Like the other
+/// `majit_macros` attributes, the attribute is consumed at expansion
+/// time and does not survive in Charon's `attr_info`, so the macro
+/// leaves a `#[doc(hidden)]` marker const
+/// `_immutable_fields_<Struct>: &str` next to the struct holding the
+/// comma-joined entry list.  Charon extracts it into `global_decls`,
+/// and `majit-translate::front::llbc_hints::
+/// harvest_immutable_fields_from_llbcs` reads it back into
+/// `SemanticProgram.immutable_fields` — the analog of RPython's
+/// translator reading `cls._immutable_fields_` off the class object.
 #[proc_macro_attribute]
-pub fn jit_immutable_fields(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    item
+pub fn jit_immutable_fields(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let entries = parse_macro_input!(attr as ImmutableFieldList);
+    let item_struct = parse_macro_input!(item as syn::ItemStruct);
+    let vis = &item_struct.vis;
+    let const_name = format_ident!("_immutable_fields_{}", item_struct.ident);
+    // `rclass.py:644-678 _parse_field_list` splits the declared list into
+    // (name, rank) pairs; the marker carries the list verbatim and the
+    // harvester does the splitting, so the suffix grammar stays in one
+    // place (`majit_translate::model::ImmutableRank::parse`).
+    let joined = entries.0.join(",");
+    quote! {
+        #item_struct
+
+        #[doc(hidden)]
+        #[allow(non_upper_case_globals, dead_code)]
+        #vis const #const_name: &'static str = #joined;
+    }
+    .into()
+}
+
+/// `_immutable_fields_` entry list as written in the attribute —
+/// either string literals (`"size?"`, `"digits[*]"`) or bare
+/// identifiers for plain immutable fields.
+struct ImmutableFieldList(Vec<String>);
+
+impl Parse for ImmutableFieldList {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut entries = Vec::new();
+        while !input.is_empty() {
+            if input.peek(syn::LitStr) {
+                entries.push(input.parse::<syn::LitStr>()?.value());
+            } else {
+                entries.push(input.parse::<Ident>()?.to_string());
+            }
+            if input.is_empty() {
+                break;
+            }
+            input.parse::<Token![,]>()?;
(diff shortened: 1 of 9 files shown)


youknowone:
The Codex P1 on `virtualize.rs` was correct, and this merged before it could be acted on — so the unsound fold is currently on main. Fixed in #920.

It was worse than the report described. `force_box` installs the forced instance with an **empty** field list (`info.rs:1113-1118`, deliberately, so heap's `do_setfield` cannot MUST_ALIAS-elide the materializing SETFIELD_GC) and emits the header write as an ordinary `SetfieldGc` that lands in `CachedField::lazy_set`. Since OptVirtualize runs before OptHeap, the `Instance` arm saw an empty slot on *every* forced instance: the recorded-value branch was dead code, and the whole 43 → 37 op win came from the unsound fallback. A retag to a user subclass would have read back as the base class.

`w_subclass2.py` does not cover that shape — it passed the entire time. #920 adds `retag_force.py`, which reads the class both before and after the forcing store.

The parity review's §2 on `unroll.rs` I'd reclassify as §4: #556 already ported upstream's own answer for an unmapped jump arg in this function (`unroll.py:425-434` defers to the outer replay loop rather than raising), and the same carrier is used 25 lines earlier for `optimizer.flush()`.

— *commented by Claude*

PR #917, opened 2026-07-31
gc: re-derive shadow-stack slots across visitor calls, and guard a walk in debug

youknowone:
Follow-up to the Codex review on #909. Acts on the finding, but not with the
attribution the finding gave it.

## What the review said, and what is actually true

> When a `walk_shadow_stack` visitor re-enters any safe shadow-stack API […]
> The previous `RefCell` implementation deterministically rejected this

The mechanism is right. `walk_shadow_stack` held `stack.iter_mut()` across
`visitor(slot)`, and `pin_root`'s `Vec::push` can reallocate underneath it.

The attribution is not. `walk_shadow_stack` is dead outside tests — its only
non-test caller, `pyre_object_root_walker`, is never registered. The walker that
actually runs is `walk_shadow_stack_area`, and before #909 that one read through
`RefCell::as_ptr()`, which bypasses the borrow flag exactly as completely as an
`UnsafeCell` does. So on the live path #909 removed no check; there was none.

What the investigation did turn up is worse than the reported regression: the
live walker had no re-entrancy check in **any** build, before or after.

## Reachability

Not reachable today. Following both visitor chains to their leaves — including
the `DictStrategy::walk_gc_refs` dyn call, whose implementors are a closed set
of raw-storage-plus-visitor bodies, and the collector-side closures
(`drag_out_root`, `enumerate_root_walker_values`) — finds no call back into any
`gc_roots` API. `majit-gc` cannot even name `pyre_object::gc_roots`; its only
dependencies are `indexmap`, `majit-ir` and `libc`. Three independent attempts
to refute that — from the collector-callback, test-tooling and
indirect-mutation (finalizer / nested collection / `RootScope::drop`) angles —
each failed to produce a chain.

`#868` landing in the base does not change this: `execute_finalizer_triggers`
fires once at the end of a major step, after the root walk, and only notifies
the death deques — "pyre has explicit death deques but no collector-run
`execute_finalizers` phase". No Python runs inside a walk.

So this is defence in depth, and the change is scoped to match.

## Change

- Both walkers re-derive the slot address and re-read the length after every
  visitor call, so the walk no longer depends on the storage staying put.
- A debug-only per-thread walk flag, asserted in `pin_root`,
  `shadow_stack_set` and `RootScope::drop`, covering both walkers. It has to be
  a separate flag from `SHADOW_STACK_ACCESS_DEPTH`: once the exclusive access
  no longer spans the visitor call, the depth counter would see nothing.
- `RootScope` held a bare `usize` and was therefore auto-`Send`; dropping one on
  another thread truncates that thread's stack to an unrelated save point. Now
  `!Send` via `PhantomData<*const ()>`.

The hot read path — `shadow_stack_get`, `shadow_stack_len`,
`shadow_stack_copy_range` — is untouched, so none of #909's measured 9.8% is at
stake.

## What this does NOT fix

It narrows the hazard; it does not remove it. The `&mut PyObjectRef` handed to a
visitor still points into the buffer for the duration of that call, so a visitor
that pins and then writes through its own slot is still unsound. That is
intrinsic to walking a growable container.

Passing the value by copy and writing it back would close it, and I rejected
that: the collector threads the real slot address into its diagnostics
(`copy_nursery_object(..., slot_addr)`), and "GC BUG: traced slot contains
nursery poison at slot_addr=…" would then point at a stack temporary.

Upstream does not have the problem at all. `shadowstack.py:344-349` sizes the
root stack once at `root_stack_depth` (`:281`) and `incr_stack` (`:80-84`) is a
bare pointer bump with no bounds test — a push never moves the storage.
Converging on that fixed-capacity shape is the real close; it carries a
per-thread memory trade-off (163840 entries ≈ 1.3 MB) that deserves its own
change and its own measurement.

## Parity correction after review

The first commit re-read `len()` each iteration, which extended a walk over
roots the visitor itself pinned. The Codex parity review flagged that as a
regression and it was right: `walk_stack_root` (`shadowstack.py:43-46`) takes
`start` and `addr` as arguments and runs `while addr != start`, so the interval
is fixed when the walk begins.

My justification for diverging had been that a root pinned mid-walk would
otherwise go unforwarded and dangle. That argument does not survive contact
with the rest of this PR: pinning during a walk is exactly what the walk guard
here declares illegal, so the divergence protected a case this change makes a
debug panic. The second commit fixes the interval at entry and keeps the
per-iteration slot re-derivation, which is what actually defends against a
reallocation.

It also sharpens the release test. Reading `0x22` correctly now happens on the
iteration *after* the visitor forced the reallocation, so that assertion is the
use-after-free discriminator on its own; previously the property was buried in
a length check.

## Verification

`gc_roots` debug tests 7 passed (including a new `should_panic` test that the
walk guard actually fires), release-only reallocation test 1 passed — confirmed
it genuinely runs rather than being compiled out — `pyre-object` 283 passed,
`cargo check -p pyrex --features dynasm` clean (this is the blast radius of the
`!Send` change), `gc_stress` 23 passed / 0 failed.

Those counts are from the first commit. `origin/main` advanced to #868
mid-session and the branch was rebased onto it (the `gc_roots.rs` blob is
byte-identical across the rebase), and the parity correction above landed
afterwards, so the whole gate is re-running on the new base against the final
commit. I will post the result.

### DIFF
--- a/pyre/pyre-object/src/gc_roots.rs
+++ b/pyre/pyre-object/src/gc_roots.rs
@@ -454,17 +454,29 @@ unsafe fn walk_shadow_stack_cell(
     cell: *const UnsafeCell<Vec<PyObjectRef>>,
     visitor: &mut impl FnMut(&mut PyObjectRef),
 ) {
+    // The interval is fixed at entry: `walk_stack_root` (`shadowstack.py:43-46`)
+    // receives `start` and `addr` as arguments and runs `while addr != start`,
+    // so a root pushed mid-walk is not part of that walk. Re-reading the length
+    // each iteration instead would extend the walk over roots pinned by the
+    // visitor — and pinning during a walk is precisely what the debug guard
+    // above declares illegal, so there is nothing to serve by diverging here.
+    let end = {
+        // SAFETY: no reference from this block outlives it.
+        let stack = unsafe { &*(*cell).get() };
+        stack.len()
+    };
     let mut index = 0;
-    loop {
+    while index < end {
+        // Re-derive the slot address every iteration rather than holding one
+        // `iter_mut()` cursor: a re-entrant push can still reallocate the
+        // buffer, which would strand a cached pointer in the old allocation.
+        // Only pushes can occur, so `end` stays within bounds.
         let slot = {
             // SAFETY: the caller guarantees exclusive access except for a
             // possible re-entrant mutation by `visitor`. No reference from a
             // previous iteration is live here, and this temporary reference
             // ends before `visitor` is called.
             let stack = unsafe { &mut *(*cell).get() };
-            if index >= stack.len() {
-                break;
-            }
             // Return only the slot address so the `Vec` reference is not held
             // across the visitor, which may push and reallocate its buffer.
             unsafe { stack.as_mut_ptr().add(index) }
@@ -677,9 +689,15 @@ mod tests {
         });
 
         assert!(pushed);
+        // The visitor's pins really did move the buffer.
         assert!(with_shadow_stack(Vec::capacity) > capacity_before);
-        assert_eq!(&seen[..2], &[0x11, 0x22]);
-        assert_eq!(seen.len(), 2 + additional);
+        // `0x22` is the discriminator: it is read on the iteration AFTER the
+        // reallocation, so getting it right means the walk followed the buffer
+        // to its new address instead of reading the freed one. The walk stops
+        // at the two roots that were live at entry — the interval is fixed
+        // there, as in `walk_stack_root` (`shadowstack.py:43-46`), so the
+        // roots the visitor pinned are not part of this walk.
+        assert_eq!(seen, vec![0x11, 0x22]);
         assert_eq!(shadow_stack_get(0) as usize, 0x11);
         assert_eq!(shadow_stack_get(1) as usize, 0x22);
     }
--- a/pyre/pyre-object/src/gc_roots.rs
+++ b/pyre/pyre-object/src/gc_roots.rs
@@ -51,6 +51,7 @@
 #[cfg(debug_assertions)]
 use std::cell::Cell;
 use std::cell::UnsafeCell;
+use std::marker::PhantomData;
 
 use crate::pyobject::PyObjectRef;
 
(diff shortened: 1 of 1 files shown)


youknowone:
Second commit pushed: the parity correction, plus the full gate re-run against it on the new base.

## The review changed the design, not just the wording

The first commit re-read `len()` on every iteration so that a root pinned by the
visitor joined the same walk. I had argued for that in the description on safety
grounds — a root pinned mid-walk would otherwise go unforwarded and dangle.

The Codex parity review called it a regression, and checking the source settles
it. `walk_stack_root` (`shadowstack.py:43-46`) takes `start` and `addr` as
arguments and runs `while addr != start`, walking downward: the interval is
fixed when the walk begins, and a push cannot be observed by a walk in progress.

The flaw in my argument was self-inflicted. The case it protected — pinning
during a walk — is exactly what the walk guard *in this same PR* declares
illegal and panics on in debug. Diverging from upstream to serve a case the
patch itself outlaws is backwards.

`da7e32ff8` fixes the interval at entry and keeps the per-iteration slot
re-derivation, which is the part that actually defends against a reallocation.
The two were independent all along.

It also sharpens the release test instead of weakening it. `0x22` is now read on
the iteration *after* the visitor forces the reallocation, so that one assertion
is the use-after-free discriminator; previously the property was buried inside a
length check.

## Verification, all of it re-run against `da7e32ff8` on the new base

| | |
|---|---|
| `gc_roots` debug (incl. the `should_panic` walk-guard test) | 7 passed |
| release-only reallocation test | **1 passed** — confirmed it runs, not filtered out |
| `pyre-object` full | 293 passed |
| `cargo check -p pyrex --features dynasm` (the `!Send` blast radius) | `Finished`, no errors |
| `gc_stress` | **23 passed, 0 failed** |

The earlier gate was discarded rather than reported: it was building while the
parity edit landed, so its result was neither pre-edit nor post-edit. This one
ran start to finish on a clean tree at the committed sha.

## Note on this PR's own review coverage

The `codex-review` check going green does not mean a parity report exists — the
job is named "Queue Codex parity review" and only the queueing succeeded. #909,
#910 and #911 were all green with no report at all after the runner's codex
token expired, and the last two merged that way. The review acted on above came
from running it locally against the real merge-base; the skill's default
`upstream/main` was 19 commits stale and would have diffed other people's merged
work into this patch.

— *commented by Claude*

PR #920, opened 2026-07-31
majit: resolve a forced instance's w_class in optimize_getfield, not virtualize (#915 P1)

youknowone:
Follow-up to #915, which merged before its Codex review could be acted on. The P1 raised there is **real**, and worse than the report described — the fold it names is live on main right now.

## The defect

#915's `f20f9925c0` extended the `w_class` header fold from `PtrInfo::Virtual` to `PtrInfo::Instance`, so the class stayed usable after an allocation escaped. But `virtualize.py:184-195 optimize_GETFIELD_GC_*` folds only under `opinfo.is_virtual()` and otherwise emits, leaving a non-virtual's read to `heap.py`. That layering is load-bearing for this field:

- `force_box` installs the forced instance with an **empty** field list (`info.rs:1113-1118` — deliberately, so heap's `do_setfield` cannot MUST_ALIAS-elide the materializing SETFIELD_GC),
- and emits the header write as an ordinary `SetfieldGc`, which lands in `CachedField::lazy_set`,
- while OptVirtualize runs a pass **earlier** than OptHeap.

So the `Instance` arm saw an empty slot on *every* forced instance and answered from the size descr's canonical class. Two consequences: the recorded-value branch was effectively dead code, and **the entire 43 → 37 op win came from the unsound fallback**. An instance retagged to a user subclass and then forced would read back as the base class — a silent miscompile of a class check, not a crash.

## The fix

Both resolutions move into `optimize_getfield`, after the cache consult: a value already in the instance's own slot wins, and only when the header was never written does the layout's canonical class answer — gated on no `w_class` store still being pending.

The pending check scans **every** `is_w_class()` descr rather than the one being read. The caches are keyed by `Arc::as_ptr` (`heap.rs cached_field_pos_for_descr`) and pyre mints more than one descr for this header (a walker singleton with no parent, and the size descr's own `all_fielddescrs` entry), so a miss on the read descr does not prove the field is untouched — the store may sit in the other spelling's `lazy_set`.

The slot is still looked up through the layout's own `w_class` fielddescr, never the read descr's index: a parent-bearing spelling has `index_in_parent == 0`, which is the first *value* field, and folding through it forwards `Ref <- Int`.

### One wrong turn worth recording

The first gate also refused on `!cached_structs.is_empty()` and dropped the loop to **40 ops**. Plain *reads* fill `cached_structs` via `register_info`, so the first `w_class` read blocked every later fold. Only `lazy_set` distinguishes a pending write; a store that has already been flushed is readable from the instance slot instead. Gating on `lazy_set` alone restores 31 ops.

## Regression coverage

`w_subclass2.py` does **not** cover this shape — it passed throughout the unsound window. Added `retag_force.py`: `MyTuple`/`MyInt` interleaved with their base types, class read both **before and after** the forcing store.

```
NO_JIT (oracle) : 25000 50000 1249975000 tuple MyTuple
dynasm JIT      : 25000 50000 1249975000 tuple MyTuple
cranelift JIT   : 25000 50000 1249975000 tuple MyTuple
```

## Also here

`majit: exit the short-preamble force loop on a non-growing arg list` — the loop reads `grown[num_short_jump_args..]` immediately after testing `grown.len() == num_short_jump_args`, so a list that ever came back shorter would slice out of bounds instead of taking the "did not grow" exit. Raised on #556, which added the surrounding guard.

## Verification

- `check.py --backend dynasm,cranelift`: **dynasm 349/349, cranelift 349/349**
- `cargo test --release -p majit-metainterp --features dynasm`: 1425 passed, 0 failed
- Append loop steady body stays at **31 ops**; `w_subclass2.py` and `synth/callee_store_global_read_after_call` (240008) unchanged on both backends and under `PYRE_NO_JIT`

## Not addressed

The parity review's §2 on `unroll.rs` (`signal_invalid_loop` where upstream's `mapping[box]` raises `KeyError`) is left as-is. #556 already ported upstream's own answer for an unmapped jump arg in this function — `unroll.py:425-434`, which **defers to the outer replay loop** rather than raising — and the same carrier is used 25 lines earlier for `optimizer.flush()`. That reads as a documented structural adaptation (§4), not a new divergence.

— *authored by Claude*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved optimization of class-field reads by reusing known stored values when safe.
  - Preserved unresolved or pending class-field reads for correct later processing.
  - Fixed short-preamble handling when jump arguments shrink, preventing invalid processing.
  - Corrected forced object reads so they are handled safely during subsequent heap processing.

### DIFF
--- a/majit/majit-metainterp/src/optimizeopt/heap.rs
+++ b/majit/majit-metainterp/src/optimizeopt/heap.rs
@@ -861,6 +861,26 @@ impl OptHeap {
         }
     }
 
+    /// Is a `w_class` store still pending under ANY descr spelling?
+    ///
+    /// The field caches are keyed by `Arc::as_ptr` identity
+    /// (`cached_field_pos_for_descr`), and pyre mints more than one descr for
+    /// the `w_class` header (a walker singleton with no parent, and the size
+    /// descr's own `all_fielddescrs` entry). A miss on the descr being read
+    /// therefore does not prove the field is untouched — the store may sit in
+    /// the other spelling's `lazy_set`. Asking across every spelling costs a
+    /// fold, never correctness.
+    ///
+    /// Only `lazy_set` is consulted: a store that has already been flushed
+    /// (`force_lazy_set` → `put_field_back_to_info`) is readable from the
+    /// instance's own field slot, and `cached_structs` also fills up from plain
+    /// reads, which say nothing about the field having been written.
+    fn w_class_set_pending(&self) -> bool {
+        self.cached_fields.iter().any(|(_, descr, cf)| {
+            descr.as_field_descr().is_some_and(|fd| fd.is_w_class()) && cf.lazy_set.is_some()
+        })
+    }
+
     fn cached_field_pos_for_descr(&self, descr: &DescrRef) -> Option<usize> {
         let identity = descr_identity(descr);
         self.cached_fields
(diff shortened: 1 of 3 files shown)


youknowone:
Both P1s and the CodeRabbit finding are correct, and this merged before they could be acted on. Addressed in #922.

They converge on one conclusion: the heap-side fold is still unsound in two distinct ways. When a store is pending under the other descr spelling, suppressing the fold only declines the constant — the `GETFIELD_GC` is then emitted ahead of the unflushed `SETFIELD_GC`, so the load reads the pre-retag word. And when a store has already been flushed, `put_field_back_to_info` slots it by `field_slot_index(write_descr)` — `Descr::index()` for the parentless walker spelling, not the `index_in_parent()` the lookup reads — so the value is invisible and `lazy_set` is empty by then.

Neither is patchable without stacking another special case on the descr split, and upstream has no counterpart to port: `jtransform.py:1004-1009 handle_getfield_typeptr` deletes this read at codewriter time. #922 removes both arms. Cost is 31 → 40 ops on the append loop — 9, not the 6 the original 43 → 37 measurement implied, so the fold was also carrying part of the `_immutable_fields_` win.

The `<=` on `unroll.rs` is reverted there too: `unroll.py:425` uses equality and the list can only grow, so it was an unreachable branch bought with a divergence in a line-by-line port.

— *commented by Claude*

Issue #108, opened 2026-05-27
Replace ConstPtrInline(GcRef) with GC-traced const refs (BoxRef + gc_table)


youknowone:
# Epic: ConstPtrInline(GcRef) → GC-traced const refs (BoxRef forwarding + per-loop gc_table)

> **Design correction (parity).** This epic was originally written around a
> `ConstRefHandle` / `ConstRefTable` indirection. That design is **not
> RPython-faithful** — RPython keeps no handle table for JIT const refs — so it
> was *not* pursued. The parity-correct solution (now implemented) is described
> under **Parity-faithful design** below; the original handle-table plan is kept
> at the bottom under **Superseded design (history)**.
>
> Status: the backend half landed in #190 (per-loop `gc_table` + `rgc.can_move`
> baking guard); the optimizer/IR half landed in #132 / #156 / #167 / #185
> (`Forwarded::Const` + `BoxRef` migration); the finalization (review fixes, the
> `CompiledTrace.constants` cleanup, and the moving-GC test) is in #213. Closing
> via #213.

## Problem

`OpRef::ConstPtrInline(GcRef)` carries a raw GC pointer inside a `Copy` type.
Every assignment or clone creates an independent `GcRef` copy. When the moving
GC forwards the object, only the storage slots reached by a root walker are
updated; all other copies become stale.

RPython's `ConstPtr` is a heap object. `op.args[i]` stores a reference to the
`ConstPtr` object, and the GC traces its `value` field through the normal object
graph. Pyre emulates this with explicit walkers over Rust containers, because
Rust `Vec`/`HashMap` storage is not GC-managed. The fragility the epic targets
is the **raw `GcRef`-inside-`Copy`** defect: a long-lived inline `ConstPtr`
that escapes walker coverage goes stale after a collection.

## RPython Parity Notes

`ConstPtr.same_constant` is value equality, not object identity. RPython can
allocate two distinct `ConstPtr(value)` boxes and still have them compare equal
by `same_constant`; `Const.same_box` also delegates to `same_constant` for
constant boxes. Do not describe const-ref identity as "same GcRef means same box
instance"; that is not a PyPy invariant. pyre's `BoxRef` const model preserves
value-equality semantics through `same_box`/`same_constant`, not box identity.

## Parity-faithful design (implemented)

There is no handle table. Const refs are made GC-safe the way RPython makes them
safe, split across the three places a `GcRef` can live:

[...]

PR #925, opened 2026-07-31
jit: guard the inlined callee's Function fields, and bound the dynasm/wasm bridge chain

youknowone:
A loop that defines a function in its own body and **calls** it deopted every
second iteration, forever. Two independent defects composed; both are fixed
here for the case that shape actually hits.

## 1. The inline lever guarded the function OBJECT

`inline_call.rs` emitted a single `GuardValue` on the callable's own address.
`MAKE_FUNCTION` allocates a fresh `Function` every iteration, so that guard can
only hold by address coincidence. It never does — 9480/9480 distinct pointers,
stepping ~+352 bytes of nursery bump.

Upstream never guards the instance. `function.py:34-42` declares
`_immutable_fields_ = ['code?', 'w_func_globals?', 'closure?[*]', 'defs_w?[*]']`
and `function.py:91-96 getcode()` promotes `self.code`. Those four names are
exactly what the inline bakes, so each is now read off the **live** function and
guarded there. `closure` is read through to its cells, because the tuple itself
is rebuilt by every `MAKE_FUNCTION` while the cells are stable.
`FUNCTION_DESCR_GROUP` grows `code`, `w_func_globals` and `closure` beside the
`defs_w` it already had.

`synth/call_loop_local_function`: **`guard_failures` 9480 → 1**,
`loops_compiled` 2 → 1. Baselines re-snapshotted on all three backends.

### Scope: a NON-constant callable only

The guards are emitted only when the callable is not a trace constant. That is
the fresh-`MAKE_FUNCTION` case — the whole point — but it is a real limit, for
two separate reasons found while landing this:

- **Loading through a baked `ConstPtr` is a GC hazard.** A baked constant object
  pointer is not GC-forwarded (gh #108 gc-table; `synth/exception_subclass_attrs.py`
  documents the same trap). Comparing against one is fine, dereferencing one
  dangles as soon as a minor collection moves the object.
  `synth/inline_subwalk_property_mutates` — a property getter that allocates
  every iteration — segfaults on cranelift on CI's macOS runner with the reads
  in place. With the restriction its jitstats are byte-identical to pre-change.
- **`callable` and `callable_guard_value` are different objects.** The
  exception-string specializer resolves `str(e)` to an exception subclass's
  `__str__` but passes the CALL's own operand — the `str` builtin — as the value
  to pin. Reading `Function.code` at offset 16 off a `PyCFunction` is a
  type-confused load; the guard compared a field that is not `code` and never
  matched, recreating this very storm (`synth/exception_subclass_attrs`:
  guard_failures 1 → 99480, 497 bridges, 0.19s → 6.18s). The emission now also
  requires that the pinned object be a `Function` whose `code` is the code this
  inline resolved.

**Consequently `f.__code__ = g.__code__` on a trace-constant callable remains a
miscompile** — a 40000-iteration loop reassigning `__code__` prints 41998
instead of 20020000, on `main` and on this branch. Re-checking it needs either
quasi-immutable `code?` with trace invalidation, or gh #108 so a baked constant
is dereferenceable at all. No perf gate is changed by this PR.

## 2. The bridge chain was unbounded on dynasm and wasm

`regalloc.py:496-499 consider_guard_value` calls
`descr.make_a_counter_per_value(op, index)` on **every** upstream backend
(`x86/regalloc.py:496-499`, `aarch64/regalloc.py:815-818`), so `must_compile`
hashes the (guard, failing value) pair (`compile.py:753-781`) and `store_hash`
skips an already-stamped descr (`compile.py:826-829`).

pyre had the helper and the metainterp decode, but called it **only from
cranelift**. dynasm and wasm left `status == 0` — one bucket per guard — so
`trace_eagerness = 200` fired forever and every 200th failure compiled another
bridge. `bridges == floor(guard_failures / 200)` held at every size: 480→2,
9484→47, 99487→497, 499489→2497.

| N | dynasm before | dynasm after | cranelift |
|---|---|---|---|
| 20 000 | 47 bridges, 0.23s | 0 bridges, 0.19s | 0 bridges |
| 200 000 | 497 bridges, 2.71s | 0 bridges, 0.77s | 0 bridges, 0.71s |

pyre's index is a fail-arg **position**, not a register index, because
`must_compile_with_values` reads `fail_values.get(index)`.

## 3. `MAJIT_NO_BRIDGE=1` was a lying control

`run_compiled_detailed_with_bridge_keyed` — the loop pyre actually reaches from
`execute_assembler` — computed `should_bridge` without consulting
`no_bridge_enabled()`, unlike its two sibling loops. It reported a bridge-free
run while still compiling all 47. It now consults the same opt-out and emits the
same `@@@GUARD` line the siblings do.

That line paid for itself immediately: it named the failing guard (`fail=17`)
in one run when the specializer type-confusion above turned up in CI. This loop
previously had no per-guard-failure trace at all.

## Also here

- `defs_w` keeps a tuple-**identity** guard. Only its length matters, but
  `GuardValue` over an `arraylen_gc` leaves the guard's only argument dead after
  it and the bridge compiled at that fail index segfaults on entry — reassigning
  `f.__defaults__` to a different length mid-loop is correct to ~1000 post-flip
  iterations and dies past ~2000, with and without `MAJIT_NO_BRIDGE`.
- The virtuals cache is rooted across `force_from_resumedata`'s materialization
  window: `getvirtual_ptr` hands back the cache slot precisely because a
  materialization collection is expected to forward it in place, so a bare
  `prepare` left every virtual written before a minor pointing into from-space.
- `call_loop_local_function` gets the `max-pypy-ratio=8` header #853's
  convention now expects (worst measured 3.7x; gate is `max(5, ceil(worst × 2))`).

## Verification

- `pyre/check.py` on the merged tree: dynasm **353/353**, cranelift **353/353**,
  wasm **349/349**.
- `cargo test --release --features dynasm` on pyre-jit-trace, majit-metainterp,
  majit-gc, majit-backend-dynasm — green; CI runs it on three OSes.
- 16 shape probes (fresh def, lambda, no-free-var, hoisted, never-called, bound
  method, fresh instance, partial application, `__globals__` / `__closure__` /
  `__defaults__` reassignment, exception `__str__` override) all match their
  `PYRE_NO_JIT=1` reference. `adv_code` (`__code__` reassignment on a constant
  callable) is the known exception described above.

Not fixed here: `types.FunctionType(f.__code__, ns)` at a hot call site aborts
with `GC BUG: invalid type_id ... site=minor_root_target`. It reproduces
identically before this branch, so it is a separate pre-existing defect.
`getattr_hook_binding` and `format_z_negative_zero` perf flakes on the macOS
runner are also pre-existing — `getattr_hook_binding` fails on `main`'s own CI.


youknowone:
CI caught a real regression in the first push, and it was mine. Fixed in `c80c413e25c`, with `origin/main` merged in so local gates measure what CI measures.

## `synth/exception_subclass_attrs`: 0.19s → 6.18s (31x)

`guard_failures` 1 → **99480**, `bridges_compiled` 0 → **497** — the exact storm signature this PR removes elsewhere, recreated by the fix for it.

It did not show up locally because that fixture's `max-pypy-ratio=160` gate landed in #853, which this branch did not have; CI builds the PR merge.

### Root cause

`try_walker_inline_resolved_user_call` takes two callable parameters that are **not the same object**:

- `callable` — the resolved callee `Function`
- `callable_guard_value` — whatever the caller wants pinned

The exception-string specializer (`inline_call.rs` ~:4305) resolves `str(e)` to an exception subclass's `__str__`, but passes the CALL's own operand — the `str` builtin — as the value to pin. Reading `Function.code` at offset 16 off a `PyCFunction` is a type-confused load, so the guard compared a field that is not `code` and could never match.

It only bit once the guards became unconditional: the old `!is_constant()` skip hid it, because these callers pass constants.

### Fix

Gate on the invariant rather than on the caller — emit the `Function.*` field guards only when the pinned object is a `Function` whose `code` is the code this inline resolved, otherwise fall back to the operand-identity guard.

| | before fix | after fix | baseline |
|---|---|---|---|
| exception_subclass_attrs | 6.18s, 497 bridges, 99480 guard failures | **0.14s, 0 bridges, 1** | 0.19s, 0, 1 |
| call_loop_local_function | 1 guard failure | 1 guard failure | 9480 |

Note it ends up *faster* than baseline — the `__str__` inline benefits from the field guards once they are applied to the right object.

The `defs_w` block reads off the same operand and has the same latent confusion, but is unreachable there: the specializer paths call with exact arity, so `positional_defaults` is `None`.

The `@@@GUARD` line this PR adds to `run_compiled_detailed_with_bridge_keyed` is what found it — it named the failing guard (`fail=17`) in one run. That loop previously had no per-guard-failure trace at all.

## Also

- `call_loop_local_function` gets the `max-pypy-ratio=8` header it was missing under #853's convention (worst measured 3.7x; gate is `max(5, ceil(worst × 2))`).
- Gates re-run on the merged tree: dynasm **353/353**, cranelift **353/353**, wasm **349/349**.

— *commented by Claude*

youknowone:
**Correction to the earlier comment and to `e5404cc02f2`'s message: the `f.__code__` re-check is NOT in this PR.** Pushed `1f6cedde9b7`; PR body updated.

## macOS CI found a GC hazard the field guards introduced

`synth/inline_subwalk_property_mutates` **SIGSEGVs on cranelift** on the macOS runner — deterministic there across both runs, and not reproducible locally in 60 runs, under `MAJIT_GC_STRESS=1`, under `MAJIT_GC_NURSERY_POISON=1`, or in a full 353/353 local gate.

The patch made exactly one structural change to that fixture's path: the field guards **dereference a baked `ConstPtr`**. A baked constant object pointer is not GC-forwarded (gh #108 gc-table — `synth/exception_subclass_attrs.py` documents the same trap in its own comment). Comparing against such a constant is fine; loading through one dangles as soon as a minor collection moves the object, and that fixture is a property getter that allocates every iteration.

Fix: restrict the guards to a **non-constant** callable. Evidence it is the right cut — that fixture's jitstats are now byte-identical to the pre-change run (1 bridge / 116 aborts / 23784 guard failures), i.e. the patch no longer touches the path.

## What this costs

`f.__code__ = g.__code__` on a trace-constant callable is **still a miscompile** — `adv_code` prints 41998 instead of 20020000, on `main` and here. `e5404cc02f2`'s message claimed that fix; it is withdrawn. Re-checking it needs quasi-immutable `code?` with trace invalidation, or gh #108 so a baked constant is dereferenceable at all.

## What it buys

- **The gate raise is gone.** `inline_helper`'s cranelift vs-pypy gate is back to 1.5 — its callees are module globals (trace constants), so they get no guard and it measures 1.1x again. This PR now changes no perf gate.
- **The storm fix is untouched.** A fresh `MAKE_FUNCTION` callee is a live residual result, never a trace constant. `call_loop_local_function` keeps `guard_failures=1`; `exception_subclass_attrs` keeps 0 bridges.

## The other macOS/ubuntu failures are not from this branch

- `getattr_hook_binding` (6.4x/6.7x vs gate 5x) — **fails on `main`'s own CI run** (job 91166301304), along with `type_metatype_data_descr`.
- `format_z_negative_zero` (8.0x vs 5x) and ubuntu's `type_immutable_reject` (154x vs a 152x gate, a 1.3% margin) are the same 0.01s-denominator flake family. Not touching those gates.

Gates re-run on the corrected tree: dynasm **353/353**, cranelift **353/353**, wasm **349/349**.

— *commented by Claude*

PR #931, opened 2026-07-31
jit-trace: bridge the codewriter's PyObject.w_class read to the walker descr (append loop 40 → 33 ops)

youknowone:
Recovers most of what #922 gave up, and this time by deleting a duplicate read rather than folding past a pending write.

## The duplicate

`make_descr_from_bh` already redirects a codewriter-lowered body's field descr to the walker's canonical one — for `W_ListObject.int_items.*`, `ItemsBlock.capacity` and the box payloads. The reason is stated in that function: the heapcache and the optimizer's heap pass both key on descr identity, so a field carrying two descrs silently breaks aliasing between them. *"One field is one descr."*

The shared `PyObject` header's `w_class` had no such arm, and it is read from both sides in the same loop:

- `orthodox_list_append_commit` (`specialize.rs:6446`) pins the appended value's class — reads offset 8 through `w_class_descr()`, guards it to a constant, `replace_box`.
- the sub-walk then evaluates `is_plain_int1(value)` (`pyre-object/src/listobject.rs`), whose `value.w_class` read comes through the modelled `PyObject` parent — **a different identity for the same field**.

So the pinned constant never reached the second read. In the steady body that showed up as the load appearing twice, the second followed by its own null test and equality test:

```
v233 = GetfieldGcR(v231) descr=<PyreFieldDescr { offset: 8 …     ← walker
GuardValue(v233, ptr(0x102954938))
v235 = GetfieldGcR(v231) descr=<SimpleFieldDescr { index: 1 …    ← layout entry
v236 = PtrEq(v235, ptr(0x0));        v237 = IntIsTrue; GuardFalse(v237)
v238 = PtrEq(v235, ptr(0x102954938)); v239 = IntIsTrue; GuardTrue(v239)
```

## The fix

Bridge `("PyObject" | "pyre_object::pyobject::PyObject", "w_class")` to `w_class_descr()`.

Placed **ahead of** the parent-group lookup, for the reason the neighbouring `int_items.*` arm documents: when the codewriter does model the parent, that lookup answers with the parent's own entry for the same offset and re-creates the split. Putting the arm in the later `(owner, name)` match — where the box-payload bridges live — measured no change at all, because the parented path returns first.

## Result

Append loop steady body **40 → 33 ops**, so 7 of the 9 #922 gave up.

The remaining pair is the single genuine read and its guard. `guard_class` cannot replace it: a builtin subclass instance shares the payload `ob_type` and passes `guard_class`, retagging only `w_class` (`walker_frame_ops.rs:196-200`). Removing it needs the object-model split behind `builtin subclass shares the base layout`, not another descr change.

## Verification

- `check.py --backend dynasm,cranelift`: **dynasm 352/352, cranelift 352/352**
- `cargo test --release -p pyre-jit-trace --features dynasm`: 314 passed, 0 failed, including a new test asserting both owner spellings bridge to the same `Arc`
- `retag_force.py` matches the `PYRE_NO_JIT` oracle on both backends (`25000 50000 1249975000 tuple MyTuple`); `w_subclass2.py` and `synth/callee_store_global_read_after_call` unchanged
- Re-measured on the current base after the branch was rebased mid-work, not on the tree the first measurement used

— *authored by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of object class descriptors for supported object layouts.
  * Ensured compatible descriptors are reused correctly while preserving distinct descriptors when their sizes differ.

* **Tests**
  * Added coverage for descriptor reuse and size-mismatch behavior.

### DIFF
--- a/pyre/pyre-jit-trace/src/descr.rs
+++ b/pyre/pyre-jit-trace/src/descr.rs
@@ -3628,33 +3628,66 @@ mod tests {
         }
     }
 
-    /// The shared `PyObject` header's `w_class` bridges to the same descr the
-    /// walker pins a value's class through, so a codewriter-lowered subclass
-    /// test (`is_plain_int1`) reads the header the walker already guarded
-    /// instead of emitting a second, uncacheable read of offset 8.
-    #[test]
-    fn make_descr_from_bh_bridges_pyobject_w_class_to_the_walker_descr() {
+    /// A `PyObject.w_class` `BhDescr` describing the same access as the
+    /// canonical header descr, for both owner spellings the codewriter emits.
+    fn w_class_bh(owner: &str, field_size: usize) -> majit_translate::jitcode::BhDescr {
         use majit_ir::descr::ArrayFlag;
         use majit_translate::jitcode::BhDescr;
 
+        BhDescr::Field {
+            offset: pyre_object::pyobject::W_CLASS_OFFSET,
+            field_size,
+            field_type: Type::Ref,
+            field_flag: ArrayFlag::Signed,
+            is_field_signed: false,
+            is_immutable: false,
+            is_quasi_immutable: false,
+            // slot 0 is `ob_type`; `w_class` is slot 1 of the header.
+            index_in_parent: 1,
+            parent: None,
+            name: "w_class".into(),
+            owner: owner.into(),
+        }
+    }
+
+    /// The shared `PyObject` header's `w_class` bridges to the same descr the
+    /// walker pins a value's class through, so a codewriter-lowered subclass
+    /// test (`is_plain_int1`) reads the header the walker already guarded
+    /// instead of emitting a second, uncacheable read of the same offset.
+    #[test]
+    fn make_descr_from_bh_bridges_pyobject_w_class_to_the_walker_descr() {
+        let canonical = w_class_descr();
+        let width = W_CLASS_FIELD_DESCR.field_size();
+
         for owner in ["PyObject", "pyre_object::pyobject::PyObject"] {
-            let descr = make_descr_from_bh(&BhDescr::Field {
-                offset: 8,
-                field_size: 8,
-                field_type: Type::Ref,
-                field_flag: ArrayFlag::Signed,
-                is_field_signed: false,
-                is_immutable: false,
-                is_quasi_immutable: false,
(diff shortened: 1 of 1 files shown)


youknowone:
The P1 is correct — and the tree already documented the trap it names.

`new_w_class_field_descr` (`descr.rs:1918-1938`) carries this on the very field in question:

> ⚠️ `WORD` on paper — the field is a `*mut PyObject`, so 4 bytes on wasm32, and the build-time descr pool already sizes it that way (`call.rs get_type_flag` → `layout::target_word_size()`). Deriving it here to match makes `synth/exception_traceback_loop_forms` lose one iteration's `e.__traceback__` on the wasm backend, so the two universes stay deliberately out of step until that is understood.

Verified both halves:

- `call.rs:7448-7452 get_type_flag` gives every pointer field `crate::layout::target_word_size()` — 4 on wasm32.
- `new_w_class_field_descr` hardcodes `field_size: 8`.

So on a 32-bit target the bridge replaced a 4-byte load with an 8-byte load at the same offset, pulling four bytes of the adjacent payload into the class pointer. Same family as the `arraydescrof_concrete` literal-8 fallback that mis-strided a pointer array on wasm32 and surfaced as an `invalid type_id` GC panic.

**Fix taken: the second of the two you offered.** Making the canonical descr target-layout-correct is the first option, and the comment above records that it was already tried and regressed `synth/exception_traceback_loop_forms` on wasm — so it is not a bounded change here. Instead the bridge now returns the canonical descr only when offset, width, and field type all match the incoming `BhDescr`:

```rust
let canonical = &*W_CLASS_FIELD_DESCR;
if canonical.offset() == *offset
    && canonical.field_size() == *field_size
    && canonical.field_type() == *field_type
{
    return w_class_descr();
}
```

Where the widths disagree the bridge declines and the target keeps exactly the descr it had before #931 — the documented "two universes out of step" state stays as-is rather than being silently merged.

Test coverage: the existing bridge test and a new decline test now both derive offset and width from the canonical descr instead of hardcoding `8`, so neither can go vacuously green if the layout moves.

Verification on this commit: `check.py` dynasm 352/352, cranelift 352/352; `cargo test --release -p pyre-jit-trace --features dynasm` green. The 64-bit win is unaffected — the append loop steady body stays at 33 ops, since the widths agree there.

— *commented by Claude*

PR #952, opened 2026-08-01
bench: commit class_attrs_methods jitstats baselines

fregataa:
## What

Commit `class_attrs_methods.{dynasm,cranelift}.jitstats` structural baselines
(plus their `pyre/.gitignore` exception line, following the existing
`comprehension_object_append_hot` / `const_arg_call_resume` /
`nested_list_comprehension_hot` pattern).

```
loops_compiled=1
loops_aborted=0
guard_failures=1
bridges_compiled=0
internal_compile_panics=0
...
```

## Why

`class_attrs_methods` is the `max-pypy-ratio=824` fixture — the loosest perf
gate in the tree, calibrated (#853) when its hot `c.add(i & 7)` method call was
declined by the method-form inline gate and ran as a full `CallMayForceR`
residual round trip per iteration (fresh callee `PyFrame`, warmstate dispatch
into a *separately compiled* function-entry trace via `execute_token`,
re-boxing — ~1.7s user on an M-series dynasm build vs 0.09s CPython).

Current main inlines the call again: `loops_compiled` 2 → 1, dynasm 0.07s /
cranelift 0.09s user. But nothing pins that trace shape — with an 824x
allowance, the callee residualizing again (and recompiling its own func-entry
loop) would sail through a bare `pyre/check.py`. The committed baselines make
check.py's jit-stats regression floor catch exactly that structural
regression, timing noise aside.

Baselines were recorded with
`pyre/check.py --synthetic-only --synthetic-pattern class_attrs_methods.py --snapshot`
and verified green on a bare re-run (dynasm 1/1, cranelift 1/1, 4.2x/6.2x vs
pypy). No wasm baseline: no wasm toolchain on this machine — can follow up
from a runner that has one, as the other baselined benches carry a wasm file.

## Summary by CodeRabbit

* **Chores**
  * Updated repository tracking configuration to retain selected benchmark result files.


youknowone (approved):
👍

PR #957, opened 2026-08-01
__pypy__: register reversed_dict/move_to_end/objects_in_repr as interp-level functions

youknowone:
Part of the appleveldefs→interpleveldefs epic (#40): move the three dict-order
primitives `collections.OrderedDict` imports from `__pypy__` out of
`identity_dict_app.py` into interp-level `functions:` entries, so they register
as non-binding `BuiltinFunction`s — matching PyPy's `interp_dict.py` /
`interp_magic.py` placement. `identity_dict` stays an app-level class.

## Changes
- **pyre-object `dictmultiobject`**: `w_dict_move_to_end_checked`, the
  object-strategy reposition primitive backing `move_to_end`
  (`nondescr_move_to_end`), with the callback-free fast path + reentrant-scan
  fallback of the `delitem_if_value_is` precedent. Skips the `keys_version`
  bump when the key is already at the target (no-op reorder), matching the
  precedent and PyPy's length-only iteration guard.
- **baseobjspace**: `dict_move_to_end` wrapper.
- **executioncontext**: `ec.py_repr` slot holding the `objects_in_repr`
  identity dict (EC-owned, thread-local like PyPy's `ec._py_repr`), init to
  `PY_NULL` and walked in `walk_builtin_roots`.
- **`__pypy__/mod.rs`**: the three interp functions + `functions:` registration.
- `move_to_end` binds `last` positionally or by keyword (`@unwrap_spec(last=bool)`)
  and rejects unknown keywords; `reversed_dict`/`move_to_end` reject a read-only
  `mappingproxy` (`isinstance(W_DictMultiObject)` parity).

## Testing
- New `bench/synth/pypy_dict_primitives_nonbinding.py` — byte-identical `OK` on
  pyre, CPython 3.14, and PyPy3.
- `test.test_ordered_dict`: failure set unchanged (pre-existing C-impl
  OrderedDict gaps + `__del__`/GC recursion tests; none in the changed paths).

## Follow-up
- #46: the module-dict branches of `move_to_end` / `delitem_if_value_is` lack a
  reentrant fallback (contrived, `__pypy__`-direct-only; shared limitation).


youknowone:

Verified each finding against the PyPy source before deciding.

**§1 Regression — typed-strategy de-specialization (`w_dict_move_to_end_checked`)** — CONFIRMED real. PyPy's typed-strategy `move_to_end` (`dictmultiobject.py:1170-1180`) reorders in the typed storage for a native-type key and switches to Object only for a foreign key; the old app-level `dict.__delitem__`+`dict.__setitem__` also preserved the typed strategy, so this is genuinely a regression. **Deferred to a follow-up task**, not dismissed, because: (a) it is not Python-observable here — pyre exposes no `__pypy__.strategy()` — and `move_to_end` still yields the correct order/values (the divergence is JIT typed-dict de-specialization only); (b) a faithful fix ports `move_to_end` across pyre's whole strategy taxonomy (Int/Unicode/Bytes/Identity + Empty/Kwargs/Map), a substantial rewrite of a hot GC-sensitive primitive that would balloon this reviewed diff. The typed stores are all ordered `IndexMap`, so the convergence path is a per-strategy `get_index_of`+`move_index` (callback-free); tracked with that recipe.

**§2** — None, agreed.

**§3 Pre-existing**
- `identity_dict` (id-only storage, no `keys`/`values`, empty iteration, address-reuse false hit) vs `W_IdentityDict`: pre-existing (this patch did not touch the class) and belongs to the broader interp-level port epic. Note: the id-reuse false-hit does not materialize for the two real consumers (`objects_in_repr` repr-guard and the pickle memo), where keys stay live for the entry's lifetime.
- `keys_version` iterator rejection on delete/reinsert-restoring-length: pre-existing; inherent to pyre's index-based dict iterator (documented adaptation). This patch's no-op guard removes the spurious-bump case where nothing moved.

**§4 Structural adaptations** — all acknowledged as necessary Rust/GC/ABI adaptations (`ec.py_repr` field, `resolve_dict_backing` for subclass backings, flat builtin-kwargs marker parsing, in-place `IndexMap` reorder). The §4 caveat that in-place reorder is acceptable only if strategy semantics stay equivalent is captured by the §1 follow-up.

— *commented by Claude*

PR #970, opened 2026-08-02
typedef: method-wrapper and classmethod_descriptor types, plus jit-stats baselines

youknowone:
Follow-up work on top of #845.

## `jit`: exempt the load-global namespace operand from the NULL-Ref walk abort

`pickle_terminal_raise_resume` produced wrong code. The residual-call walk
aborted on a NULL `Ref` operand that a `LOAD_GLOBAL` namespace slot legitimately
carries.

## `typedef`: add `method-wrapper` and `classmethod_descriptor` types

`method_descriptor_kind.py` reported `(list, '__len__', method_descriptor)`
where the descriptor is a `wrapper_descriptor`. Gating the second retag sweep on
`gateway::is_slot_wrapper` needs a bound form for the retagged
`wrapper_descriptor`, so both missing types are added:

| type | payload | published via |
|---|---|---|
| `method-wrapper` | the `Method` a bound method descriptor uses | `w_method_set_public_class` in `function::method_wrapper_new` |
| `classmethod_descriptor` | the `ClassMethod` a `@classmethod` uses | `w_class` write in `function::classmethod_retag_descriptor` |

Neither writes `ob_type`, so `py_type_check` predicates (`is_method`,
`is_classmethod`) and every binding path are unaffected — only `typedef::r#type`,
which prefers `w_class`, sees the new type.

- `classmethod_descr_get` picks `builtin_bound_method_new` for a
  `classmethod_descriptor` and `w_method_new` otherwise, so `dict.fromkeys` is a
  `builtin_function_or_method` while a user `@classmethod` binds to `method`.
- `is_bound_builtin_method` widens from "wraps a `method_descriptor`" to
  "published as `builtin_function_or_method`" — the classmethod case wraps a
  plain builtin, so the old test rejected it and `repr(dict.fromkeys)` raised.
- `wrapper_descriptor` and `method_descriptor` get `__reduce__`
  (`getattr(objclass, name)`); `copyreg._reduce_ex` refuses them otherwise.
  `classmethod_descriptor` deliberately gets none — pickling one raises
  "cannot pickle 'classmethod_descriptor' object".
- `builtins::is_native_exception_dunder` guarded on `is_function`, which excludes
  `SLOT_WRAPPER_TYPE`; once `__str__`/`__repr__` retag, the anti-recursion guard
  stopped firing and `str(SyntaxError("bad"))` overflowed the native stack. It
  now guards on `is_function_carrier`.

The `classmethod_descriptor` retag is two-phase: `init_typeobjects` runs the
sweeps inside `TYPEOBJECT_CACHE.get_or_init`, where `gettypeobject` returns
nothing, so the type is registered with `set_instantiate` at creation, the retag
reads `get_instantiate` and no-ops on null, and an idempotent closing pass over
`reg.values()` covers the namespaces built earlier (`object`, `type`, `int`,
`dict`, `mappingproxy`).

`lib-python/3/inspect.py` returns to pristine CPython 3.14 for `ismethod`,
`isbuiltin`, `ismethodwrapper`, `ismethoddescriptor`, `_NonUserDefinedCallables`
and `_signature_get_user_defined_method`. `_slot_wrapper_type` and the
`_builtin_code_type` block are deleted — `dict.update.__code__` stops resolving
after the split, so the `isfunction` override it fed was already dead.

## `bench`: record the seven missing jit-stats baselines

21 files for hot_loop_exit_then_class_stmt, inline_freevar_after_mayforce,
math_log_trig_hot, pickle_terminal_raise_resume, pypy_dict_primitives_nonbinding,
raise_reg_unbound_jitstress and tuple_unpack_array_backed_hot on all three
backends. All seven were inherited without a baseline.

Two nonzero badness values are pinned as-is: `raise_reg_unbound_jitstress`
`loops_aborted=2` on every backend, and `pickle_terminal_raise_resume`
`loops_aborted=54` on wasm against 0 on both native backends — unexplained, and
the baseline only pins it against a rise.

## `bench`: re-record the four caught-in-callee jit-stats baselines

float_div_zero_caught_loop, mutate_then_raise_caught,
inline_subwalk_property_mutates and inline_subwalk_mutating_residual. The cause
is the `try_walker_inline_resolved_user_call` decline for a branchy callee with
its own exception handler when `fbw_callee_body_replay_safety` is not `Clean` —
reverting that hunk alone puts all four back on the old numbers exactly.

The re-record raises no gated badness field:

| field | movement | gate class |
|---|---|---|
| `guard_failures` | 402→611, 401→611, 401→611, 603→813 | rise-bounded (what failed) |
| `bridges_compiled` | +1 each | ungated |
| `loops_aborted` | 1→0, 2→1, 2→0, unchanged | badness — falls |
| `loops_compiled` | 1→2, 3→4 where it moves | fall-gated — rises |

## Gates

`pyre/check.py --backend dynasm`: 5 failed, 361 passed (16 failed, 350 passed
before this branch).

The five are left red on purpose — closure_freevar_branch_resume,
exception_args_virtual, exception_multi_handler_warmup,
exception_reraise_tb_depth_jitstress and list_length_hint_validate raise
`loops_aborted` and lower `loops_compiled` with no attributed cause, and
re-recording them would disarm the floor.

`pyre/check.py --backend cranelift`: 7 failed, 359 passed — the same five plus
two perf-ratio rows that are the pypy-startup clamp, not a regression:

| bench | full suite | isolated ×2 | gate |
|---|---|---|---|
| jit_reg_const_pool_256_slot_decline | 131.2x (exec 0.66s) | 50.8x / 53.1x (exec 0.70s) | 122x |
| recursion_memo_branch | 24.8x (exec 0.12s) | 10.7x / 11.4x (exec 0.19s) | 20x |

`0.66/131.2 = 0.00503` and `0.12/24.8 = 0.00484` are both `EXEC_TIME_FLOOR_S`, so
the pypy denominator was clamped and the ratio is `pyre_exec / 5ms`. The isolated
pyre times are higher while the ratios fall.

`extra_tests/parity_tests/run.py`: green on cpython, dynasm and cranelift.
`cpython_tests/run.py --backend dynasm`: 46 pass, no regressions.
`test_inspect`: 362/362.

Descriptor kinds and bound/unbound type pairs were diffed against a CPython
3.14 oracle for `object.__init_subclass__`, `object.__subclasshook__`,
`type.__prepare__`, `int.from_bytes`, `dict.fromkeys`, `dict.__class_getitem__`,
`float.fromhex`, `bytes.fromhex` and `set.__class_getitem__`.

## Summary by CodeRabbit

- **New Features**
  - Added CPython-compatible method-wrapper and classmethod-descriptor behavior, including binding, invocation, representation, comparison, hashing, and attribute access.
  - Improved handling of built-in slot wrappers and class methods.

- **Bug Fixes**
  - Prevented recursive dispatch when formatting native exception representations.
  - Improved introspection of callable types and signatures.
  - Fixed nested global lookups involving nullable namespaces.

- **Tests**
  - Expanded JIT benchmark coverage and refreshed performance statistics across multiple execution paths.

### DIFF
--- a/lib-python/3/inspect.py
+++ b/lib-python/3/inspect.py
@@ -165,21 +165,6 @@ from operator import attrgetter
 from collections import namedtuple, OrderedDict
 from weakref import ref as make_weakref
 
-# PyPy/pyre gateway functions are ordinary ``function`` objects carrying a
-# distinct builtin-code object.  CPython identifies native descriptors by
-# their callable object types; those aliases are broader on PyPy (notably
-# ``types.WrapperDescriptorType is types.FunctionType``), so retain the code
-# type discriminator used by PyPy's inspect port.
-try:
-    _builtin_code_type = type(dict.update.__code__)
-except AttributeError:
-    _builtin_code_type = None
-
-# ``types.WrapperDescriptorType`` cannot name the slot-wrapper type here:
-# ``object.__init__`` is an ordinary method descriptor, so that alias resolves
-# to ``MethodDescriptorType``.  ``int.__pow__`` is a real slot wrapper.
-_slot_wrapper_type = type(int.__pow__)
-
 # Create constants for the compiler flags in Include/code.h
 # We try to get them from dis to avoid duplication
 mod_dict = globals()
@@ -202,10 +187,7 @@ def isclass(object):
 
 def ismethod(object):
     """Return true if the object is an instance method."""
-    if not isinstance(object, types.MethodType):
-        return False
-    func = object.__func__
-    return not isinstance(func, types.MethodDescriptorType)
+    return isinstance(object, types.MethodType)
 
 def ispackage(object):
     """Return true if the object is a package."""
@@ -226,8 +208,7 @@ def ismethoddescriptor(object):
     tests return false from the ismethoddescriptor() test, simply because
     the other tests promise more -- you can, e.g., count on having the
     __func__ attribute (etc) when an object passes ismethod()."""
-    if (isclass(object) or ismethod(object) or isfunction(object)
-            or isbuiltin(object) or ismethodwrapper(object)):
+    if isclass(object) or ismethod(object) or isfunction(object):
         # mutual exclusion
         return False
     tp = type(object)
@@ -299,10 +280,7 @@ def isfunction(object):
         __dict__        namespace which is supporting arbitrary function attributes
         __closure__     a tuple of cells or None
         __type_params__ tuple of type parameters"""
-    if not isinstance(object, types.FunctionType):
-        return False
-    code = getattr(object, "__code__", None)
-    return not (_builtin_code_type and isinstance(code, _builtin_code_type))
+    return isinstance(object, types.FunctionType)
 
 def _has_code_flag(f, flag):
     """Return true if ``f`` is a function (or a method or functools.partial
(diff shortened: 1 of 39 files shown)


youknowone:
Follow-up in #981 for the review findings here.

Fixed, each verified against a CPython 3.14.2 oracle:
- the missing `PyType_IsSubtype` check in `classmethod_get` / `classmethoddescr_call` (both the P2 on `__get__` and the Major on `__call__` were this one defect) — without it `dict.__dict__['fromkeys'](list, ['x'])` reached `dict.fromkeys` with `list`, and `type.__dict__['__prepare__'](list, 'X', ())` returned `{}`;
- the fixed arity-3 `__get__` registration, which `wrap_descr_get` unpacks as 1..2 — the same defect was in the `wrapper_descriptor` and `method_descriptor` entries, so all three now share one helper;
- the `classmethod_descriptor` getset receiver-mismatch wording.

Not applicable, with the measurements:
- **`bh_load_global_fn` namespace resolution** — the helper already discarded `namespace_ptr` before this PR (`let _ = namespace_ptr;` is untouched by the branch), so the arg-0 exemption decides only whether the walk aborts. One code object driven through 200k iterations under two different `__globals__` returns `('module', 'alt')` on dynasm with `loops_compiled=4`, matching CPython and `PYRE_JIT=0`. The custom-`__builtins__` variant diverges identically with the JIT off, so it is a pre-existing interpreter gap.
- **rooting in `descr_reduce`** — the owner is a `W_TypeObject` and `w_str_new` allocates through off-GC `lltype::malloc_typed`, both non-moving; and `w_tuple_new` roots its own items either way (`w_tuple_new_array_backed` pins each and refills from the shadow stack, as does `w_specialised_tuple_oo_new`).

New parity script `classmethod_descriptor_kind.py` pins all of it.

— *commented by Claude*

PR #972, opened 2026-08-02
jit: root the blackhole virtualizable_ptr slot, drop the re-entrant trace-too-long teardown, and give the exception-edge bridge its discarded-frame traceback nodes

youknowone:
Eight commits on top of `main`. Each is independently reproducible.

> **Rebased 2026-08-02 onto `ca1a421e847`.** Three commits that were in the
> earlier version of this PR are gone:
>
> - the quasi-immut watcher install and its `ABORT_FORCE_QUASIIMMUT` walk abort —
>   **merged to main as #977** (`717d37c5e2d`), which is a superset;
> - `jit: keep the handler-only live registers past the Ref-bank frame-slot
>   retain` — superseded by **#973** (`ae58d25cf0d`), which deletes that retain
>   outright, the orthodox fix (`liveness.py:5-12` has no frame-slot condition);
> - `jit: screen the FOR_ITER consume where it is reached instead of scanning the
>   framestack`, its repair, and the baselines they moved — **withdrawn**, see
>   "Withdrawn" below.

## 1. `virtualizable_ptr` was a bare copy across a collection

`BlackholeInterpreter::run` roots every chain level's register bank and, for a
level with a `virtualizable_info`, the array-field slots inside its
virtualizable. `virtualizable_ptr` itself was neither — it is a bare copy of the
frame red the level was bound to, so a collection forwarded the register and left
the copy naming the pre-move address. A compiled trace allocates an inlined
callee's `PyFrame` through its own `NewWithVtable`, which the GC rewriter lowers
to a nursery allocation, so a minor collection inside `run_inner` relocates it;
the propagation loop then stores the stale pointer into `PyTraceback.frame`,
whose custom trace forwards it on the next collection and reads a data word as a
type id. Register the slot, the shape `blackhole_from_resumedata` already applies
for the chain-build window.

## 2. Two epilogue blocks both handled the same abort, running the callee twice

`LoopBearingCalleeInlineUnsupported` is handled by the carrier block (which can
take `CalleeRebuild`, resuming INSIDE the rebuilt callee) and by the
`FBW_ABORT_OUTER_RESUME` block (which rewinds the outer frame to its CALL and
re-executes it). Nothing kept them apart: the second block's gates are inclusion
tests on `fbw_executed_nonpure_residual` / `fbw_has_unjournaled_effect`, and
`walk_end_resume_provable` samples `FBW_EXECUTED_EFFECT_COUNT`, which the rebuilt
callee's plain interpretation never bumps. Read `WALK_END_FLUSH_COMMITTED` first
and reset the latch.

## 3. The after-residual-call `-live-` markers were never narrowed

`get_list_of_active_boxes` reads the marker AFTER the call whenever `in_a_call`
or `after_residual_call` holds and the one before the op otherwise
(`pyjitpl.py:194-198`); `compute_liveness` is one uniform pass over every
`-live-`. Group the after-call markers alongside the per-PC ones so both marker
families go through the same pass, adding the preceding call's own Ref result
register (`pyjitpl.py:186`). Declines at `collect_callee_active_boxes` 5 → 1 over
`pyre/bench/**`.

## 4. A re-entrant interpreter run tore down the outer trace

`blackhole_if_trace_too_long` runs in the tracer's own stepping loop
(`pyjitpl.py:2861-2867 _interpret`); `pyframe.py dispatch_bytecode` has no such
call. Pyre's walk loop already runs the check, so the copy in `eval_loop_jit`'s
opcode dispatch was a second caller of the same teardown — and a residual call
executed inside an inline sub-walk runs Python through `eval_loop_jit`, whose
per-step check read the OUTER trace's op count and moved that `TraceCtx` out of
the shared slot. The in-flight walk then kept recording through a `&mut TraceCtx`
whose recorder buffer had been freed. RPython cannot reach this: `warmstate.py:437-441
bound_reached` builds a fresh `MetaInterp` per trace attempt.

## 5. The exception-edge bridge emitted one traceback node for a whole unwind

`route_exc_edge` re-points the live frame at its own handler and discards the
inlined callee frames outright, so the flat walk that follows records only the
catching frame's node. `pyopcode.py:148 record_application_traceback` runs BEFORE
the `:152` exception-table lookup, so a frame the unwind only passes through
contributes a node exactly like the one that catches. `resume_coords[1..]` is
exactly the discarded set: publish it at the routing point and emit one node per
level at the handler entry, innermost-first.

## 6. The quasi-immut watcher was installed after the value was read

`quasiimmut.py:124-126` orders `self.qmut = get_current_qmut_instance(...)` before
`self.constantfieldbox = self.get_current_constant_fieldvalue()`. Reading first
leaves a window in which the field moves with no watcher installed, so nothing
invalidates, nothing bumps the force counter, and the trace keeps a value that is
already stale. Without a GIL that window is a real interleaving. (Ordering fix on
top of #977's install.)

## 7. The discarded level's traceback pc was off by one instruction

`record_discarded_level_traceback` received `py_pc` straight out of
`resume_coords`, a `next_instr`-style coordinate — the same one `exc_table_offset`
converts with `saturating_sub(1)` and the live frame converts with
`set_last_instr_from_next_instr`. Its three consumers all want the instruction
that RAN, so nodes named the instruction AFTER the raising or calling opcode and a
bare `RERAISE` decoded as whatever follows it, gaining a node the
`RaiseWithExplicitTraceback` rule forbids. (Raised by both reviewers on this PR.)

## 8. One grouping helper for both marker families

`filter_liveness_in_place` had grown a second copy of the "push onto the entry
with this `insn_idx`, or start one" loop. Extract `group_py_pcs_by_insn`.
(Raised by CodeRabbit on this PR.)

## Withdrawn: the FOR_ITER framestack-scan deletion

The earlier version replaced `fbw_inline_callee_hazardous`' loop-bearing
framestack scan with a screen at the `for_iter_next` consume, on the measurement
that "the deleted scan claimed no abort the other two arms do not" — declines at
that site 16 → 14, synth corpus byte-identical.

**That measurement counted declines at the fbw site, not the aborts the admitted
callee goes on to cause**, and it held only because the `-live-` Ref-bank retain
narrowed away the colors that make `collect_callee_active_boxes` decline.
Once #973 deleted that retain, the same change takes
`synth/inline_subwalk_user_iterator` from `loops_aborted=1` (main's recorded
baseline, which main still meets) to **5**, and panics on wasm with
`blackhole recursive_call: jitdrivers_sd[0] carries no portal runner`. Measured
by reverting `fbw_state.rs`/`residual_call.rs` to main's version: 5 → 1.
`foriter_exempt_shared_generator`, the fixture that change existed to fix, passes
on main's version either way — the framestack scan masks that bug too. So the
commit, its repair, and the baseline re-record that recorded all three of their
consequences are withdrawn rather than re-baselined.

## Status

`cargo fmt --all -- --check` clean; `cargo test --all --no-default-features
--features dynasm` green (101 suites).

| backend | result |
|---|---|
| dynasm | 2 failed / 366 passed |
| cranelift | 2 failed / 366 passed |
| wasm | 2 failed / 362 passed |

All six failures are **inherited from `main`**, verified locally by reverse-applying
this branch's whole diff and re-running each fixture on the resulting tree — the
numbers are identical, not merely the fixture names:

| fixture | main | this branch |
|---|---|---|
| `synth/exception_args_virtual` (all 3 backends) | FAIL `loops_aborted 0 -> 3`, `guard_failures 401 -> 1002` | identical |
| `synth/list_length_hint_validate` (dynasm, cranelift) | FAIL `loops_aborted 14 -> 34`, `guard_failures 828 -> 4923` | identical |
| `synth/pickle_terminal_raise_resume` (wasm) | FAIL `loops_aborted 54 -> 59` | identical |

### DIFF
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -1482,6 +1482,26 @@ impl BlackholeInterpreter {
         unsafe {
             let mut current = Some(&mut *self);
             while let Some(frame) = current {
+                // `virtualizable_ptr` is a bare copy of the frame red this
+                // level was bound to, and the banks rooted just above do not
+                // cover it: a collection forwards the register the copy came
+                // from, never the copy.  A young virtualizable makes that
+                // difference observable — a compiled trace allocates an
+                // inlined callee's `PyFrame` with its own `NewWithVtable`,
+                // which the GC rewriter lowers to a nursery allocation, so the
+                // first minor collection `run_inner` triggers moves it and
+                // leaves this field naming the vacated block.  Every later
+                // reader takes that address: the vable opcodes below, and the
+                // traceback recorded for each frame an exception propagates
+                // through, which stores it into `PyTraceback.frame` — a slot
+                // the collector does trace, so the stale pointer surfaces as
+                // an invalid type id rather than as a wrong answer.  Root the
+                // SLOT so the walker rewrites it, the same shape
+                // `blackhole_from_resumedata` already applies to the resume
+                // reader's own `virtualizable_ptr` for the chain-build window.
+                majit_gc::shadow_stack::push_resume_ref_roots(std::slice::from_mut(
+                    &mut frame.virtualizable_ptr,
+                ));
                 if !frame.virtualizable_info.is_null() {
                     let vinfo = &*frame.virtualizable_info;
                     vinfo.push_resume_ref_roots_for_registers(&frame.registers_r);
--- a/majit/majit-metainterp/src/lib.rs
+++ b/majit/majit-metainterp/src/lib.rs
@@ -154,10 +154,12 @@ pub use pyjitpl::{
     MetaInterpStaticData, RawCompileResult, StandaloneFrameStack, build_state_field_snapshot,
     call_int_function, call_ref_function, call_void_function, counters,
     record_application_traceback_for_recording, record_application_traceback_hook_address,
+    record_discarded_level_traceback_for_recording, record_discarded_level_traceback_hook_address,
     record_inline_application_traceback_for_recording,
     record_inline_application_traceback_hook_address, set_record_application_traceback_hook,
-    set_record_inline_application_traceback_hook, struct_fields_write_effect_info, trace_jitcode,
-    trace_jitcode_from_merge_point, trace_jitcode_with_args, trace_jitcode_with_args_and_runtime,
+    set_record_discarded_level_traceback_hook, set_record_inline_application_traceback_hook,
+    struct_fields_write_effect_info, trace_jitcode, trace_jitcode_from_merge_point,
+    trace_jitcode_with_args, trace_jitcode_with_args_and_runtime,
 };
 pub use resume_box_reader::{
     BridgeVirtualCache, decode_fieldnum, default_bridge_array_descr, emit_pending_field_op,
--- a/majit/majit-metainterp/src/pyjitpl.rs
+++ b/majit/majit-metainterp/src/pyjitpl.rs
@@ -1571,10 +1571,21 @@ pub struct JitHooks {
 /// majit keeps only this crate-neutral ABI hook.
 pub type RecordApplicationTraceback = extern "C" fn(i64, i64, i64, i64);
 pub type RecordInlineApplicationTraceback = extern "C" fn(i64, i64, i64, i64, i64);
+/// Runtime callback for a frame the resume crossed and then DISCARDED.
+///
+/// `pyopcode.py:148` attaches a node on every frame an unwind passes through,
+/// which upstream gets for free because the unwind is traced code running on a
+/// rebuilt MIFrame stack.  A resume that re-points the live frame straight at
+/// its own handler has no such stack, and the only description left of the
+/// frames it skipped is the `(w_code, python pc)` pair in the resume data —
+/// so this hook takes a PYTHON pc where the two above take a jitcode one.
+pub type RecordDiscardedLevelTraceback = extern "C" fn(i64, i64, i64);
 
 static RECORD_APPLICATION_TRACEBACK: std::sync::atomic::AtomicUsize =
     std::sync::atomic::AtomicUsize::new(0);
 static RECORD_INLINE_APPLICATION_TRACEBACK: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+static RECORD_DISCARDED_LEVEL_TRACEBACK: std::sync::atomic::AtomicUsize =
     std::sync::atomic::AtomicUsize::new(0);
 
 pub fn set_record_application_traceback_hook(hook: Option<RecordApplicationTraceback>) {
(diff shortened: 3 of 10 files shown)


youknowone:
Review dispositions after the rebase onto `ca1a421e847`. Two findings fixed, one
**confirmed with a reproducer and filed as pre-existing**, one whose commit is
withdrawn, two now moot.

### Fixed

- **CodeRabbit, `state.rs:4465` — install the watcher before reading the value.**
  Correct; `quasiimmut.py:124-126` orders `get_current_qmut_instance` before
  `get_current_constant_fieldvalue`. Fixed on top of #977's install (commit 6).
- **Codex P2 `call_jit.rs:842-848` / CodeRabbit `call_jit.rs:849` — convert the
  resume PC to the failing opcode.** Both reviewers found the same defect and
  both are right: `resume_coords` carries `next_instr`-style coordinates — the
  same function's `exc_table_offset` converts them with `saturating_sub(1)` — but
  `record_discarded_level_traceback` fed `py_pc` unconverted to
  `decode_instruction_at`, `frame.last_instr` and `record_application_traceback`.
  Converted once at entry (commit 7).
- **CodeRabbit, `codewriter.rs:4449-4456` — extract a shared grouping helper.**
  Done, `group_py_pcs_by_insn` (commit 8).

### Confirmed, and it is pre-existing — filed, not fixed here

- **Codex P1 `residual_call.rs:2345-2349` — decline loop-bearing callees before
  their first mutation.** The mechanism is real and I reproduced it as an output
  divergence, but it is **not** introduced by the commit it was raised against.

  Probe: an inlined callee that runs an irreversible residual *before* reaching
  its own `FOR_ITER`, called 30000 times; `c.pos` should end at 30000.

  ```python
  def step(c, it):
      c.pos = c.pos + 1      # irreversible residual, BEFORE the FOR_ITER
      s = 0
      for x in it:           # user __next__, StopIteration after 3
          s += x
      return s
  ```

  `fbw_abort_nested_unjournaled_residual` returns `Ok` for the store (the callee
  is neither self-recursive nor `DeferredCall`-admitted), the walk aborts later,
  the effect-count gate at `fbw_state.rs:1420` correctly refuses to install the
  rewind leg, the walk exits non-committed, and the legacy replay re-runs the
  call. The store is outside the journal, so it applies twice.

  | tree | `c.pos` (correct 30000) |
  |---|---|
  | `origin/main` + this branch | 30005 |
  | + pre-#973 `pc_live_r.retain` restored | 30001 |
  | + retain restored **and** the FOR_ITER consume screen made a no-op | 30001 |

  So the double-apply is **pre-existing on `main`**; #973's retain deletion
  amplifies it ×5 by opening `collect_callee_active_boxes`' `OpRef::NONE` decline
  as a fresh abort site; the screened commit contributes 0. Corpus-wide there are
  only 7 `committed=false effects>0 journal=0` census lines over 414
  `pyre/bench` scripts and none of them diverge — the census shape alone is not
  proof, which is why this needed a constructed case. Tracked separately; the fix
  is either the `CONST_NULL` substitution `collect_outer_active_boxes` already has
  at `mod.rs:4721-4724`, or landing the abort on a flush leg so there is no replay.

- **Codex P1 `call_jit.rs:859-865` — rebuild discarded traceback frames from
  resume data.** Correct: the node's `tb_frame` is fabricated from `(w_code,
  py_pc)` and has no arguments, locals, cells, value stack or back-reference.
  Deferred with a named blocker — the recorded IR bakes `w_code`/`py_pc` as
  constants, so materializing per-frame resume state needs the resume data
  carried to the recorder, which is a separate slice.

### Moot after the rebase

- **Codex P1 `quasiimmut.rs:261-262` / `mod.rs:2412` — scope force events to the
  active walk**, and **CodeRabbit `trace.rs:3745` — do not replay after an
  effectful `ForceQuasiImmut` handoff fails.** Both name the
  `FORCED_QUASI_IMMUTS` process counter and the `FBW_FORCED_QUASIIMMUT_BASE`
  thread-local baseline, which this branch no longer carries: that work merged to
  main as #977 with a different mechanism. Re-raise against `main` if the concern
  survives there.

— *commented by Claude*

PR #980, opened 2026-08-02
jit: two exception-path fixes (callee is-None inline, too-long Ref image)

youknowone:
Two JIT fixes found while attributing the slowest exception benchmarks against pypy.

### `jit: key the callee is-None inline decline on the kept stack`

The multiframe seed declined any callee containing `POP_JUMP_IF_NONE` /
`POP_JUMP_IF_NOT_NONE`. Its stated premise — a residual `is`/`is_not` whose Ref
operand a mid-body guard resume cannot source — no longer holds: the codewriter
emits a bare `ptr_eq`/`ptr_ne` with a `Kind::Int` result feeding the exitswitch.
`LOAD_FAST_CHECK` emits the same shape (`ptr_nonzero/r>i`) and was never declined.

What actually breaks is narrower: a branch that leaves operands on the value
stack, whose guard resume must restore kept slots the inline sub-walk's mirror
does not model. The scan now asks `LiveVars::stack_depth_at` instead of matching
the opcode, so a statement-level `if x is None:` inlines while the same test in
expression position keeps the residual. Strictly narrower than the old predicate.

Measured on dynasm, interleaved A/B, median of 9 pairwise ratios:
`exception_nested_exc_info_restore` 1.72x, `gc_bug_bridge_flavor_traceback_names`
1.32x. Everything else lands inside this box's ±15% noise floor.

### `jit: fill the dead-box_bool Ref color in the too-long image`

Latent wrong-code on main. `fill_trace_too_long_register_banks` skips any Ref
color it cannot supply a concrete for, on the premise that an unfilled color is
one the interpreter does not read at that pc. That fails for the dead-`box_bool`
marker, which writes a comparison's raw truth Int straight into the Ref register:
neither fill arm can represent an Int in a Ref bank, so the color was skipped, the
blackhole read NULL, and `is_true(NULL)` sent a module-scope `while` to its exit
arm after one iteration.

Reachable only when the module-dict NAME cell fold fires, so a module carrying
any `try` never showed it — including `trace_too_long_effect_replay`, whose own
`try: import pypyjit` unfolds its whole body and left the phase-A contract it
documents untested.

Also corrects both cell-fold comments: the handler-free gate is not the
can-raise argument they claimed (that argument transfers from the ungated
LoadGlobal fold, which shares the emitter). Lifting it is held back by a retrace
storm instead — `exception_reraise_tb_depth_jitstress` 0.90s → 5.85s at
loops_compiled 305 → 1802 — so both gates stay.

### Verification

- `check.py --backend dynasm` 362 passed / `--backend cranelift` 362 passed, with the same 6 pre-existing jit-stats baseline failures as before the change
- JIT-on vs JIT-off identical across all 353 synth fixtures
- `cargo test --all --features dynasm` 97 suites ok; `cargo test -p pyre-object` ok

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved exception handling across nested calls with more accurate traceback information.
  - Prevented unsafe multi-frame optimizations in unsupported scenarios.
  - Fixed edge cases involving conditional checks, preserved values, and boolean reconstruction.
  - Preserved consistent string interning and repeated code-name access.

- **Performance**
  - Optimized runtime loading to materialize JIT code only when needed.
  - Reduced unnecessary compilation and guard failures in benchmarked workloads.

- **Documentation**
  - Clarified boolean state resets, trace recovery, and optimized name-access behavior.

### DIFF
--- a/pyre/bench/synth/exception_catching_frame_tb_node.cranelift.jitstats
+++ b/pyre/bench/synth/exception_catching_frame_tb_node.cranelift.jitstats
@@ -1,8 +1,8 @@
-bridges_compiled=3
+bridges_compiled=2
 descr_set_absent=0
 descr_set_ambiguous=0
 descr_set_stale_absent=0
-guard_failures=601
+guard_failures=401
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=4
+loops_compiled=3
--- a/pyre/bench/synth/exception_catching_frame_tb_node.dynasm.jitstats
+++ b/pyre/bench/synth/exception_catching_frame_tb_node.dynasm.jitstats
@@ -5,4 +5,4 @@ descr_set_stale_absent=0
 guard_failures=401
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=4
+loops_compiled=3
--- a/pyre/bench/synth/exception_try_call_inlined_callee_raise.cranelift.jitstats
+++ b/pyre/bench/synth/exception_try_call_inlined_callee_raise.cranelift.jitstats
@@ -1,8 +1,8 @@
-bridges_compiled=2
+bridges_compiled=1
 descr_set_absent=0
 descr_set_ambiguous=0
 descr_set_stale_absent=0
-guard_failures=408
+guard_failures=201
 internal_compile_panics=0
-loops_aborted=1
-loops_compiled=3
+loops_aborted=0
+loops_compiled=2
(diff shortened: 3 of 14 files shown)


pyre/bench/synth/exception_catching_frame_tb_node.cranelift.jitstats
```diff
@@ -5,4 +5,4 @@ descr_set_stale_absent=0
 guard_failures=601
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=4
+loops_compiled=3
```
youknowone, line 8: do not giveup this

PR #986, opened 2026-08-02
majit-gc, thread: port rgil, hold the GIL across pyre code, and release it at every blocking call

youknowone:
pyre had no GIL. Mutual exclusion for GC operations came from `gc_sync`'s own per-operation gate: every `gc_op` compare-and-swapped a claim word whose expected value was the calling thread's ident, so **every allocation** paid a `_tlv_get_addr` call (Mach-O `thread_local!`), a CAS, and a release store.

A design pass over that gate concluded it cannot be made cheaper on its own terms. A constant token instead of the ident dies on `in_gc_op()` — `gc_query_reentrant` asks "am *I* already inside a `gc_op`", and a shared token only answers "*someone* is". Both error directions are fatal, and `register_thread`'s own doc admits the reachable case: an unregistered thread may still call `gc_op`. Every repair needs per-thread state on the entry path, which on Mach-O is the TLS call being removed.

Upstream does not have this problem, and not because its gate is cheaper — it has no gate. `rg threadlocalref_addr rpython/memory/gctransform/framework.py rpython/memory/gc/incminimark.py` has **no hits**: the allocation path reads no thread identity at all. The GIL is held, so `malloc_fixedsize` just bumps `gcdata.gc.nursery_free`.

So this ports the GIL, and specifically its *lifetime*.

## What is ported

`majit/majit-gc/src/rgil.rs` is `rpython/translator/c/src/thread_gil.c` plus the `mutex2_t` half of `thread_pthread.c:559-595` and the `rgil.py` surface.

The composite GIL is two locks (thread_gil.c:2-31). `rpy_fastgil` is a plain word — 0 means unlocked, anything else is the holder's ident; release is a plain store of 0, acquire is a CAS against 0. `mutex_gil` is a mutex the fast path never unlocks; the loser of the CAS goes to `RPyGilAcquireSlowPath`, becomes the stealer via `mutex_gil_stealer`, and alternates between retrying the fast path and `mutex2_lock_timeout(&mutex_gil, 0.0001)`. `rpy_waiting_threads`, the randomised early poll (`n = rpy_early_poll_n*2+1` wrapped into `[40, 400)`), `RPyGilYieldThread`, `gil_get_holder` and `am_I_holding_the_GIL` all come across.

`pyre-interpreter/src/module/thread/gil.rs` is `pypy/module/thread/gil.py:44-50` — `GILReleaseAction`, a `PeriodicAsyncAction` with `use_bytecode_counter=True` whose `perform` calls `rgil.yield_thread()`. It is registered **per `ExecutionContext`** rather than once per space, because each pyre mutator owns its own ticker and actionflag; `clone_for_thread` clears the slot.

The lifetime: acquire in `register_thread` and on return from an external block, release in `unregister_thread` and in `before_external_block`. A thread therefore holds the GIL for as long as it runs pyre code, which is the whole point — between two external calls arbitrarily many allocations run with zero synchronisation.

One deviation, stated deliberately: a thread waiting for the GIL leaves the RUNNING census (`leave_running_for_gil` / `rejoin_running_after_gil`). pyre's collector drains that census explicitly; upstream needs no equivalent because there the collector *is* the GIL holder.

## What that lets go

Deleted from `gc_sync`: `IN_FAST_PATH`, `GATE_INSIDE`, `GateGuard`, `release_gate`, `revoke_standing_claim`, `acquire_gate`, `in_gc_op`, `gc_op_slow`. `gc_mutex` no longer guards any operation and is renamed `install_mutex`.

```rust
pub fn gc_op<R>(f: impl FnOnce(&mut MiniMarkGC) -> R) -> R {
    debug_assert!(crate::rgil::am_i_holding_the_gil(), "a GC operation needs the GIL");
    let _reentry = ReentryGuard::enter();
    f(unsafe { singleton_mut() })
}
```

In release that is a bare borrow: no atomic, no TLS read. The reentrant-`&mut` check moves to a `#[cfg(debug_assertions)]` global bool — under a GIL one global replaces per-thread state.

`revoke_standing_claim` in particular existed *because* pyre had no guaranteed release points. It now has them.

## Measurement

`synth/int_mul_ovf_bignum_promote`, 400M iterations, dynasm/aarch64. Paired per-round ratio and a sign test, because the machine is shared. Both arms are built from the same tip — the baseline is `git checkout HEAD~1 -- majit/majit-gc pyre/pyre-interpreter pyre/pyre-jit` **without** re-extracting `.ullbc`, so the only difference between the two binaries is the GIL. They produce identical benchmark output.

| run | load avg | median ratio | rounds faster |
|---|---|---|---|
| 15 rounds | 44 → 13 | 0.8790 (−12.1%) | 13/15 |
| 12 rounds | 6 | **0.8498 (−15.0%)** | **12/12** |

This is more than the −6.8% ceiling an earlier unsound probe predicted, and the reason is that the probe only stubbed `my_ident()` — it still paid the CAS and the gate-release store on every allocation. The GIL removes all three.

## The commits on top

**`thread, modules: release the GIL around the remaining blocking host calls`.** Under a GIL a missed release is a deadlock, not a delay, because the sleeper holds it. `rffi.llexternal`'s `releasegil` default is *release* (rffi.py:151 — `invoke_around_handlers = not sandboxsafe and not _nowrapper`), so an audit of every blocking host call in the interpreter was owed. Newly bracketed: `select.poll`/`select.select`, `kqueue.control`, socket `accept`/`connect`/`send`/`sendall`/`sendto`/`sendmsg`/`recv`/`recvfrom`/`recvmsg` and the `getaddrinfo`/`getnameinfo` lookups, `signal.sigwait`, `_multiprocessing`'s `sem_wait` and both timed-wait arms, the fd read and write behind file objects, posix `open`/`fsync`/`fdatasync`/`sendfile`, fcntl `fcntl`/`ioctl`/`flock`/`lockf`, the `host_seam` real-host bodies and the sandbox trampoline, the stdlib-path `python3` subprocess, and the REPL line read.

It also adds `call_external_function`, which is `rffi.py:193-211`: release, call, **read errno, then** re-acquire. Upstream runs `_errno_after` ahead of `rgil.acquire()` (rffi.py:207-210) and pyre's existing brackets did not — `posix.read` and `posix.write` read `last_os_error()` after their guard had dropped, so a re-acquire that went through the stealer's mutex and condvar waits could overwrite the value they then reported.

Demonstrated with a smoke test that runs a counter thread while the main thread blocks: every one of `select.select`, `select.poll`, `os.read` on an empty pipe, `socket.accept` and `socket.recv` now lets the peer advance ~240k iterations across a 0.2s wait, where before it would have advanced none.

**`thread: take the GIL at the points Rust enters pyre code`.** This is what turned CI red on all three platforms, and it is not a test artefact. The registration — and the GIL that comes with it — was owned by pyre-jit's `init_gc_subsystem`, which only the launcher thread and spawned Python threads reach. A thread entering pyre code anywhere else allocated without ever becoming a mutator, so `gc_op`'s assertion fired in debug and, in release, the operation had **no exclusion at all**; before this port, `gc_op`'s own gate covered exactly that thread.

The fix cannot live inside `gc_op`: `am_i_holding_the_gil()` needs `my_ident()`, the `_tlv_get_addr` call that *was* the entire −6.8% ceiling, so a per-operation check costs back about half the win. The invariant has to be established by the threads instead. `module/thread::ensure_runtime_thread` is `rgil.py:186-193 acquire_maybe_in_new_thread`: idempotent per thread, it registers the mutator, takes the GIL, and arms a thread-local whose destructor unregisters and gives the GIL back at thread exit. That destructor is the load-bearing half — without it the first test thread would hold the GIL forever and every later one would deadlock instead of assert. the compile entries and `PyFrame::execute_frame` name it, and `init_gc_subsystem` now delegates rather than registering itself, so the registration has one owner.

**`majit-gc: keep the published nursery top live with more than one mutator`.** `register_thread` pinned the `gc.py:525-531` published nursery-top slot to zero on the 1→2 registration transition, so generated inline allocation stopped using the non-atomic bump once a second mutator registered. `llsupport/gc.py:525-531 get_nursery_free_addr`/`get_nursery_top_addr` carry no thread-count gate — the bump is serialised by the GIL. The transition goes, and with it `inline_alloc_enabled`, `set_inline_alloc_enabled` and its `GcHandle` forwarder, which had no other caller.

**`majit-gc, thread: address the GIL-port review findings`.** The fixes the bot reviews earned. `rgil::after_fork_child` re-created only `mutex_gil`'s `locked` flag; a thread that vanished in `fork()` inside `acquire_slow_path` leaves `mutex_gil_stealer` locked too, so both are now overwritten the way `rpy_init_mutexes` (thread_gil.c:92-98) `pthread_mutex_init`s them. `cfuncptr_call` held the GIL across `host_ctypes::call`, but `clibffi.py:350-352` declares `ffi_call` with the default `releasegil='auto'`, i.e. released. Under `sandbox`, `interp_time::sleep` entered an external block and then called `ops::sleep`, whose trampoline enters another one — the nested guard trips the RUNNING-census assertion; the redundant outer bracket goes. `RealHost::read` read `errno` after `blocking` had re-acquired the GIL. Plus: `BlockingGuard` is `#[must_use]` and `!Send`, `isenabled` stops being the one `&self` query routed through `gc_op`, the runtime-thread entry moves into `compile_source`/`compile_source_with_opts`, and `gil::shutdown` gives back the `GILReleaseAction` each execution context allocated. The full triage of all 22 findings, including four refuted with their disproofs, is in a comment below.

**`_socket: hold a writable-buffer export across the recv_into family`**, **`thread: answer setup_threads from the space, not the execution context`** and **`majit-gc: port RPyGilReleaseSignal`**. The remaining review findings. The first is demonstrable: a thread blocked in `recv_into(bytearray(64))` while the main thread appends to the same bytearray used to have the append accepted and the peer's five bytes land in the freed allocation (`buf[:5] == b'\x00\x00\x00\x00\x00'`); it now raises `BufferError` and reads back `b'hello'`. The third is from the Codex parity review: `RPyGilReleaseSignal`'s `_WIN32` guard is the point of it, not a reason to skip it — without it a Windows stealer waits out its timed wait at the 10-15 ms timer quantum.

## Verification

`cargo test --all --no-default-features --features dynasm`: 7380 passed, 0 failed. `cargo check --workspace` clean, and the CI clippy gate (`pyre-interpreter --features sandbox,dynasm`) clean.

`pyre/check.py` after `scripts/extract-llbc.py`: **dynasm 370/370, cranelift 370/370, wasm 366/366 — all three backends green.**

## What was dropped

`majit-gc: delete the stop-the-world protocol the GIL replaced` was on this branch and has been taken off it. It bisects to a SIGBUS in `gc_table_extra_root_walker` under `finalize_runtime`'s full collection — `MAJIT_GC_NURSERY_POISON=1 ./pyre-dynasm pyre/bench/synth/mutate_then_raise_caught.py`, interleaved A/B: 0/100 at its parent, 10/100 with it, 0/140 at the branch base. The argument for the deletion still holds; what it exposes does not, so it is being finished separately rather than landed here. The safepoint harness therefore stays underneath the GIL for now, and `pyre/design.md` says so.

`pyre/design.md`'s §3.3 declared free threading as present policy and listed a GIL among the non-options. It now records the GIL as the shipped mechanism, free threading as the target, and the measurement above.

— *opened by Claude*

## Summary by CodeRabbit

* **New Features**
  * Added coordinated interpreter locking for safer execution across runtime threads.
  * Blocking networking, filesystem, subprocess, synchronization, and input operations now yield execution appropriately.
  * Runtime and compiled code initialize consistently on the correct interpreter thread.
  * Crash diagnostics now include macOS reports and more complete captured output.

* **Bug Fixes**
  * Improved garbage-collection coordination, root preservation, allocation safety, signal interruption, and error reporting.
  * Improved fork and finalization behavior during concurrent execution.

* **Documentation**
  * Updated runtime and garbage-collection guidance for the new locking and synchronization behavior.

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -430,14 +430,10 @@ pub struct MiniMarkGC {
     /// The nursery (young generation).
     nursery: Nursery,
     /// gc.py:525-531 published nursery-top slot read by generated inline
-    /// allocation paths. Kept separately from the real top so free-threaded
-    /// execution can pin the published limit to zero without changing the
-    /// Rust allocator's nursery bounds. The atomic publishes limit changes to
-    /// generated code and Rust-side readers. The Box keeps its baked address
-    /// stable even if the containing GC value moves.
+    /// allocation paths. The atomic publishes limit changes to generated code
+    /// and Rust-side readers. The Box keeps its baked address stable even if
+    /// the containing GC value moves.
     published_nursery_top: Box<AtomicUsize>,
-    /// Whether generated code may use the shared nursery bump fast path.
-    inline_alloc_enabled: bool,
     /// The old generation.
     oldgen: OldGen,
     /// Type registry for tracing objects.
@@ -670,7 +666,6 @@ impl MiniMarkGC {
         let mut gc = MiniMarkGC {
             nursery,
             published_nursery_top,
-            inline_alloc_enabled: true,
             oldgen: OldGen::new(),
             types: TypeRegistry::new(),
             roots: RootSet::new(),
@@ -735,17 +730,10 @@ impl MiniMarkGC {
     }
 
     /// Refresh the gc.py:525-531 published nursery-top slot from the real
-    /// allocator bound, unless free-threading has disabled the non-atomic
-    /// shared bump fast path.
+    /// allocator bound.
     fn refresh_published_nursery_top(&mut self) {
-        self.published_nursery_top.store(
-            if self.inline_alloc_enabled {
-                self.nursery.top_ptr() as usize
-            } else {
-                0
-            },
-            Ordering::Release,
-        );
+        self.published_nursery_top
+            .store(self.nursery.top_ptr() as usize, Ordering::Release);
     }
 
     // ── incminimark.py:1292-1308 card marking geometry ──
@@ -4876,11 +4864,6 @@ impl GcAllocator for MiniMarkGC {
         self.published_nursery_top.as_ptr() as usize
     }
 
-    fn set_inline_alloc_enabled(&mut self, enabled: bool) {
-        self.inline_alloc_enabled = enabled;
-        self.refresh_published_nursery_top();
-    }
-
     fn max_nursery_object_size(&self) -> usize {
         self.config.large_object_threshold
     }
@@ -5525,34 +5508,6 @@ mod tests {
         let obj = gc.alloc_nursery(16);
         assert!(!obj.is_null());
         assert!(gc.is_in_nursery(obj.0));
-    }
-
-    #[test]
-    fn disabled_inline_alloc_keeps_published_top_zero_without_disabling_allocator() {
-        let mut gc = test_gc(4096);
-        gc.register_type(TypeInfo::simple(16));
-        let published_addr = gc.nursery_top_addr();
-
-        gc.set_inline_alloc_enabled(false);
-        assert_eq!(
-            unsafe { &*(published_addr as *const AtomicUsize) }.load(Ordering::Acquire),
-            0
-        );
-        assert!(!gc.alloc_nursery(16).is_null());
-        assert!(!gc.alloc_with_type(0, 16).is_null());
-
-        gc.collect_nursery();
-        assert_eq!(
-            unsafe { &*(published_addr as *const AtomicUsize) }.load(Ordering::Acquire),
-            0
-        );
-        assert!(!gc.alloc_nursery(16).is_null());
-
-        gc.set_inline_alloc_enabled(true);
-        assert_eq!(
-            unsafe { &*(published_addr as *const AtomicUsize) }.load(Ordering::Acquire),
-            gc.nursery_top() as usize
-        );
     }
 
     #[test]
--- a/majit/majit-gc/src/gc_sync.rs
+++ b/majit/majit-gc/src/gc_sync.rs
@@ -237,7 +237,7 @@ pub fn register_thread() {
         !GC_THREAD.with(|t| t.registered.get()),
         "GC mutator thread registered twice"
     );
-    let old = REGISTERED_THREADS.fetch_add(1, Ordering::SeqCst);
+    REGISTERED_THREADS.fetch_add(1, Ordering::SeqCst);
 
     let mut state = GC_SYNC.quiesce.lock().unwrap();
     state = GC_SYNC
@@ -253,14 +253,6 @@ pub fn register_thread() {
     // `rgil.acquire_maybe_in_new_thread` (rgil.py:186-193). Everything below
     // this line already runs pyre code and so needs it held.
     crate::rgil::acquire_maybe_in_new_thread();
-
-    if old > 0 && is_initialized() {
-        // gc.py:525-531 publishes the nursery slots used by generated inline
-        // allocation. A shared free-threaded nursery cannot safely use its
-        // non-atomic bump sequence, so the 1→2 transition pins that path off.
-        // This is sticky: unregistering threads never republishes the top.
-        gc_op(|gc| gc.set_inline_alloc_enabled(false));
-    }
 }
 
 /// Unregister the current thread. It stops running pyre code, so it gives the
@@ -863,18 +855,23 @@ mod tests {
         unregister_test_mutator();
     }
 
+    /// gc.py:525-531 publishes the nursery slots unconditionally, and
+    /// `llsupport/gc.py:525-531 get_nursery_free_addr/get_nursery_top_addr`
+    /// have no thread-count gate: the non-atomic bump in generated code is safe
+    /// because the GIL serialises the mutators that run it.  A second mutator
+    /// must therefore leave the published top alone.
     #[test]
-    #[ignore = "requires exclusive process — mutates the sticky singleton inline-allocation state"]
-    fn second_registered_thread_pins_published_nursery_top() {
+    #[ignore = "requires exclusive process — registers process-global mutators"]
+    fn second_registered_thread_leaves_published_nursery_top_live() {
         ensure_gc();
         register_test_mutator();
-        assert_ne!(
-            gc_query(
-                |gc| unsafe { &*(gc.nursery_top_addr() as *const AtomicUsize) }
-                    .load(Ordering::Acquire)
-            ),
-            0
-        );
+        fn published_top() -> usize {
+            gc_query(|gc| {
+                unsafe { &*(gc.nursery_top_addr() as *const AtomicUsize) }.load(Ordering::Acquire)
+            })
+        }
+        let before = published_top();
+        assert_ne!(before, 0);
 
         let worker = std::thread::spawn(|| {
             register_test_mutator();
@@ -882,13 +879,7 @@ mod tests {
         });
         blocking(|| worker.join()).unwrap();
 
-        assert_eq!(
-            gc_query(
-                |gc| unsafe { &*(gc.nursery_top_addr() as *const AtomicUsize) }
-                    .load(Ordering::Acquire)
-            ),
-            0
-        );
+        assert_eq!(published_top(), before);
         unregister_test_mutator();
     }
 
--- a/majit/majit-gc/src/gc_sync.rs
+++ b/majit/majit-gc/src/gc_sync.rs
@@ -6,47 +6,29 @@
 //! unchanged inside the STW window — it already assumes a single-threaded
 //! world during collection.
 //!
-//! # The operation gate
+//! # The operation gate is the GIL
 //!
-//! Every GC operation (alloc, collect, barrier, query) runs under one word,
-//! `IN_FAST_PATH`, holding the claimant's ident — `rpy_fastgil`
-//! (thread_gil.c:1-31), down to the spelling of its protocol: a
-//! compare-and-swap claims it (`_rpygil_acquire_fast_path`,
-//! threadlocal.h:153-156), a plain store drops it (`_RPyGilRelease`,
-//! :163-167), and a thread recognises its own claim by comparing against its
-//! ident (thread_gil.c:19-21).
+//! GC operations (alloc, collect, barrier, query) take no lock of their own.
+//! Exclusion comes from [`crate::rgil`]: a thread holds the GIL for as long as
+//! it runs pyre code, so between two external calls arbitrarily many
+//! allocations run with no synchronisation, and [`gc_op`] is a bare borrow of
+//! the singleton — `malloc_fixedsize` bumps `nursery_free` under exactly the
+//! same terms (incminimark.py, framework.py:361-402).
 //!
-//! A claim **stands between operations**, as `rpy_fastgil` stays set for as
-//! long as its thread runs RPython code, so a repeated allocation only marks
-//! the claim it already holds instead of re-taking a free word. Only the sole
-//! registered mutator claims that way; once a second registers, claims are per
-//! operation and taken under `gc_mutex`.
-//!
-//! Upstream can leave a standing claim to its owner because every blocking
-//! call is `releasegil=True` and drops it. Pyre cannot assume that of arbitrary
-//! Rust blocking, so `revoke_standing_claim` takes it back instead, the way
-//! `RPyGilAcquireSlowPath` overwrites the holder's ident once it owns the real
-//! lock (:214-223). Waiting therefore never depends on the holder running
-//! again.
-//!
-//! What upstream reads off `mutex_gil` — whether the claimant is *using* the
-//! GIL or merely holding it — pyre keeps in `GATE_INSIDE`, the low bit of the
-//! same word. Upstream needs no such distinction (holding the GIL is being
-//! inside the GC); pyre does, because its collector re-enters for read-only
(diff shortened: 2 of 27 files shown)


youknowone:
All 22 review findings (20 inline + 2 outside-diff) triaged. Two structural notes first:

- **`majit-gc: delete the stop-the-world protocol the GIL replaced` has been dropped from this PR.** It bisects to a SIGBUS in `gc_table_extra_root_walker` under `finalize_runtime`'s full collection (`MAJIT_GC_NURSERY_POISON=1 ./pyre-dynasm pyre/bench/synth/mutate_then_raise_caught.py`: 0/100 at its parent, 10/100 with it). The RCA continues separately; the commit is kept locally. What remains is the GIL port itself, which keeps the safepoint harness underneath it.
- Consequently `shadow_stack.rs`, `eval_breaker_word.rs`, `pyre-interpreter/src/eval.rs` and the backend assemblers are no longer in this PR's diff, and the findings against them are moot.

## Fixed

| # | Finding | What changed |
|---|---|---|
| codex P1 `rgil.rs:151` | Reinitialize both GIL mutexes after fork | Correct. `mutex_gil_stealer` and `mutex_gil` now live in one cell that `init_mutexes` overwrites, which is what `rpy_init_mutexes` (thread_gil.c:92-98) does through `pthread_atfork`: `pthread_mutex_init` on an already-locked mutex resets it and abandons the old state. Only re-flagging `locked` left the child inheriting a mutex locked by a thread that died in `fork()`. |
| codex P1 `gc_sync.rs:220` | Release the GIL around ctypes foreign calls | Correct. `clibffi.py:350-352` declares `ffi_call` with the default `releasegil='auto'` (`rffi.llexternal`: "Default is yes"), so `cfuncptr_call` now brackets `host_ctypes::call`. |
| CR `gc_sync.rs:247` | `BlockingGuard` `#[must_use]` + `!Send` | Both taken. `#[must_use]` went on the type rather than the function so `pyre-interpreter`'s re-export inherits it. |
| CR `lib.rs:885` | `isenabled` uses `gc_query` | Correct, and it was the only `&self` method that did. Now `gc_query_reentrant`. |
| CR `rgil.rs:152` | Assert the fork-child GIL invariant | Added, in `rgil::after_fork_child` — the function that depends on it. |
| CR `rgil.rs:282` | Strengthen the exit assertion | Correct: both steal-loop exits leave *this* thread's ident in the word, so the assertion is `am_i_holding_the_gil()`. |
| CR `rgil.rs:346` | The two tests race on the process-global GIL | Correct. Serialized on a local mutex rather than `#[ignore]`d. |
| CR `host_seam.rs:656` | Capture `errno` before releasing the GIL in `RealHost::read` | Correct, and it was the only `blocking(...)` caller reading `errno` outside the closure. It now returns `(got, last_os_error())` from inside, the shape `call_external_function` already uses. |
| CR `host_seam.rs:404` | Nested external-block guards | Confirmed: under `sandbox`, `interp_time::sleep` bracketed and then called `ops::sleep`, whose trampoline brackets again — the inner call trips `"GC mutator entered an external block twice"`. Fixed by removing the redundant outer bracket (every trampolined seam op already brackets its round trip) rather than by adding depth counting, which has no upstream counterpart. It was the only such site. |
| CR `compile.rs:211` | Public compile entries and the runtime thread | Moved into `compile_source` and `compile_source_with_opts`, the two funnels every public entry reaches; removed from `compile_eval`/`compile_exec`. |
| CR `thread/gil.rs:69` | Reclaim the worker GIL action | Correct — one leak per thread. `gil::shutdown` gives the box back from `thread_is_stopping`, last, after which nothing runs bytecode before the execution context and its actionflag are dropped. (An owning `Box` in the field is blocked by `ExecutionContext: Clone`, which `clone_for_thread` uses.) |
| CR `design.md:198` | §3.3 contradicts the mechanism | Correct. §3.3 now states free-threading as the target and the GIL as what ships, drops the "no production call sites" claim, and removes "a GIL" from the non-options list. |
| CR `gc_sync.rs:141-171` (outside diff) | Aliasing invariant for reentrant queries | The note now says what actually makes the reborrow sound: the shared borrow dies with `f` and `f` reads only; the GIL is what rules out a second *thread*. |

## Not applicable

- **codex P1 `gc_sync.rs:300` — blocking host operations.** Reviewed against `1fcf18cecd`, before `thread, modules: release the GIL around the remaining blocking host calls`. All four cited sites are bracketed on the current tip: `interp_socket.rs:2669` (`accept`), `:2930` (`recv`), `interp_select.rs:181` (`poll`), `_multiprocessing/mod.rs:154` (`sem_wait`).
- **CR `interp_posix.rs:1429` — kernel reads a Python-owned buffer.** Refuted. `file_write_buffer_bytes` returns an owned `Vec<u8>` (`bytes_like_data(obj).to_vec()`), so `libc::write` reads an interpreter-owned copy on the Rust heap; no collection or Python-side mutation can reach it.
- **CR `thread/mod.rs:71` — TLS destructor order.** Refuted. The exiting thread holds the GIL for its whole teardown and gives it back only inside `unregister_thread`, so no collector can walk the registry between `RootStackOwner::drop` and `unregister_mutator`. The ordering is arranged deliberately (`RUNTIME_THREAD` is armed after `register_mutator`) and documented at the declaration.
- **CR `gc_sync.rs:195` — acquire the GIL before publishing the mutator.** Targets `other_mutators_registered`, which exists only in the dropped commit. On this branch registration goes through the quiesce protocol: it waits out an in-flight STW request, and a thread that then blocks for the GIL leaves the RUNNING census for the wait.
- **CR `shadow_stack.rs:710`, `eval_breaker_word.rs:17`, `eval.rs:1992-1994`.** All in the dropped commit's files; out of this PR's diff.

## Valid, tracked separately

- **codex P1 `interp_socket.rs:3122` — hold writable buffer exports across the released-GIL call.** Real: `socket_writebuf` hands out a raw `&'static mut [u8]` into the exporter's `Vec` with no export count, so a concurrent resize reallocates under the syscall. Not introduced here, though — before this PR the same `recv` ran with no GIL at all and any other thread could resize it at any moment. It wants a `WritableBuffer` export across `recv_into` / `recvfrom_into` / `recvmsg_into`, which is its own change.
- **CR `thread/gil.rs:79` — `setup_threads` reports "first" per execution context.** Reading of `gil.py:25-34` is right; `gil_ready` is space-level and quasi-immutable upstream. pyre's actionflag is per-execution-context, so the flag has no space-level home yet and the sole caller discards the value.

— *commented by Claude*

youknowone:
The two findings I had put down as "valid, tracked separately" are fixed too, plus the one the parity reviewer raised on the new tip.

**`_socket: hold a writable-buffer export across the recv_into family`** (codex P1 `interp_socket.rs:3122`). This turned out to be demonstrable, so here is the demonstration rather than an argument. A thread blocked in `recv_into(bytearray(64))` while the main thread appends to the same bytearray:

| | resize while blocked | `buf[:5]` after the peer sends `b"hello"` |
|---|---|---|
| before | `allowed` | `b'\x00\x00\x00\x00\x00'` |
| after | `BufferError: Existing exports of data: object cannot be re-sized` | `b'hello'` |

The five bytes went into the freed allocation. Same binary, same script, one file reverted.

`SocketWritableBuffer` is `builtins::WritableBuffer` for the socket paths — the socket ones keep their own `writebuf_w` rejection messages, so they could not just switch over. It roots the requested object and the storage owner, takes the export for the whole call including the released-GIL window, and gives it back on drop. `socket_writebuf` now reports that owner alongside the slice: the object itself for a `bytearray`/`array.array`, the view for a `memoryview` — whose own construction already holds an export on its backing, so the storage underneath is covered either way. `recvmsg_into` holds one per iovec.

**`thread: answer setup_threads from the space, not the execution context`** (CR `thread/gil.rs:79`). `gil_ready` becomes a process-global, pyre running one object space per process, and `rgil.allocate()` runs only on the transition, ahead of publishing the flag as at `gil.py:29-31`. `threads_initialized` (`:37-38`, reached through `os_thread.py:155-156`) reads it. Registering the release action stays per-execution-context, because the ticker it hangs on is.

**`majit-gc: port RPyGilReleaseSignal`** — from the Codex parity review of `e5b5283`, and it was right. I had dismissed `RPyGilReleaseSignal` as having no counterpart because its body is `_WIN32`-only, which is backwards: the `_WIN32` guard is the *point*. thread_gil.c:258-276 says a stealer asleep in `mutex2_lock_timeout` otherwise waits out its timed wait, and Windows' 10-15 ms default timer resolution makes that "severe latency for any code that releases the GIL" — while on POSIX the 0.1 ms timeout makes the wakeup unnecessary and its scheduling interference hangs `test_interrupt_non_main`. `release()` now ends on a `release_signal()` that is empty off Windows.

Local gates on the new tip: `cargo test --all --no-default-features --features dynasm` 7380 passed / 0 failed; `cargo check` clean for the workspace, `pyre-wasm --target wasm32-unknown-unknown --features web`, and `majit-gc --target x86_64-pc-windows-msvc`; `cargo fmt --check` and the clippy gate clean; `pyre/check.py` after `scripts/extract-llbc.py` **dynasm 370/370, cranelift 370/370, wasm 366/366**.

That leaves nothing outstanding from the four reviews.

— *commented by Claude*

PR #991, opened 2026-08-02
jit: decode the int and float register banks in the multi-frame bridge recipe

youknowone:
Three commits on top of `main` (rebased onto `949b530211c`).

## 1. The int and float register banks are now decoded in the multi-frame bridge recipe

`reconstruct_inline_recipe` declined with `P2Recipe::UnboxedLiveRegister`
whenever the resume frame carried a live int or float register, seeding only
`registers_r`. `resume.py rebuild_from_resumedata` hands `consume_boxes` all
three banks (`f.registers_i, f.registers_r, f.registers_f`), and a JitCode int
temp with no `locals_cells_stack_w` home is an ordinary live register there.

Three things this needed, each one a silent miscompile if missed:

- The int/float banks stay **COLOR-indexed** while `registers_r` is inverted to
  `locals_cells_stack_w` slots. That array is a W_Root array
  (virtualizable.py:86-98), so an unboxed temp has no slot to invert a color
  into — which is exactly why it lives in `registers_i` upstream.
  `PyreSym::registers_i` is read by color, and `assemble_bridge_inline_pending`
  already copies the bank across unchanged.
- Each decoded failarg's value is stamped onto the OpRef-keyed shadow, the way
  the frame red's already was. Without it the register reaches the callee and
  the sub-walk stops at `GotoIfNotValueNotConcrete { value: InputArgInt(4) }`.
- The portal-callee branch indexed `values[frame_pos]` with a Ref-bank position
  while `values` is `[int | ref | float]`-ordered. That read correctly only
  while the removed decline guaranteed an empty int bank; those positions now
  index a `ref_values` slice.

The decode also moves ahead of the portal-callee branch so both of its returns
carry the banks, and the banks are threaded through
`setup_reconstructed_callee_frame`'s callers into `drive_bridge_frame_subwalk`,
which seeded `regs_i` from `constants_i` alone.

`synth/or_chain_fresh_alloc_arg` under `PYRE_FBW_CALLEE_VSTACK=1` is the
witness. Its kept-stack guard used to reach `2 P2Recipe::UnboxedLiveRegister` ->
`P2Drain::NoRecipes` -> no bridge, and spin:

| | before | after |
|---|---|---|
| recipe census | `2 UnboxedLiveRegister`, `NoRecipes` | `3 PortalSiblingAdmit`, `2 StreamSlotOverImage`, `CompileRoot` |
| `bridges_compiled` | 0 | **1** |
| `guard_failures` | 9653 | **201** |
| `loops_aborted` | 0 | 0 |

Reachability of the lifted decline, counted over `pyre/bench/synth` on the
parent commit: `P2Recipe::UnboxedLiveRegister` fires **0 times** in the default
configuration, and the admit census is identical either side of this commit
(274 `PortalSiblingAdmit`, 266 `StreamSlotOverImage`, 5
`MaterializedFrameAdmit`). The shape is reached only under
`PYRE_FBW_CALLEE_VSTACK=1`, so by default this admits nothing new.

## 2. Two comments this session's measurements refuted

`fbw_abort_nested_unjournaled_residual` claimed the legacy path's journal
rollback makes a declined rewind's replay exactly-once. It does so only for
journaled effects; `residual_call.rs` bumps `FBW_EXECUTED_EFFECT_COUNT` for a
non-pure residual that wrote live heap or entered a user Python frame, and
neither is journaled. An 8-line reproducer — a callee that stores an attribute
before an or-chain whose kept-stack branch guard aborts — returns 30001 where
the interpreter returns 30000.

`reconstruct_inline_recipe`'s decline cited "the encoder confirms this
empirically (both reconstructed function_calls frames decode 0 int / 0 float
registers)". `or_chain_fresh_alloc_arg` reaches it twice per run under
`PYRE_FBW_CALLEE_VSTACK=1`. (Commit 1 then removes the decline itself.)

## 3. The callee-vstack flag's two remaining costs, re-measured

Commit 1 invalidated the flag's own doc comment, which still claimed
`or_chain_fresh_alloc_arg` died at `UnboxedLiveRegister` / `NoRecipes` with
9653 guard failures and summarised the flag as "349 -> 347 passing, no fixture
improved, CPU time unchanged". Re-measured here (dynasm, fresh release build,
370 fixtures): `PYRE_FBW_CALLEE_VSTACK=1` is **2 failed / 368 passed**, both
jit-stats gates, and the two are different in kind.

**`or_chain_fresh_alloc_arg` — `loops_compiled 2 -> 1` is a merge, not a
coverage loss.** `MAJIT_LOG=1`: OFF compiles two loops, one of them keeping
`pick` residual (`final ops: total=20 call_may_force=1`), and reports `Traces
aborted: 2`, one of them `permanent=true`. ON compiles a single 164-op loop
with the callee inlined (164 -> 124 after opt), `Traces aborted: 0`, the same
one bridge and 201 vs 200 guard failures. min-of-7 interleaved child CPU:
0.165s -> **0.092s**. This is exactly the case `check.py`'s own
`JITSTATS_FALL_FIELDS` comment describes.

**`getframe_inline_subwalk_multiframe` — `guard_failures 0 -> 8948` is real,
and it is not the mirror's.** `MAJIT_GUARDLOG=1` puts 8947 of the 8948 on one
guard: `fail_index=4` = `op_index=40 opcode=GuardNotForced`, the guard after
the middle of the trace's three `CALL_MAY_FORCE` — the `sys._getframe(2)`
itself. Such a failure is never compiled ("but always just blackholed",
`compile.py:950-953 ResumeGuardForcedDescr.handle_fail`), so no bridge is
reachable for it.

`PYPYLOG=jit-summary:- pypy3` on the same fixture reports `Total # of loops: 1`,
`Total # of bridges: 0`, **`forcings: 0`**, **`virtualizables forced: 0`** and
zero aborts — PyPy emits none of the three calls, because it traces *through*
`_getframe`/`f_locals` and the vable materialization becomes in-trace
setfields rather than a runtime forcing. So the 8948 is pyre's
residual-`_getframe` deviation; the mirror only reaches it sooner (OFF never
compiles that loop: `loops_compiled 0 -> 2`, `loops_aborted 15 -> 1`).
min-of-7 child CPU 0.129s -> 0.136s.

Re-recording that baseline would bless a measured deviation, so the flag's
default stays **OFF** until `sys._getframe` is transparent to the walker. This
commit is comment-only.

## Status

`cargo fmt --all -- --check` clean; `cargo test --all --no-default-features
--features dynasm` green (101 suites ok).

| backend | result |
|---|---|
| dynasm | **370 / 370 ALL PASSED** |
| cranelift | **370 / 370 ALL PASSED** |
| wasm | **366 / 366 ALL PASSED** |

The six inherited reds reported on the previous push (`exception_args_virtual`,
`list_length_hint_validate`, `pickle_terminal_raise_resume`) are gone with the
rebase — they were `main`'s, not this branch's, and `main` has since moved past
them. The `sandbox build + e2e` red seen on this PR's earlier run was likewise
`main`'s: `949b530211c` carries *"sandbox: admit prctl(PR_SET_VMA) so mimalloc
can tag a post-lockdown mapping"*, which is exactly that failure
(`blocked syscall 157` / `_exit(159)`, intermittent because whether mimalloc's
first post-lockdown `mmap` happens at all depends on how much the workload
allocates). Rebasing onto it is the fix.


youknowone:
Thanks — I tried to reproduce the P1 and it does not hold as stated. Measurements below, all dynasm on this branch.

**The example is measurably the other way round.** The finding predicts "an attribute update before such a guard becomes 30001 instead of 30000". That reproducer exists (a callee that stores `c.pos` and then evaluates an or-chain whose kept-stack branch guard aborts):

| config | result |
|---|---|
| default (`PYRE_FBW_CALLEE_VSTACK` off) | `(50000, 30001)` — **on the parent commit too** |
| `PYRE_FBW_CALLEE_VSTACK=1` (the only config that reaches the new admission) | `(50000, 30000)` — correct |

The 30001 is a pre-existing defect with a different root: with the callee operand-stack mirror off, every inline sub-walk runs `vstack_valid == false`, Hazard (1) at `jitcode_dispatch/mod.rs` fires at any kept-stack guard, `trace.rs`'s branch flush declines on `abort_in_subwalk`, and the legacy drop replays the iteration. It reproduces identically without this PR.

**The newly-admitted shape is unreachable in the default configuration.** Counted over `pyre/bench/synth` (369 scripts) on the parent commit, `P2Recipe::UnboxedLiveRegister` fires **0 times**, and the admit census is identical either side of this commit — 274 `PortalSiblingAdmit`, 266 `StreamSlotOverImage`, 5 `MaterializedFrameAdmit`. So by default this commit admits nothing that was not already admitted. Under `PYRE_FBW_CALLEE_VSTACK=1` the reproducer above reaches `P2Drain::CompileRoot`, not an effectful abort.

**The hazard the finding describes is real but general, and already has its handler.** `trace.rs` gates the drain abort on the odometer:

```rust
let adopted = fbw_executed_effect_count() != effects_at_entry
    && try_adopt_blackhole(ctx, cf_addr, live_root_addr, WalkEndCommitLeg::CarrierAbort);
```

with the comment that once the odometer has moved the drain "may not hand the guard back to a blackhole resume from `rd_numb`". That exposure is shared by all 274 recipes admitted per corpus run; nothing about the int/float bank makes it worse, and `UnboxedLiveRegister` was never a safety property — it was a "we only seed `registers_r`" limitation, which is why upstream's `consume_boxes` fills all three banks unconditionally.

Keeping the decline would re-block the parity port to compensate for a hazard it does not cause and does not bound. The two things that do bound it are filed and named in the commit message: the bridge sub-walk's `vstack_valid: false` (which this PR's own measurement surfaced) and the branch-flush decline on `abort_in_subwalk`.

— *commented by Claude*

Issue #50, opened 2026-05-15
majit-translate auto-emit SSA per opcode handler


youknowone:
RPython:    JitCode encodes the ENTIRE interpreter (dispatch loop +
              every bytecode handler).
              → ec / frame flow naturally as SSA Register uses.
  
  pyre today: codewriter only encodes the dispatch-loop skeleton into
              SSARepr. The per-Python-bytecode handlers are emitted by
              the tracer (`pyre-jit-trace/src/trace_opcode.rs`, 7857
              lines, hand-transcribed) directly into trace IR at trace
              time.
              → ec / frame never enter the codewriter's SSA, so
                `compute_liveness` cannot see them as Register uses.
  
  Workarounds pyre currently carries (and which are not RPython-orthodox):
  
  1. emit_live_placeholder! (codewriter.rs:3805) — pushes the portal reds (frame, ec) as force-alive args of every PC's -live- op. In RPython
  the natural SSA Register uses already keep them alive; the force-alive args mechanism (liveness.py:11-12) is only needed because pyre's SSA
  never sees those uses.
  2. Encoder-side substitution at trace_opcode.rs:1475 — when the encoder encounters portal_frame_reg / portal_ec_reg it substitutes reads of
  sym.frame / sym.execution_context. RPython has no counterpart; the value is just a regular Ref-bank register.
  3. The PyJitCodeMetadata.portal_frame_reg / portal_ec_reg side channel — a pyre-only field used to remember which Ref-bank color holds each
  portal red. RPython encodes the same information implicitly via the regular inputarg position.
  
  All three only exist because the opcode handlers are not in the codewriter's SSA. M4 removes that gap, which makes all three workarounds
  deletable.

youknowone:
Static parity/design note related to this issue:

The dead-forwarder / `eliminate_empty_blocks` case supports this issue's direction. PyPy has one orthodox path:

`flow graph -> simplify_graph -> regalloc -> flatten_graph -> liveness -> assemble`

Pyre currently has a split path: the Python-bytecode walker emits SSARepr inline, while graph dual-writes are later used for regalloc/canonical checks. That split means a PyPy-equivalent graph rewrite is not always sufficient by itself. After collapsing graph links through dead mergeblock forwarders, Pyre may still need to patch already-emitted walker `TLabel`s so the byte stream and graph agree.

[...]

PR #995, opened 2026-08-02
jit: fix two backend-independent jitstats floor reds (catch-landing exc_value, str_len_descr parent)

youknowone:
Closes the two remaining backend-independent jit-stats floor reds on `main`, both by fixing JIT code to conform to the committed baselines — no baseline re-recording.

## Commit 1 — `list_length_hint_validate` (catch-landing producerless exc_value)

The catch-landing branch in `jit/codewriter.rs` pushed the caught exception value with a producerless `fresh_ref_value`. Regalloc coalesced it onto a body colour (the local list `r`); on a guard-failure bridge that re-walks the handler, that colour's def sits in the skipped prefix, so the push reads unbound (`RegisterReadUnbound{pc:456, reg:3}` ×20) and the walk aborts.

Fix: reuse the landing's own `last_exception` value Variable (which has a `generate_last_exc` producer and is re-derived on any bridge), mirroring the existing `PUSH_EXC_INFO` twin.

Measured: `list_length_hint_validate` loops_aborted 34→14, guard_failures 4923→828 — exact match to baseline.

## Commit 2 — `exception_args_virtual` (parentless const-str getfield)

`str_len_descr()` minted the `len(str)` FieldDescr parentless via `make_immutable_field_descr`. When a trace constant-folds `len(const_str)`, `protect_speculative_field` fails closed on the missing parent (`get_parent_descr() == None`), signals `InvalidLoop`, the bridge is discarded and `compile_loop` gives up with `ABORT_BAD_LOOP`.

Fix: add `W_UNICODE_DESCR_GROUP` (vtable `&STR_TYPE`, type_id `W_UNICODE_GC_TYPE_ID`, single positional `usize` `len` field), mirroring the `W_Int`/`W_Bool`/`W_Long` single-field groups, and route `str_len_descr` through `field_descr_from_group` so the FieldDescr carries a parent SizeDescr whose subclassrange validates the folded constant. This restores PyPy's `get_field_descr` invariant (descr.py:218-239, `parent_descr` set unconditionally). No `BhDescr`/serde change, no LLBC re-extraction.

Measured: `exception_args_virtual` loops_aborted 3→0 on dynasm, cranelift and wasm.

## Verification

- Full synth jit-stats gate: **353/353 on dynasm, 353/353 on cranelift**, zero collateral.
- wasm: both fixtures verified (exception_args_virtual 3→0). `pickle_terminal_raise_resume` is provably inert to commit 2 (main and branch both read 59 on macOS; the committed wasm baseline `54` is the Linux-CI value, so the local +5 is a platform read, not a regression).
- Codex parity review (scoped to the changed files): no regressions, no introduced mismatches; the str descr group's omitted GC slots and unsigned-`usize` length are intentional Rust-layout adaptations.

— *commented by Claude*

### DIFF
--- a/pyre/pyre-jit-trace/src/descr.rs
+++ b/pyre/pyre-jit-trace/src/descr.rs
@@ -853,6 +853,36 @@ static W_LONG_DESCR_GROUP: LazyLock<PyreObjectDescrGroup> = LazyLock::new(|| {
     )
 });
 
+static W_UNICODE_DESCR_GROUP: LazyLock<PyreObjectDescrGroup> = LazyLock::new(|| {
+    // `str` needs a SizeDescr parent for its `len` FieldDescr so that a folded
+    // `len(const_str)` passes protect_speculative_field: descr.py:238 sets
+    // `parent_descr` unconditionally, and the fold reaches into the parent's
+    // vtable subclassrange (str's `&STR_TYPE`) to validate the constant.
+    // Single positional field mirrors the W_Int/W_Bool/W_Long groups; str is
+    // never inline-`NewWithVtable`'d (its value buffer is a residual-call
+    // allocation), so the GC-pointer slots traced by the runtime TypeInfo
+    // (value/w_slots/index_storage) stay out of the positional list.
+    build_object_descr_group_with_def_path(
+        std::mem::size_of::<pyre_object::unicodeobject::W_UnicodeObject>(),
+        W_UNICODE_GC_TYPE_ID,
+        &pyre_object::pyobject::STR_TYPE as *const _ as usize,
+        &[(
+            // `len` (codepoint count) is a `usize`: read at usize width so a
+            // wasm32 `GetfieldGcI` does not fold in the adjacent field's high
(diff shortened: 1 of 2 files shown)


youknowone:
Superseded — both fixes already landed on `main` independently:

- **#51 str_len_descr parent SizeDescr** → landed as #989 (`b44c30bae3b`): publishes the Unicode layout as a `PyreObjectDescrGroup` and takes the length field from it, same RCA (`parent_descr` None → `protect_speculative_field` fails closed → InvalidLoop). Brings `exception_args_virtual` 3→0 and `list_length_hint_validate` 34→14.
- **#50 catch-landing exc_value** → covered by #993 (`8d5e76e5268`): names the handler's live-in (color 3, local `r`) in every `-live-` the exception table covers, fixing `list_length_hint_validate`'s 20× `RegisterReadUnbound`.

`origin/main` (949b530) already has both fixtures at their committed baselines, so this branch adds nothing (and would conflict on the duplicate `W_UNICODE_DESCR_GROUP`). Closing and deleting the branch.

— *commented by Claude*

PR #998, opened 2026-08-03
jit: recognize shared-reference niche Option<&T> with a thin ADT pointee (gh#346)

youknowone:
## Summary

Two JIT front-end slices for the gh#346 two-phase-rtyper census campaign (`rtyper-legacy`), both stacked on current `main` (#993).

### 1. `jit: fold object_items_slice from_raw_parts to the items-block base` (`c5324700590`)
Ports the `object_items_slice` `from_raw_parts` items-block-base fold onto this branch (prior task, previously reviewed).

### 2. `jit: recognize shared-reference niche Option<&T> with a thin ADT pointee` (`3a49e3ac9df`)
`tyref_is_niche_option_ptr` previously accepted only `Option<&mut T>` / `Option<NonNull<T>>`. A shared `Option<&T>` over a **Sized ADT** pointee is equally a one-word null niche (Rust guarantees `&T` non-null, identically to `&mut T`), so its `Some` payload should alias the base pointer rather than emit an aggregate `__pos_0` `FieldRead` that the annotator cannot resolve. This widening lifts the read-only `&self` interpreter accessors (`__pyre_wrap_*` getset/dunder wrappers via `from_obj(...).map(|m| &*m)`, and `getdebug_data() -> Option<&FrameDebugData>` combinator getters) that were blocking the prepass.

Fat shared refs (`&str` / `&[T]` / `&dyn` / `&<?Sized>`) are **excluded** by the `adt_node_def_id` Sized gate — their two-word layout has no one-word niche payload, and folding them would read the metadata word. One recognizer edit flips all 12 niche call sites uniformly.

Front-only change (`majit/majit-translate/src/front/mir.rs`), so no LLBC re-extract.

## Evidence

- **Census (two-phase rtyper prepass):** phaseA `1213 → 1187` (−26 real drop). `__pos_0` "Blocked block" heads `144 → 5` (the 5 residual are out-of-scope genuine two-word `Option::Some`). `collected 207 source paths` unchanged (no truncation); no new failure symbol class.
- **`check.py`:** dynasm 370/370, cranelift 370/370 — zero correctness/SNAPDIFF mismatch on both backends.
- **Unit tests:** `cargo test -p majit-translate --lib` green (synthetic Sized-gate test + two `#[ignore]` real-LLBC anchors: the `get_w_locals` win and the `set__CHUNK_SIZE` `&mut` regression, both pass on the current corpus).

## Reviewer note (documented, non-blocking)

A whole-branch review flagged that the shared-ref widening makes the `Rvalue::Discriminant` fold emit a pointer-null bool switch for `Option<&T>`, which would defeat `front::iter_next`'s literal `__discriminant`-diamond matcher. This is **latent and safe-failing**: the worst case is a census Skip (graph falls to the legacy walker = correct execution, not a miscompile), and it is currently unreached — every `.iter()` for-loop reaching `iter_next` iterates `*mut PyObject`/tuple/int (none a thin nominal ADT), confirmed by 0 `slice::iter::Iter::next` census heads and the clean check.py. The two governing doc comments (`tyref_is_niche_option_ptr` and `type_node_is_mut_ref`) are updated to record the collision, its dormancy, and the prescribed follow-up (teach `iter_next` the null-test shape) for the day a thin-ADT `.iter()` loop appears; the per-slice census is the existing tripwire.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of container-backed integer and list values during compilation.
  - Enhanced detection of optional values held through shared references while correctly rejecting unsupported wide references.
  - Preserved existing behavior for mutable-reference processing.
  - Added regression coverage to improve reliability for shared-reference payloads and sized data types.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -6859,6 +6859,22 @@ impl<'a> Lowering<'a> {
                     self.graph.set_goto(bb_id, target_bb, link_args);
                     return Ok(());
                 }
+                // `Option::<&T>::copied` / `Option::<T>::cloned` on an `Option`
+                // receiver — reading the referent value out of an `Option<&T>`
+                // is an identity in the lifted value model (`&T` and `T` are one
+                // GC pointer word, and `Option<&T>` / `Option<T>` share the
+                // niche discriminant + word layout), so alias the destination to
+                // the receiver instead of emitting a `copied` method call the
+                // rtyper cannot route on the classdef-less `Option` receiver.
+                if args.len() == 1
+                    && self.is_option_copied_identity(&reg, first_arg_ty.as_ref(), &call.dest.ty)
+                {
+                    self.local_var[dest_local] = Some(args[0].clone());
+                    let target_bb = self.block_id[target];
+                    let link_args = self.edge_args(mir_bb, target)?;
+                    self.graph.set_goto(bb_id, target_bb, link_args);
+                    return Ok(());
+                }
                 // Workspace `Index::index` / `IndexMut::index_mut`
                 // impls (`FixedObjectArray` and friends) bottom out at
                 // raw-slice construction (`as_mut_slice` →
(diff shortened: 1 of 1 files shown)


youknowone:
Both review findings were verified against source + the built LLBC corpus and fixed (2 commits pushed).

**P2 — `Option<&Wtf8>` fat-ref mis-fold (Codex `mir.rs:11491`, CodeRabbit `mir.rs:11494`).** Confirmed valid: `rustpython_wtf8::Wtf8 { bytes: [u8] }` is a DST (corpus layout `size: null`), so `&Wtf8` is a two-word fat pointer, yet it carries a nominal ADT def-id — the `adt_node_def_id(...).is_some()` gate wrongly accepted it. `str_slice_args -> Option<&Wtf8>` (matched by `startswith`/`endswith`) mis-folded to one word, dropping the length. The premise "every nominal ADT is Sized" was wrong for DST newtypes. Fixed in `0202483` by gating the shared-ref arm on the pointee's resolved layout `size.is_some()`; every `W_*`/`FrameDebugData` niche win carries a concrete size, so those still fold. Added a real-LLBC decline anchor (`str_slice_args_option_wtf8_not_niche_folded`, verified to fail on the old gate) + a Sized-gate anchor.

**P1 — list logical length lost (Codex `mir.rs:9516`).** Confirmed valid: `object_items_slice() = from_raw_parts(base, self.length)` (logical length, `< capacity` after `pop`) was folded to the `ItemsBlock` header whose gcarray descr reports **capacity**, so `w_list_getitem` would accept an index in `[length, capacity)` and return a stale slot. Fixed in `690f038` by dropping the `object_items_slice{,_mut}` arms from the fold set. `RBigInt::digits{,_mut}` uses `from_raw_parts(base, capacity)` (length == capacity), so its header alias is sound and stays. Anchors added for both.

Verification: `cargo test -p majit-translate --lib` green (+ ignored real-LLBC anchors); two-phase census phaseA 1187→1188 (+1 = the two declined **unsound** folds re-surfacing into existing orthodox classes, no new failure symbol — correctness over lift); `check.py` dynasm 370/370 + cranelift 370/370, zero mismatch.

— *commented by Claude*

PR #1010, opened 2026-08-03
interpreter: suppress the module-miss __spec__ diagnosis on swallowed attribute lookups

fregataa:
## Problem

`python3 pyre/check.py`'s `synth/import_from_hot` ran ~100x slower than pypy. Profiling showed the compiled loop's residual `IMPORT_NAME` call spending most of its time constructing an AttributeError that is immediately discarded: `dunder_import`'s fast path probes the module for `__path__` via `findattr_result`, and every miss ran the full `__spec__` shadowing diagnosis — including the `has_location` property getter, which executes a Python frame — just to phrase a message nobody sees.

## Fix

Mirror CPython's `_PyObject_LookupAttr` / `module_getattro_impl(suppress)` semantics: thread a `suppress` flag through `getattr_str_impl` → `module_getattr_fallback` → `module_getattr_hook_or_err`. Suppressed lookups still run the full protocol — the PEP 562 module `__getattr__` hook is called as before — but a terminal module miss keeps the plain error instead of building the diagnosis.

Suppressed callers (all of which swallow the AttributeError, matching CPython's use of the suppressed lookup at each site):
- `findattr` / `findattr_result` (covers `dunder_import`'s `__path__` probe)
- `builtin_hasattr`
- `builtin_getattr` when a default is supplied

Surfaced misses (`math.nope`, explicit `ModuleType.__getattribute__`, from-import failures) are unchanged; verified message parity against CPython, including the PEP 562 hook paths.

## Results

- `synth/import_from_hot`: 0.57s → 0.23s wall; vs pypy ~100x → **40.0x (dynasm) / 59.6x (cranelift)**; gate tightened `max-pypy-ratio=764` → `180`
- Full `pyre/check.py`: **ALL PASSED — dynasm 371/371, cranelift 371/371**

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `hasattr` so missing attributes return `False` without triggering unnecessary module diagnostics.
  - Improved `getattr` with a default value while preserving normal error reporting when no default is provided.
  - Refined attribute lookup for descriptors, iterators, and length hints to handle suppressed lookup consistently.

- **Chores**
  - Updated benchmark configuration and clarified related diagnostic guidance.

### DIFF
--- a/pyre/bench/synth/import_from_hot.py
+++ b/pyre/bench/synth/import_from_hot.py
@@ -1,9 +1,11 @@
-# pyre-check: max-pypy-ratio=764
+# pyre-check: max-pypy-ratio=180
 # IMPORT_FROM in a hot loop: `from math import pi, e` runs IMPORT_FROM for
 # each imported name every iteration.  The compiled per-CodeObject jitcode
 # walks the import_from residual (getattr on the peeked module, with a
 # submodule-import fallback) instead of an abort_permanent marker, so the
-# hot body JIT-compiles rather than declining to the trait leg.  Output is
+# hot body JIT-compiles rather than declining to the trait leg.  The
+# residual's `__path__` probe runs suppressed (`_PyObject_LookupAttr`), so
+# the discarded miss skips the `__spec__` shadowing diagnosis.  Output is
 # verified against CPython/PyPy.
 N = 200000
 
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -4102,12 +4102,13 @@ pub fn not_(obj: PyObjectRef) -> Result<PyObjectRef, PyError> {
     Ok(w_bool_from(!is_true(obj)?))
 }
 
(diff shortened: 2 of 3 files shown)


youknowone:
@fregataa it makes sense, thank you! other failures seems irrelevant to this change, but could you check about the cargo test failure?

fregataa:
Looked into the cargo test failure. It was the `pyre-object` lib-test binary aborting on ubuntu with `memory allocation of 8 bytes failed` (SIGABRT) partway through its 308 tests, with no single test attributable (parallel harness, the aborting test never printed its name).

I believe it's a one-off runner OOM rather than something from this PR:

- This PR only touches `pyre-interpreter` (plus a bench header); `pyre-object` is unchanged, and the `-p pyre-object --lib` test binary doesn't link `pyre-interpreter` at all.
- The same job passes on `main` (including runs newer than this PR's base), on other recent PRs, and on macOS/Windows within this very run.
- `cargo test -p pyre-object --lib` passes locally on this branch.

I've rebased onto current main and repushed, so CI is running again — will report if the failure reproduces.

youknowone:
Thank you for confirming!

PR #1013, opened 2026-08-03
jit: type-id-less field-descr parents, the missing half of verify_types, a root-walk write-back test

youknowone:
Five commits. Two are real bugs in `field_descr_ref_from_bh`, one ports a missing upstream assert, one adds a root-walk test, and the last replaces a guess in a comment with a measurement.

## 1. Every type-id-less parent shared one size descr and one field namespace

`field_descr_ref_from_bh` reconstructs a `FieldDescr` from a serialized blackhole descr by publishing its parent into the gccache under `LLType::Struct(p.type_id)`.

`type_id == 0` is not a type. It is the no-STRUCT-identity sentinel stamped by `bh_size_spec_from_callcontrol` when the analyzer-time callcontrol has no host-type carrier. Publishing under `Struct(0)` collapses **every** type-id-less parent onto whichever one was inserted first:

- `_cache_size` already holds that slot, so a second parent of a different size validates against the first one's size and vtable.
- `get_field_descr` keys `_cache_field` by **field name**, so a second parent sharing a field name resolves to the first parent's descr outright.

The trace-side twin `simple_descr_group_from_bh_size` (`pyre-jit-trace/src/descr.rs`) already carves zero out on this exact input; the metainterp side did not. Both arms now mint fresh per resolution — without an identity carrier the orthodox reading is "each of these is a distinct STRUCT".

The two arms landed separately because the second is the more damaging one: the empty-layout arm publishes one field, the `all_fielddescrs` arm publishes a whole **field table** under `Struct(0)`.

Discriminator, with the guard removed: a second parent of size 48 resolves to the first parent's size descr of 16 (24/56 in the empty-list twin). Three tests cover both arms plus the keyed path; the keyed test passes either way, which is what shows the other two target the zero path and not caching in general.

Audited the other readers of `BhSizeSpec.type_id` that key a descr cache — `pyre-jit-trace/src/descr.rs` (two sites) and `jitcode_runtime.rs` — and all three already filter zero.

Also recorded, at the keyed arm, why it passes `immutable_flag = false`: `descr.py:112` derives it from `heaptracker.is_immutable_struct(STRUCT)`, and a serialized parent spec carries no immutability hint to derive it from.

## 2. `verify_types` was half-ported

`descr.py:614-620` has two halves. `collect_call_args` carried only the second — the per-class argument counts at `:616-620`. The first, `:615 assert self.result_type in return_type`, had no counterpart anywhere in the tree.

It is not redundant with the argument check. **Nothing on the call path reads `result_type` to pick a dispatcher.** The blackhole reaches one of `bh_call_i/r/f/v` because the codewriter emitted `residual_call_*_i/_r/_f/_v`; the opcode and the descr are two independent records of the return register, and they disagree silently — the dispatcher would read the callee's result out of the register file the *opcode* named while the callee returned it in the one `result_type` names.

`verify_result_type` is called from all four entry points with the same `return_type` literals `llmodel.py:816/822/828/835` pass: `"iS"`, `"r"`, `"fL"`, `"v"`.

Two placement decisions:

- **`debug_assert`**, mirroring the `if not we_are_translated():` guard at each upstream call site. `majit-backend` is not one of the LLBC-extracted crates, so a release build genuinely drops it.
- **After** the `func == 0` early return, where upstream verifies first. That guard has no upstream counterpart — it is pyre's "no callee is installed" sentinel, and it pairs with a `BhCallDescr::default()` whose `result_type` is `'\0'`. Verifying the return register of a call that is not made would report that pairing instead of an ABI disagreement.

Note that the `#[should_panic]` tests are `#[cfg(debug_assertions)]`: `[profile.release]` in this tree sets only `lto`/`codegen-units`, so `cargo test --release` drops every `debug_assert!` and would report those two as failures-to-panic.

## 3. The root-stack walk's write-back was untested across a grow

`walk_shadow_stack_survives_push_that_grows_the_buffer` covers only the read side — that the slot visited *after* a visitor-triggered grow is read from the new buffer.

A moving collector also writes the forwarded address back, and that write happens after the visitor has had its chance to grow, so the store has to resolve the buffer address a **second** time. The new test forwards each root inside the visitor and asserts on slot 0, which was read from the old buffer before the grow.

Restoring the store to an in-buffer `&mut` leaves both existing walk tests green and fails only this one. Release-only, like its read twin — a reentrant push during a walk panics by design in debug builds.

## 4. The parentless fallback: a guess, replaced by a census

The branch carried the claim that pyre "still emits a `getfield_gc` here" and that porting `jtransform.py:942 rewrite_op_getsubstruct` would settle it. That was written without ever counting the population.

Censused the built descr table instead: **65 of 1514 Field slots are parentless**, in three families, every one a raw non-GC aggregate —

- `__pos_N` positional slots of closures / `Tuple<K,V>` / `core::alloc::layout::Layout` (~57)
- the four fields of `Arguments`, a plain Rust struct with no `#[jit_struct]`, so no GC header and no type id
- one owner-less `Adt_0`

That is exactly `rewrite_op_getsubstruct`'s domain: it accepts only `gckind == 'raw'` and lowers to `int_add(ptr, ofs)` with **no descr at all**. So the parent is not lost — upstream never mints the descr in the first place.

They are also cold: `MAJIT_LOG=1` counts **zero** resolutions through this branch on `fib_recursive`, `nbody` and `spectral_norm`. The standing cost is descr-table entries, not a declined speculation.

And synthesising a parent would be the wrong fix. Absent one, `protect_speculative_field` (`llmodel.py:555-567`) returns the decline its own `SpeculativeError` models, which is correct for a pointer that carries no type id. A fabricated parent would instead hand the `actual_tid != sizedescr.type_id()` compare an invented id, turning a safe decline into an accept whenever it happened to match.

Disposition: **won't-fix (documented)**, not deferred. The lowering stays the named convergence path if the branch ever becomes hot. Comment-only commit.

## Method note

The first attempt at (4) went into the wrong assembler. `JitCodeBuilder` in `majit-metainterp` is the *runtime* jitcode builder and emits **zero** Field descrs for pyre itself — pyre's field descrs come from the build-time `descrs.bin` side, in `majit-translate`. Instrumentation showed `fields=0 parentless=0 pending=0`, so the whole side table was reverted rather than shipped unexercised.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of call result types to prevent incompatible integer, reference, floating-point, and void dispatches.
  * Fixed field descriptor isolation for layouts without type identifiers, preventing unrelated layouts from sharing metadata.
  * Ensured relocated garbage-collection roots are correctly written back after shadow stack growth.

* **Tests**
  * Added coverage for result-type validation, descriptor caching behavior, and relocated shadow-stack roots.

### DIFF
--- a/majit/majit-backend/src/call_stub.rs
+++ b/majit/majit-backend/src/call_stub.rs
@@ -1156,6 +1156,42 @@ impl CallArgs {
     }
 }
 
+/// `descr.py:615` `assert self.result_type in return_type` — the half of
+/// `verify_types` that [`collect_call_args`] does not cover.
+///
+/// `accepted` is the caller's `return_type`, the literal each `bh_call_*`
+/// passes in `llmodel.py:816/822/828/835`: `"iS"` (`history.INT + 'S'`),
+/// `"r"`, `"fL"` (`history.FLOAT + 'L'`) and `"v"`.
+///
+/// This is not redundant with the argument-class check. Nothing on the call
+/// path reads `result_type` to pick a dispatcher: the blackhole reaches one of
+/// the four entry points because the codewriter emitted
+/// `residual_call_*_i/_r/_f/_v`, so the two records can disagree. They
+/// disagree silently — the dispatcher would read the callee's result out of
+/// the register file the *opcode* named while the callee returned it in the
+/// one `result_type` names, e.g. an f64 result taken from `rax`. Comparing
+/// them here is what makes that a failure rather than a garbage value.
+///
+/// `debug_assert` mirrors upstream's `if not we_are_translated():` guard at
+/// each call site — a translated JIT does not run this. `majit-backend` is not
+/// one of the LLBC-extracted crates, so a release build really does drop it.
+///
+/// The callers check this *after* their `func == 0` early return, where
+/// upstream verifies first. That guard has no upstream counterpart: it is
+/// pyre's "no callee is installed" sentinel, and it pairs with a descr that has
+/// no result type either — `jitdriver.rs` builds a `BhJitDriverSd` whose
+/// `portal_runner_ptr` is `None` and whose `mainjitcode_calldescr` falls back
+/// to `Default`, so `result_type` is `'\0'`. Verifying the return register of a
+/// call that is not made would report that pairing instead of an ABI
+/// disagreement.
+pub fn verify_result_type(result_type: char, accepted: &str) {
+    debug_assert!(
+        accepted.contains(result_type),
+        "BhCallDescr.verify_types: result_type {result_type:?} is not one of {accepted:?}; \
+         the emitted residual_call opcode and the calldescr disagree about the return register"
+    );
+}
+
 /// Build the C-ABI class sequence and positional argument list from
 /// `args_i` / `args_r` / `args_f`, following `calldescr.arg_classes` order.
 ///
@@ -1547,5 +1583,40 @@ mod tests {
                 &[1, 2, 3, 4, 5, 6.0_f64.to_bits() as i64],
             );
         }
+    }
+
+    /// The four `return_type` literals `llmodel.py:816/822/828/835` pass, each
+    /// against every result class `type_to_argclass` can produce. `'L'` and
+    /// `'S'` ride along with the float and int sets the way upstream spells
+    /// them (`history.FLOAT + 'L'`, `history.INT + 'S'`).
+    #[test]
+    fn verify_result_type_accepts_exactly_its_caller_s_return_classes() {
+        for (accepted, ok) in [("iS", "iS"), ("r", "r"), ("fL", "fL"), ("v", "v")] {
+            for c in ok.chars() {
+                verify_result_type(c, accepted);
+            }
+        }
+    }
+
+    /// The mis-route this exists to catch: a float-returning callee reached
+    /// through the integer entry point would take its result from `rax`.
+    ///
+    /// Only meaningful where the assertion is compiled in — `verify_result_type`
+    /// is a `debug_assert`, so a release test binary would not panic.
+    #[cfg(debug_assertions)]
+    #[test]
+    #[should_panic(expected = "is not one of \"iS\"")]
+    fn verify_result_type_rejects_a_float_result_on_the_int_entry_point() {
+        verify_result_type('f', "iS");
+    }
+
+    /// `BhCallDescr::default()` leaves `result_type` at `char::default()`.
+    /// `jitdriver.rs` reaches that through `unwrap_or_default()`, so the
+    /// sentinel has to be rejected rather than silently matching some class.
+    #[cfg(debug_assertions)]
+    #[test]
+    #[should_panic(expected = "is not one of \"r\"")]
+    fn verify_result_type_rejects_the_default_descr_s_null_result_type() {
+        verify_result_type('\0', "r");
     }
 }
--- a/majit/majit-backend/src/lib.rs
+++ b/majit/majit-backend/src/lib.rs
@@ -2701,6 +2701,8 @@ pub trait Backend: Send {
         if func == 0 {
             return 0;
         }
+        // llmodel.py:818 `calldescr.verify_types(..., history.INT + 'S')`.
+        crate::call_stub::verify_result_type(calldescr.result_type, "iS");
         // SAFETY: `func` is a valid funcptr matching the ABI recovered from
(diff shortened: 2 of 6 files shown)


youknowone:
Dispositions for the parity review's section 4. Sections 1–3 came back empty.

**`func == 0` bypasses the debug verification** — **won't-fix (documented).** Correct as-is, and the placement is deliberate: the sentinel has no upstream counterpart and pairs with a `BhCallDescr::default()` whose `result_type` is `'\0'`, so verifying the return register of a call that is never made would report that pairing rather than an ABI disagreement. Already argued in `verify_result_type`'s doc comment.

**Zero `type_id` fresh-mints instead of caching by a real `STRUCT`** — **won't-fix (documented).** This is the patch's own fix, not a residue: zero is the no-identity sentinel, so caching by it aliases unrelated parents onto one size descr and one field namespace. Already argued at both arms.

**Name-miss synthesises a field descr where upstream lowers `getsubstruct` to `int_add`** — **fixed now, as documentation.** The claim is right, and the site was the one place that did *not* cite it: its comment deferred to "the note in the empty-list branch", but that note lives in the keyed continuation of the same arm, and the empty-list branch below carries no such note. Now names `heaptracker.py:68-69` and `jtransform.py:942 rewrite_op_getsubstruct` in place (`488bddfbf9e`, comment only).

For the record, the parentless-fallback commit in this PR is the measurement behind that third item: 65 of 1514 Field slots are parentless and every one is a raw non-GC aggregate, which is exactly `rewrite_op_getsubstruct`'s domain, and `MAJIT_LOG=1` counts zero resolutions through the branch on `fib_recursive`, `nbody` and `spectral_norm`.

— *commented by Claude*

youknowone:
The re-run of the parity review found something the first pass did not, and it was right.

**§2 — `type_id == 0` mint drops `is_gc_managed`/`headerless` → fixed (`65345a21119`).**

Confirmed and real. Both zero arms called `make_simple_descr_group`, whose signature has no place for either flag and whose body hardcodes `(true, false)`, while the keyed arms beside them pass both. So the zero path was losing information the spec carried, and a raw or header-less parent came back GC-managed and headered — a `GUARD_GC_TYPE` and a header offset the object does not have.

The justification the branch shipped with was wrong, and this is the part worth recording. It claimed "the raw and header-less producers all carry a real type id because they publish through `register_struct_layout`". That holds for `bh_size_spec_from_callcontrol`, which hardcodes `is_gc_managed: true, headerless: false`. It does **not** hold for the descr→spec inverse: `bh_size_spec_from_descr` reads both flags straight off an arbitrary descr while taking `type_id` from `cache_key()`, so an unkeyed raw parent reaches these arms. I asserted the closure of that set without enumerating the producers.

`make_simple_descr_group_with_flags` now takes the shape explicitly and `make_simple_descr_group` becomes it with `(true, false)`, so every legacy mint site is byte-identical. `descr.py:105-127 get_size_descr` reads both off the STRUCT itself, so they describe the parent rather than the cache slot — which is why a missing key is not a reason to drop them.

The **trace-side twin had the same split** and is fixed in the same commit: `simple_descr_group_from_bh_size` passed both flags on its keyed arm and dropped them on its zero arm. That one is pre-existing rather than new, but it is a one-line instance of the same defect in the exact code this patch mirrors and cites, so fixing it here rather than filing it was the cheaper and less confusing outcome.

Discriminator, since a test nobody has seen fail proves nothing: with the flag pass-through removed the new test fails on the first assert, and the three sibling zero-path tests pass either way — so it targets the flags, not the carve-out.

**§3 — `'S'` (SingleFloat) arguments → deferred, blocker verified.**

`collect_call_args` panics on `'S'` where `descr.py:545-569` reads the argument out of the `args_i` bank via `int2singlefloat` and calls a `lltype.SingleFloat` signature. The blocker is checked rather than assumed: the dispatch table has **zero** `extern "C" fn(...)` arms taking `f32`, so this needs an f32 dimension across the whole matrix, not one arm. Current behaviour is a loud panic carrying its own explanation, not a miscompile. Filed with the convergence path. Note the result side already accepts it — `verify_result_type(result_type, "iS")` mirrors `llmodel.py:818` — so only the argument side is missing.

**§3 — parentless fields have no `parent_descr`** — unchanged from the earlier comment: won't-fix (documented), with the census behind it.

**§4** — the four adaptations are the ones already argued in place; the `debug_assert!` ↔ `we_are_translated()` item is the language-gap mapping this tree already uses as convention.

— *commented by Claude*

PR #1018, opened 2026-08-04
gateway, pycode: split the fsencode boundary; report the byte-exact co_filename from every reader

youknowone:
Follow-up to #1001, closing the two items its review left open.

## 1. Split the fsencode boundary into the converter and the raw encode

Upstream has two distinct things here and the difference is load-bearing:

- `baseobjspace.py:1970-1983 space.fsencode_w` — the argument converter. It warns
  `path should be %s, not %s` as a `DeprecationWarning` for a readable buffer
  that is not `bytes`, and ends in `bytesbuf0_w` (`:2005-2018`), which rejects an
  embedded null with `ValueError: embedded null byte`.
- `baseobjspace.py:1962 space.fsencode` — the plain unicode→bytes filesystem
  encode, with neither. `interp_socket.py:157` takes this one deliberately:
  *"Not using space.fsencode_w since Linux allows embedded NULs."*

pyre collapsed both into one encoder, so it had neither the warning nor the null
check anywhere, and could not express the socket carve-out at all.

`fsencode_bytes_w` now follows `fsencode_w`: it resolves a readable buffer before
`__fspath__`, warns for a buffer that is not `bytes`, names the types in both
`TypeError`s (`interp_posix.py:3043` and `:3053`), and applies the null check.
`gateway::fsencode` is the raw form, and the AF_UNIX address packer takes it.

`open()` drops its own hand-rolled str / bytes-like / `__fspath__` resolution for
the converter, the way `interp_fileio.py:134-137 _open_fd` reaches
`interp_posix.py:64 space.fsencode_w` through `dispatch_filename`.

The four resulting messages match CPython 3.14 character for character:

```
TypeError: expected G.__fspath__() to return str or bytes, not int
TypeError: expected str, bytes or os.PathLike object, not object
ValueError: embedded null byte          # open("a\0b")
ValueError: embedded null byte          # os.system("a\0b")
```

## 2. Report the byte-exact `co_filename` from every reader

`PyCode` owns the raw path (`pycode.py:135 self.co_filename = filename`) and the
`co_filename` getter decodes it with `objspace.py:438 newfilename`. Every other
reader still went to `CodeObject::source_path` — the lossy UTF-8 spelling, where
a path byte with no UTF-8 form has already become U+FFFD. So a module imported
from `x\udcff.py` showed `x<?>.py` in its traceback, its warnings and its frame
repr while `code.co_filename` showed the correct name.

`w_code_filename_obj` realizes the filename the way `w_code_name_obj` realizes
`co_name`. It deliberately does **not** cache on the `PyCode`: the wrapper is
prebuilt-immortal and the decode hands back a movable managed string, which is
exactly the combination the `filename_bytes` field doc rules out.

Routed through it: the warning location, a coroutine's `cr_origin` summary, the
rendered traceback, and `pyframe.py:849-853 descr_repr`.

The last two were the half that looked blocked, because a Rust `String` cannot
hold the lone surrogate:

- the traceback writer is `std::io::Write`, so it keys its
  `StackSummary.format` dedup on the filename **bytes** and emits them with
  `write_all`. `read_source_line` keeps the lossy spelling, which simply fails to
  resolve — `error.py:150 linecache.getline` renders nothing then too.
- `descr_repr` returns a `Wtf8Buf`, and `py_repr_wtf8` takes it directly instead
  of folding through `py_repr`'s `String`. Upstream interpolates the filename
  (`"%s" % code.co_filename`) rather than its repr, so the surrogate reaches the
  result unescaped; CPython spells the same field with `%R`. Both keep the byte.

## Verification

All four readers now report the exact name (`x\udcff.py`), and the traceback's
bytes are `78 ff 2e 70 79`:

```
frame_repr     EXACT "<frame at 0x…, file 'x\udcff.py', line 2, code cap>"
warn_filename  EXACT 'x\udcff.py'
coro_origin    EXACT 'x\udcff.py'
```

New parity test `traceback_filename_surrogateescape.py`, plus assertions added to
`os_boundary_surrogateescape.py` for the null boundary and for the AF_UNIX
abstract-namespace carve-out. Both green on CPython 3.14 and on pyre.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved non-standard filesystem filenames, including surrogate-escaped characters, in tracebacks, warnings, frame displays, and coroutine diagnostics.
  * Improved filesystem path handling, including clearer validation and rejection of embedded NUL bytes.
  * Added support for Unix abstract socket names and more accurate socket address handling.
* **Tests**
  * Added coverage for path validation, abstract sockets, and surrogate-escaped filenames across error and diagnostic output.

### DIFF
--- a/pyre/extra_tests/parity_tests/os_boundary_surrogateescape.py
+++ b/pyre/extra_tests/parity_tests/os_boundary_surrogateescape.py
@@ -49,14 +49,17 @@ for call in (
 
 # `interp_socket.py:157-159` deliberately uses raw `space.fsencode`: a leading
 # null byte names Linux's abstract namespace and must reach the OS unchanged.
+name = "\x00pyre-os-boundary-%d" % os.getpid()
 s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
 try:
     try:
-        s.bind("\x00pyre-os-boundary-%d" % os.getpid())
+        s.bind(name)
     except ValueError:
         raise AssertionError("AF_UNIX abstract path rejected its null") from None
     except OSError:
         pass
+    else:
+        assert s.getsockname() == name.encode(), s.getsockname()
 finally:
     s.close()
 
--- a/pyre/extra_tests/parity_tests/os_boundary_surrogateescape.py
+++ b/pyre/extra_tests/parity_tests/os_boundary_surrogateescape.py
@@ -35,6 +35,31 @@ os.waitpid(pid, 0)
 pid = os.posix_spawn(TRUE, [TRUE], {"PYRE_X": "v\udcff"})
 os.waitpid(pid, 0)
 
+for call in (
+    lambda: os.system("true a\x00b"),
+    lambda: os.posix_spawn(TRUE, [TRUE, "a\x00b"], {}),
+    lambda: open("a\x00b"),
+):
+    try:
+        call()
+    except ValueError as exc:
+        assert "embedded null byte" in str(exc), str(exc)
+    else:
+        raise AssertionError("embedded null byte was accepted")
+
+# `interp_socket.py:157-159` deliberately uses raw `space.fsencode`: a leading
+# null byte names Linux's abstract namespace and must reach the OS unchanged.
+s = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+try:
+    try:
+        s.bind("\x00pyre-os-boundary-%d" % os.getpid())
+    except ValueError:
+        raise AssertionError("AF_UNIX abstract path rejected its null") from None
+    except OSError:
+        pass
+finally:
+    s.close()
+
 # AF_UNIX: bind to a path whose last byte has no UTF-8 spelling.  Whether the
 # kernel accepts it is a platform question — macOS refuses a non-UTF-8 path
 # with EILSEQ — so the assertion is that the *encoding* is not what refuses:
--- a/pyre/extra_tests/parity_tests/traceback_filename_surrogateescape.py
+++ b/pyre/extra_tests/parity_tests/traceback_filename_surrogateescape.py
@@ -1,0 +1,80 @@
+"""Every reader of `co_filename` reports the byte-exact filesystem spelling.
+
+`pycode.py:135 self.co_filename = filename` keeps the raw path, so the frame
+repr (`pyframe.py:849-853 descr_repr`), the warning location, a coroutine's
+origin and the rendered traceback all name what the `co_filename` getter hands
+out — a path byte with no UTF-8 spelling must not fold to U+FFFD on the way.
+"""
+
+import _imp
+import io
+import subprocess
+import sys
+import traceback
+import warnings
+
+FILENAME = "x\udcff.py"
+
+
+def stamped(source):
+    """A code object carrying a filename no UTF-8 encoder can spell."""
+    code = compile(source, "orig.py", "exec")
+    _imp._fix_co_filename(code, FILENAME)
+    return code
+
+
+# The stdlib formatter reads `co_filename` directly, so this is the anchor the
+# rest are measured against.
+try:
+    exec(stamped("raise ValueError('boom')"), {})
+except ValueError:
+    buf = io.StringIO()
+    traceback.print_exc(file=buf)
+    text = buf.getvalue()
+assert FILENAME in text, ascii(text)
+assert "�" not in text, ascii(text)
+
+# `descr_repr` interpolates the filename rather than its repr, so the surrogate
+# reaches the result unescaped; CPython formats that field with `%R` and spells
+# it `x\udcff.py`.  Both keep the byte, so only the folding is asserted here.
+namespace = {"sys": sys}
+exec(stamped("def capture():\n    return sys._getframe()\nframe = capture()"), namespace)
+frame_repr = repr(namespace["frame"])
+assert "�" not in frame_repr, ascii(frame_repr)
+
+# The warning's location comes from the frame's own code object.
+with warnings.catch_warnings(record=True) as caught:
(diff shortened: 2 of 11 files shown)


youknowone:
Adjudicated the Codex parity review. Three findings fixed, one refuted, one deferred with the blocker named.

### §1 — refuted

> the new shared `fsencode_bytes_w` rejects NUL bytes … PyPy's `pathname='fsencode'` uses raw `space.fsencode`

`unwrap_spec` `'fsencode'` does not reach `space.fsencode`. `gateway.py:366` and
`gateway.py:578` both expand `visit_fsencode` to `space.fsencode_w(...)`, which
ends in `bytesbuf0_w` — so `_imp._fix_co_filename(code, "a\0b")` raises
`ValueError: embedded null byte` upstream too, exactly as it now does here.

CPython does diverge: it accepts the name and stores `'a\x00b'`. That is a
CPython-vs-PyPy difference, and this repo follows PyPy.

### §3 — two fixed, one deferred

- **`__fspath__` returning `bytearray`** — fixed. `interp_posix.py:3049-3051`
  narrows the result to `str` or `bytes`; a readable buffer is not a path. The
  message now matches upstream and CPython character for character:
  `expected B.__fspath__() to return str or bytes, not bytearray`.

- **AF_UNIX abstract namespace** — fixed, and it was the larger of the two.
  `unpack_inet_addr` had no address length at all, so it could not compute
  `rsocket.py:432`'s `maxlength` and the NUL scan stopped at byte 0. It now
  takes the length, every one of the nine callers passes what the OS wrote back
  (`slen`, `msg_namelen`, `ai_addrlen`), and `interp_socket.py:45` returns the
  abstract name as `bytes`. The packing side gained the mirror fix:
  `rsocket.py:409-410`'s `baseofs + len(path)`, with the terminator counted only
  for a regular name, and the offset taken from `sockaddr_un` rather than from
  `sa_family_t` — not the same on every platform.

- **traceback source lookup** — fixed. The lossy spelling reached the source
  provider, so a surrogate-named file could never render its offending line.
  The bytes are carried through and turned into a `Path` with
  `OsStr::from_bytes`.

- **`memoryview` as a path** — deferred. `baseobjspace.py:1971` accepts any
  `BUF_FULL_RO` exporter; this boundary has no buffer-protocol entry point to
  call, only the `bytes | bytearray` predicates. Closing it needs that seam
  first, which is more than this diff should carry.

### §4 — documented, no change

Non-stable `co_filename` object identity is the deliberate adaptation the
`filename_bytes` field doc describes: the wrapper is prebuilt-immortal and the
decode hands back a movable managed string, so caching one on it would need
per-collection forwarding. It predates this diff — the previous getter also
built a fresh string per read — and the value is unchanged either way.

### Verification

`cargo fmt --check` clean; full `extra_tests` dynasm run is 232/241 both before
and after, with the same nine pre-existing failures (NaN ordering ×3, ast
identity, `dict.fromkeys.__qualname__`, `sys._getframemodulename`, absent
`_sqlite3`, absent `__pypy__.builders`, the known unpickler desync). The LLBC
was re-extracted after the rebase, and the numbers are identical across that
too. The abstract-namespace branch is compiled out on macOS, so only Linux CI
executes it; the regular AF_UNIX bind/`getsockname()` round trip is verified
locally and matches CPython.

— *commented by Claude*

youknowone:
Rebased onto `origin/main` and re-verified. The `pyre/check.py` reds on all three OSes are base-owned.

### Attribution

main's own CI run at the **same base sha** (`ff42f0e3317`, #950) fails the same three jobs. The failure sets are identical, not merely similar:

| | main @ base | this branch |
|---|---|---|
| ubuntu | `dynasm 8 / cranelift 16 / wasm 14` — 38 FAIL lines | identical, same 38 lines |
| macOS | `dynasm 8 / cranelift 16` | identical |
| windows | `dynasm 8 / cranelift 17`, same 25-bench list | character-identical |

Diffing the two ubuntu FAIL lists leaves nothing on either side — same benches, same counters, same numbers (`loops_compiled 407 -> 409`, `guard_failures 19702 -> 802`, and so on).

That is what you would expect from #950: it moved `PyCode` wrappers from Box-immortal to managed stable-oldgen, which changes trace shape across the exception, gc and raise benches, and the recorded baselines predate it. Nothing in this branch touches the optimizer or adds a JIT-visible struct field.

I deliberately did not re-record any baseline here — that would absorb main's drift into this diff.

### Rebase notes

Two conflicts, both blended rather than taken from one side:

- **`pycode.rs`** — main reworded the wrapper-family doc ("Bootstrap/prebuilt"), we inserted `w_code_filename_obj` at the same point. Kept both. The function's doc justified not caching by calling `PyCode` prebuilt-immortal, which #950 made stale, so it now states the reason that survives the model change: the filesystem decode returns a movable managed string, so a retained slot would need a managed wrapper's custom trace *and* `eval::walk_raw_code_roots` for the bootstrap family. Corrected the same stale claim in the `filename_bytes` field doc it cites.
- **`error.rs`** — main switched the traceback source read to `read_source_bytes` + `decode_source_bytes`, since `linecache` opens with `tokenize.open` and the BOM/PEP 263 cookie decide the encoding. We had changed the same call to take a byte-derived `Path`. Kept main's tokenize-aware decode over our path, so a non-UTF-8 source name both resolves *and* decodes by its declared encoding.

The other three commits are byte-identical across the rebase (same patch-ids).

### Verification after the rebase

LLBC re-extracted, release rebuilt, full `extra_tests` dynasm run: **234/241**, up from 232 — `builtin_compile.py` and `pickle_pypy_module.py` now pass on the new base. The remaining seven are the same pre-existing failures as before (NaN ordering ×3, `dict.fromkeys.__qualname__`, `sys._getframemodulename`, absent `_sqlite3`, the known unpickler desync). No new failure.

— *commented by Claude*

PR #1020, opened 2026-08-04
gc: separate a failed managed allocation from an absent GC

youknowone:
`try_gc_alloc*` returns `None` when no hook is installed and `Some(null)` when an
installed hook could not satisfy the request. Four sites collapsed the two with
`.filter(|p| !p.is_null())` and fell back to `lltype::malloc_raw` for both, so a
*failed* managed allocation silently produced a headerless object.

| site | traced edge | walker |
|---|---|---|
| `alloc_rbigint_nursery_impl` | `_digits` | `W_LongObject`'s `LONG_VALUE_OFFSET` — plain offset walker, **no `try_gc_owns_object` guard** |
| `alloc_rbigint_nursery_collecting_impl` | same | same |
| `alloc_rbigint_stable` | same | same |
| `alloc_frame_block` | `previous` | chain head is guarded, but the interior `previous` is the unguarded offset walker |

Reachable, not theoretical: `try_alloc_with_type_no_collect_body` returns
`GcRef(0)` once the nursery bump and `spill_to_oldgen_or_null` both fail. The
result is a `Box::into_raw` payload forwarded as though it carried a header —
the failure class behind the SIGBUS fixed in #994.

## Change

`GcAllocOutcome::{Allocated, Failed, NoRoute}` with `from_hook` and
`allocated_or_abort`, plus `try_gc_alloc_stable_or_abort` for the
`dont_look_inside` raw-return callers. `Failed` reaches `gc_alloc_failed`
(`#[cold] #[inline(never)] -> !`), which calls `handle_alloc_error` exactly as
`alloc_typed_items_block_nursery` already does for the digit array — so an
rbigint payload now fails the same way its own digits do. `NoRoute` alone
returns the caller to its `malloc_raw` path.

Upstream has neither state: the GC is a prebuilt constant
(`framework.py:254`), and a nursery that cannot satisfy the request reaches
`collect_and_reserve` (`incminimark.py:981-985`), which raises MemoryError
rather than returning null. Abort rather than panic because these run under JIT
frames that cannot unwind.

What makes the abort safe: `init_gc_subsystem` runs `install_gc_into_backend()`
before `install_pyre_object_hooks`, so once a pyre-object hook is visible the
`set_active_*` cells are too — `Some(null)` can never mean "no backend".

## Deliberately unchanged

Sites whose mixed GC/raw population is supported and guarded, judged by the
walker rather than by the fallback spelling:

- `gc_alloc_storage_box` — `set_object_custom_trace` guards `set.items` with
  `try_gc_owns_object` ("A no-GC-hook fallback allocation is not
  collector-owned"). I changed this one first and reverted it.
- `clone_debugdata_ptr` / `getorcreate_debug_data` — guarded, with a deliberate
  in-place walk for Box payloads.
- `alloc_dict_object`, `w_long_new` — fall back to `malloc_typed`, which
  prepends a `GcHeader`.

Also replaces the module doc's claim that the `Box::into_raw` fallback is
dropped "as the hook's reliability is verified under the full bench suite". That
criterion cannot be met: the state that matters is `Some(null)`, which a green
benchmark never exercises.

## Verification

Green on base `7b0aca4bc8e`: `check.py` dynasm 371/371, cranelift 371/371, wasm
367/367, with HEAD unchanged across the run.

On the current base, `extract-llbc.py`, the release build, and `pyre-object`
312 / `pyre-interpreter` 490 / `majit-gc` 231 all pass. The full local
`check.py` was **not** completed on this base: the shared working tree was
rebased twice mid-run (13:07 and 15:17), the second time pulling in #1019, which
re-records jit-stats baselines — so the comparison's baseline changed under a
binary built from the previous base. Those runs were discarded rather than
reported as results. CI is the gate for this base.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory allocation safety by preventing managed allocations from silently falling back to untracked memory.
  * Allocation failures under active garbage collection now terminate safely instead of risking incorrect memory tracking.
  * Preserved fallback behavior when no garbage-collection route is available.

* **Tests**
  * Added coverage for allocation outcomes, failure handling, and fallback eligibility.

### DIFF
--- a/pyre/pyre-interpreter/src/pyframe.rs
+++ b/pyre/pyre-interpreter/src/pyframe.rs
@@ -1080,11 +1080,15 @@ unsafe fn alloc_frame_block(
     allocation: FrameLocalsArrayAllocation,
 ) -> *mut FrameBlock {
     if allocation == FrameLocalsArrayAllocation::OldGenGc {
-        if let Some(raw) = pyre_object::gc_hook::try_gc_alloc(
-            FRAME_BLOCK_GC_TYPE_ID,
-            std::mem::size_of::<FrameBlock>(),
+        // `FRAME_BLOCK_GC_TYPE_ID` registers `previous` as a traced edge, so
+        // the walker forwards the rest of the chain unconditionally once it
+        // reaches a managed block. A `malloc_raw` block spliced in after a
+        // failed managed allocation would be forwarded without a header.
+        let payload_size = std::mem::size_of::<FrameBlock>();
+        if let Some(raw) = pyre_object::gc_hook::GcAllocOutcome::from_hook(
+            pyre_object::gc_hook::try_gc_alloc(FRAME_BLOCK_GC_TYPE_ID, payload_size),
         )
-        .filter(|raw| !raw.is_null())
+        .allocated_or_abort(payload_size)
         {
             unsafe { std::ptr::write(raw as *mut FrameBlock, block) };
             return raw as *mut FrameBlock;
--- a/pyre/pyre-object/src/gc_hook.rs
+++ b/pyre/pyre-object/src/gc_hook.rs
@@ -13,11 +13,18 @@
 //! the same pointer to every thread, and a collector running on an arbitrary
 //! thread must see it even if that thread never ran the install path.
 //!
-//! Callers use [`try_gc_alloc`] which returns `None` when no hook is
-//! installed — they fall back to the `Box::into_raw` path in
-//! that case. Incremental migration drops the `Box::into_raw`
-//! fallback at each call site as the hook's reliability is verified
-//! under the full bench suite.
+//! Callers use [`try_gc_alloc`], whose two non-pointer answers mean
+//! different things: `None` is *no GC owns the heap*, `Some(null)` is *a GC
+//! owns the heap and this allocation failed*. Only the first licenses the
+//! `Box::into_raw` path. [`GcAllocOutcome`] names the two so a call site
+//! cannot collapse them: substituting a raw object for a failed managed one
+//! puts a headerless payload into the traced graph, where an owner field
+//! registered as a gc-pointer offset forwards it as though it had a header.
+//!
+//! Dropping the `Box::into_raw` fallback at a call site is therefore a
+//! question of whether that site can still run before `init_gc_subsystem`,
+//! not of how reliable the hook has proven under a bench suite: the state
+//! that matters is `Some(null)`, which a green benchmark never exercises.
 //!
 //! Layering: this module defines the function-pointer slots only. Wire-up
 //! lives in `pyre-jit`.
@@ -127,6 +134,70 @@ pub fn clear_gc_alloc_hook() {
 #[inline]
 pub fn try_gc_alloc(type_id: u32, payload_size: usize) -> Option<*mut u8> {
     GC_ALLOC_HOOK.get().map(|f| f(type_id, payload_size))
+}
+
+/// What an allocation hook answered, with the two non-pointer states kept
+/// apart.
+///
+/// `malloc_fixedsize` (incminimark.py:640-693) has neither state. The GC is a
+/// prebuilt constant (framework.py:254), so a route always exists, and a
+/// nursery that cannot satisfy the request reaches `collect_and_reserve`
+/// (incminimark.py:981-985), which raises MemoryError rather than handing a
+/// null back. Both states are pyre's own, and they are not interchangeable:
+///
+/// * [`NoRoute`](Self::NoRoute) — nothing owns the heap yet: a bare unit
+///   test, the pre-`init_gc_subsystem` bootstrap, or a build with no backend.
+///   The caller's `malloc_raw` path *is* the whole heap there, so taking it
+///   keeps the object graph consistent.
+/// * [`Failed`](Self::Failed) — a GC owns the heap and could not satisfy the
+///   request. `malloc_raw` is the wrong answer now: the object would hold
+///   managed references the collector never traces or forwards, and its
+///   missing header lets a type-id witness misread the words before it.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum GcAllocOutcome {
+    Allocated(*mut u8),
+    Failed,
+    NoRoute,
+}
+
+impl GcAllocOutcome {
+    /// Classify a hook result: `None` is [`NoRoute`](Self::NoRoute),
+    /// `Some(null)` is [`Failed`](Self::Failed).
+    #[inline]
+    pub fn from_hook(result: Option<*mut u8>) -> Self {
+        match result {
+            Some(raw) if !raw.is_null() => Self::Allocated(raw),
+            Some(_) => Self::Failed,
+            None => Self::NoRoute,
+        }
+    }
+
+    /// The allocated pointer, or `None` for [`NoRoute`](Self::NoRoute) so the
+    /// caller takes its own non-GC path. A [`Failed`](Self::Failed) does not
+    /// return: see [`gc_alloc_failed`].
+    #[inline]
+    pub fn allocated_or_abort(self, payload_size: usize) -> Option<*mut u8> {
+        match self {
+            Self::Allocated(raw) => Some(raw),
+            Self::Failed => gc_alloc_failed(payload_size),
+            Self::NoRoute => None,
+        }
+    }
+}
+
+/// A GC that owns the heap could not satisfy an allocation.
+///
+/// `collect_and_reserve` (incminimark.py:981-985) raises MemoryError at this
+/// point, so no caller of `malloc_fixedsize` observes a null. The pyre callers
+/// return a bare pointer and run under JIT frames that cannot unwind, so the
+/// failure aborts instead — the answer `alloc_typed_items_block_nursery`
+/// (`object_array.rs`) already gives for a digit array whose allocation fails.
+#[cold]
+#[inline(never)]
+pub fn gc_alloc_failed(payload_size: usize) -> ! {
+    let layout = std::alloc::Layout::from_size_align(payload_size, std::mem::align_of::<usize>())
+        .unwrap_or_else(|_| std::alloc::Layout::new::<usize>());
+    std::alloc::handle_alloc_error(layout)
 }
 
 /// Install the `malloc_fast` allocation callback.
@@ -242,6 +313,19 @@ pub fn try_gc_alloc_stable(type_id: u32, payload_size: usize) -> Option<*mut u8>
 #[majit_macros::dont_look_inside]
 pub fn try_gc_alloc_stable_raw(type_id: u32, payload_size: usize) -> *mut u8 {
     try_gc_alloc_stable(type_id, payload_size).unwrap_or(core::ptr::null_mut())
+}
+
+/// [`try_gc_alloc_stable_raw`] for a caller whose fallback is `malloc_raw`.
+///
+/// Returns null only for [`GcAllocOutcome::NoRoute`]; a GC that owns the heap
+/// and then fails aborts rather than letting the caller substitute an untraced
+/// object. Keeps the raw return so the residualised call carries no
+/// discriminant, for the reason [`try_gc_alloc_stable_raw`] documents.
+#[majit_macros::dont_look_inside]
+pub fn try_gc_alloc_stable_or_abort(type_id: u32, payload_size: usize) -> *mut u8 {
+    GcAllocOutcome::from_hook(try_gc_alloc_stable(type_id, payload_size))
+        .allocated_or_abort(payload_size)
+        .unwrap_or(core::ptr::null_mut())
 }
 
 majit_gc::global_hook!(static GC_ALLOC_COLLECTING_HOOK: GcAllocHookFn);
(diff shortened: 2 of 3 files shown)


youknowone:
`pyre/check.py (macos-latest)` is red here, and it is **inherited from the base, not introduced by this PR**.

The same job is already `completed/failure` on three consecutive `main` commits, including this PR's base `ff42f0e3317`:

| commit | `pyre/check.py (macos-latest)` |
|---|---|
| `ff42f0e3317` (this PR's base) | failure |
| `7fd8d544b74` (#1019) | failure |
| `4d7588324eb` (#1014) | failure |

Comparing the failing sets keyed on backend + fixture + **the counter delta itself**, not just the fixture name:

```
main @ ff42f0e3317 : 24 fails    this PR : 24 fails
only in PR   : (empty)
only in main : (empty)
```

Byte-identical, e.g. `cranelift synth/exception_bridge_traceback_head  jit-stats change: loops_aborted 1 -> 0, bridges_compiled 6 -> 4, guard_failures 1203 -> 802` appears on both sides with the same numbers. This PR adds no jit-stats change of its own, which is what you would expect: the only new code is a `#[cold] #[inline(never)] -> !` arm taken when a managed allocation fails.

I have **not** re-recorded any baseline to make this green. The counters moved for reasons that belong to the base, and re-recording would bury a real signal that is not mine to bury.

Locally this branch was green on base `7b0aca4bc8e` — `check.py` dynasm 371/371, cranelift 371/371, wasm 367/367, with HEAD verified unchanged across the run.

— *commented by Claude*

youknowone:
All checks have settled. Extending the earlier control to **all three platforms**: every `pyre/check.py` failure here is inherited from the base `ff42f0e3317`, and this PR adds none of its own.

| platform | base `ff42f0e3317` | this PR | only in PR | only in base |
|---|---|---|---|---|
| macos-latest | 24 fails | 24 fails | — | — |
| ubuntu-24.04 | 38 fails | 38 fails | — | — |
| windows-latest | 25 fails | 25 fails | — | — |

Sets compared on backend + fixture + the counter delta text, not the fixture name alone. All 87 entries are a single category, `jit-stats change` — no crash, no test failure, no new category. The `pyre/check.py` job is `completed/failure` on `ff42f0e3317`, `7fd8d544b74` (#1019) and `4d7588324eb` (#1014), so `main` cannot currently take a PR through this gate green.

Everything else passes (15/15): `cargo test` on all three platforms, CPython suite gate, sandbox e2e, both wasm builds, Charon/LLBC prepare on all three, fmt, pre-commit, and both review gates.

No baseline was re-recorded. The counters moved for reasons owned by the base, and re-recording them here would bury a signal that is not mine to bury.

— *commented by Claude*

youknowone:
Force-pushed: rebased onto `a8f3cdb36ed` and dropped `a06cbb4fe04` ("check: record managed-allocation JIT stats"). The branch now carries the GC commit only.

Why the re-record went away rather than getting narrowed to its improvements: main's `7a7d9174088` (#945) re-recorded 108 `.jitstats` baselines after this branch was cut, and it already covers almost all of that commit. Comparing its 38 files against current `origin/main`:

* **34 of 38 are byte-identical to main** — a no-op after the rebase.
* The other 4 differ only because they were observed on the pre-#945 base:

| file | current `origin/main` | dropped commit |
|---|---|---|
| `exception_inline_callee_tb_frames.wasm` | `guard_failures=860` | `975` — rise |
| `exception_traceback_frame_lineno.cranelift` | `guard_failures=811`, `loops_compiled=17` | `808`, `14` — loops fall |
| `raise_reg_unbound_jitstress.cranelift` | `loops_aborted=1`, `loops_compiled=6` | `2`, `8` — abort rise |
| `raise_reg_unbound_jitstress.dynasm` | `loops_aborted=1`, `loops_compiled=6` | `2`, `8` — abort rise |

Each of those was measured against a tree without #945, so writing it on top of #945 would pin a number that does not describe the tree — and two of the four move in the bad direction anyway.

This also supersedes the inherited-red report I posted earlier: the 24/38/25 `jit-stats change` rows at base `ff42f0e3317` were main's own stale baselines, and #945 is what recorded them. Any counter that still moves at the new base will be recorded from a run at *that* base, improvements only.

— *commented by Claude*

youknowone:
**Correction to my previous comment.** I wrote that all four DIFF files "were measured against a tree without #945, so writing them on top of #945 would pin a number that does not describe the tree." That is wrong for three of the four.

`git log ff42f0e3317..a8f3cdb36ed -- <file>` per file:

| file | touched by #945? | last changed by |
|---|---|---|
| `raise_reg_unbound_jitstress.dynasm` | no | `c1e2c017550` (#1003) |
| `raise_reg_unbound_jitstress.cranelift` | no | `c1e2c017550` (#1003) |
| `exception_inline_callee_tb_frames.wasm` | no | `a0314863dd5` (#1009) |
| `exception_traceback_frame_lineno.cranelift` | **yes** → `guard_failures=811, loops_compiled=17` | `7a7d9174088` (#945) |

A file #945 never touched is not stale by construction, so only the last row was the case I described.

And a bracketed local run at the new base (`HEAD_BEFORE == HEAD_AFTER == a4ce0376fbe`, `check.py --synthetic-only`) observes exactly the dropped numbers: `raise_reg_unbound_jitstress` `loops_aborted 1 -> 2, loops_compiled 6 -> 8` on both dynasm and cranelift, and `exception_inline_callee_tb_frames.wasm` `guard_failures 860 -> 975`. Those three do describe the current tree.

What that does not change: they are still not recorded on this branch. Each moves a counter the wrong way — `loops_aborted` rise (×2), and a `guard_failures` rise with `loops_compiled`/`bridges_compiled` unchanged at 6/4 — and this branch records improvements only.

They are inherited regardless: main's own run at `a8f3cdb36ed` (ubuntu, job `91949366146`) reports all of them, 46 rows in total. 42 of those 46 are `loops_compiled`-fall rows on dynasm/cranelift (`arith_int_bool 8 -> 7`, `ast_compile_roundtrip 1 -> 0`, `sre_*`, `type_dotted_name`, …) that do **not** reproduce on macOS, so main also has a host-divergence class open on that gate. Windows reports 43 rows.

— *commented by Claude*

youknowone:
Set diff against main's own run at this branch's base (`a8f3cdb36ed`), all three platforms, keyed on backend + fixture + counter-delta text:

| platform | main `a8f3cdb36ed` | PR `a4ce0376fbe` | only in PR | only in main |
|---|---|---|---|---|
| macos-latest | 2 | 2 | — | 4 (perf, not jit-stats) |
| ubuntu-24.04 | 46 | 46 | — | — |
| windows-latest | 43 | 43 | — | — |

Per-backend counts match exactly too (ubuntu dynasm 21 / cranelift 21 / wasm 4; windows dynasm 21 / cranelift 22). `cargo test` passes on all three platforms. The four macOS rows main has and this branch does not are all `[pypy exec clamped to floor; ratio not a measurement]`, i.e. the collapsed-denominator flake, not a signal.

**Nothing was re-recorded, because at this base no row is an improvement.** Every moving row carries at least one counter in the bad direction:

* `raise_reg_unbound_jitstress` (dynasm, cranelift, all 3 OS) — `loops_aborted 1 -> 2` alongside `loops_compiled 6 -> 8`.
* `exception_inline_callee_tb_frames.wasm` — `guard_failures 860 -> 975` with `loops_compiled`/`bridges_compiled` unchanged at 6/4, so the rise is not paid for by anything.
* `pickle_terminal_raise_resume.wasm` — `loops_compiled 74 -> 73`, a plain fall.
* `list_append_write_barrier_gc.wasm`, `recursion_memo_branch.wasm` — `loops_aborted` rises (0 -> 1, 1 -> 2) alongside large loop/bridge gains.
* the 42 dynasm/cranelift rows on ubuntu and windows — all `loops_compiled` falls (`arith_int_bool 8 -> 7`, `ast_compile_roundtrip 1 -> 0`, `sre_*`, `type_dotted_name`, …), and they do not reproduce on macOS at all, so main has a host-divergence class open on this gate independently of this PR.

I verified `list_append_write_barrier_gc` specifically rather than assuming it: this branch renames the `dont_look_inside` symbol `try_gc_alloc_stable_raw` -> `try_gc_alloc_stable_or_abort`, which can move a trace boundary, and that fixture is a GC/write-barrier one. main's own run reports the identical delta, so it is not this branch's.

A local bracketed run (`HEAD_BEFORE == HEAD_AFTER == a4ce0376fbe`, `check.py --synthetic-only`, macOS) reproduces the same six host-independent rows byte for byte.

— *commented by Claude*

youknowone:
**Revised — two rows were recordable, and are now recorded** (`32f0b3239af`). My earlier "no row at this base is an improvement" was a direction-only reading of the counters, which is too coarse. Adjudicating each row by what caused it and by the upstream oracle instead:

Attribution, by set-diffing `pyre/check.py (ubuntu-24.04)` logs across consecutive main commits (keys normalised — #945 added the `(observed …)` suffix to the message, so raw string diffs mis-attribute every row):

| row | first appears at | cause |
|---|---|---|
| `raise_reg_unbound_jitstress` dynasm+cranelift `1→2 / 6→8` | `ff42f0e3317` (#950) | builtins parity reaching a qmut-bumping namespace write |
| `exception_inline_callee_tb_frames.wasm` `guard 860→975` | `ff42f0e3317` (#950) | — |
| `list_append_write_barrier_gc.wasm`, `recursion_memo_branch.wasm` | `7a7d9174088` (#945) | #945 itself |
| `pickle_terminal_raise_resume.wasm` `74→73` | `7a7d9174088` (#945) | #945 itself |
| the 42 `loops_compiled −1` rows on ubuntu+windows | `7a7d9174088` (#945) | see below |

The decisive fact for the two recorded rows: **#945 wrote the identical move into the dynasm and cranelift files of both fixtures and left the wasm file at #1009's values.**

| fixture | #945 wrote (dynasm/cranelift) | wasm file | wasm run |
|---|---|---|---|
| `list_append_write_barrier_gc` | `1→12` loops, `1→5` bridges, `0→1` aborted, `200→1335` guards | untouched, `1 / 1 / 0 / 200` | `12 / 5 / 1 / 1335` |
| `recursion_memo_branch` | `2→3` loops, `13→16` bridges, `0→1` aborted, `1757→2613` guards | untouched, `2 / 11 / 1 / 2227` | `3 / 14 / 2 / 3083` |

So the wasm rows are the same event on the third backend, not a new one. `PYPYLOG=jit-summary:- pypy3` (7.3.22) on the same scripts agrees on direction: `list_append_write_barrier_gc` 13 loops / 5 bridges / 0 aborts, `recursion_memo_branch` 3 loops / 21 bridges / 0 aborts — 12/5 and 3/14 are closer to upstream than the recorded 1/1 and 2/11.

Left unrecorded, deliberately:

* `pickle_terminal_raise_resume.wasm` `74 → 73` — #945 moved the same fixture **+2** loops on dynasm and cranelift while wasm reads one fewer. The divergence is the finding, not the fall.
* the 42 `loops_compiled −1` rows — #945 re-recorded 21 fixtures × 2 backends at values that hold on macOS and not on ubuntu or windows (`ast_compile_roundtrip 1 → 0`, `arith_int_bool 8 → 7`, `sre_*`, `type_dotted_name`, …; every one is exactly −1 loop, and macOS is green for all of them). No single number satisfies all three hosts, so this cannot be closed by re-recording — it needs whoever owns #945.

Since my earlier comments, main's #1016 (`ca1d3d4a8de`) recorded `raise_reg_unbound_jitstress.{dynasm,cranelift}` to `2/8` and `exception_inline_callee_tb_frames.wasm` to `975`, so those rows are gone from this branch independently.

Verification, bracketed (`HEAD_BEFORE == HEAD_AFTER == f6bd8497f6f`, rebased onto `ca1d3d4a8de`): `check.py --synthetic-only` reads **dynasm 358/358, cranelift 358/358** and wasm 3 failures before the record, 1 after — the remaining one is `pickle_terminal_raise_resume` above.

— *commented by Claude*

youknowone:
Reply to the Codex P1 and the two CodeRabbit findings, all three of which land on the same arm.

## 1. "Propagate MemoryError instead of aborting" — the premise does not hold for the arm this diff changes

Before this PR that arm did not raise `MemoryError`. It was `.filter(|p| !p.is_null())` followed by `lltype::malloc_raw`, so a **failed managed** allocation produced a headerless payload and handed it back as an ordinary object. `W_LongObject.value` is registered as a plain gc-pointer offset — `object_subclass_with_gc_ptrs(size_of::<W_LongObject>(), object_tid, vec![LONG_VALUE_OFFSET])` at `pyre/pyre-jit/src/eval.rs:2125-2129`, no custom trace and no `try_gc_owns_object` guard — so the collector forwards that slot unconditionally and reads a header that was never written. The change is **silent corruption → abort**, not **MemoryError → abort**.

The repo already has a two-tier convention for exactly this question, and the change follows it rather than inventing one:

* **infallible surface, aborts** — `alloc_typed_items_block_nursery` (`pyre/pyre-object/src/object_array.rs:748`) is `try_… .unwrap_or_else(|| handle_alloc_error(…))`. Its own doc says: *"Once the GC hook is installed, allocation failure must remain a failure: a raw fallback would leave `RBigInt._digits` pointing outside the managed heap even though its descriptor traces that field as `GcArray(Signed)`."*
* **fallible companion, carries the edge** — `try_alloc_typed_items_block_nursery` (`object_array.rs:768`) already returns `None` on `Some(null)` and only falls back to `alloc` on `None`, feeding `RBigInt::try_new` / `RBigInt::try_with_size` → `RBigIntError::Memory`. `object_array.rs:837-841` states the rule: *"The ordinary translated allocation path above remains infallible at the Rust type level, matching RPython's implicit exception edge; this entry point makes that edge explicit for Rust methods that already return `Result`."*

So after this PR the RBigInt **payload** fails the same way its own **digit array** already failed. Threading `MemoryError` out of the infallible form would need an error channel these functions do not have — they return a bare pointer, are `#[dont_look_inside]`, and run under JIT frames that cannot unwind. Where a channel exists it is already wired, and it already reports the failure rather than substituting a raw block.

Upstream never has this state: `malloc_fixedsize` cannot return null, because `collect_and_reserve` (incminimark.py:981-985) raises MemoryError itself. pyre's `try_alloc_with_type_no_collect_body` returns `GcRef(0)` once the nursery bump **and** `spill_to_oldgen_or_null` both fail, which is why the state needs an answer here at all.

## 2. "An absent `GC_ALLOC_STABLE_HOOK` while an independent GC is active is classified as NoRoute"

Not reachable in production. `install_pyre_object_hooks` (`pyre/pyre-jit/src/eval.rs:3948-3996`) registers the fast, regular, placement, stable, collecting and rooted-collecting hooks in one function body, and `init_gc_subsystem` runs `install_gc_into_backend()` at `eval.rs:4062` before `PYRE_OBJECT_HOOKS_INSTALLED.call_once(install_pyre_object_hooks)` at `eval.rs:4070`. There is no ordering in which the fast hook answers while the stable hook is absent — only a unit test registering hooks individually can build that state.

## 3. Deliberately not changed: `try_alloc_typed_items_block` (the stable one)

`object_array.rs:842` does still conflate the two states — `try_gc_alloc_stable_raw` collapses them and it falls through to `alloc_zeroed`. That site is **not** the corruption class, so it is out of scope rather than overlooked: its consumers are `W_ListObject.int_items.block` / `float_items.block`, and `list_object_custom_trace` guards both with `try_gc_owns_object` (`eval.rs:933-940`), so a `std::alloc` block there is a supported mixed population the walker leaves in place. Turning that fallback into `MemoryError` would convert a survivable managed-heap exhaustion into an exception while the system heap still has room.

The test that decides any site here is whether the walker that reaches the pointer carries an ownership guard — not how the fallback is spelled. That is why three sites in `rbigint.rs` and one in `pyframe.rs` changed and the guarded ones (`gc_alloc_storage_box`, `FrameDebugData`, `alloc_dict_object`) did not.

— *commented by Claude*

PR #1027, opened 2026-08-04
posix, compile, display: take filesystem names as bytes across the path, compile, repr and environment boundaries

youknowone:
Four independent fixes at the filesystem-name boundary. The first is a
wrong-behaviour bug, not a reporting one.

## posix paths aliased onto one another

`gateway::fsencode_w` ended in `String::from_utf8_lossy`, and the posix module
used it for every path argument, so a path byte with no UTF-8 spelling folded to
U+FFFD **before the syscall was made**. Two names differing only in such a byte
became one name:

```python
# a file named '�victim' exists (U+FFFD is valid UTF-8)
os.stat(b"\xffvictim")          # was: succeeds, stats the U+FFFD file
                                # now: FileNotFoundError, as CPython and PyPy
os.path.exists(b"\xffvictim")   # was: True    now: False
```

Upstream never folds: `space.fsencode_w` (`baseobjspace.py:1970`) answers with a
byte string and `dispatch_filename` (`interp_posix.py:56-115`) hands those bytes
to the call unchanged. The extraction now yields bytes, and `OsStr::from_bytes`
carries them to the syscall on unix; the text form stays only where the platform
has no byte spelling of a path.

`fsencode_w` and `fsencode_w_with_kind` are deleted rather than left reachable,
and the `PyPath` / `PyPathOrNone` macro bindings — which stand for
`gateway.py:365` `unwrap_spec` `'fsencode'` — bind `Vec<u8>`.

`OSError.filename` now carries the path object the call was given
(`interp_posix.py:332 wrap_oserror2(space, e, w_path)`), so a bytes path reports
bytes and a str keeps its surrogate escapes. For an `os.PathLike`, that object is
the result of the single `__fspath__` call, not the wrapper that supplied it
(`interp_posix.py:211-219 _unwrap_path`), so the two travel together in one
record. All eight rows across `os.stat` / `os.open` / `open` / `io.FileIO`
match CPython and PyPy.

## compile() dropped its filename, silently

`compiling.py:13` declares `unwrap_spec(filename='fsencode', mode='text')`, but
`builtin_compile` bound both as `if is_str { ... } else { <default> }`:

| call | CPython | PyPy | before |
|---|---|---|---|
| `compile(s, b"ascii.py", "exec").co_filename` | `'ascii.py'` | `'ascii.py'` | `'<string>'` |
| `compile(s, b"\xff.py", "exec").co_filename` | `'\udcff.py'` | `'\udcff.py'` | `'<string>'` |
| `compile(s, '\udcff.py', "exec")` | `'\udcff.py'` | `'\udcff.py'` | UnicodeEncodeError |
| filename with `__fspath__` | `'\udcff.py'` | `'\udcff.py'` | `'<string>'` |
| `compile(s, 42, "exec")` | TypeError | TypeError | `'<string>'` |
| `compile(s, bytearray(b"ba.py"), "exec")` | TypeError | `'ba.py'` + warning | `'<string>'` |
| `compile(s, "a\x00b.py", "exec")` | ValueError | ValueError | accepted |
| `compile(s, "a.py", 42)` | TypeError | TypeError | silently `exec` |

The first row needs no exotic encoding: a plain ASCII `bytes` filename was lost,
and every traceback, warning and frame built from that code object pointed at
`<string>`. Routing filename through `fsencode_bytes_w` and mode through
`text_w` yields the whole table, since the converter already carries the
deprecation warning, both TypeError spellings and the NUL rejection. The
`bytearray` row is where CPython and PyPy differ; we follow PyPy.

The filename names the whole compilation unit, so nested constants inherit it
when they are boxed — otherwise `outer.co_filename` was exact while
`outer.co_consts[0]` and the function's `__code__` stayed lossy. That is what
separates it from `pycode.py:431`, whose constructor and `replace` filenames
rename only the object being built, so the two get separate entry points.

## repr built in a buffer that could not hold what it was given

`code_repr` formatted `source_path`, so after `c.replace(co_filename=...)` it
reported the name the object was compiled under, not the current one. It now
builds from `code_filename_bytes` as WTF-8, the way `pycode.py:570 repr` reaches
for the filename's `utf8_w`.

That made a code object the second thing whose repr can carry a lone surrogate,
and `py_repr_wtf8` was a hand-maintained prelude listing such types before
delegating to `py_repr`, whose `String` cannot hold one. Frames were on the list
and code objects were not, so `repr(code)` **aborted** in `w_str_get_value`.
Any new such type repeats that, and the delegation also lost the surrogate for
every container — reachable on `main` today, independent of this change:

```python
class C:
    def __repr__(self): return "s\udcff"
repr(C())     # 's\udcff'   exact
repr([C()])   # was '[s�]'; CPython and PyPy give '[s\udcff]'
```

So the two are inverted: `py_repr_wtf8` / `py_str_wtf8` carry the whole
dispatch, `py_repr` / `py_str` are lossy views of them for the ~109 callers that
genuinely want a Rust `String` (error-message interpolation and the like), and
the prelude's per-type carve-outs are deleted rather than extended. `dict_repr`,
`list_repr`, `tuple_repr`, `array_repr_wtf8`, the deque / set / frozenset /
slice / range / structseq / exception formatting, and the `!r` conversion in
f-strings, `str.format` and the descriptor slots build a `Wtf8Buf` and recurse
through `py_repr_wtf8` — the same shape as `listobject.py:206-225
_listrepr_inner`, which appends `space.utf8_len_w(space.repr(w_item))` into a
`rutf8.Utf8StringBuilder`.

No repr's text changes. `ReprGuard::enter` and the `stack_check()` are
untouched, so `[[...]]`, `{1: {...}}` and the `RecursionError` on a
200000-deep list all still hold, and every slot keeps the allocation kind it
had (a per-call `__repr__` result stays GC-managed rather than becoming an
immortal `malloc_raw` buffer).

Buffer assembly did cost one thing, caught by the Codex parity review: the
`parts.join(", ")` it replaced separates by **position**, while emitting a
separator whenever the buffer is non-empty separates by *what was written*. An
item whose `__repr__` answers `""` writes nothing, so the next item lost its
separator — `repr([E(), E()])` gave `[]`, and `repr(ValueError(E(), E()))`
gave `ValueError()`. All four sites now go by index, `dict_repr` included
(which was accidentally safe, since its `": "` always lands).

## putenv wrote its value through a lossy decode

`os.putenv` / `os.unsetenv` take both halves through the same `'fsencode'`
converter, and `posix.environ` already hands entries back as bytes, so a value
spelling a byte with no UTF-8 form has to survive the write. They now carry
`Vec<u8>` to `set_var` / `remove_var` — which already accept `impl AsRef<OsStr>`,
so no change outside pyre was needed — with the NUL rejection and the
`interp_posix.py:1762-1769` key rule (`=` illegal only after index 0, so the
`=C:` form Windows uses for a working directory stays legal).

## Verification

- `pyre/extra_tests/parity_tests` (176): **no `dynasm=FAIL`**. The one failure is
  `type_members_python314.py cpython=FAIL`, the system CPython, not pyre.
- `pyre/extra_tests` dynasm: **235/241**, up from 234; the remaining six are a
  subset of the previous seven.
- `cargo test -p pyre-interpreter --features dynasm`: 490 passed, 0 failed.
- New `compile_filename_boundary.py`, `repr_surrogate_wtf8.py`,
  `repr_empty_item_separator.py` and the extended
  `os_boundary_surrogateescape.py` pass under CPython and PyPy too; the
  `bytearray` filename row is left out because CPython and PyPy disagree there.

## Not in this change

- `_path_splitroot` and the Windows `arg_path` still go through text. Both are
  text-level operations on a name rather than a syscall boundary, and each
  carries the reason in place.
- `os.execve` reports no filename, which is what `interp_posix.py:1812,1817`
  does and what this branch's base did. CPython reports the program path there;
  the divergence predates this change.
- A `String`-returning `#[pymethod]` is wrapped with `w_str_new`, which is the
  immortal `malloc_raw` path — so `deque.__repr__` and its neighbours leak per
  call. That predates this change and is a macro-level fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved non-UTF-8 filesystem paths and surrogate characters across file operations, process execution, environment handling, and related errors.
  * Improved `compile()` filename handling, including path-like values, invalid inputs, syntax errors, and code-object representations.
  * Corrected string and representation formatting for nested containers, exceptions, deques, arrays, and empty items.
* **Tests**
  * Added comprehensive coverage for path boundaries, surrogate escaping, WTF-8 representations, and filesystem error behavior.

### DIFF
--- a/pyre/extra_tests/parity_tests/compile_filename_boundary.py
+++ b/pyre/extra_tests/parity_tests/compile_filename_boundary.py
@@ -87,6 +87,11 @@ replaced = code.replace(co_filename="\udcff.py")
 assert replaced.co_filename == "\udcff.py"
 assert 'file "\udcff.py"' in repr(replaced), ascii(repr(replaced))
 
+# `pycode.py:570-572` reports the zero sentinel as line -1.
+assert 'line -1>' in repr(code.replace(co_firstlineno=0)), ascii(
+    repr(code.replace(co_firstlineno=0))
+)
+
 # and a filename that never had a UTF-8 spelling reaches repr() the same way.
 from_bytes = compile(SOURCE, b"\xff.py", "exec")
 assert 'file "\udcff.py"' in repr(from_bytes), ascii(repr(from_bytes))
--- a/pyre/extra_tests/parity_tests/compile_filename_boundary.py
+++ b/pyre/extra_tests/parity_tests/compile_filename_boundary.py
@@ -1,0 +1,103 @@
+"""`compile()` takes its filename through the filesystem-encoding converter.
+
+`compiling.py:13` declares `@unwrap_spec(filename='fsencode', mode='text')`, so
+the filename accepts `bytes` and `os.PathLike` as well as `str`, keeps a byte
+with no UTF-8 spelling, and rejects an embedded NUL — while a non-`str` mode is
+a TypeError rather than a silent fall back to "exec".  Every code object reader
+then has to report that filename, including `repr()`, which reads the same
+field `co_filename` does rather than a separate compiler-side spelling.
+
+A `bytearray` filename is deliberately absent: it is a readable buffer that
+`baseobjspace.py:1975` accepts with a DeprecationWarning but CPython rejects,
+and this suite has to pass under CPython too.
+"""
+
+SOURCE = "x = 1"
+
+
+class Path:
+    def __init__(self, name):
+        self.name = name
+
+    def __fspath__(self):
+        return self.name
+
+
+# A str filename is handed through unchanged.
+assert compile(SOURCE, "plain.py", "exec").co_filename == "plain.py"
+
+# bytes decode with the filesystem encoding.  The ASCII case matters on its own:
+# it needs no surrogate to expose a filename that never reaches the code object.
+assert compile(SOURCE, b"ascii.py", "exec").co_filename == "ascii.py"
+assert compile(SOURCE, b"\xff.py", "exec").co_filename == "\udcff.py"
+
+# A str already carrying a surrogate escape encodes back to that same byte
+# instead of raising UnicodeEncodeError.
+assert compile(SOURCE, "\udcff.py", "exec").co_filename == "\udcff.py"
+
+# `__fspath__` is honoured, and may answer with either str or bytes.
+assert compile(SOURCE, Path(b"\xff.py"), "exec").co_filename == "\udcff.py"
+assert compile(SOURCE, Path("spelled.py"), "exec").co_filename == "spelled.py"
+
+# An object that is neither a path nor a string is a TypeError, not a filename
+# quietly replaced by "<string>".  The two runtimes word it differently, so only
+# the class is under test.
+for bad in (42, object(), None):
+    try:
+        compile(SOURCE, bad, "exec")
+    except TypeError:
+        pass
+    else:
+        raise AssertionError("compile() accepted %r as a filename" % (bad,))
+
+# `bytesbuf0_w` rejects an embedded NUL.
+for nul in ("a\x00b.py", b"a\x00b.py"):
+    try:
+        compile(SOURCE, nul, "exec")
+    except ValueError:
+        pass
+    else:
+        raise AssertionError("compile() accepted a NUL in %r" % (nul,))
+
+# mode='text' — a non-str mode is rejected rather than treated as "exec".
+try:
+    compile(SOURCE, "a.py", 42)
+except TypeError:
+    pass
+else:
+    raise AssertionError("compile() accepted a non-str mode")
+
+
+# The filename a SyntaxError reports is the same one the successful compile
+# would have recorded.
+try:
+    compile("(", b"\xff.py", "exec")
+except SyntaxError as exc:
+    assert exc.filename == "\udcff.py", ascii(exc.filename)
+else:
+    raise AssertionError("compile() accepted an unterminated '('")
+
+
+# `repr()` reads `co_filename`, so it follows a replacement instead of showing
+# the name the object was compiled under.
+code = compile(SOURCE, "before.py", "exec")
+assert 'file "before.py"' in repr(code), repr(code)
+
+replaced = code.replace(co_filename="\udcff.py")
+assert replaced.co_filename == "\udcff.py"
+assert 'file "\udcff.py"' in repr(replaced), ascii(repr(replaced))
+
+# and a filename that never had a UTF-8 spelling reaches repr() the same way.
+from_bytes = compile(SOURCE, b"\xff.py", "exec")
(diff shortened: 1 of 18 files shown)


youknowone:
Acted on the Codex parity review.

**§1 is real and it was mine.** `OSError.filename` retained the argument object,
which is right for a `str` or `bytes` path but wrong for an `os.PathLike`:
`interp_posix.py:211-219 _unwrap_path` calls `__fspath__` and keeps its *result*
as `Path.w_path`. Verified rather than assumed —

| | CPython | PyPy | pyre before | pyre now |
|---|---|---|---|---|
| `os.stat(P('absent'))` | `'absent'` | `'absent'` | `P('absent')` | `'absent'` |
| `os.stat(P(b'absent'))` | `b'absent'` | `b'absent'` | `P(b'absent')` | `b'absent'` |
| `open(P('absent'))`, `os.listdir(P('absent'))` | `'absent'` | `'absent'` | `P('absent')` | `'absent'` |

The bytes and the resolved object now travel together in one record, the shape
`Path` has upstream, so the single `__fspath__` call serves both. The record owns
its root scope: a resolved `PathLike` name is freshly allocated, and unlike the
caller's argument array nothing else keeps it reachable while the call runs.

**§3 — three of the four are pyre against both runtimes, so they are fixed here
rather than deferred:**

| | CPython | PyPy | pyre before |
|---|---|---|---|
| `os.execv` failure filename | `None` | `None` | `'/nonexistent'` |
| `posix_spawn` env key `=C:` | accepted | accepted (`'=' in key[1:]`) | ValueError |
| `repr(code)` at `co_firstlineno == 0` | `line -1` | `line -1` | `line 0` |

The `posix_spawn` rule now matches the `execve` path in the same file rather
than being a third spelling of it.

**`execve`'s filename is deliberately left alone.** Codex grouped it with
`execv`, but the two runtimes disagree there — CPython reports the path, PyPy
reports `None` — so pyre is not wrong against both, and making it "consistent"
with `execv` would just pick a side.

Also folded in: `os.putenv` / `os.unsetenv` take their halves as bytes, which the
PR text had listed as out of scope. It turned out not to need the
`rustpython-host_env` change I expected — `os::set_var` is already generic over
`AsRef<OsStr>`, so the bytes go straight through.

Re-verified at `cbcaa14e4c3`: parity_tests has no `dynasm=FAIL` (the lone failure
is `type_members_python314.py cpython=FAIL`, the system CPython); extra_tests
dynasm 235/241; `cargo test -p pyre-interpreter --features dynasm` 490 passed, 0
failed. The new assertions — PathLike filenames, the `execv` and `posix_spawn`
rules, the `-1` line, and the putenv round trip — pass under CPython too.

— *commented by Claude*

PR #1028, opened 2026-08-04
type: identify object's class methods by the callable, not the qualname

youknowone:
`method_class_bound_qualname` selected `object.__subclasshook__` and
`object.__init_subclass__` by comparing the wrapped function's `__qualname__`
against those two strings. `__qualname__` is writable on an ordinary function,
so a classmethod over one that spelled itself `object.__subclasshook__` was
rewritten to name its own owner instead of answering the string it was given.

`init_object_type` now registers the two callables it publishes in slots held as
process-lifetime GC roots, and the selection compares addresses against those.
The address is read back out of the classmethod, since allocating it can move
the function.

The read off a class was reaching neither carrier: the delta sat inside
`descr__getattribute__`'s `!on_method_type` branch, and the two are published as
`builtin_function_or_method`, whose type defines `__qualname__` — so the branch
was skipped for exactly the objects the delta is about, and ran only for the
user functions it is not about. It now runs ahead of that test.

```python
class A:
    def f(cls): pass
    f.__qualname__ = "object.__subclasshook__"
    f = classmethod(f)
```

| read | before | after / CPython 3.14 |
| --- | --- | --- |
| `A.f.__qualname__` | `'A.__subclasshook__'` | `'object.__subclasshook__'` |
| `A.__subclasshook__.__qualname__` | `'object.__subclasshook__'` | `'A.__subclasshook__'` |

Both directions were wrong before, in opposite ways.

## Verification

Measured on this branch (`origin/main` + this commit), after re-extracting LLBC
for the new base:

- `pyre/check.py --backend dynasm --synthetic-only`: **357 passed, 1 failed**.
  The single failure is `int_lshift_memoryerror  pypy crash`, which is the
  *oracle* aborting: that fixture drives an allocation past the limit to prove
  the result is a catchable `MemoryError`, and the run had `PYPY_GC_MAX` set in
  its environment, under which PyPy answers `using too much memory, aborting`
  instead. Without it `pypy3` runs the fixture clean. Not attributable to this
  change.
- Parity suites, `PYRE_NO_JIT=1` and JIT, all OK: `classmethod_protocol`,
  `exception_context_generator_python314`, `object_surface_python314`,
  `function_python314`, `type_members_python314`,
  `type_descriptor_surface_python314`.
- `classmethod_protocol.py` gains the four-way check plus a re-read after enough
  allocation to drive a collection, since the registered slots have to survive
  one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `__qualname__` reporting for inherited class methods, including `__init_subclass__` and `__subclasshook__`.
  * Ensured class method names remain accurate when accessed through subclasses or after memory-management activity.
  * Preserved existing behavior for keyword handling and unsupported subclass hooks.

* **Tests**
  * Added coverage for class method qualification and inheritance scenarios.

### DIFF
--- a/pyre/extra_tests/parity_tests/classmethod_protocol.py
+++ b/pyre/extra_tests/parity_tests/classmethod_protocol.py
@@ -9,6 +9,7 @@ PyPy's ``__reduce_ex__`` entry and neither implementation makes the wrapper
 callable.
 """
 
+import gc
 import sys
 
 
@@ -251,11 +252,15 @@ assert Forger.__init_subclass__.__qualname__ == "Forger.__init_subclass__"
 assert int.__subclasshook__.__qualname__ == "int.__subclasshook__"
 assert int.__init_subclass__.__qualname__ == "int.__init_subclass__"
 
-# The same reads after enough allocation to drive a collection: whatever
-# records the two carriers has to survive one.
-churn = [[object() for _ in range(64)] for _ in range(2048)]
-assert len(churn) == 2048
-del churn
+# The same reads across a full collection: whatever records the two carriers
+# has to survive one, and survive it as an address rather than as a value the
+# collection is free to relocate. A small live set is allocated first so the
+# collection has something young to promote; `gc.collect()` then forces the
+# major pass without holding a large heap. Its return value differs between
+# implementations, so nothing here reads it.
+promoted = [object() for _ in range(64)]
+gc.collect()
+assert len(promoted) == 64
 assert Forger.hook.__qualname__ == "object.__subclasshook__"
 assert Forger.__subclasshook__.__qualname__ == "Forger.__subclasshook__"
 assert Forger.__init_subclass__.__qualname__ == "Forger.__init_subclass__"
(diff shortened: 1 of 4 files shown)


youknowone:
Applied in 3dae5e1, though the stated mechanism does not hold.

The abort quoted from the PR description is `int_lshift_memoryerror`, a different fixture, which drives a shift past the allocator on purpose. This block retains 131,072 *small* `object` instances — measured peak RSS for the whole test was 76.7 MB, and it runs clean under `PYPY_GC_MAX=4000000000`:

| | minor | major | peak RSS | result |
| --- | --- | --- | --- | --- |
| with `churn`, `PYPY_GC_MAX=4e9` | 93 | 23 | 76.7 MB | OK |
| without `churn` | 92 | 23 | 54.8 MB | OK |

So lines 259-261 did execute.

The recommendation is still right for a different reason: 22 MB bought exactly **one minor** collection, and nothing asserted that a collection had run at all. `gc.collect()` forces a full **major** pass deterministically (major count 3 → 4 on an otherwise empty script), which additionally walks the old generation the two registered slots live in — a strictly stronger check. A 64-element live set is allocated first so the pass has something young to promote.

After the change: CPython OK, pyre `PYRE_NO_JIT=1` OK, pyre JIT OK, peak RSS 52.9 MB.

— *commented by Claude*

PR #1029, opened 2026-08-04
jit(#346): carry usize slice bounds through annotation; dominance-threaded operands; oparg signedness

youknowone:
Four rtyper-prepass census commits (gh#346).

## Commits

- **`dont_look_inside` on `push_roots` / `increase_root_stack_depth` / `root_stack_depth`** — residualises the root-stack calls instead of tracing into them.
- **Thread undefined operands by dominance, not the single-predecessor chain** — adds the iterative dominator-set computation and threads an operand only when its definition dominates the use.
- **Fix oparg signedness lowering** — `tyref_to_value_type` yields `ValueType::Unsigned` for transparent `u32` opargs, so `SomeInteger(unsigned=True)` carries the non-negative fact through annotation.
- **Carry usize slice bounds through annotation; keep the sub-slice capstone dormant** — preserves the checked `usize` arithmetic tuple's element type through flowspace lowering as `r_uint`, and adds the `RangeTo` recognizer (`&s[..k]` → `getslice(s, 0, k)`) plus the `prune_dead_phis` wiring it needs. Both capstone arms stay dormant.

## Why the sub-slice capstone stays dormant

`front/slice_index.rs` was already a built, unit-tested, deliberately unwired capstone. Activating the `RangeTo` arm was tried and measured, and it does not hold:

- With both arms active, the `RangeFrom` rewrite's Void `ConstNone` stop had no regalloc coloring once `_io::buffered_rwpair::<Impl>::readinto` dropped to the legacy walker.
- With `RangeTo` alone, a bare `getslice/rii>r` reached the codewriter from `builtins::split_builtin_kwargs` and `_warnings::do_warn_explicit`, failing `production_bh_builder_covers_every_build_emitted_opname` and `default_bh_builder_unwired_set_matches_task_85_snapshot` (`pyre/pyre-jit-trace/src/jitcode_runtime.rs:2031,2179`). No blackhole handler dispatches `getslice`, and none should: upstream `rtype_getslice` (`rlist.py:409-414`) always replaces it with a `gendirectcall`, so a surviving bare op means the graph did not rtype.
- Both stops are statically unsigned, so signedness does not separate the sites whose planted op is consumed from an rtyped graph from those that drop.

Neither assertion was widened and no `bhimpl` alias was added. The recognizers and their tests stay in the tree, and the module doc now records the measured reason with the observed graph names.

## Measurements

phaseA **1272**, phaseB **3** — unchanged. The `RangeTo` activation did reach phaseA 1271 (clearing `split_builtin_kwargs`), but that is not kept, because it emitted the bare `getslice` above. A correct build beats −1.

`cargo test --all --no-default-features --features dynasm` — green (the CI command; a `-p majit-translate` run does not cover the two `pyre-jit-trace` tests above).

## Note on `core::slice::index`

While the activation was live it removed the `slice::index` wall from 140 of 172 census records, yet only `split_builtin_kwargs` left the census — the caller graphs re-rooted onto walls behind it, mostly the per-wrapper `__PYRE_PARAM_NAMES` statics (117 records, 101 of them single-wall). Blast radius is a ceiling, not a yield.

## Summary by CodeRabbit

* **New Features**
  * Added support for open-ended slice ranges, including start-to-end and beginning-to-end forms.
  * Slice operations now validate constant non-negative bounds and safely preserve original behavior when bounds cannot be handled.
* **Bug Fixes**
  * Improved handling of unsigned arithmetic and subtraction, preserving correct `usize` results.
  * Corrected lowering of unsigned operations to maintain expected runtime behavior.
* **Documentation**
  * Documented supported and deferred slice-range handling, including safety constraints.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -14868,11 +14868,14 @@ fn tyref_to_value_type(ty: &TyRef, llbc: &Llbc) -> ValueType {
         return inner;
     }
     // `OpArg` and compiler-core's `newtype_oparg!` wrappers are transparent
-    // u32 bytecode operands.  Their dependency declarations are opaque in
-    // interpreter LLBC, so model the exact upstream family as bare integers
-    // (`tyref_is_oparg`).
+    // `u32` bytecode operands.  Their dependency declarations are opaque in
+    // interpreter LLBC, so model the exact upstream family as unsigned
+    // integers (`tyref_is_oparg`).  This is not merely a register-bank
+    // choice: `SomeInteger(unsigned=True)` carries the non-negative fact
+    // through annotation, which is required when an oparg-derived count is
+    // passed to a `u32`/`usize` helper.
     if tyref_is_oparg(ty, llbc) {
-        return ValueType::Int;
+        return ValueType::Unsigned;
     }
     // A fieldless (C-like) enum is represented by-value as its
     // discriminant integer — RPython has no enum type; a fieldless enum
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -2159,26 +2159,6 @@ fn lower_unstructured_with_static_addrs_and_attrs(
                 &lo.next_call_results,
             );
         }
-        // The consumer-gated RangeTo sub-slice capstone remains built but
-        // dormant: aggregate lowering records no sites. If capture is safely
-        // restored, it becomes `getslice(s, 0, k)`; a structural decline
-        // leaves the residual ctor + index chain intact.
-        // `prune_dead_phis` to reclaim the dead threaded RangeTo inputarg /
-        // link args (the getslice never reads them) — run it unconditionally
-        // on any rewrite rather than relying on the gated sweep in
-        // `simplify_lowered_graph`. Fail-safe: a structural mismatch leaves
-        // the residual ctor + index call (rtyper Skip) and touches nothing.
-        let slice_index_rangeto_rewritten = if lo.slice_index_rangeto_sites.is_empty() {
-            0
-        } else {
-            crate::front::slice_index::rewire_slice_index_rangeto_sites(
-                &mut lo.graph,
-                &lo.slice_index_rangeto_sites,
-            )
-        };
-        if slice_index_rangeto_rewritten > 0 {
-            crate::model::prune_dead_phis(&mut lo.graph);
-        }
         let next_rewritten = if lo.next_call_results.is_empty() {
             0
         } else {
@@ -2810,9 +2790,6 @@ struct Lowering<'a> {
     /// after the body lowering completes (see
     /// [`crate::front::slice_first::SliceFirstSite`]).
     slice_first_sites: Vec<crate::front::slice_first::SliceFirstSite>,
-    /// `RangeTo` aggregates recorded for the orthodox sub-slice copy rewrite
-    /// in `front::slice_index`.
-    slice_index_rangeto_sites: Vec<crate::front::slice_index::SliceIndexRangeToSite>,
     /// `{uN}::saturating_sub(a, b)` call sites recorded for the unsigned clamp
     /// diamond (`if a < b { 0 } else { a - b }`) the
     /// `front::saturating_sub` post-pass synthesizes after body lowering (see
@@ -3050,7 +3027,6 @@ impl<'a> Lowering<'a> {
             option_try_sites: Vec::new(),
             bool_then_sites: Vec::new(),
             slice_first_sites: Vec::new(),
-            slice_index_rangeto_sites: Vec::new(),
             saturating_sub_sites: Vec::new(),
             range_inclusive_new_sites: Vec::new(),
             range_iter_new_sites: Vec::new(),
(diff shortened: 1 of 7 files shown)


youknowone:
## Review triage — verdicts with evidence

### ⛔ REFUTED — "preserve the unsigned width" (`flowspace_adapter.rs:1297`)

The premise does not hold in this tree: **there is no narrow unsigned width to preserve.**
`front/mir.rs:1301` collapses `u8 | u16 | u32 | u64 | usize` into the single `ValueType::Unsigned`,
and every downstream consumer treats that as one machine word:

| site | evidence |
|---|---|
| `front/mir.rs:15041` | `ValueType::Unsigned => "u64"` (residual-call return token) |
| `jtransform.rs:9974,10226` | renders `Unsigned` as `"u64"` |
| `codewriter/assembler.rs:3084` | 8 bytes |
| `annotator/annotation_state.rs:49` | `SomeInteger(unsigned=true)` |

`r_uint` is RPython's machine-word unsigned — i.e. exactly that kind — so the post-result cast names
the type the value already has. Removing it would drop the annotation, not preserve a width.
No change made.

### ✅ CONFIRMED — `RootScope` residualised as a one-word ref (`gc_roots.rs:395`)

Verified end to end:
- `RootScope` is two words (`save_point: usize` + `stack_slot: *const RootStack`), and
  `gc_roots.rs:797 root_scope_payload_is_two_words` asserts exactly that.
- `dont_look_inside_return_token` maps a non-primitive ADT to `ValueType::Ref(None)`
  (`front/mir.rs:14823`) and then to the `"ref"` token (`front/mir.rs:15052`). The `OBJECTPTR`
  carve-out at `:15049` does not apply.
- `TermKind::Drop` does lower to a plain `set_goto` (`front/mir.rs:6465-6470`).

It is latent today — every call site binds `let _roots = push_roots();` and never reads the guard —
but the ABI is wrong the moment anything reads it. Reverting the attribute on `push_roots` only;
`increase_root_stack_depth` (unit → void) and `root_stack_depth` (`usize` → `u64`) have sound
scalar ABIs and stay.

### ✅ CONFIRMED — dominator storage has no upstream owner (`model.rs:3056`)

The upstream owner is `SSA_to_SSI` (`rpython/translator/backendopt/ssa.py:135-196`), which uses **no
dominators at all**: it walks `mkentrymap` backward with a worklist, appending the variable to every
incoming link and re-queueing `(link.prevblock, v)` until it reaches a definition. `majit-translate`
already carries a faithful port in `model_ssa.rs`, but only for
`flowspace::model::FunctionGraph`; `model.rs:3934` already records that it needs porting to
`crate::model`. My dominator map filled that gap ad hoc. Restructuring to the entrymap worklist shape.

That shape also subsumes the separate report that dead blocks enter `reachable` before the `dead`
check (`model.rs:3047`): `SSA_to_SSI` skips unreachable blocks structurally via
`if block not in entrymap: continue` (`ssa.py:151`), because `mkentrymap` only records blocks
reached by a link.

### ✅ CONFIRMED — RangeTo would erase the out-of-bounds failure (`slice_index.rs:275`)

Rust's `&s[..end]` panics when `end > len`, while the emitted `GetSlice` reaches
`build_ll_listslice_startstop_helper_graph`, which clamps `stop` to the length. This is a stronger
blocker than the missing non-negativity gate reported separately, and it is why the arm stays
dormant rather than being gated and activated. Documenting both blockers on the recognizer.

— *commented by Claude*

youknowone:
## Rebased onto `origin/main` (`4fecf865268`, +8 commits) — two commits dropped as superseded

`origin/main` `ce37de6645f` (#1024) rewrote the same function this branch was changing.
It replaced `thread_undefined_op_operands`'s single-predecessor walk with
`available_on_every_predecessor_path` + `ancestor_has_definition` — memoized, back-edge aware,
and rejecting source-less SCCs — and its doc already cites
`rpython/translator/backendopt/ssa.py:171-190`.

That covers the join case my dominator commit existed for, so both of this branch's
`thread_undefined_op_operands` commits were **dropped during the rebase**:

- `bde979f0bca` *thread undefined operands by dominance* — superseded.
- `aaeb1ba0fc0` *thread undefined operands with an entrymap worklist* — superseded. It measured
  phaseA 1261 → 1261 (neutral); its value was parity, and `#1024`'s `preds: HashMap<BlockId,
  Vec<BlockId>>` is already `mkentrymap`'s shape.

**This also resolves the "port dominator storage from the upstream shape" P1** — the ad hoc
`HashMap<BlockId, HashSet<BlockId>>` dominator side structure it objected to no longer exists on
this branch at all.

### One known item does NOT carry over, and is not fixed here

The dead-block report (`model.rs:3047`) was written against the dominator code that is now gone.
The underlying concern survives in a different place: `#1024`'s `preds` is still built from every
block without a `dead` check, and `available_on_every_predecessor_path` requires **all**
predecessors to supply the value, so a dead predecessor of a live join makes it decline a valid
thread. Direction is over-conservative (missed threading), not unsound. Left for a separate,
measured change against the new shape rather than blind-rebasing a fix onto code it no longer
matches.

### Base-move caveat on the numbers

The phaseA figures in the commit messages were measured on the previous base (`073d9e5e0d6`) and
are historical. In particular `Remove push_roots residual-call annotation` measured 1242 → 1261 —
that +19 is the deliberate price of dropping a wrong residual-call ABI, not a regression. Post-rebase
totals will differ; the pre-push check on the new base was `cargo check --all --no-default-features
--features dynasm`, clean.

— *commented by Claude*

PR #1031, opened 2026-08-04
bench: restore the forty baselines #945 recorded from a macOS-only run

youknowone:
`loops_compiled` for these benches is not one number, and #945 recorded the
macOS one.

## What the counter actually measures

The script's own directory is `sys.path[0]`, so the first import fills
`FileFinder._fill_cache`. That function's listing loop is platform-conditional
(`lib-python/3/importlib/_bootstrap_external.py:1419-1437`):

| platform | Python-level loops in `_fill_cache` | extra compiled loops |
|---|---|---|
| linux | `set(contents)` — C level | +0 |
| darwin | `+ {fn.lower() for fn in contents}` (`_CASE_INSENSITIVE_PLATFORMS`) | +1 |
| win32 | `+ for item in contents: … .add(…)` as well | +2 |

`pyre/bench/synth` holds 1435 entries, over the hot threshold, so those loops
compile. Byte-identical source, one dynasm binary:

```
same file in a dir of    1 entry   -> loops_compiled=7  guard_failures=2211
same file in a dir of  501 entries -> 7 / 2211
same file in a dir of 1001 entries -> 8 / 2213
pyre/bench/synth/arith_int_bool.py -> 8 / 2213
```

`MAJIT_LOG=1` names the eighth loop `[bcenc-audit] "_fill_cache"`. Compiling it
under `PYPY_GC_NURSERY=2G` changes nothing, so it is not a collection artifact.
Splitting `arith_int_bool.py` into its seven functions and summing gives exactly
7 / 10 / 2211 — the linux value — i.e. the bench itself contributes seven loops
on every platform and the eighth is the import machinery.

## Not a regression

`loops_aborted` 0 -> 0, `internal_compile_panics` 0 -> 0, `bridges_compiled`
10 -> 10. What moves is `loops_compiled` up and the new loop's two guard
failures, which stay under `trace_eagerness` so no bridge attaches. Compiling
the loop is `1c41d1a56c4 jit: admit SET_ADD and MAP_ADD in the FOR_ITER body
scan` working as intended, measured on that commit and its parent — one
worktree, one `CARGO_TARGET_DIR`, the same probe file, and the two revisions
differ by that one file's fifteen lines:

| arm | `loops_compiled` / `bridges` / `aborted` / `guards` |
|---|---|
| `ce42b0d5dc2` (`1c41d1a56c4^`, `SetAdd` absent from `eval.rs`) | 7 / 10 / 0 / 2211, 3/3 |
| `1c41d1a56c4` (`SetAdd` admitted) | 8 / 10 / 0 / 2213, 3/3 |

The mechanism: `_fill_cache` carries `SET_ADD` in its own code
object (the comprehension is inlined, no nested code object), and `SetAdd` does
not appear anywhere in `pyre/pyre-jit/src/eval.rs` before that commit, so the
frame was rejected whole by `unsupported_jit_shape` — a gate rejection, which
counts no abort, matching `loops_aborted=0` on both sides.

## What this PR does

Restores the forty rows (twenty benches, dynasm and cranelift) to the values
main measured identically on all three runners. #945 recorded the macOS
observation for them, so they went red on the other two: the ubuntu job moved
from 38 to 46 changed rows and windows from 25 to 44, while macOS moved from 24
to 2.

The macOS and windows counts stay unrecorded, because no single value is green
on all three — the same conclusion `ae895c30c6c` reached for
`imp_lock_rlock_semantics`, which is one of these forty.

## Follow-up, not in this PR

Making all three agree means keeping `_fill_cache` out of the measurement — the
direct lever is splitting `pyre/bench/synth/` into directories under the hot
threshold, after which every platform runs no Python-level loop there. It also
removes the standing fragility that adding or deleting bench files can move
every baseline at once.


youknowone:
Superseded by #1034, which fixed this at the source instead of at the baseline.

#1034 sets `PYTHONSAFEPATH=1` on the bench child, taking the fixture's own
directory off `sys.path`, so the import machinery's scan of a 1400-entry
directory no longer crosses the compile threshold and no longer lands in the
counters. Verified on this host with a single dynasm binary and
`pyre/bench/synth/arith_int_bool.py`:

```
PYTHONSAFEPATH=1  ->  loops_compiled=7  bridges=10  aborted=0  guard_failures=2211
(unset)           ->  loops_compiled=8  bridges=10  aborted=0  guard_failures=2213
```

7 / 2211 is ubuntu's column, now reproduced on macOS — the platform ladder is
gone rather than papered over, which is strictly better than restoring the
linux rung by hand as this PR did.

All forty files this PR touches are byte-identical to `main` after #1034's
re-record, so nothing here is left to land. Its base has also gone stale: the
diff against `main` now carries 815 deletions across `check.py`, `trace.py`'s
Rust counterparts and the wasm runner, i.e. it would revert #1023 / #1032 /
#1034. Closing.

— *commented by Claude*

PR #1036, opened 2026-08-04
builtins: fold sum() over the live iterator, and carry a wide int through the compensated fast paths

youknowone:
`builtin_sum` materialised the iterable with `collect_iterable` and then
indexed the `Vec`. `app_functional.py:54 _regular_sum` is a fold over the
**live** iterator, so `next` and `__add__` interleave.

## The ordering is observable

Both oracles agree with each other and disagree with pyre:

```
pypy3      3 ['next0', 'add0', 'next1', 'add1', 'next2', 'add2']
python3.14 3 ['next0', 'add0', 'next1', 'add1', 'next2', 'add2']
pyre       3 ['next0', 'next1', 'next2', 'add0', 'add1', 'add2']
```

Three consequences: user-visible side effects run in the wrong order, an
iterable the fold consumes in O(1) costs O(n) (`sum(itertools.count())` never
returns a `Stop` from a user `__radd__`), and — the one that matters most —
the running total and the whole item `Vec` were **raw locals across
`baseobjspace::add`**, which allocates. `collect_iterator`'s own doc comment
warns about exactly that hazard for the `Vec` it builds; the fold reintroduced
it by keeping that `Vec` alive across every `add`.

Pump the iterator instead, holding the total, the pending item and the
iterator in shadow-stack slots. The four phases (exact int, compensated
float, compensated complex, generic fold) and the "an item that leaves a
phase stays pending for the next one" handoff are preserved exactly.

## A second, pre-existing defect

Porting CPython's own `test_sum` and `test_sum_accuracy` (54 checks) into a
standalone fixture found two failures that are not caused by the rewrite —
established from the committed source, where the float phase has only
`is_float` / `is_int` arms and `break`s otherwise:

    sum([1.0, 10**100, 1.0, -(10**100)])     -> 0.0   (3.14: 2.0)
    sum([2j, 1.0, 10**100, 1.0, -(10**100)]) -> 2j    (3.14: 2+2j)

The fast paths took only a machine-word `int`, so a wider one left the
compensated loop and flushed the compensation term *before* the large value
arrived. Convert an `int` of any width to a double in the loop, raising
`OverflowError` only past f64 range.

Neither oracle gates this: `test_sum_accuracy` is `@support.cpython_only`,
and the `pypy3` oracle is 3.11, which has no compensated sum at all.

Over-range is signalled through a non-finite `f64` rather than a `Result`,
because `descroperation.rs` documents that the JIT codewriter cannot flatten
a payload mixing `Float` (`Ok`) and `Ref` (`Err`) into one register kind.

## `RootScope::set`

`RootScope` caches the thread-local root-stack cell precisely so a loop need
not re-resolve it per access, but had only `get` / `pin_root`; the free
`shadow_stack_set` resolves the thread local three times per call. Added the
symmetric `set` and put the fold on the cached-cell accessors.

## Cost

Measured control-free, since a sibling-worktree control arm proved invalid:
**+21ns per `sum()` call** (rooting `start` and the item slot) and **+3.8ns
per item** (one forwarding query for the now-rooted accumulator). `sum` on a
large list remains ~40x `pypy3` either way — untouched here.

## Verification

- 54/54 of the ported `test_sum` + `test_sum_accuracy` checks.
- A 43-case behavioural battery byte-identical to CPython 3.14, error
  messages included.
- `check.py` **374/374 on dynasm and cranelift**, with no jit-stats movement;
  `loops_comprehension` reads 10.1x against its 144x ratio gate.
- `cpython_tests` 46/46, no regressions. `cargo test` 799 pass. `cargo fmt`
  clean; the `clippy` errors in `pyre-object` are pre-existing and in files
  this branch does not touch.
- Regression coverage added to `parity_tests/builtin_sum_python314.py`, green
  on cpython, dynasm and cranelift.

Note: the check.py run above was made against the pre-rebase base. #1034 has
since re-recorded 43 baselines and changed the gate, so CI's run is the
authoritative one at this base.

— *authored by Claude*

## Summary by CodeRabbit

* **New Features**
  * Updated `sum` to process values incrementally, supporting large and unbounded iterables more efficiently.
  * Improved compensated accumulation for floating-point and complex-number calculations.
  * Added support for summing arbitrarily large integers with appropriate conversion handling.

* **Bug Fixes**
  * Improved accuracy and consistency when combining integers, floats, and complex values.
  * `sum` now raises `OverflowError` when values exceed floating-point limits.
  * Improved handling of numeric subclasses and mixed-type additions.

### DIFF
--- a/pyre/extra_tests/parity_tests/builtin_sum_python314.py
+++ b/pyre/extra_tests/parity_tests/builtin_sum_python314.py
@@ -66,6 +66,18 @@ assert math.isinf(sum([1e308, 1e308]))
 assert sum([1.0, 10**100, 1.0, -(10**100)]) == 2.0
 assert sum([2j, 1.0, 10**100, 1.0, -(10**100)]) == 2 + 2j
 assert sum([1.0, 2**63]) == 1.0 + float(2**63)
+
+# ...but a wide *running total* ends the integer fast path for good, the way
+# `PyLong_AsLongAndOverflow` signalling overflow does: what leaves that path is
+# still an `int`, so no compensated phase can claim it.
+assert sum([2**63, 0.1, 1, -(2**63)]) == 0.0
+assert sum([0.1, 1, -(2**63)], 2**63) == 0.0
+assert sum([2**63, 0.1, -(2**63)]) == 0.0
+assert sum([2**63, 1, -(2**63)]) == 1
+# A wide int reached *after* the total is already a float stays on the
+# compensated path.
+assert sum([0.1, 2**63, 1, -(2**63)]) == 1.1
+
 for values in ([1.0, 10**1000], [1j, 10**1000]):
     try:
         sum(values)
@@ -83,6 +95,33 @@ assert sum(values) == complex(
     sum(value.real for value in values),
     sum(value.imag for value in values),
 )
+
+# The complex fast path takes `PyFloat_Check` / `PyLong_Check`, not the
+# exactness test the float phase uses: a `float` or `int` subclass stays on it
+# and its reflected addition is skipped, while a `complex` subclass leaves.
+class SubFloat(float):
+    def __radd__(self, other):
+        return "radd-float"
+
+
+class SubInt(int):
+    def __radd__(self, other):
+        return "radd-int"
+
+
+class SubComplex(complex):
+    def __radd__(self, other):
+        return "radd-complex"
+
+
+assert sum([1j, SubFloat(1.0)]) == 1 + 1j
+assert sum([1j, SubInt(1)]) == 1 + 1j
+assert sum([1j, True]) == 1 + 1j
+assert sum([1j, SubComplex(1)]) == "radd-complex"
+# The float phase does test exactness, so a `float` subclass leaves it there.
+assert sum([1.0, SubFloat(1.0)]) == "radd-float"
+assert sum([1, SubFloat(1.0)]) == "radd-float"
+assert sum([1.0, SubInt(1)]) == 2.0
 
 for values in (
     [complex(1, -0.0), 1],
--- a/pyre/extra_tests/parity_tests/builtin_sum_python314.py
+++ b/pyre/extra_tests/parity_tests/builtin_sum_python314.py
@@ -1,5 +1,55 @@
+import itertools
 import math
 import random
+
+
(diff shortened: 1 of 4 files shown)


youknowone:
### Review responses

**Codex P1 — valid, fixed in `709aef73ce6`.**

Confirmed against 3.14 before changing anything:

```
                   3.14    before
codex_case         0.0     1.0     sum([2**63, 0.1, 1, -(2**63)])
wide_start         0.0     1.0     sum([0.1, 1, -(2**63)], 2**63)
```

The mechanism is the one Codex names. The C fast path accumulates into a
`long`, and `PyLong_AsLongAndOverflow` signalling overflow ends it for good:
the value that leaves is still an `int`, so neither compensated phase can
claim it and the rest of the fold is generic. pyre's int loop ran on the
arbitrary-precision `add`, so a wide total stayed in it and a later float item
handed the fold to the compensated float phase.

This was introduced by the widening in the first commit, so it is mine, not
pre-existing. The int loop now leaves once the running total is no longer a
machine-word `int`. A wide item reached when the total is *already* a float is
a different case and stays on the compensated path — `sum([0.1, 2**63, 1,
-(2**63)])` is `1.1`, matching 3.14. All eight probes now match the oracle
exactly, and the cases are pinned in `parity_tests`.

**CodeRabbit "Require an exact `float` in the complex fast path" — refuted, no change.**

Measured on 3.14:

```python
class SubFloat(float):
    def __radd__(self, other): return "radd-float"

sum([1j, SubFloat(1.0)])   # 3.14 -> (1+1j),  pyre -> (1+1j)
```

3.14's complex loop takes `PyFloat_Check`, not `PyFloat_CheckExact`, so a
`float` subclass stays on the fast path and its reflected addition *is*
skipped — the asymmetry is deliberate, and pyre already matched it. Adding the
exactness test would have introduced a deviation. The eight-case probe covering
`float` / `int` / `complex` subclasses and `bool` in both the float and complex
phases is byte-identical to 3.14 either way; those cases are now pinned in
`parity_tests` so the asymmetry cannot be "fixed" by accident later.

**CodeRabbit test-coverage suggestion — taken**, including the float-subclass
and float-then-complex cases.

Re-verified at the current base: `check.py` **377/377 dynasm**, `cpython_tests`
46/46 no regressions, the 54-check `test_sum`/`test_sum_accuracy` port and the
43-case battery still byte-identical to 3.14.

— *commented by Claude*

PR #1039, opened 2026-08-04
launch flags: share the option block with the wasm guest, which had no environment to fold

youknowone:
`LaunchFlags` and the environment fold over it move from `pyrex` into a new
`pyre-interpreter::launch_env`. `importing::set_runtime_flags` takes the block by
reference instead of fifteen positional arguments.

## The gap

wasm32 has no environment — every environment lookup in the guest returns unset.
So `-P`, `-O`, `PYTHONWARNINGS`, `PYTHONUTF8`, `PYTHONIOENCODING`,
`PYTHONDONTWRITEBYTECODE` and the `LC_ALL`/`LC_CTYPE`/`LANG` chain all read as
absent in the guest and `sys.flags` reported the defaults. Measured before the
change, native vs wasm on the same script:

| set | field | native vs wasm |
|---|---|---|
| `PYTHONSAFEPATH=1` | `safe_path` | `True` vs `False` |
| `PYTHONOPTIMIZE=2` | `optimize` | `2` vs `0` |
| `PYTHONDONTWRITEBYTECODE=1` | `dont_write_bytecode` | `1` vs `0` |
| `PYTHONWARNINGS=...` | `warnoptions` | `[...]` vs `[]` |
| `PYTHONUTF8=1` | `utf8_mode` | `1` vs `0` |
| `LC_ALL=C` | `utf8_mode` | `1` vs `0` |

The runner already forwarded `PYTHONSAFEPATH` alone, through a dedicated
`pyre_set_safe_path` export.

## The shape

Rather than teach the runner to reproduce `config_read_env_vars` semantics
(`_Py_get_env_flag`'s unset/empty/clean-int/anything-else ladder, the fatal
`PYTHONUTF8`, the locale chain), the fold moves to a crate both the launcher and
the guest already link, so there is one copy.

`launch_env::set_launch_env` installs an explicit environment table; once
installed it is authoritative and the process environment is not consulted. The
guest gains `pyre_set_launch_env(ptr, len)`, taking NUL-separated `NAME=VALUE`
records, and `pyre_launch_env_names()`, which returns the names the fold reads —
so the host does not keep its own copy of the list and cannot drift from it.

`run_python_impl` folds them over `LaunchFlags::default()` and calls
`set_runtime_flags` before `install_builtin_modules`, i.e. before the first
`import sys`. The `SAFE_PATH` cell is gone; the `sys.path[0]` decision now reads
`importing::safe_path_flag()`, the same value `sys.flags.safe_path` reports.

The runner forwards the named variables as raw bytes and resolves nothing.
`pyre_set_safe_path` stays for a module predating the new exports; the runner
falls back to it when either new export is absent.

An invalid `PYTHONUTF8` is a fatal pre-init error. `finalize` returns it instead
of exiting, so `pyrex` keeps printing the `preconfig_init_utf8_mode` banner and
exiting, and the guest writes the same banner to fd 2 and returns status 1.

## Values travel as bytes, not `String`

Whether a variable is *set* is decided on undecoded bytes, so the table and the
host/guest transport both carry `Vec<u8>`, and reads split by what each fold
needs:

- **presence** (`PYTHONSAFEPATH`, `PYTHONDEVMODE`, `PYTHONINSPECT`) and **exact
  spelling** (`PYTHONUTF8`) compare bytes, so a value that does not decode stays
  distinguishable from an absent one — it is *set*, and for `PYTHONUTF8`
  *invalid*, which is fatal rather than a silent fall-through to the locale.
- **integer flags** (`PYTHONOPTIMIZE`, `PYTHONNOUSERSITE`, …) treat undecodable
  bytes the way `_Py_str_to_int` treats trailing junk: as 1, not as unset.
- **free text** (`PYTHONWARNINGS`, `PYTHONIOENCODING`, the locale chain) keeps the
  `env::var` contract and decodes as UTF-8.

Reads route through `host_seam::getenv`, which hands back raw bytes —
`pyre-interpreter` forbids reaching the process environment directly.

## Four names the launcher never read

`PYTHONNOUSERSITE`, `PYTHONUNBUFFERED`, `PYTHONDEVMODE` and `PYTHONINSPECT` were
absent from `pyrex` before this change as well, so those four `sys.flags` fields
were reachable only from the command line. They fold now.

Which shape each takes is per-variable rather than a category, and was measured
rather than ported: PyPy's `parse_env` promotes every nonempty value to at least
1, which disagrees with 3.14 on two of them.

| name, value `0` | PyPy 3.11 | CPython 3.14 | implemented as |
|---|---|---|---|
| `PYTHONSAFEPATH` | on | on | presence |
| `PYTHONDEVMODE` | on | on | presence |
| `PYTHONINSPECT` | on | on | presence |
| `PYTHONNOUSERSITE` | on | **off** | integer flag |
| `PYTHONUNBUFFERED` | on | **off** | integer flag |

`PYTHONINSPECT` sets only `inspect`, not `interactive`, so a run that owes the
flag to the environment still reports `interactive` as false.

## Verification

25 environment cases diffed three ways — CPython 3.14 vs pyre-native vs
pyre-wasm — covering each name's `=1`, `=0`, junk-value and empty-value arm plus
`-E`/`-I`/`-P`/`-s`/`-OO` precedence. All 25 agree. The rows discriminate:
`PYTHONSAFEPATH=0` enables while `PYTHONSAFEPATH=` does not, and
`PYTHONNOUSERSITE=0` does not enable while `PYTHONNOUSERSITE=x` does.

`check.py` — dynasm 378/378, cranelift 378/378, wasm 374/374.

## Known limits

- `sys.flags` is materialized once per process, so a second `pyre_run_python` in
  the same instance with a different environment would split state. Latent — the
  runner does one run per instance.
- `utf8_mode` still infers the locale from the `LC_ALL`/`LC_CTYPE`/`LANG` strings
  rather than from the effective `LC_CTYPE`. Pre-existing, and it diverges when
  the environment names a locale the system cannot set: on
  `LC_ALL=xx_YY.invalid`, 3.14 falls back to C and reports `utf8_mode 1`. Filed
  separately — the fix needs a `setlocale` entry point on the host seam and a
  defined wasm answer, neither of which exists today.

### DIFF
--- a/pyre/pyre-interpreter/src/importing.rs
+++ b/pyre/pyre-interpreter/src/importing.rs
@@ -1815,9 +1815,9 @@ pub fn quiet_flag() -> bool {
     SYS_QUIET.load(Ordering::Relaxed)
 }
 
-/// `-i`, which `make_flags` reports as both `inspect` and `interactive`.
-/// `PYTHONINSPECT` would set only the former, but the launcher does not
-/// read it, so `-i` is the whole of this flag's input.
+/// `-i`, which `make_flags` reports as both `inspect` and `interactive`, folded
+/// with `PYTHONINSPECT`. The variable sets only `inspect`, so a run that owes
+/// the flag to the environment still reports `interactive` as false.
 pub fn inspect_flag() -> bool {
     SYS_INSPECT.load(Ordering::Relaxed)
 }
--- a/pyre/pyre-interpreter/src/importing.rs
+++ b/pyre/pyre-interpreter/src/importing.rs
@@ -1760,42 +1760,26 @@ pub fn no_site_flag() -> bool {
     SYS_NO_SITE.load(Ordering::Relaxed)
 }
 
-/// Record the command-line flags consumed by `app_main.py` before `sys` is
+/// Record the launcher option block consumed by `app_main.py` before `sys` is
 /// initialized.  The codec paths also read `dev_mode` directly, matching
-/// `space.sys.get_flag('dev_mode')` in `unicodeobject.py`.
-#[allow(clippy::too_many_arguments)]
-pub fn set_runtime_flags(
-    quiet: bool,
-    inspect: bool,
-    no_user_site: bool,
-    ignore_environment: bool,
-    isolated: bool,
-    dev_mode: bool,
-    utf8_mode: i64,
-    safe_path: bool,
-    optimize: i64,
-    bytes_warning: i64,
-    dont_write_bytecode: bool,
-    unbuffered: bool,
-    xoptions: Vec<String>,
-    warnoptions: Vec<String>,
-    stdio_encoding: Option<String>,
-) {
-    SYS_QUIET.store(quiet, Ordering::Relaxed);
-    SYS_INSPECT.store(inspect, Ordering::Relaxed);
-    SYS_NO_USER_SITE.store(no_user_site, Ordering::Relaxed);
-    SYS_IGNORE_ENVIRONMENT.store(ignore_environment, Ordering::Relaxed);
-    SYS_ISOLATED.store(isolated, Ordering::Relaxed);
-    SYS_DEV_MODE.store(dev_mode, Ordering::Relaxed);
-    SYS_UTF8_MODE.store(utf8_mode, Ordering::Relaxed);
-    SYS_SAFE_PATH.store(safe_path, Ordering::Relaxed);
-    SYS_OPTIMIZE.store(optimize, Ordering::Relaxed);
-    SYS_BYTES_WARNING.store(bytes_warning, Ordering::Relaxed);
-    SYS_DONT_WRITE_BYTECODE.store(dont_write_bytecode, Ordering::Relaxed);
-    SYS_UNBUFFERED.store(unbuffered, Ordering::Relaxed);
-    *SYS_XOPTIONS.lock().unwrap() = xoptions;
-    *SYS_WARNOPTIONS.lock().unwrap() = warnoptions;
-    *SYS_STDIO_ENCODING.lock().unwrap() = stdio_encoding;
+/// `space.sys.get_flag('dev_mode')` in `unicodeobject.py`.  The block must have
+/// been through `launch_env::finalize`, which is what fills `utf8_mode`.
+pub fn set_runtime_flags(flags: &crate::launch_env::LaunchFlags) {
+    SYS_QUIET.store(flags.quiet, Ordering::Relaxed);
+    SYS_INSPECT.store(flags.inspect, Ordering::Relaxed);
+    SYS_NO_USER_SITE.store(flags.no_user_site, Ordering::Relaxed);
+    SYS_IGNORE_ENVIRONMENT.store(flags.ignore_environment, Ordering::Relaxed);
+    SYS_ISOLATED.store(flags.isolated, Ordering::Relaxed);
+    SYS_DEV_MODE.store(flags.dev_mode, Ordering::Relaxed);
+    SYS_UTF8_MODE.store(flags.utf8_mode.unwrap_or(0), Ordering::Relaxed);
+    SYS_SAFE_PATH.store(flags.safe_path, Ordering::Relaxed);
+    SYS_OPTIMIZE.store(flags.optimize, Ordering::Relaxed);
+    SYS_BYTES_WARNING.store(flags.bytes_warning, Ordering::Relaxed);
+    SYS_DONT_WRITE_BYTECODE.store(flags.dont_write_bytecode, Ordering::Relaxed);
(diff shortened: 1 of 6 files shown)


youknowone:

Every behavioural finding was probed against `python3.14` (the compat target) and
`pypy3` (the porting reference) before acting. Three of them turned out to be
places where the two disagree, so porting PyPy's spelling would have introduced a
divergence rather than closed one.

**§1 — 1 finding, 1 fixed.** Non-UTF-8 `PYTHONSAFEPATH` dropped by
`into_string().ok()`. Real. The table and the host/guest transport now carry
`Vec<u8>`, and presence is tested on the undecoded bytes. `PYTHONSAFEPATH=<0xff>`
reports `safe_path True` on 3.14 and now here too.

**§2 — 4 findings, 4 reclassified → then fixed anyway.** All four
(`PYTHONNOUSERSITE`, `PYTHONUNBUFFERED`, `PYTHONDEVMODE`, `PYTHONINSPECT`) were
filed as *introduced by this patch*, but they are pre-existing: at the parent
commit each appears **0 times** in `pyrex/src/lib.rs`, and `PYTHONINSPECT`
appears only inside an `importing.rs` comment that says "the launcher does not
read it". The report's own §3 says the same. Reclassified to §3 — and then fixed,
since the fold is self-contained and sits in code this patch created.

Which shape each takes was measured, not ported. PyPy's `parse_env` promotes any
nonempty value to at least 1, which disagrees with 3.14 on two of them:

| name, value `0` | pypy3 | python3.14 | implemented as |
|---|---|---|---|
| `PYTHONDEVMODE` | on | on | presence |
| `PYTHONINSPECT` | on | on | presence |
| `PYTHONNOUSERSITE` | on | **off** | integer flag |
| `PYTHONUNBUFFERED` | on | **off** | integer flag |

**§3 — 5 findings: 1 fixed, 3 won't-fix (documented), 1 deferred.**

- *`PYTHONDONTWRITEBYTECODE=0` / `PYTHONOPTIMIZE=0` should clamp to 1* — **won't-fix**.
  This is the PyPy/3.14 split above: `PYTHONOPTIMIZE=0` gives `optimize 1` on
  pypy3 and `0` on 3.14; pyre already matches 3.14. `env_int_flag`'s comment
  names `_Py_get_env_flag` as the rule it implements.
- *invalid `PYTHONUTF8` should quote the value* — **won't-fix**. 3.14 prints
  `Fatal Python error: preconfig_init_utf8_mode: invalid PYTHONUTF8 environment
  variable value` with **no** value; only pypy3 appends `'bogus'`. pyre already
  matched 3.14 byte-for-byte.
- *the four missing names* — duplicate of §2, fixed.
- *locale inferred from `LC_ALL`/`LC_CTYPE`/`LANG` instead of the effective
  `LC_CTYPE`* — **deferred**, real and observable: on `LC_ALL=xx_YY.invalid`,
  3.14 falls back to C and reports `utf8_mode 1` while the string cascade reports
  0. Blocked on a `setlocale` entry point for `host_seam` (the crate forbids
  touching process state directly) plus a defined wasm answer, since wasm32 has
  no libc locale. Note pypy3 is not a usable model here — it aborts with
  `unsupported locale setting` on that input. Filed as a follow-up.

**§4 — 3 findings, 3 won't-fix (documented).** The mutex-protected supplied-env
table, the NUL-separated host/guest ABI, and the runner querying the guest for
the name list are all wasm-boundary adaptations with no RPython counterpart,
which is the point — wasm32 has no process environment. Each is commented at its
definition.

### Beyond the report

Probing the byte-handling boundary turned up two more instances of the same class
as §1, both fixed here:

- `PYTHONUTF8` with an undecodable value read as *absent* and silently fell
  through to the locale. On 3.14 it is *invalid*, which is fatal. It is compared
  as bytes now.
- integer flags with an undecodable value read as unset. `_Py_str_to_int` rejects
  undecodable input the way it rejects trailing junk, so 3.14 yields 1
  (`PYTHONOPTIMIZE=<0xff>` → `optimize 1`). They read bytes now.

### Verification

25 environment cases diffed three ways — 3.14 vs pyre-native vs pyre-wasm — each
name's `=1`, `=0`, junk and empty arm plus `-E`/`-I`/`-P`/`-s`/`-OO` precedence.
All 25 agree, and the rows discriminate: `PYTHONSAFEPATH=0` enables while
`PYTHONSAFEPATH=` does not; `PYTHONNOUSERSITE=0` does not enable while
`PYTHONNOUSERSITE=x` does.

`check.py` on the current base: dynasm 378/378, cranelift 378/378, wasm 374/374.
The sandbox clippy gate that was red is green under CI's exact invocation.

Two unrelated defects found while verifying, filed separately rather than pulled
into this diff: the `utf8_mode` locale item above, and the wasm guest discarding
buffered stdout on `os._exit` (native emits it; reproduces with no environment
variable set, so it is not from this change).

— *commented by Claude*

PR #1047, opened 2026-08-05
jit: inline a user __hash__ under the hash() builtin call residual

fregataa:
## Problem

Second half of the `synth/build_set_hashability` investigation (first half: #1037). `hash(x)` over a user instance runs the resolved `__hash__` through the generic call residual — allocating a PyFrame and entering `eval_loop` on every call — because the builtin `hash` callable declines the user-function inline route (`resolve_inlinable_callee` takes plain Python functions only). A `hash(obj)` hot loop measured ~75x slower than PyPy from the per-call frame/dispatch overhead alone, and the same cost hits any dict/set workload keyed by user objects going through the builtin.

## Fix

Recognize the canonical `hash` builtin at the CallFn residual site (new `is_builtin_hash_function`, keyed on builtin-code identity like `is_builtin_repr_function`) and inline the receiver's `__hash__` the way the exception `__str__`/`__repr__` override does:

- pin the receiver class + type version tag, then sub-walk the resolved `__hash__` body through `try_walker_inline_resolved_user_call`;
- emit `hash_call_normalize`'s digest handling in-trace: unbox the machine-int digest and apply the `-1 -> -2` map as `raw - (raw == -1)` (two int ops, virtualization-friendly);
- an effect-free body is sampled before any IR so a bool/long/non-int digest declines cleanly to the residual; a side-effecting body cannot be sampled or re-run, so an observed bool/long digest (legal but rare) aborts the walk rather than recording a normalization shape that contradicts its own operand;
- every other shape — non-instance receiver, `__hash__ = None`, non-Python or branchy body — declines to the residual unchanged.

## Results

`hash(obj)` hot loop, 2M iterations, dynasm, same quiet machine: **2.8s → 0.85s (3.3x)**. `synth/build_set_hashability` itself is mostly unchanged: its element hashing happens inside the opaque `bh_build_set_from_array` residual, which this PR does not decompose (see below).

## Verification

- `pyre/check.py` all green: dynasm 377/377, cranelift 377/377, no jit-stats gate movement.
- Digest edge cases byte-identical to CPython on dynasm under `MAJIT_STRICT=1`: `-1 -> -2` in a hot loop, `True` and `2**70` digests, raising `__hash__`, `__hash__ = None`, non-int digest TypeError, side-effecting `__hash__` called exactly once per `hash()` call, and a mid-loop digest type flip (int → big int) deopting with exact values.
- `cargo test` for the touched crates: 489 passed; the one failure (`jit_fnaddr::registered_paths_sharing_an_address_are_alias_spellings`) reproduces on clean `upstream/main` locally and is unrelated.

## Follow-up (not in this PR)

Decomposing `bh_build_set_from_array` so set-literal element hashing inlines too has a soundness wall: a guard failure mid-literal resumes at the BUILD_SET opcode and re-executes earlier elements' `__hash__`, duplicating side effects (the fixture's `HASH_CALLS.append` is exactly that shape). The sound scope without mid-opcode resume coordinates is effect-free `__hash__` bodies only; left as follow-up.

## Summary by CodeRabbit

* **Performance**
  * Improved execution speed for `hash()` calls on user-defined instances through safe inlining.
  * Added safeguards to preserve correct behavior when classes or methods change.
* **Bug Fixes**
  * Ensured hash results are validated and normalized correctly, including the reserved `-1` value.
  * Added a safe fallback when hashing cannot be optimized.

### DIFF
--- a/pyre/pyre-interpreter/src/builtins.rs
+++ b/pyre/pyre-interpreter/src/builtins.rs
@@ -11765,6 +11765,12 @@ fn hash_call_normalize(
     w_type: PyObjectRef,
 ) -> Result<i64, crate::PyError> {
     let r = unsafe { crate::baseobjspace::get_and_call_function(method, obj, w_type, &[]) }?;
+    normalize_hash_digest(r)
+}
+
+/// The result arms of [`hash_call_normalize`], shared with the JIT's digest
+/// residual: accept `bool`/`int`/`long`, reject other types, map `-1` to `-2`.
+pub fn normalize_hash_digest(r: PyObjectRef) -> Result<i64, crate::PyError> {
     let h = unsafe {
         if is_bool(r) {
             pyre_object::w_bool_get_value(r) as i64
--- a/pyre/pyre-interpreter/src/builtins.rs
+++ b/pyre/pyre-interpreter/src/builtins.rs
@@ -4053,6 +4053,24 @@ pub fn is_builtin_repr_function(callable: PyObjectRef) -> bool {
     }
 }
 
+/// True iff `callable` is the canonical builtin `hash` function object; the
+/// walker's user-`__hash__` inline keys on this identity.
+pub fn is_builtin_hash_function(callable: PyObjectRef) -> bool {
+    unsafe {
+        if callable.is_null() || !crate::is_function(callable) {
+            return false;
+        }
+        let code = crate::function_get_code(callable) as PyObjectRef;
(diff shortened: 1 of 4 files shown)


youknowone:
@fregataa could you check parity review section 1 & 2? Feel free to ignore them if it doesn't help, but they are usually worth to check

youknowone:
Thank you so much!

PR #1048, opened 2026-08-05
Filesystem names with no UTF-8 spelling, and the symbolic fnaddr tag's co-tenancy with the synthetic space

youknowone:
Two defect families, both of which abort the process and neither of which
macOS CI can observe.

## 1. Keep the symbolic fnaddr tag out of the synthetic fnaddr space

Follow-up to #1030. That PR replaced the `(fnaddr as u64) >> 47 != 0` heuristic
with an explicit high-16 tag, `SYMBOLIC_FNADDR_BASE = 0xFADD_0000_0000_0000`.
That constant sets bit 63, and `BhDescr::JitCode.fnaddr` is shared with a second
encoding whose *entire space* is a sign test:

```rust
// majit-backend::synthetic_cpu
pub const SYNTHETIC_FNADDR_BASE: i64 = i64::MIN;
pub fn is_synthetic_fnaddr(x: i64) -> bool { x < 0 }
pub fn decode_synthetic(x: i64) -> usize { (x & i64::MAX) as usize }
```

Every symbolic hash therefore landed inside the synthetic space, where
`decode_synthetic` would answer it with a fabricated jitcode index. The tag
moves to `0x7ADD_0000_0000_0000` — the shape the sibling
`STR_CONST_SENTINEL_BASE = 0x7E57_…` already uses in `codewriter/assembler.rs` —
with bit 63 left clear and the reason recorded at the constant.

No test or CI leg could have caught this: the colliding consumer is currently
`unimplemented!()`, so the collision stays latent until a `SyntheticCpu`
dispatch goes live. `fnaddr_classifiers_match_the_walker_symbolic_gate` now
asserts the invariant directly (`symbolic > 0`, `!is_synthetic_fnaddr(symbolic)`)
next to the aarch64-Linux real-address cases #1030 added.

## 2. Filesystem names that have no UTF-8 spelling

Every filesystem name pyre reads is decoded with the surrogateescape handler
(`fs_name_obj` → `fsdecode_filename_bytes` → `charp2uni` →
`decode_utf8_with_errors(data, "surrogateescape")`), so the resulting `str` can
hold lone surrogates in U+DC80..U+DCFF. `w_str_get_value` *panics* on such a
buffer — a process abort, not a raise — and seven boundaries reached it, or
misreported it, with a name that came straight off the filesystem.

| site | was | now |
| --- | --- | --- |
| `importing` `python_sys_path_dirs`, `parent_package_path` | `&str` → `PathBuf` | `gateway::fspath_buf`, which folds the escapes back to the original bytes through `fsencode` |
| `importing` sys.path entry compare, bootstrap-filename check | `w_str_get_value` | `w_str_get_value_opt` — a non-UTF-8 entry compares unequal |
| `posix.DirEntry.__repr__` | `&str` interpolation | the name's own `repr`, matching `DirEntry_repr`'s `"<DirEntry %R>"` |
| `_posixsubprocess.fork_exec` | `w_str_get_value(o).as_bytes()` | `gateway::fsencode` |
| `baseobjspace::realunicode_w` | `w_str_get_value` | `str_utf8_w` → `UnicodeEncodeError: surrogates not allowed` |
| `posix.stat` | every `fsencode_path_w` error remapped to one `TypeError` | only a genuine `TypeError` is re-reported under `stat`'s wording |
| `write_syntax_error_object` (`SyntaxError.filename`, `.text`) | `w_str_get_value` | rendered with `backslashreplace`, as stderr already does |

Two of those fall out on the way past:

- `DirEntry.__repr__` read the name as a `str` unconditionally, so
  `os.scandir(b'.')` — whose `name` is `bytes` — was a type confusion.
  Rendering by `repr` spells a bytes name `b'…'`, as it should.
- `posix.stat` collapsed the embedded-null `ValueError` and the surrogate
  `UnicodeEncodeError` into `TypeError: stat: path should be string, bytes,
  os.PathLike`. CPython 3.14 answers `os.stat('a\0b')` with `ValueError:
  stat: embedded null character in path` and `os.stat('\ud800')` with
  `UnicodeEncodeError`.

**A path entry that cannot be encoded is now reported, not skipped.**
`fspath_buf` first answered `None` there and both import path readers dropped
the entry and searched on. CPython 3.14 raises `UnicodeEncodeError` for
`sys.path.insert(0, '\ud800')` and for the same entry in a package `__path__`,
so the error travels out through `python_sys_path_dirs` → `find_in_sys_path` →
`find_module` to `load_part`, which already returned `Result`. The non-unix arm
of `fspath_buf` also stopped round-tripping through
`String::from_utf8_lossy`, which substituted U+FFFD and made distinct entries
alias onto one another; it reads the text with `str_utf8_w` instead.

**Why none of this shows up on macOS.** Exposure is filesystem *capability*, not
a code branch: APFS and HFS+ reject a non-UTF-8 filename with `EILSEQ`, so the
repro cannot even be planted, while ext4 and tmpfs accept arbitrary bytes.

Two parity tests, deliberately split by what they need:

- `os_undecodable_name_boundaries.py` plants `b"pyre_undecodable_\xff"` in a
  tempdir and covers `os.listdir`, both `DirEntry` reprs, an import off
  `sys.path`, the exec boundary, and the `SyntaxError` traceback. It self-skips
  where the filesystem refuses the name (and on win32), so in practice it runs
  on Linux.
- `import_unencodable_path_entry.py` needs no filesystem support at all — the
  entry never reaches a syscall — so it runs on every platform.

## 3. State the full cpu surface the per-thread `BackendImpl` backs (comment only)

Adjudicating a §3 finding from #1030's Codex review: `pyre_production_cpu` leaks
one `BackendImpl` per thread, so `bh.cpu` is a *different* object from the
`MetaInterp::backend` that recorded the trace, and the finding was that the
blackhole could observe backend state distinct from the metainterpreter's.

It cannot, today. The surface reachable through `bh.cpu` is
`bh_getfield_gc_{i,r,f}`, `bh_setfield_gc_{i,r,f}`,
`bh_getarrayitem_gc_{i,r,f}`, `bh_arraylen_gc`, `bh_new`,
`bh_new_with_vtable`, `bh_new_array{,_clear}`, `bh_call_{i,r,f,v}` and
`clear_stored_exception` — on dynasm every one is a pure function of its
arguments (descr-supplied offsets into `read_int_at_mem`/`write_int_at_mem`, or
a free function such as `jit_exc_clear` / `call_stub::bh_call_*_dispatch`), none
reading `&self`.

The old comment justified the split with "vable handlers only invoke stateless
reads" while listing four of those methods. It now records the whole list, the
invariant, and what breaks if a `&self`-reading method joins it. The
architectural port — hand the metainterp's `BackendImpl` to
`BlackholeInterpBuilder` and delete the TLS — stays open in the in-code TODO,
blocked on the `Backend` trait's `Send + Sync` bound and on `BH_BUILDER3`
outliving any single `MetaInterp`.

## A boundary this does *not* close

Running pyre with a script path that is not valid UTF-8 aborts before Python
starts, in the launcher rather than in anything above:

```
panicked at library/std/src/env.rs:876:51:
called `Result::unwrap()` on an `Err` value: "/tmp/…/broken_pyre_undecodable_\xFF.py"
```

`pyrex/src/lib.rs:656` reads argv with `std::env::args()`, which panics on any
argument that is not valid Unicode (internally `OsString::into_string().unwrap()`
— the `Err` payload above is exactly that). Swapping in `args_os()` is not the
fix: the launcher carries argv as `String` throughout (`RunMode`,
`parser.raw_args()`, `set_sys_argv(&[String])`), and a Rust `String` cannot hold
a surrogate at all. The convergence path is to carry argv as `OsString`/`Vec<u8>`
and build the str objects at the `sys.argv` seam with `fsdecode_filename_bytes`,
the same producer `readdir` names use.

That is the process-entry boundary, not a name read back from disk, so it is
left out rather than folded in here. `os_undecodable_name_boundaries.py`
*imports* its broken module instead of running it precisely to keep the
undecodable name off the command line, and says so where it does it.

## Verification

- Linux (aarch64 Debian trixie, ext4) A/B on the boundary test — before:
  `panicked at pyre/pyre-object/src/unicodeobject.rs:579: w_str_get_value:
  backing Wtf8Buf is not valid UTF-8 (lone surrogate)`; after: `OK`; CPython
  control `OK`. This is where the boundary test's assertions actually run;
  on APFS it skips itself.
- Linux full parity suite on the branch tip: 178 scripts, **0 dynasm FAILs**.
  (The 41 `cpython=FAIL` rows there are the rig — Debian ships python3.13, so
  the `*_python314.py` fixtures fail on the *control* side.)
- macOS `python3 pyre/check.py`: **dynasm 378/378, cranelift 378/378,
  wasm 374/374, 3/3 backend runs**.
- Both new parity tests were validated against a real CPython 3.14.5 first and
  only then against pyre, so each asserts upstream behaviour rather than
  pyre's.

### DIFF
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -3992,6 +3992,13 @@ mod tests {
                 assert!(super::is_callable_fnaddr(real_high));
             }
 
+            // The symbolic tag shares `BhDescr::JitCode.fnaddr` with the
+            // synthetic tag, whose space is every negative `i64`
+            // (`synthetic_cpu::is_synthetic_fnaddr`).  The two must not
+            // overlap.
+            assert!(symbolic > 0, "symbolic fnaddrs must stay out of the synthetic (negative) space");
+            assert!(!majit_backend::synthetic_cpu::is_synthetic_fnaddr(symbolic));
+
             assert!(!super::is_symbolic_fnaddr(0));
             assert!(!super::is_callable_fnaddr(0));
         }
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -7804,8 +7804,7 @@ fn handler_residual_call_r_v(
 // `blackhole.py:107`).  All `*_pyre_u16` width adapters have been
 // retired; canonical `handler_*` decoders own every dispatch slot.
 
-/// Per-thread Backend instance for blackhole's `bh_getfield_gc_*` /
-/// `bh_setfield_gc_*` / `bh_getarrayitem_gc_*` / `bh_arraylen_gc` reads.
+/// Per-thread Backend instance backing every `bh.cpu` call.
 ///
 /// TODO: RPython `blackhole.py:55-56,286` reads
 /// `self.cpu = builder.cpu`, where `builder.cpu` is the metainterp-shared
@@ -7815,11 +7814,24 @@ fn handler_residual_call_r_v(
 ///
 /// pyre's `MetaInterp::backend` is per-instance (not `'static`) and
 /// owns trace-compilation state (descr registries, etc.) that would be
-/// inappropriate to share with the blackhole `bh.cpu` field which only
-/// needs the stateless GC memory-access methods.  We leak ONE
-/// `BackendImpl` instance per thread — its `bh_getfield_gc_*` and
-/// related methods do direct unsafe pointer arithmetic
-/// (`runner.rs:2244-2271` for dynasm).
+/// inappropriate to share with the blackhole `bh.cpu` field.  We leak ONE
+/// `BackendImpl` instance per thread instead, so `bh.cpu` is a *different*
+/// object from the one the metainterp that recorded the trace holds.
+///
+/// ## The invariant that makes the split observationally inert
+///
+/// Every `Backend` method this field reaches must be a pure function of
+/// its arguments — `&self` may not be read.  The full reachable surface is
+/// `bh_getfield_gc_{i,r,f}`, `bh_setfield_gc_{i,r,f}`,
+/// `bh_getarrayitem_gc_{i,r,f}`, `bh_arraylen_gc`, `bh_new`,
+/// `bh_new_with_vtable`, `bh_new_array{,_clear}`, `bh_call_{i,r,f,v}`, and
+/// `clear_stored_exception`.  On dynasm each is descr-supplied offsets fed
+/// to `read_int_at_mem` / `write_int_at_mem` (`runner.rs:1386,1402`) or a
+/// free function (`jit_exc_clear`, `call_stub::bh_call_*_dispatch`); none
+/// touches `&self`.  A method added to that list that *does* read backend
+/// state would make the blackhole observe a fresh, empty `BackendImpl`
+/// where the metainterp observes the populated one, and nothing here would
+/// catch it.
 ///
 /// Convergence path: align with RPython's `builder.cpu` invariant by
 /// passing the metainterp-owned `BackendImpl` to `BlackholeInterpBuilder`
@@ -7827,8 +7839,7 @@ fn handler_residual_call_r_v(
 /// `BH_BUILDER3` outlives any single MetaInterp invocation).  Open
 /// architectural item; not in scope here because the change cascades
 /// through the Backend trait's `Send + Sync` bound and every callsite
-/// that holds `&'static dyn Backend` today.  The leak is functionally
-/// correct because vable handlers only invoke stateless reads.
+/// that holds `&'static dyn Backend` today.
 #[cfg(any(feature = "dynasm", feature = "cranelift"))]
 pub fn pyre_production_cpu() -> &'static dyn majit_backend::Backend {
     // Per-thread leak: `BackendImpl` is not `Sync`, so we keep one
--- a/majit/majit-metainterp/src/blackhole.rs
+++ b/majit/majit-metainterp/src/blackhole.rs
@@ -3996,7 +3996,10 @@ mod tests {
             // synthetic tag, whose space is every negative `i64`
             // (`synthetic_cpu::is_synthetic_fnaddr`).  The two must not
             // overlap.
-            assert!(symbolic > 0, "symbolic fnaddrs must stay out of the synthetic (negative) space");
+            assert!(
+                symbolic > 0,
+                "symbolic fnaddrs must stay out of the synthetic (negative) space"
+            );
             assert!(!majit_backend::synthetic_cpu::is_synthetic_fnaddr(symbolic));
 
             assert!(!super::is_symbolic_fnaddr(0));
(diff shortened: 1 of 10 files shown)


youknowone:

**§1: none. §2: 2 findings, both reclassified — one refuted, one pre-existing. §3: 2, both documented/deferred. §4: 2, agreed.**

### §2.1 `_posixsubprocess/mod.rs:85` — refuted

> the new `fsencode(o)?` makes a surrogate-containing `str` in `executable_list`/`env_list` acceptable […] PyPy's `seqstr2charpp` requires `bytes0_w`

Two problems with the premise.

**The change did not widen what is accepted.** The `is_str(o)` branch was already there; only its encoder changed:

```diff
-                w_str_get_value(o).as_bytes().to_vec()
+                crate::gateway::fsencode(o)?
```

A `str` was accepted before and is accepted now. What changed is that a surrogate no longer **aborts the process** — `w_str_get_value` panics on a WTF-8 buffer holding a lone surrogate, which is the whole subject of this PR.

**The oracle is CPython 3.14, not PyPy here.** Measured, pyre vs CPython, same three inputs:

| argv element | CPython | pyre (this PR) |
| --- | --- | --- |
| `os.fsdecode(b'sur\xff')` | `b'sur\xff\n'` | `b'sur\xff\n'` |
| `'plain'` | `b'plain\n'` | `b'plain\n'` |
| `'\ud800'` | `UnicodeEncodeError` | `UnicodeEncodeError` |

(measured on Linux/ext4; before this PR the first row aborted the interpreter). CPython's `_posixsubprocess` runs each item through `PyUnicode_FSConverter`, which is exactly `fsencode` — so this moves *toward* the target, not away.

### §2.2 `gateway.rs:1692` — pre-existing, reclassified to §3

The finding is that `fsencode` hard-codes UTF-8/surrogateescape instead of the configured filesystem encoding. That is a property of `fsencode` itself, which landed in #1018 and #1027 — this PR only *routes through* it, which is the repo's established filesystem-encoding boundary. Same reclassification as the pre-merge run of this review made for the identical finding. The locale-codec divergence is worth its own issue; introducing a second encoding path here would be the wrong fix.

### §3.1 `blackhole.rs:7820` — won't-fix (documented), deliberately

Correct that the identity split is unchanged. It is inert today and the PR now says why: the full surface reachable through `bh.cpu` is `bh_getfield_gc_{i,r,f}`, `bh_setfield_gc_{i,r,f}`, `bh_getarrayitem_gc_{i,r,f}`, `bh_arraylen_gc`, `bh_new`, `bh_new_with_vtable`, `bh_new_array{,_clear}`, `bh_call_{i,r,f,v}`, `clear_stored_exception` — on dynasm every one is a pure function of its arguments, none reads `&self`. The port stays open in the in-code TODO with its blocker named.

### §3.2 `interp_posix.rs:2426` — deferred, unrelated

`stat`/`lstat` not supporting `dir_fd` or fd-as-path is real and predates this PR; nothing here touches that path. Filed separately rather than pulled into a diff about name encoding.

### §4 — agreed, both

— *commented by Claude*

PR #1050, opened 2026-08-05
jit, gc: traced traceback nodes non-moving; virtualizable debugdata and token fixes; journal the walk-time traceback attaches

youknowone:
Thirteen commits: three wrong-code fixes, three crash fixes, one GC diagnostic,
and the fixtures and baselines they need.

## Wrong code

* **`debugdata` carried as a box, not a recording-time constant.**
  `MIFrame::flush_to_frame` republished all six virtualizable statics as
  constants read out of the live frame. `DELETE_NAME` resolves the name through
  `debugdata.w_locals`, so a constant pins the recording frame's mapping and the
  compiled loop then deletes through it once `exec(code, globals)` reuses the
  same code object with a fresh namespace, raising `NameError`.
  `virtualizable.py:86-93 read_boxes` carries every static as a loop-carried box
  for exactly this reason; the remaining pointer statics keep their
  promoted-constant representation because the bytecode path cannot rebind them
  mid-loop.

* **The pointer flag survives the `BhFieldSpec` round trip.**
  `from_field_descr` mapped every pointer field to `ArrayFlag::Unsigned`, and
  `field_spec_from_bh` copies that flag into the rebuilt
  `SimpleFieldDescrSpec`, whose `is_pointer_field()` is
  `flag == ArrayFlag::Pointer` (`descr.py:173`). A pointer field that traversed
  the round trip therefore reported itself as non-pointer and
  `handle_write_barrier_setfield` declined its write barrier.

* **Walk-time traceback attaches are journaled.**
  Three of the four attach sites mutated the live exception outside the fbw undo
  log, so a walk that did not commit left its node attached and the
  interpreter's replay recorded the same frames again. A freshly raised
  exception hides this — the discarded walk's object is unreachable — so the
  observable case is a `raise` that re-uses one exception object.

## Crashes

* **The virtualizable token is settled on the top-level raise exit.**
  `compile_exit_frame_with_exception` (`pyjitpl.py:3261`) opens with
  `store_token_in_vable()`, the same as `compile_done_with_this_frame`
  (`:3223`). pyre's value and void arms settle the token; the raise arm only
  published `last_instr`, so the exception exit left the frame naming a jitframe
  that dynasm frees in `execute_token`, and the next `is_force_token_armed`
  walked `jf_forward` off freed memory.

* **The root walk skips the PyObject-shaped raw walks for a virtual ref in a
  frame slot.** `walk_pyframe_roots_area` handed every `locals_cells_stack_w`
  slot to `walk_raw_exception_roots`, which dereferences the slot's leading word
  as an `ob_type`. A slot can hold a `JitVirtualRef`, whose leading word is the
  `JIT_VIRTUAL_REF_VTABLE` magic. The `f_backref` arm one block earlier already
  tested for this; it is extracted and applied to the value-stack slots too.

* **Traced traceback nodes are allocated non-moving.**
  `w_pytraceback_new` allocates through the stable old-gen allocator and states
  the obligation in place: raw `*mut PyTraceback` readers and the exception
  `w_traceback` chain hold bare pointers no root walker reaches.
  `PYTRACEBACK_DESCR_GROUP` never stamped `set_non_moving`, so the same object
  type was movable or not depending on which allocator built it, and a node
  built by compiled code was a nursery object a minor collection could relocate.
  The `W_ObjectObject` group and `mapdict_storage_gcarray_descr` already carry
  the stamp for the same reason; a new test asserts all three, since a missing
  flag shows up only as a GC abort at particular nursery sizes and `check.py`
  pins `PYPY_GC_NURSERY` to 4MB, which is not one of them.

  Measured over a 512K–32M `PYPY_GC_NURSERY` sweep: the witness and four
  stack-depth variants abort at 18 of 320 (fixture, size) pairs before the
  change and none after.

## Diagnostic

* `trace_and_update_object` now threads its caller's `site` down to
  `copy_nursery_object` and `assert_traced_slot_initialized` as `parent_site`,
  so the invalid-type_id and nursery-poison panics name the root path that
  reached the object rather than only the slot kind. `&'static str`, so a traced
  slot allocates nothing.

## Fixtures

* `exception_reused_object_tb_not_doubled.py` and
  `exec_fresh_globals_delete_name.py` gate the first and third wrong-code fixes.
* `exception_tb_f_locals_vref_root_walk.py` moves out of `_pending` and into the
  suite — the traceback stamp above unblocked it — and gates the virtual-ref
  walker guard on dynasm and cranelift.
* `gc_varsize_item_const_shape_witness.py` stays under `_pending`; see below.

## Known open, and NOT fixed here

Two divergences are documented in the tree rather than fixed, each with the
measurement that scopes it:

* **cranelift `GC BUG ... site=minor_varsize_item_target`**
  (`_pending/gc_varsize_item_const_shape_witness.py`). The traceback stamp hides
  it, but the stamp is not its root cause: with the stamp removed and the same
  source, cranelift aborts at 3 of 64 nursery sizes while **dynasm aborts at 0
  of 64**. Three hypotheses were tested and refuted this cycle — reloading every
  ref root at every collecting call, treating every residual call as collecting,
  and a probe on every cranelift Ref-store emitter (zero hits, gate proven
  live).

* **wasm loses `e` from a catching frame's `f_locals`.** Found by the promoted
  fixture; compiled-code only (clean at N=4000, wrong from N=8000) and
  reproduces with the whole source tree reset to the merge base, so it predates
  this branch. Same class as the exemption
  `getframe_caller_locals_nested_compiled_callee` already carries, and exempted
  the same way.

## Verification

* `pyre/check.py`: dynasm and cranelift all-pass; wasm all-pass. Re-run after
  the fixture promotion changed the suite composition.
* The traceback-stamp claim is a 64-size `PYPY_GC_NURSERY` sweep per arm, not a
  single run — this abort is a ~5% duty-cycle resonance over nursery size, so
  one clean run is not evidence.
* Codex parity review: sections 1, 2 and 3 empty; three section-4 structural
  adaptations, each already carrying its upstream citation in the code.


youknowone:
# Update note for PR #1050 (draft — post after the suites are green)

Rebased onto `origin/main` @ 22c62a78023 and added two commits.

## New in this push

**`jit: replay the folded inline-callee locals when a traceback node takes the frame`**

A `STORE_FAST` on a strictly-inlined callee's own fresh frame takes the fold in
`setarrayitem_vable_via_metainterp` and records no `SETARRAYITEM_GC`.
`disarm_folded_inline_callee_after_escape` is the replay that writes those slots
back, and it had a single call site — the branch where a residual call forced
the virtualizable. `emit_traceback_node` storing the callee frame box into
`PyTraceback.frame` is a second escape, so the frame outlived the trace carrying
only its seeded parameters.

Witness `synth/exception_inline_callee_tb_frame_locals`: dynasm and cranelift
both answered `('i',)` on 2758 of 4000 iterations before, `('i', 'marker')` on
all 4000 after — matching pypy3, CPython and `PYRE_NO_JIT=1`.

**`bench: park the virtual-ref root-walk fixture back under _pending`**

Its cranelift `guard_failures` is not the same on every host: macos-latest and
ubuntu-24.04 both report 5980, windows-latest reported 5979. That is the band
#1043 removed the `closure_per_call` overlay over, so an overlay cannot hold it
either. The walker guard consequently has no suite gate; the file stays as the
reproduction with the measurement in its header.

## The three CI failures on the previous push

- `cargo fmt --check` — the rebase carries the fix. Verified locally against
  CI's own rustfmt (1.97.1); the earlier local pass was 1.96.1.
- `pyre/check.py (windows-latest)` `dynasm synth/recursive_call_frame_relocation
  guard_failures 638 -> 637` — main fails identically at 295b3fcad04. Not this
  branch.
- `pyre/check.py (ubuntu-24.04)` `cranelift synth/exception_value_op_caught
  ratio 30.5x > gate 24x [pypy exec clamped to floor; ratio not a measurement]`
  — same class as main's own red at the same base (`inline_subwalk_mutating_residual`
  55.0x > gate 40x on ubuntu, `metatype_property_dunder` 5.9x > gate 5x on
  macOS). Not this branch, and the fixture is untouched here.

— *commented by Claude*

PR #1053, opened 2026-08-05
descr: catch the field-position invariant where it is produced, not where it is repaired

youknowone:
## Summary

A field descr's `index_in_parent` (`descr.py:228`) must name the slot its parent
puts the field in — upstream gets this for free because `heaptracker.py:60-72
get_fielddescr_index_in` and `:96-112 all_fielddescrs` are one walker sharing one
skip set. pyre has several independent producers, and
`all_fielddescrs()[index_in_parent]` is a load-bearing lookup
(`optimizeopt/info.rs force_box`), so a stale rank either runs off the end or
emits the store against a different field.

Two producers were leaving that rank stale, and the counter that looks like it
guards this could not see either of them.

## The counter could not have caught this

`field_pos_rederived` was documented as "the one to watch". It cannot be:

* `GcCache::derive_index_in_parent` is the judge **and** the repairman — it
  replaces the caller's number with the parent's in the same expression, so a
  nonzero reading is a log of repairs already applied, never a defect still
  present.
* It mostly never asks. `make_simple_descr_group_keyed_with_headerless` mints
  every field **before** `register_keyed_size` publishes the parent those fields
  will be indexed against, so the dominant path takes the `FIELD_PARENT_ABSENT`
  early return. Measured on this tree: `parent_absent` 1202 against `rederived`
  0.

General shape, worth keeping: *a counter placed inside a normalising reader
measures repairs, not defects, over whatever subset the reader chose to examine.*

## What replaces it

Two censuses taken before anything can normalise, each with a real denominator:

| counter | what it reads | this tree |
| --- | --- | --- |
| `field_pos_spec_{checked,misplaced}` | a parent's positional list as handed to either group factory, including the `cache_key == 0` groups that never enter `_cache_size` | 1712 / 0 |
| `field_pos_attached_{checked,misplaced}` | a standalone field descr against the parent its own producer attached | 247 / 0 |

Both `*_misplaced` join `JITSTATS_BADNESS_FIELDS`. A field absent from a baseline
reads as 0, so they gate from the first run with no re-record.

## …and why a census was still not enough

Negative control, fix reverted, full corpus: **all four** of `field_pos_rederived`,
`field_pos_spec_misplaced`, `field_pos_attached_misplaced` and
`positional_misplaced` still read 0, with `attached_checked` unchanged at 248.
The defect is input-order dependent — it needs a field minted before a
lower-offset sibling registers — so a corpus that registers in offset order
exercises the producer without exercising the bug. Building a better census did
not help; that is the finding.

So the load-bearing check is at the construction site.
`field_descr_position_disagreement` runs at `try_finish` under
`cfg!(debug_assertions)` over every emitted `Field` descr, covering all three
states the pass can leave behind: the offset resolves (index and name are that
slot's), the offset is absent (the `(0, "")` fallback is intact — those are the
inline aggregates `heaptracker.py:68-69` mints no descr for), and the parent's
`type_id` is absent from `struct_size_specs` (the `continue` that patched
nothing). Removing the re-resolution panics with
`field descr at offset 16 … claims slot 0 named "hi", but that offset is slot 1`.

## The two producers

1. **`jitcode/assembler.rs`** — `patch_field_descr_parents` already swapped the
   snapshotted `parent` for the final merged spec but left `index_in_parent` /
   `name`, which came from that same snapshot, so one descr's two halves
   described different fields. `register_struct_layout` re-indexes on every
   merge, which is what invalidates the snapshot.
2. **`codewriter/assembler.rs`** — `fielddescrof`'s `struct_layout_for` and
   `heuristic_field_layout` arms overwrite `offset` from a second source and
   leave `index_in_parent` at its `0` initialiser while `parent` carries the
   whole flattened list. The `found_parent_field` arm is deliberately left alone:
   it picked its spec by name, and re-resolving by offset would return the wrong
   one for a struct with two fields at one offset.

Both fixes are pinned by a test that constructs the state directly, each with the
revert-and-fail control shown.

## wasm

The two new badness fields would have read absent-and-therefore-zero on wasm, the
exact hole the `descr_set_*` exports were added to close. Exported the same way.
Not redundant with the native readings — the invariant is stated in byte offsets
and wasm32 lays structs out on a 4-byte word (`symbolic.py:12`); the wasm run
reports its own denominators (`spec_checked=1665`, `attached_checked=245` against
1712/247 on the dev host), so it is measuring the wasm-side universe.

## Note on the first three commits

`738a524` / `39bd8a9` add an executed-allocation GC root list and then patch a
crash that root itself introduced; `e8204d9` removes both. They net to zero and
are kept as the record: the finding they answered ("the `execute_new` result
lives only in a borrowed slice that is not a GC root") is **false** — the result
is stamped onto the recorded op's `value` cell (`history.py:803-807
*FrontendOp(pos, value)`) and `MetaInterp::walk_active_trace_refs` has forwarded
those cells since `d3bfd17`. Say the word and I will fold them out.

## Verification

`check.py` dynasm 378/378, cranelift 378/378, wasm 374/374;
`cargo test --workspace --no-default-features --features dynasm` 101 suites, no
failures; `cargo fmt --all --check` clean. No jitstats baseline was re-recorded.

— *authored by Claude*

## Summary by CodeRabbit

* **New Features**
  * Added JIT field-position diagnostics with checked and misplaced counts.
  * Exposed field-position metrics through runtime, WebAssembly, and JIT statistics interfaces.
  * Structural regressions now report increases in field-position mismatches.

* **Bug Fixes**
  * Improved field index and name resolution for reordered or merged layouts.
  * Preserved fallback behavior for ambiguous or unmatched fields.

* **Tests**
  * Added coverage for reordered, shared-offset, and zero-sized fields.
  * Added allocation tests verifying recorded object references.

### DIFF
--- a/majit/majit-ir/src/descr.rs
+++ b/majit/majit-ir/src/descr.rs
@@ -690,6 +690,63 @@ static FIELD_INDEX_REDERIVED: std::sync::atomic::AtomicUsize =
 static FIELD_INDEX_UNRESOLVED: std::sync::atomic::AtomicUsize =
     std::sync::atomic::AtomicUsize::new(0);
 
+/// Counters behind [`GcCache::spec_position_census`] — the same question the
+/// four above ask, put to the producer instead of to the reader.
+static FIELD_SPEC_CHECKED: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+static FIELD_SPEC_MISPLACED: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+
+/// Counters behind [`GcCache::attached_position_census`].
+static FIELD_ATTACHED_CHECKED: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+static FIELD_ATTACHED_MISPLACED: std::sync::atomic::AtomicUsize =
+    std::sync::atomic::AtomicUsize::new(0);
+
+/// One field descr's two halves, compared against each other.
+///
+/// `expected` is the slot of the attached parent's `all_fielddescrs` that this
+/// field's own offset occupies; `actual` is the `index_in_parent`
+/// (`descr.py:228`) the producer stamped on it. Upstream cannot disagree —
+/// `heaptracker.py:60-72` and `:96-112` are one walker — so every disagreement
+/// is a descr declaring itself to be at a slot its parent fills with a
+/// different field.
+///
+/// Caller supplies `expected`, because only it holds the parent the producer
+/// actually attached. Once the descr reaches `get_field_descr` the parent is
+/// whatever `_cache_size` holds for the key, which is a different question.
+pub fn census_attached_index(expected: usize, actual: usize) {
+    use std::sync::atomic::Ordering::Relaxed;
+    FIELD_ATTACHED_CHECKED.fetch_add(1, Relaxed);
+    if expected != actual {
+        FIELD_ATTACHED_MISPLACED.fetch_add(1, Relaxed);
+    }
+}
+
+/// `field_specs[i].index_in_parent == i`, counted on the list as handed in.
+///
+/// This is [`GcCache::positional_invariant_census`]'s predicate moved from the
+/// published list to the submitted one, and the move is the whole point.
+/// `get_field_descr` re-derives every index against the parent before the list
+/// is frozen, so a census taken afterwards reads the reader's answer; the
+/// producer's own number survives nowhere else. Taken here it needs no parent,
+/// no name and no cache — the caller supplies both halves of the claim.
+///
+/// Both factories feed this, so it also covers the `cache_key == 0` fresh-mint
+/// groups that never enter `_cache_size` and are therefore outside
+/// `positional_invariant_census`'s domain entirely.
+fn census_spec_positions(field_specs: &[SimpleFieldDescrSpec]) {
+    use std::sync::atomic::Ordering::Relaxed;
+    FIELD_SPEC_CHECKED.fetch_add(field_specs.len(), Relaxed);
+    let misplaced = field_specs
+        .iter()
+        .enumerate()
+        .filter(|(i, spec)| spec.index_in_parent != *i)
+        .count();
+    if misplaced > 0 {
+        FIELD_SPEC_MISPLACED.fetch_add(misplaced, Relaxed);
+    }
+}
+
 /// descr.py:14-23 GcCache.
 ///
 /// Per-type descriptor caches keyed by LLType (structural equality).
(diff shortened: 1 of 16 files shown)


youknowone:
## Codex parity review — §1 and §2 confirmed and fixed

Both findings were about the same thing and both were **real**, so I measured
rather than argued: a temporary probe counting, per parent field list, how many
specs sit at an offset another field of the same parent also occupies.

```
[probe] spec_dup_offset=7      (of field_pos_spec_checked=1714)
```

Offset is not an identity. A flattened layout puts an inline aggregate and its
first leaf at one address (`heaptracker.py:68-69` recurses into a nested
`lltype.Struct` without minting a descr for the container), so
`position(|fd| fd.offset == offset)` can name a sibling — exactly as the review
says.

### What changed

`field_slot_in` (`jitcode/assembler.rs`) keys on the field **name**
(`heaptracker.py:60-72 get_fielddescr_index_in(STRUCT, fieldname)`) whenever the
descr carries one, and falls back to the byte offset only where there is no name
— and then only when exactly one field sits there. Where the offset is
ambiguous it resolves to nothing and the caller's number stands, the same
refusal `descr.rs find_index_in_parent` already documents.

Applied at all four sites the review named:

* `patch_field_descr_parents` (§1) and its debug postcondition (§1, §2) — the
  check is now stated over `field_slot_in`, so it can no longer panic on a descr
  that names its field correctly and merely shares an address with a sibling.
* `fielddescrof`'s fallback arm and its census (§2) — `unique_slot_at_offset`.
  Reaching that arm means the name lookup already missed, so offset is all there
  is; declining on ambiguity is the only sound option.
* `field_descr_ref_from_bh`'s census (§2) — name first, unambiguous offset
  otherwise. Counting an ambiguous one would have put a false reading behind
  `JITSTATS_BADNESS_FIELDS`.

The `pyre-jit-trace make_descr_from_bh` census already keyed on
`(offset, field_key)`, so it was unaffected.

### §3 — one of the two pre-existing items is now closed

> `add_struct_field_descr` has no field-name input and has always selected the
> first matching offset.

Correct, and it was feeding a **guessed name** into the resolution above, which
is the damaging half: that name becomes the descr's `_cache_field` key
downstream, so naming the aggregate hands the leaf's access the aggregate's
descr. It now takes the documented `(0, String::new())` fallback on an ambiguous
offset instead of the first match. The missing `fieldname` parameter itself is
still a real gap and is not something this PR closes.

The nested-struct flattening item (§3, second) is untouched — that is the same
`rewrite_op_getsubstruct` lowering the 152 `field_pos_unresolved` are waiting
on, and it is out of scope here.

### Tests

* `an_ambiguous_offset_is_left_alone_rather_than_arbitrated` — control:
  arbitrating fails it `(0, "agg")` vs `(0, "")`.
* `a_named_field_resolves_by_name_through_an_ambiguous_offset` — control:
  `Some(1)` vs `None`.

### Verification

check.py dynasm 379/379, cranelift 379/379, wasm 375/375;
`cargo test --workspace` 101 suites, no failures. Counters unchanged and healthy
(`field_pos_spec_misplaced=0`, `field_pos_attached_misplaced=0`,
`field_pos_unresolved=152`).

## The macos-latest `check.py` failure is not from this PR

Five `max-pypy-ratio` rows in the getattr/metatype family, three of which
check.py itself labels `[pypy exec clamped to floor; ratio not a measurement]`.
`main` at `295b3fcad` failed its macOS `check.py` with the **same fixture, the
same 5.9x, the same clamped 0.01s denominator**:

```
FAIL cranelift synth/metatype_property_dunder  exec 0.04s > pypy 0.01s  ratio 5.9x > gate 5x
```

The gated counters this PR adds read 0 on that CI run
(`field_pos_spec_misplaced=0 field_pos_attached_misplaced=0`).

— *commented by Claude*

youknowone:
## §1 and §2 are now `None`, and §3's first item is closed

The re-run reports no regressions and no new mismatches. §3's first item was
written against the previous push and is addressed by `e7e04dde53c`:

> Pyre's `add_struct_field_descr(offset, field_type, type_id)` has no
> `fieldname` … cannot select the field PyPy's name-keyed
> `heaptracker.get_fielddescr_index_in()` selects.

The name was already there and simply not passed. `jitcode_lower` hands
`register_struct_layout` a `(offset, is_ref, stringify!(#member))` triple and
then, **in the same `quote!` expansion**, called `setfield_gc_*` /
`getfield_gc_*` with the offset alone. Threading `stringify!(#member)` through
the five emitters makes read and write agree by construction and turns
`field_slot_in` into a real `get_fielddescr_index_in(STRUCT, fieldname)`. The
unique-offset fallback stays for a caller that has no name to give.

`an_ambiguous_offset_is_arbitrated_by_name_and_otherwise_declined` now pins both
halves: the named access resolves to slot 1 (`"leaf"`) through an offset two
fields share, the unnamed one still declines to `(0, "")`.

### The two remaining §3 items are deferred, with reasons

* **`all_fielddescrs` ordered by byte offset vs `STRUCT._names` traversal
  order** — a real ordering substitution, and the sharper of the two: it can
  assign a different `index_in_parent` than upstream wherever declaration order
  and offset order disagree, which no amount of name-keying fixes.
  `field_specs_from_layout` sorts by offset deliberately (a Rust struct literal
  has no stable source order across literals of the same type), so closing it
  needs the declared field order carried from the front end — the same channel
  the accessed-subset residual needs. Not something to bolt on inside this diff.
* **inline aggregate vs leaf at one address** — this is the
  `jtransform.py:942 rewrite_op_getsubstruct` lowering, the same one the 152
  `field_pos_unresolved` are waiting on. Until a raw substruct access lowers to
  `int_add(ptr, offset)` with no descr, an aggregate access has no exact
  `FieldDescr` to be.

### Verification

check.py dynasm 379/379, cranelift 379/379, wasm 375/375;
`cargo test --workspace` 101 suites, no failures. `field_pos_*` and
`positional_*` unchanged (`spec_misplaced=0`, `attached_misplaced=0`,
`unresolved=152`).

— *commented by Claude*

PR #1060, opened 2026-08-05
jit: emit the lazy virtualizable store, and publish walk-time locals at a traceback escape

youknowone:
Three fbw fixes, plus the regression fixture and the jit-stats re-records they imply.

**`OptHeap::emit_lazy_setfield` dropped a virtualizable store.** It returned
without emitting whenever the stored value was still virtual and the target was
the standard virtualizable frame, so the per-slot `SETARRAYITEM_GC` that
`gen_store_back_in_vable` (`rpython/jit/metainterp/pyjitpl.py:3489-3520`) writes
at a trace exit was deleted for every slot holding a virtual box. Upstream
`force_lazy_set` (`optimizeopt/heap.py:122-145`) emits unconditionally — the
virtual-rhs skip belongs to `force_lazy_sets_for_guard` (heap.py:608-637), which
routes the op into `rd_pendingfields` instead of dropping it.

**The walk-time traceback attach named a frame it had not published.** The
recording walk steps a snapshot, so a `STORE_FAST` it performs does not reach
the live frame; storing that frame into a `PyTraceback` escapes the
virtualizable, so both attach sites now write the locals region first. The
inlined level needed a `CalleeLocalsShadow`-backed twin of
`flush_locals_region_to_frame` — the neighbouring
`disarm_folded_inline_callee_after_escape` covers only the strict fresh-frame
fold and is inert on the multiframe path.

Symptom for both: a callee that binds a local on one branch and raises answered
`tb_frame.f_locals` with its parameters alone — 859 of 4000 iterations before,
0 after, identical on dynasm and cranelift, 0 under `PYRE_NO_JIT=1`.
`pyre/bench/synth/exception_vable_frame_virtual_local.py` guards it.

**The immortal-singleton enumeration test raced its siblings.** It compared two
consecutive enumerations for equality over process-global `OnceLock` slots while
the test binary runs cases concurrently; it failed that way on the ubuntu
runner.

`pyre/check.py`: dynasm 387/387, cranelift 387/387, wasm 383/383.
`cargo test --workspace --features dynasm`: 101 suites, 0 failed.
Re-records `closure_per_call.wasm` (guard_failures 471 -> 468),
`pickle_ctor_args.cranelift` (436 -> 201), and adds the missing
`exception_escape_hot_callee_tb_node_once.wasm` baseline.

— *commented by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved traceback accuracy by preserving virtualizable and inline-frame local variables during exception reporting.
  * Corrected garbage-collection handling for virtual references in frame value slots.
  * Unified lazy field and array store behavior for more reliable optimized execution.

* **Tests**
  * Added regression coverage for exception traceback locals, frame-root scanning, and callee-local publication.
  * Updated benchmark statistics and expanded exception-handling performance coverage.
  * Improved singleton-enumeration tests for concurrent execution.


youknowone:
## Codex parity review — disposition

**§1 Regressions:** none reported.

**§2 Other mismatches (1) — fixed in-session.**

`mod.rs ↔ virtualizable.py:101-138 write_boxes` — the callee flush skipped absent shadow entries.

The finding is real. Upstream's `virtualizable_boxes` is *total*, so `write_boxes` can write every locals-array element unconditionally; `CalleeLocalsShadow` is *sparse*, so an absent slot was ambiguous:

- absent from both `opref` and `concrete` — the walk never wrote the slot, so the frame already holds the value `write_boxes` would have written. Skipping is correct.
- present in `opref` but not in `concrete` — the walk *did* write it, with no concrete half (`set_concrete` drops a `Value::Void`). The frame's value is stale and unreconstructable.

Only the second case is a defect, and `opref` is the write-witness that separates them (every `set_concrete` call site is paired with a `set_opref` at the same slot). The validation pass now declines the whole flush on that case, restoring the all-or-nothing discipline of the top-level `flush_locals_region_to_frame`. It also declines on `Value::Ref(GcRef::NO_CONCRETE)`, which would otherwise store `usize::MAX - 1` into a GC-traced slot. Extracted as `callee_locals_region_is_publishable` with a unit test covering all four cases.

**§3 Pre-existing (2) — both deferred, blocker named.**

1. `CalleeLocalsShadow` HashMap side tables vs `MIFrame` register arrays — blocked: pyre lowers callee `LOAD_FAST`/`STORE_FAST` to `getarrayitem_vable`/`setarrayitem_vable(frame, slot)` while the `WalkContext` register banks are post-regalloc colour-indexed. Making frame-owned register state authoritative means allocating `MIFrame`-style per-slot registers for inlined callees in the codewriter, which cascades through the jitcode lowering, regalloc and every vable op. The §2 fix narrows the harm meanwhile: a sparse entry now declines instead of publishing a stale value.
2. Unseeded inline calls fabricate a traceback-only frame from code/globals — blocked: a branchless-leaf inline level never materialises a `PyFrame` at all (that is what the inline fold buys), so its shadow has `concrete_frame == 0` and there is nothing to flush into. Filling those locals requires porting `virtualizable.py force_now` for fresh inline frames.

**§4 Structural adaptations (2) — both won't-fix, documentation confirmed present.**

1. `eval.rs` explicit GC root walk + `JitVirtualRef` tag check — required by pyre's movable GC / raw-pointer representation; the in-code comment cites `virtualref.py:94-98 is_virtual_ref(gcref)`.
2. Boxing `Int`/`Float` and the explicit remembered-set barrier — the Rust-runtime form of RPython's typed `unwrap` plus GC write barrier; cited at the store.

— *commented by Claude*

youknowone:
## Codex parity review, round 2 (`b5becf1`) — disposition

§1 and §2 are now both **None**; the round-1 §2 finding is closed.

**§3 (1) — deferred, blocker named, reachability measured.**

`mod.rs ↔ virtualizable.py:101` — when `flush_callee_locals_region_to_frame` declines, the traceback is still attached to a frame whose locals were not published. The classification is right: this is the residual of the pre-existing sparse-shadow adaptation, and the patch narrows it rather than removing it.

Measured before deciding. A temporary decline counter built into `pyre-dynasm` and run over the whole synthetic corpus:

```
scanned=370 benches_with_declines=0 total_declines=0
```

plus 0 declines on `exception_vable_frame_virtual_local` itself (2000 / 2000 both buckets correct). The decline path is currently unreached; the instrumentation was reverted after the measurement and is not in the diff.

Eliminating it — rather than narrowing it — means making the shadow **total**, which is the same blocker as round 1's §3 item 1: callee `LOAD_FAST`/`STORE_FAST` lower to `getarrayitem_vable(frame, slot)` while the `WalkContext` register banks are post-regalloc colour-indexed, so a total `virtualizable_boxes` equivalent requires `MIFrame`-style per-slot registers for inlined callees in the codewriter, cascading through the jitcode lowering, regalloc and every vable op. Out of scope for this diff; deferred with that dependency named.

The two alternatives available without that port were both rejected as inventions rather than ports: falling back to the fabricated code/globals frame loses frame identity as well as the locals, and NULL-ing the unreproducible slots substitutes a different wrong answer for the stale one. The all-or-nothing decline at least matches the discipline the top-level twin `flush_locals_region_to_frame` already follows.

**§4 (3) — all won't-fix, documentation confirmed present** (sparse-shadow validation, `JitVirtualRef.forced` GC edge, and the relaxed singleton-enumeration test for concurrent access to process-global slots).

— *commented by Claude*

PR #1066, opened 2026-08-05
The host's own spelling of an argument: launcher argv, -W/-X option values, and os.stat's descriptor and dir_fd forms

youknowone:
Three argument boundaries where the host's own spelling was being lost.

## The launcher's argv

`std::env::args()` panics on an argument that is not valid Unicode, so
`pyre script.py 'bad<0xff>'` aborted the process with exit 101 before the
interpreter started. CPython decodes the argument with surrogateescape and
hands it to the program.

The launcher now carries arguments as `OsString` end to end — `drain_args`
collects `parser.raw_args()`, the four `argv[0]` sites build `OsString`, and
`sys.argv` / `sys.orig_argv` are decoded at the boundary by
`gateway::fsdecode_os_str`. On Windows that is `Wtf8Buf::from_wide` over
`encode_wide`, which is lossless; elsewhere it is the existing
`fsdecode_filename_bytes`. Only `--inspect`'s interactive path still needs a
`String`, and it narrows there with `lexopt::Error::NonUnicodeValue`.

`pyre-wasm-runner` had the same panic on its own `args()` call and is closed
the same way.

Measured against CPython 3.14: `pyre script.py 'bad<0xff>'` now reports
`['bad\udcff']` and exits 0. A non-UTF-8 `-m` argument is a separate case —
CPython accepts it and fails later with `No module named ba\udcffd` (exit 1),
where pyre still rejects the argument at exit 2; that half is not addressed
here.

## `-W`, `-X` and PYTHONWARNINGS

The same problem one layer up: `app_main.py:785-786` splits an `-X` value on the
first `=` and stores both halves in `sys._xoptions` verbatim, and `:892-906`
appends the `-W` values and the PYTHONWARNINGS pieces to `sys.warnoptions`
verbatim. None of them is an identifier, so none is required to have a UTF-8
spelling — `-W $'ignore\xff'` reaches `sys.warnoptions` as `'ignore\udcff'` on
3.14, where pyre exited 2 out of `lexopt`'s `.string()?`.

`LaunchFlags.warnoptions` / `xoptions` and the `importing` statics behind them
became `OsString`, decoded in `sys` by the same `fsdecode_os_str` the argv above
goes through. `-X` still matches the options pyre acts on, through `to_str()`:
every one of them is ASCII, so a value with no UTF-8 form cannot be one.
`_xoptions` splits at the first `=` over the encoded bytes — `OsStr` documents
that as sound at an ASCII byte — and puts *both* halves through WTF-8, because
`-X $'k\xff=v'` is `{'k\udcff': 'v'}`: the key carries an escape as readily as
the value.

PYTHONWARNINGS moved from `read` to `read_raw`. `read` is
`String::from_utf8(..).ok()`, so one undecodable byte was discarding the whole
variable rather than the single comma-separated entry that carried it — a
silent drop this fixes on the way past.

The **script path** is a fourth boundary and is *not* fixed here: it still
narrows to `String`. Its `__file__` / `co_filename` cannot round-trip until
those are WTF-8, and it cannot be tested on macOS at all — APFS refuses to
create a non-UTF-8 filename (`OSError: [Errno 92]`). Tracked separately.

## `os.stat`'s descriptor and `dir_fd` forms

`os.stat` advertised itself in `os.supports_fd` while raising `TypeError` on a
descriptor, and both `stat` and `lstat` rejected `dir_fd` with
`NotImplementedError` — `lstat` while being listed in `os.supports_dir_fd`
through `HAVE_LSTAT` (`os.py:121`).

`FsEncodedPath` now carries `Path.as_fd` (`interp_posix.py:140-152`), set only
by the entry points that pass `allow_fd`, and `fsencode_path_or_fd_w` names the
caller in the type error so the allowed-type list can widen with it: `stat`
answers `"string, bytes, os.PathLike or integer"` where `lstat` answers
`"string, bytes or os.PathLike"`. The descriptor arm is probed with `__index__`
and sits before the PathLike arm, so an object carrying both is read as a
descriptor.

`stat_entry` is now the three arms of `do_stat` (`interp_posix.py:633-649`): a
descriptor goes to `fstat`, a `dir_fd`-relative name to `fstatat`, and a bare
name to `stat`/`lstat`. The `stat_result` assembly moves into
`stat_result_from_fields` so `std::fs::Metadata` and the raw `libc::stat` that
`fstatat` fills both reach it. `_have_functions` gains `HAVE_FSTATAT` under the
cfg where that is implemented, which is what makes the `supports_dir_fd` entry
for both calls true.

One message follows CPython rather than PyPy: `os.stat(fd, dir_fd=...)` reports
`"stat: can't specify dir_fd without matching path"` (3.14) where
`interp_posix.py:639` says `"can't specify both dir_fd and fd"`. The parity
suite's oracle is CPython.

## Tests

`pyre/extra_tests/parity_tests/argv_undecodable_argument.py` passes an
undecodable argument to a child through `subprocess`, so it needs no such file
on disk, and self-skips on win32.

`pyre/extra_tests/parity_tests/os_stat_file_descriptor.py` covers the
descriptor form, both `ValueError`s, the two exact `TypeError` strings, and —
guarded on `os.stat in os.supports_dir_fd` — the `fstatat` arm including
`AT_SYMLINK_NOFOLLOW` agreeing with `lstat`, an absolute name ignoring the
descriptor, `ENOENT` reporting the name, and `ENOTDIR` for a non-directory
descriptor. It also pins the argument unwrap *order* by observation, not only
by message: `path`, then `dir_fd`, then `follow_symlinks`, each of which can run
user code (`__fspath__`, `__index__`, `__bool__`).

`pyre/extra_tests/parity_tests/option_value_undecodable.py` covers `-W`, the
four `-X` shapes (value half, key half, bare key, and a value carrying further
`=`), and the comma-separated PYTHONWARNINGS case.

All three scripts pass under CPython 3.14 first, and on both backends.

## Known-red checks

`pyre/check.py` is red on this PR, and every failure is inherited from `main`,
which is itself red on `pyre/check.py (ubuntu-24.04)` and `(windows-latest)` at
its current tip. There are **no correctness failures** on any of the three
platforms:

* ubuntu — synth perf *ratio* gates, most annotated by check.py itself as
  `[pypy exec clamped to floor; ratio not a measurement]`. The failing set
  differs from main's in *both* directions (6 here main lacks, 4 there this
  lacks), which is the signature of noise rather than a regression; several are
  *minimum*-ratio failures where pyre is faster than the floor.
* windows — one bench, `pickle_ctor_args`, reporting a jit-stats *improvement*
  (`guard_failures 436 -> 201`). Byte-identical to main's failure on the same
  bench, same tally.
* the missing `exception_escape_hot_callee_tb_node_once.wasm.jitstats` baseline,
  which is deterministic and already recorded on another branch.

Tracked separately; none of it is reachable from this diff.

### DIFF
--- a/pyre/extra_tests/parity_tests/argv_undecodable_argument.py
+++ b/pyre/extra_tests/parity_tests/argv_undecodable_argument.py
@@ -1,0 +1,67 @@
+"""A command-line argument with no UTF-8 spelling reaches `sys.argv` as itself.
+
+`targetpypystandalone.py:76-80` builds `sys.argv` with `space.newfilename`,
+which is `fsdecode(newbytes(s))`, so an argument carrying a byte the filesystem
+encoding cannot spell arrives as the surrogate escape that re-encodes to that
+byte — not rejected, not replaced. `sys.orig_argv` carries the same value.
+
+The argument is passed to a child, so the test needs no such name on disk: the
+filesystem never sees it, only `execve` does. Windows has no byte argv at all
+and takes the wide command line, so this shape does not exist there.
+"""
+
+import os
+import subprocess
+import sys
+
+if sys.platform == "win32":
+    print("OK")
+    raise SystemExit
+
+UNDECODABLE = b"pyre_undecodable_\xff"
+ESCAPED = os.fsdecode(UNDECODABLE)
+
+# The escape is what the filesystem decode produces, and it round-trips.
+assert ESCAPED.endswith("\udcff"), ascii(ESCAPED)
+assert os.fsencode(ESCAPED) == UNDECODABLE, ascii(ESCAPED)
+
+CHILD = r"""
+import os, sys
+assert sys.argv[1:] == [os.fsdecode(%r), "plain"], ascii(sys.argv)
+assert os.fsencode(sys.argv[1]) == %r, ascii(sys.argv[1])
+# `orig_argv` is the launcher's own line, so the argument appears there too,
+# with the same escaping.
+assert sys.argv[1] in sys.orig_argv, ascii(sys.orig_argv)
+assert sys.orig_argv[-2:] == sys.argv[1:], ascii(sys.orig_argv)
+print("child ok")
+""" % (UNDECODABLE, UNDECODABLE)
+
+result = subprocess.run(
+    [sys.executable, "-c", CHILD, ESCAPED, "plain"],
+    stdout=subprocess.PIPE,
+    stderr=subprocess.PIPE,
+)
+assert result.returncode == 0, (result.returncode, result.stderr)
+assert result.stdout == b"child ok\n", result.stdout
+
+# A script run the same way answers with the argument in argv[1], and the
+# script's own path stays argv[0].
+import tempfile
+
+with tempfile.TemporaryDirectory() as tmp:
+    script = os.path.join(tmp, "show_argv.py")
+    with open(script, "w") as f:
+        f.write(
+            "import os, sys\n"
+            "print(os.fsencode(sys.argv[0]) == os.fsencode(%r))\n" % script
+            + "print(ascii(sys.argv[1]))\n"
+        )
+    result = subprocess.run(
+        [sys.executable, script, ESCAPED],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
+    assert result.returncode == 0, (result.returncode, result.stderr)
+    assert result.stdout == b"True\n" + ascii(ESCAPED).encode() + b"\n", result.stdout
+
+print("OK")
--- a/pyre/extra_tests/parity_tests/option_value_undecodable.py
+++ b/pyre/extra_tests/parity_tests/option_value_undecodable.py
@@ -1,0 +1,69 @@
+"""`-W`, `-X` and PYTHONWARNINGS carry a value with no UTF-8 spelling.
+
+These are free text, not identifiers: `app_main.py:785-786` splits an `-X`
+value on the first `=` and puts both halves into `sys._xoptions` verbatim, and
+`:892-906` appends the `-W` values and the PYTHONWARNINGS pieces to
+`sys.warnoptions` verbatim. None of them is required to be spellable in UTF-8,
+so a byte the filesystem encoding cannot spell arrives as the surrogate escape
+that re-encodes to that byte — in the `_xoptions` key as much as in its value.
+
+An option value never reaches the filesystem, so like
+`argv_undecodable_argument.py` this needs no such name on disk and passes the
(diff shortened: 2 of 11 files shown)


youknowone:
Codex parity review dispositions, and one finding refuted.

**§2 — descriptor-plus-`dir_fd` wording · won't-fix (documented).** Measured: 3.14 says `stat: can't specify dir_fd without matching path`, `interp_posix.py:639` says `can't specify both dir_fd and fd`. The parity suite's oracle is CPython, so the 3.14 wording stays — but the divergence is now cited in a comment at the raise site rather than only in the commit message.

**§2 — argument unwrap order · fixed (0ed154d).** Real defect, thank you. Measured on 3.14: `os.stat(1.5, dir_fd=1.5)` answers with the *path* type error, and `os.stat(PathLike, dir_fd=1.5)` calls `__fspath__` before rejecting `dir_fd`, with `follow_symlinks.__bool__` never read in either case. Signature order it is. The parity test now observes the call order (`order == ["path"]`), not just the message, so the arrangement cannot regress silently.

**§2 — `fstatat` field spellings "will not compile for macOS libc's `stat` layout" · refuted.** libc 0.2.186 exposes `st_atime_nsec` / `st_mtime_nsec` / `st_ctime_nsec` on Apple targets too (`src/unix/bsd/apple/mod.rs:321,323,325`); the crate normalises the `st_atimespec` spelling RPython has to branch on. The branch was built `--release` for both backends and ran the full parity suite and `check.py` on Darwin arm64 before this comment, so the claim is refuted by construction as well as by the source.

**§3 — script path still narrowed to `String` (`lib.rs:181`) · deferred, tracked.** Correct and still real. It is out of this PR because it reshapes `RunMode::Script(String)` and `RunMode::Stdin { argv0 }` and needs its own oracle measurement and parity test; it is now tracked together with the `-W` / `-X` value boundaries, which have the same cause and the same fix (`gateway::fsdecode_os_str`, added here).

**§4** — both adaptations confirmed as intended; each already cites its upstream decision point in-code.

---

Unrelated to this diff, found while verifying it: `synth/pickle_ctor_args` fails `check.py` on macOS. `d1fef848351` (#1033) tightened its header from `max-pypy-ratio=336` to `36`; this Darwin arm64 host measures 37.6-43.5x (dynasm) and 47.1-50.4x (cranelift) across load averages 16.6 to 7.6, so load is not the variable. It is not a regression either — the same host measured 47.9x / 58.9x / 80.7x on the pre-#1033 base and passed, because the ceiling was 336 then. CI is green, so the ceiling holds on the Linux runners; it appears to have been sized from Linux numbers, and macOS gates harder. Tracked separately.

— *commented by Claude*

PR #1069, opened 2026-08-05
mapdict: wire the four `?` fields as real quasi-immutables

youknowone:
`pypy/objspace/std/mapdict.py` declares four quasi-immutable fields —
`_immutable_fields_` with a `?` suffix on `allow_unboxing` (`:305`),
`ever_mutated` (`:421`), and `attr` / `typ` (`:669`). Pyre carried all four as
plain `Cell`s with doc comments saying no watcher was installed.

## The measured result

`pyre/bench/synth/mapdict_unboxed_type_change_attr` — the fixture that writes
`obj.x = 1.5` to an int-unboxed slot inside a hot loop:

| | before | after |
|---|---|---|
| dynasm / cranelift `guard_failures` | 404 | **4** |
| dynasm / cranelift `bridges_compiled` | 2 | **0** |
| dynasm / cranelift `loops_compiled` | 2 | 4 |
| wall clock (dynasm) | 0.14s (10.3x pypy) | **0.07s (~7.7x)** |
| wasm | 404 / 2 / 2 | 204 / 1 / 3 |

The `allow_unboxing` freeze used to leave the compiled loop running on a stale
unboxed shape, so the promoted-map `GuardValue` failed 200 times per loop until
a bridge compiled. It now revokes the loops out of line. `loops_compiled 2 -> 4`
is exactly one recompile per invalidated loop — that number not growing is the
setters' value-change early return holding.

## What is here

1. **`mapdict.rs`** — a `QuasiImmutField` per field (**two** on
   `CachedAttributeHolder`; `rclass.py:534-536` synthesises one `mutate_<name>`
   per declared name and `mapdict.py:669` declares two), one notify-then-store
   setter each, the accessor surface, and every write site routed through a
   setter. Setter order is `rclass.py:1010-1012` — `hook_setfield` then
   `genop('setfield')` — which `celldict.rs::mutated` now also follows.
2. **`descr.rs`** — four singleton quasi-immutable `FieldDescr`s under a reserved
   `0x5000_0000` tag. Not `make_quasi_immutable_field_descr`: its index comes
   from `stable_field_index(offset, field_size, field_type, signed)`, and
   `allow_unboxing` and `ever_mutated` are both `Cell<bool>`, so they collide
   whenever their `repr(Rust)` offsets do. The index picks the pointer cast in
   `install_quasiimmut_field`, so that would be a type confusion.
3. **`state.rs` / `eval.rs`** — `install_quasiimmut_field` and
   `register_quasi_immutable_deps` were two-way if/elses whose `else` cast the
   pointer to `W_TypeObject` unconditionally. A headerless `*const Terminator`
   down that arm reads offset 0 of a leaked `Box`: in bounds, so no fault, and
   every check reads `false` forever with no diagnostic. Both are now exhaustive
   over the six quasi-immutable descrs with a fail-loud default.
4. **`mod.rs` / `specialize.rs`** — `walker_pin_*` markers at the folds that read
   the fields, and the `ever_mutated` mark moved out of the two store resolvers
   (they marked while merely *asking* whether a store was foldable, and the
   caller can still decline on the `is_bool` / type rejections or abort on the
   shape guard).
5. **`residual_call.rs`** — the tracer-side force check, gated **opt-in**,
   forcing only where the abort's flush leg can adopt the walk's operand-stack
   mirror (`9ed58953e88`; see below).

## Two deviations from the original plan, both source-driven

**No post-write backstop.** The design called for a force counter bumped by the
setters and snapshotted around every residual, aborting *after* the write ran.
That is unnecessary and unfaithful: `heap.py:798-804` already raises
`InvalidLoop('quasi immutable field changed during tracing')`, and
`quasiimmut_field_still_valid` is already ported in
`majit-metainterp/src/optimizeopt/heap.rs`. It covers every route the
predicates do not model — `setattr()`, `obj.__dict__[k] = v`, the
`PlainAttribute.delete` -> `_copy_attr` -> `add_attr` -> `pick_attr` re-add
chain. Upstream never aborts after a write, and neither does this.

**The `ever_mutated` marker is at the write site, not in a restored
`_pure_direct_read`.** Pyre's `node_read` is ordinary Rust executed concretely
inside a `#[dont_look_inside]` residual — there is no `jit.isconstant` there, so
an upstream-shaped three-term gate would make the pure arm uncallable and the
marker dead.

## The tracer force check DOES fire — and reaching it was wrong code

**This section supersedes the earlier "could not be demonstrated firing".** Four
hand-built probes said the check was unreachable and I wrote a structural
argument for why. The corpus falsified both: with `PYRE_QMUT_MAPDICT_FORCE=1`,
`pickle_terminal_raise_resume` raised

```
struct.error: unpack requires a buffer of 8 bytes
```

from `pickle.load_frame`'s `self.read(8)`. The census names it:

```
end=ForceQuasiImmutable committed=false leg=0 unj_val=false unj_sym=false effects=4
```

`unj_val` and `unj_sym` are both false, so the flush declined on the second term
of `fbw_has_unjournaled_effect() || session.abort_in_subwalk`: the forcing
`self.x = v` sits inside an inlined callee, whose resume coordinate names the
callee's code object, so the flush leg cannot adopt it. The abort fell to the
legacy replay-from-entry, which re-ran the four residuals the walk had already
executed — two of them Python-frame entries — desynchronising the unpickler's
read position.

`9ed58953e88` fixes it by hoisting the latch's own precondition to the top of
the force helper as an early return: force only where the abort can hand the
flush leg a mirror it can adopt.

```rust
if !ctx.vstack_valid || ctx.fbw_mode.inline_subwalk || ctx.trace_ctx.is_bridge_trace {
    return None;
}
```

Declining is sound rather than a hole — without the abort the trace keeps
recording, and `quasiimmut_field_still_valid` discards any loop whose recorded
`?` value moved.

## Why the gate stays opt-in — measured, not cautious

After the fix, with `PYRE_QMUT_MAPDICT_FORCE=1` across all 386 `bench/synth`
benches:

- the check fires **3 times, all committing**, with no `committed=false
  effects>0` walk anywhere;
- `pickle_terminal_raise_resume` returns `checksum = 216` again;
- the corpus is output-correct and exactly one counter moves, that bench's
  `loops_aborted 1 -> 3`;
- **0 firings in the other 385 benches.**

So the benefit is negligible, while turning it on would make it the first common
producer of `DispatchError::ForceQuasiImmutable` — today's only producer needs a
module-namespace watcher — and three flush-decline paths still end in the legacy
replay, where that error's leg in `trace.rs` drops an in-flight FOR_ITER item.

## Verification

Three LLBC re-extractions (once per change to a `pyre-interpreter` struct) and a
rebuild of both release binaries each time.

- **dynasm 386/386**, **cranelift 386/386**.
- wasm 381/382; the one failure is `exception_escape_hot_callee_tb_node_once`,
  whose `.wasm.jitstats` #1033 never recorded. Pre-existing on `main`.

Two baselines re-recorded:
- `mapdict_unboxed_type_change_attr` on all three backends — the win above.
- `pickle_ctor_args` cranelift `guard_failures 436 -> 201`. **`6a173f90e91`'s
  commit message attributes this to the watcher fields widening three structs
  and moving `intern_node`'s addresses. That attribution is wrong.** The same
  436 -> 201 was measured independently at this same base (`da5e6fb38c7`) with
  that other work's change reverted, so 201 is what the base produces and the
  committed 436 is stale — the recurring "cranelift jitstats red is BASE, not
  branch" shape. The re-record is correct either way; only the explanation was.
  I am leaving the commit text as-is rather than force-pushing a published
  branch.

— *authored by Claude*

## Summary by CodeRabbit

- **Performance**
  - Improved JIT optimization of map-based attribute access, updates, and transitions.
  - Reduced guard failures and increased compiled-loop coverage in affected workloads.

- **Reliability**
  - Strengthened invalidation tracking when attribute metadata or unboxing behavior changes.
  - Improved consistency for module dictionaries and cached attribute information.

- **Compatibility**
  - Enhanced support for boxed and unboxed attribute reads, writes, transitions, and deletions.
  - Improved handling of attribute changes that require compiled code to be refreshed.


youknowone:
Two corrections and one new fix, after measuring the opt-in check across the corpus.

### Correction: the `pickle_ctor_args` attribution in `6a173f90e91` is wrong

That commit says the cranelift `guard_failures 436 -> 201` follows from the
watcher fields widening three structs and moving `intern_node`'s addresses.
It does not. The same 436 -> 201 was observed independently at this same base
(`da5e6fb38c7`) by other work, **with that work's change reverted** — the
recurring "cranelift jitstats red is BASE, not branch" shape. The committed
`436` is stale relative to what the code at that sha produces.

The re-recorded value is right either way; only the explanation was wrong. I am
leaving the commit as-is rather than force-pushing over a published branch.

### The tracer force check IS reachable, and reaching it was wrong code

`could not be demonstrated firing` in the PR description is superseded. My four
probe fixtures missed it, but the corpus did not: with
`PYRE_QMUT_MAPDICT_FORCE=1`, `pickle_terminal_raise_resume` raised

```
struct.error: unpack requires a buffer of 8 bytes
```

from `pickle.load_frame`'s `self.read(8)`. The census names it exactly:

```
end=ForceQuasiImmutable committed=false leg=0 exec_mf=7 exec_pl=2 effects=4
```

`unj_val` and `unj_sym` are both false, so the flush declined on the second term
of `fbw_has_unjournaled_effect() || session.abort_in_subwalk`: the forcing
`self.x = v` is inside an inlined callee, whose resume coordinate names the
callee's code object. The abort fell to the legacy replay-from-entry, which
re-ran the four residuals the walk had already executed — two of them
Python-frame entries — desynchronising the unpickler's read position.

`9ed58953e88` fixes it: force only where the abort can hand the flush leg a
mirror it can adopt, which is the predicate the latch at the end of the function
already applied — `vstack_valid && !inline_subwalk && !is_bridge_trace`, hoisted
to the top, the last point where declining is still free. Declining is sound
rather than a hole: without the abort the trace keeps recording and
`is_still_valid_for` discards any loop whose recorded `?` value moved.

### The measurement that answers "should the gate be flipped on"

After the fix, with `PYRE_QMUT_MAPDICT_FORCE=1`:

- the check fires **3 times, all committing**, no `committed=false effects>0`
  walk anywhere;
- `pickle_terminal_raise_resume` returns `checksum = 216` again;
- corpus-wide the gate-on run is output-correct and moves exactly one counter,
  that bench's `loops_aborted 1 -> 3`;
- **0 firings in the other 385 benches.**

So the gate stays opt-in on evidence rather than on caution: the benefit is
negligible, and the three remaining flush-decline paths still end in the legacy
replay.

— *commented by Claude*

PR #1071, opened 2026-08-06
bench, parity: widen six floored-denominator ratio gates; scope the platform-specific parity fixtures

youknowone:
Three follow-ups to #1060, all found by reading its CI run.

## `bench`: six `max-pypy-ratio` ceilings derived from a floored denominator

Ten bench-gate failures on the ubuntu runner, on six fixtures. Seven of the ten carry check.py's own `[pypy exec clamped to floor; ratio not a measurement]` marker: pypy's startup-subtracted exec time is pinned at `EXEC_TIME_FLOOR_S`, so the ratio divides by the floor rather than by a measurement. Raising `N` does not lift it — the bodies are loop-invariant and pypy hoists them whole.

None of the six was touched by this branch; all were last modified by `d1fef848351`, which tightened them (e.g. `pickle_ctor_args` 336 → 36).

The clamped fixtures get their pre-tightening ceilings back and lose their `min-pypy-ratio` floors, which came from the same non-measurement. Two exceptions:

- `type_immutable_reject` — the restored 15 is below the 18.2x observed on the runner, so it gets twice that instead.
- `pickle_ctor_args` — a real measurement (pypy 0.04s, never clamped), but spanning 18.6x to 70.3x across hosts. Ceiling 121 (twice the slowest) with an explicit floor of 9, because the default ceiling/5 = 24.2 rejects the fastest host at 13.5x.

## `parity`: platform-scoped fixtures, and one real Windows divergence

The windows job's benches now pass 387/387, which is what let it reach `parity_tests` for the first time — main's own windows run died at the bench step. Six fixtures fail there.

Five fail under the reference CPython too, so the comparison measures nothing: a `b"\xff.py"` filename only decodes under a UTF-8 filesystem encoding with `surrogateescape`; `os.fsencode` only rejects a surrogate outside U+DC80..U+DCFF under that same handler; `os.DirEntry`'s qualified names are the `posix` module's; and `type_members_python314`'s absolute C-struct sizes are one ABI's — the reference itself reports `OSError.__basicsize__` 120 on Windows and `BaseExceptionGroup.__basicsize__` 88 on macOS arm64.

`run.py` gains a `# pyre-check: platforms=` header naming the `sys.platform` values a script's subject holds on; an unmarked script still runs everywhere.

The sixth is a real divergence. `W_SyntaxError.descr_str` calls `os.path.basename`, which is the platform's — `ntpath` splits on `\` as well as `/` and drops the drive `splitdrive` peeled off first. Splitting on `/` alone left the full temporary path in the rendered message.

## `bench`: four wasm jitstats baselines

`guard_failures` moves by one or two against the values recorded at `58fcd373e05` and `da5e6fb38c7`, with `loops_compiled` and `bridges_compiled` unchanged on all four. `closure_per_call` returns to the 468 that this branch and its own CI run both observed before the rebase took main's 470.

— *opened by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `SyntaxError` filename display across Windows and Unix-like platforms, including drive-letter paths and mixed separators.
  * Corrected platform-specific handling of filenames, surrogate characters, and exception metadata in parity coverage.

* **Tests**
  * Added platform-aware filtering and reporting for parity tests.
  * Updated benchmark thresholds and runtime guidance to reflect stable, measurable results.
  * Refreshed JIT statistics with additional field-position counters and current guard-failure values.


youknowone:
## Codex parity review — disposition

**§1 Regressions: None. §2 Other mismatches: None.**

**§3 Pre-existing (1) — fixed now** (`b902570a38e`).

`display.rs ↔ interp_exceptions.py:875` — Windows UNC roots. The finding is right: `ntpath.splitdrive` treats the whole `\\server\share` as the drive, so `os.path.basename` of a bare UNC root is `""`, and a separator-only scan answers `share`. It sits in the code this patch just touched, so it is a fix-now rather than a defer.

Split the drive off first, as `ntpath.split` does, then take what follows the last separator in the remainder. `nt_drive_len` covers both prefix shapes — a two-character `X:` drive and a UNC root that runs to the second separator after the leading pair, or to the end of a path with fewer components. Cross-checked against `ntpath.splitdrive` / `ntpath.basename` on 11 inputs, all in the unit test:

```
'C:\\dir\\enc.py'            2  'enc.py'
'C:enc.py'                   2  'enc.py'
'\\\\server\\share'         14  ''
'\\\\server\\share\\enc.py' 14  'enc.py'
'\\\\server'                 8  ''
'\\\\server\\'               9  ''
'\\\\'                       2  ''
'//server/share/enc.py'     14  'enc.py'
'\\dir\\enc.py'              0  'enc.py'
'enc.py'                     0  'enc.py'
''                           0  ''
```

**§4 Structural adaptations (2) — both won't-fix, documentation confirmed present.** Doing the platform-dependent basename in Rust rather than through an app-level `import os` is the translation-boundary adaptation, and the helper's doc cites `interp_exceptions.py:875` for it. Preserving an exact empty `SyntaxError.filename` instead of `self.filename or "???"` is the 3.14-vs-3.11 difference already documented at the call site.

## Verification at this head

```
ALL PASSED: dynasm 388/388
ALL PASSED: cranelift 388/388
ALL PASSED: wasm 384/384
ALL PASSED: 3/3 backend run(s)
```

`parity_tests/run.py` on macOS: 187 scripts, `all parity tests pass`, with `type_members_python314.py` skipped as `platforms=linux`. `cargo +1.97.1 fmt --all -- --check` clean.

— *commented by Claude*

PR #1104, opened 2026-08-07
parity: say why a runner failed, then fix the five Windows failures it was hiding

youknowone:
A failed `pyre/extra_tests/parity_tests` job named none of what failed in it.
[This one](https://github.com/youknowone/pyre/actions/runs/31159459145/job/92811514550)
ends:

```
  winreg_key_lifecycle.py              cpython=OK dynasm=OK cranelift=OK
  zip_identity_python314.py            cpython=OK dynasm=OK cranelift=OK

##[error]Process completed with exit code 1.
```

So the runner now says why, and then everything it was failing to say — five
scripts, red on every Windows runner — is fixed.

## 1. The runner reports where a CI log is read

The reason was in the log, 200 lines *above* the run's own header. The
per-failure detail and the `N failure(s)` count went to stderr while the
213-row table went to stdout, and a piped stdout is block-buffered where
stderr is not, so the report overtook everything it was meant to explain and
the last line before the non-zero exit was a passing row.

* **The report is printed last, on stdout**, and echoes each child's stderr
  verbatim rather than a `repr` of the whole thing on one line — a traceback
  only reads as one when its line breaks survive.
* **Each failing row carries its one-line verdict beneath it.**
* **`::error file=` annotations** under `GITHUB_ACTIONS`, one per failure,
  carrying the verdict and the exception line. They show on the pull request
  without opening the log.
* **stdout is line-buffered**, so a run that dies mid-way names the scripts it
  got through.
* **stdout is pinned to UTF-8.** This one bit during verification: the report
  echoes a child's stderr, and these scripts are largely about names no console
  codepage can spell, so printing one raised `UnicodeEncodeError` *out of the
  runner* instead of the failure it was explaining — the same no-reason death,
  one level up.

`_run` returns the reason and the stderr as separate values, which the new
`Failure` tuple carries; nothing about what counts as a pass moves.

It works: the run this branch was rebased over reported its own failure as
`surrogate_name_messages.py [cpython]: exited 1 … AssertionError: '<function
q\udcffn at 0x000002A594587E20>'`, traceback and all, which is §5 below.

## 2. The failure it was hiding

Two independent things, both Windows:

**CPython has no `os.pathconf` there.** The reference failed the script at
line 96 before any backend could, and comparing a backend against a failing
reference measures nothing. That section is now gated on the name existing —
`hasattr`, not a platform list, because what it needs is the name.

**`os.utime` refused every time before 1970.** The host call takes its times
as a `Duration`, which has no second below its epoch at all, so
`u64::try_from(sec)` was the whole pre-epoch range's refusal.

#1107 landed a Windows `SetFileTime` path over the same lines while this
branch was out, and the rebase takes its structure — `CreateFileW` with
`FILE_WRITE_ATTRIBUTES` and `FILE_FLAG_BACKUP_SEMANTICS`, which is what
`rposix.py:1558-1575` opens with and what lets the name open a directory or a
read-only file. What is kept from this side is the arithmetic: `r_longlong`
in `rwin32file.py:320-323`, and the `__int64` of the C it transcribes, both
*wrap*. A checked conversion answers a `ValueError` for a tick count neither
runtime produces, and by then anything no `time_t` can hold has already been
refused with the `OverflowError` below.

## 3. What measuring it turned up

Every row below was measured against CPython 3.14 on the same host, before
and after:

| | was | now (== CPython 3.14) |
|---|---|---|
| `utime(p, times=(-5.0, -6.0))` | `ValueError: utime: timestamp out of range` | writes it; `st_mtime_ns == -6_000_000_000` |
| `utime(p, ns=(-1, -1))` | same refusal | `-100` — a FILETIME's granularity |
| `utime(p, ('a', 'b'))` | `ValueError: could not convert string to float` | `TypeError: argument must be int or float, not str` |
| `utime(p, (1e30, 0))` | `ValueError: utime: timestamp out of range` | `OverflowError: timestamp out of range for platform time_t` |
| `utime(p, (2**200, 0))` | the same, via a float that rounded it | the same OverflowError, from the exact integer |
| `utime(p, (nan, 0))` | `ValueError: utime: timestamp out of range` | `ValueError: Invalid value NaN (not a number)` |
| `utime(p, (1,))` | `utime: 'times' must be a tuple of two ints` | `... must be either a tuple of two ints or None` |
| `utime(p, ns=(2**80, 0))` | `OverflowError` | written |
| `truncate(p, 2**64)` | `int too large to convert to int` | `int too big to convert` |
| `st_mtime_ns` of a file dated 30000 | `-5443715538058477568` | `880000000000000000000` |

Three of those are worth their own sentence:

* **`ns=2**80`.** `split_py_long_to_s_and_ns` splits with `divmod` *before* it
  narrows anything, so a nanosecond count too wide for a `time_t` is refused
  only when the second it names is — and 2\*\*80 ns is a second that fits.
  Dividing after the narrowing turned away the whole range. Going through
  `divmod` is also what answers for `ns=('a', 'b')`.
* **`truncate`.** The Windows arms read their length with a bare `int_w`: no
  `__index__`, and the wrong message. Both now go through `truncate_length_w`,
  hoisted out of the unix arm, which names the C type the platform's own
  `Py_off_t_converter` names.
* **`st_*_ns`.** The `sec * 1_000_000_000` product was taken in `i64`, which
  runs out in **2262** — not an exotic date for a filesystem whose FILETIME
  reaches the year 30828. It is taken in `i128` now and the field is an int of
  whatever width it needs.

`parity_tests/os_utime_pathconf_truncate` pins all of it and no longer skips
its negative-time section on Windows; only the exact-nanosecond value is
platform-dependent there.

## 4. The three failures it was hiding everywhere else

The same job on [#1109](https://github.com/youknowone/pyre/actions/runs/31233554489/job/93043853949)
ends the same way — a passing row, then exit code 1 — over three more scripts,
all Windows, all main's and independent of each other:

* **`keyboard_interrupt_exit_status`.** A Win32 process has no SIGINT to die
  of. `app_main.py` restores `SIG_DFL` and calls `raise(SIGINT)` saying the MSVC
  runtime then exits with `STATUS_CONTROL_C_EXIT`; measured, that pair *returns*
  and the process ends with status 3 — under CPython too. `raise` never
  terminating meant `terminate_by_sigint` fell through to `process::abort`,
  whose status 3 was what the fixture read. Windows exits with
  `STATUS_CONTROL_C_EXIT` directly now.
* **`builtin_module_loader_spec`.** `_sysconfig.config_vars()` answered an empty
  dict, but `sysconfig._init_non_posix` *subscripts* `Py_GIL_DISABLED` and
  `Py_DEBUG`, so under `os.name == 'nt'` an absent key is a `KeyError` and not
  the `None` the `.get()` readers take. Behind it, `sys._vpath` and
  `sys.getwindowsversion` were both absent — the latter now a five-field
  sequence with five named-only fields over it, off
  `host_env::windows::get_windows_version`, matching CPython 3.14 field for
  field bar `build` (see the commit for why kernel32's file version is the
  honest one for an unmanifested binary).
* **`frame_clear_finalization`** imported `resource` for one CPU-time bound at
  the end; it is POSIX-only, absent from CPython on Windows too, so the import
  took all six checks off the platform and left the reference failing beside the
  backends. `time.process_time` is the same measurement and is everywhere.

## 5. Every repr's address was wrong on Windows

`surrogate_name_messages` was red on Windows with **`cpython=FAIL`** — the
reference failing it, so the backends were being compared against a failing
oracle. It asserts `"<function %s at 0x%x>" % (S, id(f))`, and that is the glibc
spelling.

`PyUnicode_FromFormat`'s `%p` hands the pointer to the platform's own `printf`
and normalizes only the prefix — guaranteed to start with a literal `0x`
"regardless of what the platform's printf yields". Everything after it is the
platform's: the MSVC runtime pads to the pointer width and uppercases, glibc
does neither. Rust's `{:p}`, `{:?}` on a raw pointer, and a hand-written
`0x{:x}` are all only ever the second spelling, so **every** address-bearing
repr was wrong there. Measured against CPython 3.14 on the same host, all of
them disagreed:

| | was | now (== CPython 3.14) |
|---|---|---|
| `repr(f)` | `<function f at 0x1b7af7ffcc0>` | `<function f at 0x000001B7AF7FFCC0>` |

…and the same for object, generator, coroutine, async_generator, bound and
built-in method, method-wrapper, cell, weakref, memoryview, ContextVar, Token,
code, frame and both `_thread` locks — fifteen kinds, now through one
`display::repr_addr`, verified shape-identical. `_pickle`'s cyclic-object
message names an address the same way and is the one such site that is not a
repr.

The frame repr's `file '...'` is left raw: `pypy/interpreter/pyframe.py:851`
interpolates `'%s'` where CPython's `%R` escapes the backslashes a Windows path
is full of. That is a parity-source disagreement rather than this fix's
business.

## 6. And the failure that was already there

`synth/pypy_type_surface` is red on `main` itself, on all three runners and
byte-identically on all three backends — `bridges_compiled 102 -> 5,
guard_failures 20497 -> 1011` in
[ubuntu](https://github.com/youknowone/pyre/actions/runs/31258456255/job/93106657036),
[macos](https://github.com/youknowone/pyre/actions/runs/31258456255/job/93107449271)
and [windows](https://github.com/youknowone/pyre/actions/runs/31258456255/job/93107640525)
— so the branch could not be green while it stood.

The recorded 102/20497 are what the `Cls.__name__` fold measured while its
metaclass guard could not be discharged: the guard reads the raw `w_class`
slot, the fast path answers through `typedef::type`'s `gettypefor(ob_type)`
fallback, and a receiver reached through that fallback got a
`guard_value(NULL, type)` that failed on every execution and that nothing later
discharged. Taking the fixture apart puts every one of the 97 excess bridges in
`check_descriptor_kinds`, which reads `type(type.__dict__[name]).__name__`; the
`getset_descriptor` type object is the one class that reaches it, built lazily
from inside the init loop so it never enters the registry the post-loop sweep
walks to stamp the slot.

#1106 landed the fold's narrowing while this branch was rebasing, so the
cherry-pick dropped out; what is left is the baseline, re-recorded here from
what the three runners and a local windows run all report rather than from a
prediction.

Also left of it is a `debug_assert` on the helper: `is_exact_builtin_instance`
and `is_plain_int1` both read a null `w_class` as a second spelling of "exact
builtin", so an operand admitted under that spelling and pinned against the
canonical gets a guard its own recorded operand fails. Measured across
`bench/synth`, `bench` and `parity_tests` — 1217 pins over 131 files — every
site already establishes what it pins, and nothing checked that they did.

## 7. The review round

* **`sys.getwindowsversion` built its structseq type inside the call**, so
  `type(sys.getwindowsversion()) is type(sys.getwindowsversion())` read False
  where CPython reads True. It was the one of pyre's ten structseq types not
  already caching in a `OnceLock`; a probe over every structseq both runtimes
  carry reports the types equal for all nine others and for all of CPython's,
  so this constructor was the whole divergence. `structseq_type_identity` pins
  the property and fails on the binary built before the fix.
* **A timed-out child's stderr was bytes.** `subprocess.run(text=True)` decodes
  what `communicate()` returns; the timeout path raises with the raw chunks it
  had joined. Reporting those would have printed one `b'...\n...'` line — the
  shape §1 exists to remove.
* **Two `utime` assertions could only hold on NTFS**, and the reason neither
  was caught is that the parity step has never reached ubuntu or macos:
  `check.py` runs first in that job and has been failing, so the suite never
  starts. `ns=(2**62, 2**62)` expected 2\*\*62 rounded to a FILETIME tick, and
  `(8.8e11, 8.8e11)` names the year 29880 — an APFS timestamp is itself an
  int64 of nanoseconds so 2262 bounds it too, and ext4 stops in 2446. Both go
  through one `storable()` now, and the far date is held to the identity the
  `i128` widening buys wherever the write succeeds.

## Verification

Rebased onto `a7eb493079`; re-extracted LLBC before every measurement a stale
image could have fabricated.

* `extra_tests/parity_tests` on Windows — **214/214 on both backends**, bar
  `builtin_module_loader_spec` under a local CPython with no `test` package
  installed; CI's reference has it and reported `cpython=OK`. It was 208/213
  before, five scripts red on every Windows runner.
* `pyre/check.py` — **410/410 on dynasm and 410/410 on cranelift**, both
  all-green. `synth/type_name_attr_fold` still reads its own `5/0/1/4`, so
  #1106's decline is narrow enough that the fold keeps applying wherever its
  guard holds.
* `cargo test --all --no-default-features --features dynasm`, and the cranelift
  subset CI re-runs after it — both clean. `cargo fmt --all -- --check` clean.
* four ad-hoc probe scripts covering 30 `utime`/`truncate`/`stat` shapes,
  diffed line by line against CPython 3.14 — **0 differences**.

## Summary by CodeRabbit

- **New Features**
  - Added Windows support for `sys.getwindowsversion()`, locale operations, and platform-specific filesystem behavior.
  - Improved handling of negative, high-precision, and far-future file timestamps.
  - Standardized object representations across platforms, including Windows address formatting.

- **Bug Fixes**
  - Improved timestamp validation, overflow handling, truncation, and filesystem metadata reporting.
  - Windows interrupt termination now exits with the expected control-C status.
  - Added clearer test failure details, timeout diagnostics, and CI annotations.

- **Tests**
  - Expanded parity coverage for locale behavior, timestamp APIs, and struct-sequence types.


youknowone:
Thanks — three of the four were real, and one of them was worse than reported.

**`sys.getwindowsversion` rebuilding its type: fixed, and it was a behaviour bug.** Building the struct-sequence type inside the call meant `type(sys.getwindowsversion()) is type(sys.getwindowsversion())` read **False** where CPython reads True. Probing every structseq both runtimes carry: pyre reports the types equal for all nine others and CPython for all of its, so this one constructor was the whole divergence — and it was the only one of pyre's ten that did not already cache in a `OnceLock` (`stat_result`, `terminal_size`, `uname_result`, `statvfs_result`, `waitid_result`, `times_result`, `struct_time`, `_ExceptHookArgs`, `UnraisableHookArgs` all do). It now does the same. The cache is a bare `usize` the GC cannot see, which is the established pattern rather than a new bet — checked rather than assumed: 200 `os.stat` answers dropped across forced collections plus 60MB of churn leave the cached type identical and its fields readable.

New `parity_tests/structseq_type_identity.py` pins the property for every structseq the host carries. It fails on the binary built before the fix, naming exactly `sys.getwindowsversion`.

**Timed-out child's stderr: fixed.** `subprocess.run(text=True)` decodes what `communicate()` *returns*; the timeout path raises with the raw chunks it had joined, so it is bytes on POSIX and str on Windows. It is decoded now — reporting the bytes would have printed one `b'...\n...'` line, which is the unreadable shape this PR exists to remove.

**The 8.8e11-second `utime`: fixed, and there was a second one beside it.** Both assertions could only ever have held on NTFS, and the reason neither had been caught is that **the parity step has never run on ubuntu or macos** — `check.py` runs first in that job and has been failing, so the suite never starts.

* `ns=(2**62, 2**62)` expected `4611686018427387900`, which is 2\*\*62 rounded down to a FILETIME's 100ns tick; ext4 and APFS keep the nanosecond and answer `...904`.
* `utime(p, (8.8e11, 8.8e11))` names the year 29880. No filesystem but NTFS reaches it — an APFS timestamp *is* an int64 of nanoseconds, so 2262 bounds it too, and ext4 stores a 34-bit second and stops in 2446.

Both spellings now come from one `storable()`, which also replaces the `-1`/`-100` conditional above them. The far-date block asserts the identity the `i128` widening actually buys — `st_mtime_ns == int(st_mtime) * 1_000_000_000` — wherever the write succeeds, and the exact value only where the second survived the round trip.

**f-strings in `surrogate_name_messages`: left alone.** The file builds every expected repr with `%` formatting, including the assertions this branch did not touch, and nothing lints for it; converting one helper would make it the odd one out.

Verified after the change, with LLBC re-extracted first: `check.py` 410/410 on dynasm and 410/410 on cranelift, parity 213/213 on both backends (bar `builtin_module_loader_spec` under a local CPython with no `test` package), `cargo test --all --features dynasm` clean, `cargo fmt --check` clean.

PR #1106, opened 2026-08-07
jit(wasm): drop the wasm32 arm of the self-recursive root-bridge inline

youknowone:
`bridge_rec_root_selfrec` (`inline_call.rs`) was gated on `cfg!(not(target_arch = "wasm32"))`, so a guard-failure bridge reaching a self-recursive callee's `CALL` took the root-bridge admission on dynasm and cranelift and residualized on wasm. This drops the cfg term, leaving the last policy-level `wasm32` cfg in the JIT frontend behind.

## The claim did not survive an audit

The cfg's stated reason — *"the wasm always-portal path type-confuses the self-recursive inline (`setintbound: got Ref`)"* — is phrasing carried over from a different site:

* `setintbound: got Ref` enters the tree in #643 `920367da965`, attached to `fbw_inline_callee_hazardous` — a decline that is **not** cfg-gated and rests on the `CALL_ASSEMBLER` moving-nursery trampoline frame. That slice did measure wasm.
* The **cfg gate** is a separate site introduced by #749 `71726847a10`, whose commit message reports `dynasm + cranelift 294/294` and no wasm number at all.
* `always-portal` appears in three comments and names nothing in the tree.

Both mechanisms that could make such a type confusion wasm-specific are refuted:

* **Word-size Int→Ref promotion.** `W_IntObject { ob_header, intval: i64 }` is i64 on every target. `ALLOW_UNBOXING_INTS = LONG_BIT == 64` gates the mapdict unboxed attribute slot, which is pointer-sized — not the integer's width.
* **A missing wasm `CALL_ASSEMBLER` capability.** General wasm CA landed in #564 `c89254a6211` on 2026-07-15 and was hardened through #575/#609 by 07-17 — before both #643 and #749. Unlike Hazard 3, this is not a capability that arrived after the claim.

`fbw_inline_callee_hazardous`'s own self-recursive decline is deliberately **unchanged** — it is all-backend and rests on the CA-trampoline reason. Only the unsupported wasm clause is removed from its doc.

## Coverage the slice was missing

`bridge_recursion_overflow` already exercised this admission, but only in its easiest form: tail recursion whose live set is two machine integers. `synth/selfrec_bridge_nontail_promote` adds the three ingredients a "Ref reached an int operation" actually needs —

* **non-tail** recursion (`inner + len(tag)`), so the guard's resume stream is multi-frame;
* a **Ref local live across the recursive CALL** beside the int;
* an accumulator crossing `2**63` at level 13 of 24, so it genuinely **promotes to a long — a Ref** — partway down and the overflow guard fires inside the recursive frame.

| | loops / bridges / aborted / guard_failures |
|---|---|
| wasm, decline kept | 2 / 4 / 1 / 606 |
| wasm, decline removed | 2 / **6** / **0** / **806** |
| dynasm = cranelift | 2 / 6 / 0 / 806 — byte-identical |

806 − 606 = 200 = exactly one `trace_eagerness` bucket.

## Verification

`check.py` with this commit alone: **dynasm 405/406, cranelift 405/406, wasm 401/402.** All five wasm fixtures the admission moves are re-recorded and now carry the native values, except `recursion_memo_branch` at one guard failure apart — previously twenty-one.

The one failure, `synth/pypy_type_surface`, is not this change:

* it reports the same `bridges_compiled 5 -> 102, guard_failures 1011 -> 20497` with both touched files restored to `origin/main`;
* it fails identically on all three backends, while this change is a no-op wherever the cfg already read `true`;
* the fixture contains no self-recursive function for `code_is_self_recursive` to answer.

It is a red at the main tip; the second commit below diagnoses and fixes it, after which all three backends are fully green.

## Performance

wasm user+sys CPU, min/median over 31 interleaved samples, decline removed vs kept (all arms OUTPUT-MATCH):

| fixture | min | median |
|---|---|---|
| `wasm_ca_trampoline_decline` | −20.1% | −20.7% |
| `selfrec_bridge_nontail_promote` | −12.9% | −11.9% |
| `recursion_memo_branch` | −1.2% | −1.6% |
| `ca_bridge_multiframe_resume_double_call` | +1.2% | +2.7% |
| `foriter_call_resume_drops_iteration` | +2.4% | +4.0% |

The two that get slower now report exactly what the dynasm baseline records — 16 bridges / 0 aborts and 27 bridges / 1 abort — where the decline left them at 16/1 and 26/2. The cost is wasm taking the native decision and paying its compile toll on a short bench, not a shortcut being removed.

---

## Second commit: a main-tip red this branch had to clear first

`synth/pypy_type_surface` fails on **all three backends** at `origin/main`
(`bridges_compiled 5 -> 102, guard_failures 1011 -> 20497`). Verified not to be
the first commit's doing: it reports identically with both touched files
restored to `origin/main`, it fails where this branch is a no-op (the cfg
already read `true` on the natives), and the fixture holds no self-recursive
function. So it had to be diagnosed rather than worked around.

**Cause.** `try_walker_specialize_load_type_name_attr` (#1097) takes the
metaclass from `type_name_obj_fast_path` → `typedef::type`, which falls back to
`gettypefor(ob_type)` when the receiver's `w_class` slot is null. The fold then
guards the metaclass by reading the **raw `w_class` field**, so such a receiver
gets `guard_value(NULL, type)` — unfailable-to-discharge, and nothing writes the
slot afterwards, so it fails once per iteration forever.

One class reaches it: the `getset_descriptor` type object is built lazily inside
the init loop as a builder for other typedefs' descriptors, so it never enters
the registry whose post-loop sweep stamps `w_class = type`. `typedef::type`
still answers `type` via the fallback, so the null is invisible from Python.

**Localisation.** `check_descriptor_kinds` alone carries all 97 excess bridges;
the receiver value and the `FOR_ITER` are both irrelevant; `int`, `str`, `list`,
`NoneType`, `object`, `type`, `method_descriptor`, `wrapper_descriptor`,
`builtin_function_or_method` and a user class are clean; and
`getattr(C, "__name__")` — identical semantics, but the fold declines because
the site's name is not in the code's name table — is clean.

**After.** check.py **dynasm 406/406, cranelift 406/406, wasm 402/402**, fully
green. `pypy_type_surface` reads its recorded 11/5/0/1011 exactly and #1097's own
`type_name_attr_fold` still reads 5/0/1/4, so the fold keeps applying wherever
its guard holds.

Happy to split this into its own PR if you would rather review it separately —
it is a single self-contained commit.

---

## Third commit: the Windows check.py red, also from the base

`pyre/check.py (windows-latest)` fails at the merge-base `d936eb4be42` with
`os_utime_pathconf_truncate.py cpython=FAIL dynasm=FAIL cranelift=FAIL` — the
same row on #1102, whose merge is that base. Two halves of that script are
POSIX-only: the `times=` keyword check reused a time before the epoch, which
the block above it already excludes Windows from, and the `pathconf` section
reads `os.pathconf_names`, registered under `#[cfg(unix)]` here and absent from
CPython's `os` on Windows. The keyword spelling now uses a time the platform
holds and the `pathconf` section runs where the names table exists;
`pyre/extra_tests/parity_tests/run.py` is green on macOS.

## Fourth commit: the `test.test_re` CPython-suite red, cherry-picked

`CPython suite (gate)` also fails at the base — `test.test_re` raises
`TypeError: list indices must be integers or slices, not list_iterator` from
`_compiler.py:504 _get_charset_prefix`, where `p[0]` inside `for p in av[1]`
received the loop's iterator as its subscript index.

`reconstructed_all_ref_call_stack` read the residual op's Ref var-list at a
fixed operand offset 1. That holds for the Ref-only shape `iRd>r` but not for
the mixed `iIRd>r` the method-form CALL helpers lower through, whose leading
Int list is variable-width — so the read landed on the Int list's length byte
and resolved its register indices through the Ref bank. The composed stack is
`kept vstack prefix ++ this list`, whose height is `vstack_boxes.len()` for any
list length, so the abort-flush's `depth_at_py_pc` check could not fail.

The fix derives the offset from the op's argcodes. It is `624d547875d`,
authored on `rbigint`; cherry-picked here (`-x`) because that branch has no PR
yet and this one inherits the red. **Drop this commit when `rbigint` lands.**
The one conflict was in `tests.rs`, where the incoming hunk also carried a test
belonging to an earlier `rbigint` commit; only the new
`ref_var_list_offset_follows_the_argcodes_not_a_fixed_byte` was kept.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JIT handling for self-recursive calls, mixed integer/reference arguments, and resumed multi-frame execution across native and WebAssembly builds.
  - Prevented invalid `__name__` specialization guards and corrected operand reconstruction after `FOR_ITER` call aborts.
  - Improved platform-specific behavior for timestamp and configuration-limit tests on Windows.

- **Tests**
  - Added coverage for recursive promotion, reference-list layouts, and iterator-call recovery.
  - Updated benchmark statistics to reflect current bridge, guard, loop, and recovery behavior.


youknowone:

### §2 — reclassified, not fixed

> `inline_call.rs:2911 ↔ pyjitpl.py:1376` … The removed baseline documentation identifies this exact wasm path as `setintbound: got Ref`; no wasm typing/lowering repair is included in the patch.

The finding takes the deleted comment as authority, and that comment is exactly what this patch audits. Its provenance: `setintbound: got Ref` enters the tree in #643 `920367da965` attached to `fbw_inline_callee_hazardous` — a **different site**, not cfg-gated, resting on the `CALL_ASSEMBLER` trampoline frame, and that slice did measure wasm. The cfg at `inline_call.rs:2911` is #749 `71726847a10`, whose commit message reports `dynasm + cranelift 294/294` and no wasm number at all. `always-portal`, the mechanism the sentence names, appears in three comments and nothing else in the tree.

The cited upstream line argues the other way. `_opimpl_recursive_call` (pyjitpl.py:1376-1422) decides purely from a green-key frame count against `memmgr.max_unroll_recursion`: below it, `perform_call` **inlines** the recursion; at or above it, `dont_trace_here` plus `do_recursive_call` → `do_residual_call(assembler_call=True)`. There is no target-architecture term anywhere on that path, so a per-target policy term is the thing without upstream basis — keeping the cfg is the deviation, not removing it.

The typed-residual half of the finding is also unaffected: the non-inlined case still reaches `fbw_inline_callee_hazardous`, which this patch leaves in place on every backend and only edits the doc of.

Evidence that the named confusion does not occur, rather than that it merely was not hit: `synth/selfrec_bridge_nontail_promote` is built for it — non-tail recursion (multi-frame resume stream), a `Ref` local live across the recursive `CALL`, and an accumulator crossing `2**63` at level 13 of 24 so a long, a `Ref`, genuinely reaches the int operations. wasm reads 2/6/0/806 byte-identical to dynasm and cranelift, output matches, and `internal_compile_panics=0`. That counter is a real assert detector here: `jit_strict_mode()` is `cfg!(debug_assertions) || env MAJIT_STRICT`, and the guest has no environment, so an optimizer panic is caught and counted rather than re-raised.

### §3 — both confirmed real, both deferred with a named blocker

Neither is "pre-existing, therefore skip" — both are genuine divergences worth a follow-up, and both exceed what this diff can safely carry.

1. **`fbw_state.rs:1446 ↔ pyjitpl.py:1995`** — confirmed. pyre aborts and permanently declines the nested residual; PyPy executes and records it. Upstream never aborts for recursion: it counts same-green-key frames and, past `max_unroll_recursion`, emits the residual `CALL_ASSEMBLER`. Convergence path: port the green-key recursion count and the recorded-residual shape into the walker so the decline becomes a bounded unroll limit instead of a permanent denial. Blocker: that replaces `fbw_abort_nested_unjournaled_residual`'s whole contract, which the journal/rollback layer is currently written against.

2. **`fbw_state.rs:1274 ↔ warmstate.py:679`** — confirmed; upstream calls `warmrunnerstate.dont_trace_here(greenboxes)` at pyjitpl.py:1413, immediately beside the decision above. pyre leaves the declined callee on the normal threshold path, so it is traced later than upstream would trace it. Blocker: `JC_DONT_TRACE_HERE` lives in a JitCell keyed by green arguments; pyre's denial is a per-thread `HashSet` of code-object addresses (§4's own second item), so this needs the JitCell keying to land first.

### §4 — both accurate, kept as documented adaptations.

— *commented by Claude*

PR #1109, opened 2026-08-07
gate brake both directions, and the 108 opcodes behind the codewriter catch-all

youknowone:
Follows #1105, which merged while these were being written. Four commits: two
close `rework.md`'s F5, one re-measures and shrinks F4, one renames a local whose
name is taken elsewhere in the tree.

## `487d825` — the brake now scans the harness Python

`gate_triage_complete.rs` walked only workspace-member Rust. Adding `environ.get`
to `READ_FORMS` and `pyre/` + `scripts/` to the walk made **six** gates visible
that no `.rs` file reads — `PYRE_CHECK_PYPY3`, `PYRE_CHECK_PYTHON3`,
`PYRE_SHARED_BUILD`, `PYRE_SYNTH_PYPY`, `PYRE_SYNTH_PYRE`, `PYRE_SYNTH_PYTHON`.
Every census to date, including §6's own, had missed all six.

An allow-list, not a tree walk: the repository vendors ~4600 upstream `.py` files
under `lib-python/`, `pypy/`, `rpython/`, `lib_pypy/`, and untracked scratch sits
at the root. Only unambiguous reads count — the harness also *writes* gates into
a child's environment (`env[…] = …`), a subscript cannot be told from a read
without parsing, and writing a gate for a child is not owning it.

Two names came back undocumented and both turned out to be documented all along:
`gate-triage.md` wrote them as `_PYRE`, `_PYTHON` in its run-on shorthand, which
whole-token matching cannot resolve. They are spelled out now, and §6 carries the
rule. A census can only find what the document is willing to say.

Also from #1105's review: §1d's heading reads *Parity verdicts*, so the section
counts as live, yet its table marks `PYRE_FBW_VABLE_SCALAR_CA` **RETIRED** — a
mixed section that section-level granularity cannot express. A row that says
"retired" now documents nothing wherever it sits, anchored on that gate.

## `fb9f53f` — the catch-all was not what the finding said

`_other => emit_abort_permanent!(py_pc)` was recorded in `rework.md` as "an
opcode nobody has looked at". Measured against the enum: **227** `Instruction`
variants, **119** named by the dispatch, **108** in the catch-all — and all 108
are opcodes this compiler cannot emit.

| class | n |
|---|---|
| adaptive specializations (`BinaryOpAddInt`, `LoadAttrSlot`, `ToBoolBool`, …) | 82 |
| `Instrumented*` (`sys.monitoring`) | 21 |
| interpreter/JIT-internal (`EnterExecutor`, `InterpreterExit`, both `JumpBackward` forms, `Reserved`) | 5 |

pyre's eval loop does not quicken — nothing calls `replace_op` outside a
corruption test — so the specializations cannot reach the walk, and
`deoptimize()` is deliberately not called on them: a stream carrying a
specialization this interpreter never wrote is corrupt, not something to fold
back. So the unlisted cliff was never a live cliff. What mattered was the
*silence*: had one appeared, the loop declined with no record of which opcode did
it.

Classifying all 108 and deleting `_other` makes the match exhaustive, so a
variant added upstream fails to compile here instead of vanishing into a
catch-all. `majit-translate`'s `flowspace/flowcontext.rs` already classifies the
same three groups; this brings the walker in line with it.

Behaviour is unchanged — same arms, same macro, only the patterns are spelled
out. The coverage claim is load-bearing rather than decorative: dropping one
pattern fails with `error[E0004]: non-exhaustive patterns:
Instruction::ToBoolBool not covered`.

## `4b4ef3b` — the brake's other direction, and one retirement

The brake failed on a read with no live entry. A live entry whose reader was
deleted stayed green, and that is the quieter half: it survives the code it
describes, and the next sweep spends its time re-deriving that the name is
already gone. `PYRE_FBW_REC_UNROLL` sat in §5's config-switch list from
2026-07-06, when PR#374 (`9a97c47f6e9`) deleted `fbw_unroll_bound()`; its
successor knob `PYRE_FBW_REC_UNROLL_DEPTH` was never listed at all and lost its
reader too (PR#887, `e5546b2ed36`).

`every_live_gate_triage_entry_still_has_a_reader` reports the difference the
other way, and points at §1c rather than at deleting the row — deleting it loses
why the gate existed. Planting a live row for a name nothing reads fails the test
by name, which is what keeps it from passing vacuously.

The document writes `PYRE_*` and `PYRE_FBW_*` for families and the token scan
stops at the `*`; a documented token ending in `_` is a wildcard stem, not a
name. With those set aside the two sets are equal at **111**.

## `6c0c1ad` — `pc_map` names something else in this tree

`snapshot_map_from_trace_snapshots` called its `SnapshotFramePcs` local `pc_map`,
and three of five call sites bound it as `snapshot_pc_map`. In
`pyre-jit/src/jit/codewriter.rs`, `pc_map` is the live `-live-` marker table
keyed by Python pc (`pc_map[py_pc]`). This one is keyed by snapshot id and holds
one `(jitcode_index, pc, py_pc)` per frame. Renamed to `frame_pcs_map` /
`snapshot_frame_pcs`, which is what the other two call sites already used.

## Documents

`rework.md`: F5 deleted (closed), its two costly lessons kept in *Settled*; F4
rewritten to what the measurement showed, so what remains of it is the eight
unported opcodes. `gate-triage.md`: §1c 10 → 11, not-gates 11 → 12 (§2's heading
already said 12), config switches ~18 → ~17.

## Verification

`cargo test -p pyre-jit --features dynasm` 33 passed · brake 3 passed ·
`cargo fmt --all -- --check` rc=0 · `check.py --backend dynasm` **404 passed, 1
failed** — the one failure being `synth/pypy_type_surface`, which is main's own
(#1099 names it) and reproduces without either commit.

Two local-only failures appeared first and were **not** real:
`synth/make_function_inline` **wrong output** and `synth/del_cellvar_walk_commit`
jit-stats. Both came from extracting one crate's LLBC — `extract-llbc.py
pyre-jit` leaves `pyre-object`/`pyre-interpreter` at their previous base, and
this branch's base had moved. Re-extracting the whole set cleared both. CI never
sees that mixture because it extracts every crate fresh, which is exactly what
made the failure look like a regression CI was missing.

These numbers were measured before the rebase onto `0b1dee0`; re-measurement on
the new base is running.

### DIFF
--- a/majit/majit-metainterp/src/pyjitpl.rs
+++ b/majit/majit-metainterp/src/pyjitpl.rs
@@ -514,7 +514,10 @@ fn snapshot_map_from_trace_snapshots(
     let mut size_map = Vec::new();
     let mut vable_map = Vec::new();
     let mut vref_map = Vec::new();
-    let mut pc_map = Vec::new();
+    // Not `pc_map`: that name belongs to the `-live-` marker table keyed by
+    // Python pc (`pc_map[py_pc]`, `pyre-jit/src/jit/codewriter.rs`). This is
+    // keyed by snapshot id and holds one `(jitcode_index, pc, py_pc)` per frame.
+    let mut frame_pcs_map = Vec::new();
     // opencoder.py:603 _encode: trace snapshot recorder only emits Box
     // (live deadframe slot) and Const (compile-time pool) payloads.
     // TAGVIRTUAL belongs to resume numbering (resume.py:_number_boxes)
@@ -575,9 +578,9 @@ fn snapshot_map_from_trace_snapshots(
         snapshot_insert(&mut size_map, id, frame_sizes);
         snapshot_insert(&mut vable_map, id, vable_boxes);
         snapshot_insert(&mut vref_map, id, vref_boxes);
-        snapshot_insert(&mut pc_map, id, frame_pcs);
+        snapshot_insert(&mut frame_pcs_map, id, frame_pcs);
     }
-    (box_map, size_map, vable_map, vref_map, pc_map)
+    (box_map, size_map, vable_map, vref_map, frame_pcs_map)
 }
 
 struct PreparedBridgeTrace {
@@ -6025,7 +6028,7 @@ impl<M: Clone> MetaInterp<M> {
             snapshot_frame_size_map,
             mut snapshot_vable_map,
             mut snapshot_vref_map,
-            snapshot_pc_map,
+            snapshot_frame_pcs,
         ) = snapshot_map_from_trace_snapshots(&trace_snapshots, &mut constants);
         // history.py:220/261/307 — `Const{Int,Float,Ptr}.type` is an
         // intrinsic attribute on the Box itself, so no raw-u32 type
@@ -6035,7 +6038,7 @@ impl<M: Clone> MetaInterp<M> {
         unroll_opt.snapshot_frame_sizes = snapshot_frame_size_map.clone();
         unroll_opt.snapshot_vable_boxes = snapshot_vable_map.clone();
         unroll_opt.snapshot_vref_boxes = snapshot_vref_map.clone();
-        unroll_opt.snapshot_frame_pcs = snapshot_pc_map.clone();
+        unroll_opt.snapshot_frame_pcs = snapshot_frame_pcs.clone();
         // The original snapshot maps are re-cloned into `simple_opt` on the
         // InvalidLoop retry below, so they must stay rooted across the WHOLE
         // unroll. Each phase's `replace_compile_snapshot_roots` overwrites the
@@ -6169,7 +6172,7 @@ impl<M: Clone> MetaInterp<M> {
                         simple_opt.snapshot_frame_sizes = snapshot_frame_size_map;
                         simple_opt.snapshot_vable_boxes = snapshot_vable_map;
                         simple_opt.snapshot_vref_boxes = snapshot_vref_map;
-                        simple_opt.snapshot_frame_pcs = snapshot_pc_map;
+                        simple_opt.snapshot_frame_pcs = snapshot_frame_pcs;
                         simple_opt.call_pure_results = call_pure_results.clone();
                         // Forward the recorder's operand pool — the retry path
                         // uses the same upstream `Rc<Box>` allocations from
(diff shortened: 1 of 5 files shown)


youknowone:
Rebased onto `4555e3d` and force-pushed. The `cargo test` red was mine and is fixed; the other two are main's.

## `cargo test` — the brake caught a real case, on a rule that was too narrow

`every_live_gate_triage_entry_still_has_a_reader` failed on
`PYRE_CARRIER_EXC_RESUME`. Its reader (`carrier_exc_resume_enabled`) was deleted
2026-08-06 and §5 records it retired — but §1e is a **live** section that
discusses the seam that gate used to guard and names it three more times, on
lines that do not say "retired". Skipping retirement *lines* is not enough when
the same gate is written elsewhere.

Retirement is a property of the gate, not of the line: the new commit collects
the retired names and drops them from the live set wherever else they appear.

Only the **first** name on a retirement line is its subject. §1c's row for
`PYRE_AUTHORITATIVE` ends "`PYRE_PROBE_AUTHORITATIVE` is separate and remains
live" — that clause exists precisely because the second name is not the one being
retired, and retiring every name on the line took `PYRE_PROBE_AUTHORITATIVE` and
`PYRE_JIT` with it when I first tried the blunt rule.

Both directions are empty on this base: **110 read, 110 documented live**. Four
anchors now pin the classification (`PYRE_SINGLE_PASS`,
`PYRE_FBW_VABLE_SCALAR_CA`, `PYRE_CARRIER_EXC_RESUME` must not count;
`PYRE_JD1` must). Both negative controls fail as designed: a planted live row
with no reader fails naming it, and marking a read gate retired fails the other
direction naming `PYRE_JD1`.

## The other two reds are main's own

- **`pyre/check.py`** — `synth/pypy_type_surface`, `bridges_compiled 5 -> 102`,
  `guard_failures 1011 -> 20497`, identically on all three backends. Measured
  locally on this base: dynasm **1 failed, 407 passed**; cranelift **1 failed,
  407 passed** — the same single bench. main's own CI reports it, and #1099 named
  it without re-recording this one.
- **`CPython suite (gate)`** — `test.test_re: PASS -> FAIL`. main's own run at
  this base (`4555e3d`) fails the same job.

Neither is reachable from this branch: two commits are a test and two markdown
files, one is an identifier rename, and the codewriter commit only spells out
patterns that already fell to the same `_other` arm.

## One trap worth recording

Two local-only failures appeared earlier and were **not** real —
`synth/make_function_inline` **wrong output** and `synth/del_cellvar_walk_commit`
jit-stats. Both came from `extract-llbc.py pyre-jit`: extracting one crate leaves
`pyre-object`/`pyre-interpreter` at their previous base, and the JIT then
compiles from the mixture. Re-extracting the whole set cleared both. CI never
sees that mixture because it extracts every crate fresh, which is exactly what
made it look like a regression CI was missing.

— *commented by Claude*

PR #1112, opened 2026-08-08
sys: audit hooks and the sys._getframe event; jit: feed portal_trace_positions, and force only the frame _getframe returns

youknowone:
Five commits, all gated with `PYRE_LLBC_STRICT=1 python3 pyre/check.py --synthetic-only` on dynasm, cranelift and wasm: **395 / 395 / 394, all passed.**

## `sys.audit` / `sys.addaudithook` (2 commits)

`2e485d58f7c` ports the audit-hook machinery from `pypy/module/sys/vm.py:438-504` — `AuditHolder`, `trigger_audit_events`, `audit`, `addaudithook`. The registration that existed was a no-op stub. Reading the port against upstream turned up a real parity defect in it: `hooks_w` was being pushed in place, where upstream *replaces* the list (`holder.hooks_w = holder.hooks_w + [w_hook]`, with `debug.make_sure_not_resized` as the other half of that declaration). `Box<[PyObjectRef]>` makes it structurally unresizable.

`7452b43fa83` emits `vm.py:51 audit(space, "sys._getframe", [f])` and makes the constant-depth walker fold accountable for it. With a hook already armed the fold declines, because upstream's `trigger_audit_events` is `@objectmodel.dont_inline`; with none, it records a 7th quasi-immutable descr on the `hooks_w is None` read plus one `GUARD_NOT_INVALIDATED`, so `addaudithook` revokes the loop instead of silently eliding the event. Both `descr.index()` censuses gained the arm.

The fixture for this was wrong on the first try, in a way worth naming: installing the hook *between* two `loop(N)` calls made GREEN and RED print identically, because a fresh call re-traces anyway and never re-enters the compiled loop. Moving the install *inside* the loop split them (`9999` → `1`, `loops_aborted` 5 → 0).

## `portal_trace_positions` (1 commit)

`6f1d7ef4650`. The log was empty in every production run, so `find_biggest_function` always answered `None` and **every too-long abort took the segmenting arm by default rather than by finding**. The FBW walker inlines a Python callee by walking its body directly, with no `MIFrame`, so `newframe`'s append never ran. The sub-walk's `walk` call — and nothing else — is now bracketed, and `note_root_trace_too_long` runs upstream's two arms separately (`FORCE_BRIDGE_SEGMENTING` moves to the segmenting arm alone). On `trace_too_long_inline_multiframe`, 10 of 23 aborts now take the huge-function arm.

The first census read `ptp_push=12 ptp_pop=1` and looked like a leak. It was not: the counters sat inside the append, which early-returns once `blackhole_if_trace_too_long` retires the log mid-sub-walk. Moved to the call sites they read 12 == 12, and a mismatch now means one thing only.

## `sys._getframe` forced the top of the stack at every depth (1 commit)

`e9a29d81740`. `getframe` called `ec.gettopframe()` ahead of the walk and **discarded the result** — it was there for the force. `force_pyframe` clears the traced virtualizable's `TOKEN_TRACING_RESCALL` for any frame `flush_active_frame_escape` matches, and at the top walk level that top frame *is* the portal, so the residual reported `VableEscapedDuringResidualCall` at every depth — including the depths whose answer lies below the portal, where `pyjitpl.py:2159-2161 _do_jit_force_virtual` answers `topframeref` with `virtualizable_boxes[-1]` and forces nothing.

The trailing `force_frame(current)` already covers the frame handed out, which is the access `rvirtualizable.py:49-53 hook_access_field` would have fired for. The walk itself needs neither force: `hide()` reads `pycode` (written at construction), `f_backref` is not a virtualizable field, and a level the JIT kept virtual is materialised by the vref force `gettopframe_nohidden` already takes.

Identical on all three backends, with **no output changed anywhere in the corpus**:

| fixture | before | after |
| --- | --- | --- |
| `getframe_root_loop_force_while_merge_declined` | aborted 5, compiled 0 | aborted 0, compiled 1 |
| `getframe_root_loop_force_blackhole_crn_declined` | aborted 5, compiled 0 | aborted 0, compiled 1 |
| `getframe_root_loop_force_blackhole_crn_nonidempotent_declined` | aborted 5, compiled 0 | aborted 0, compiled 1 |
| `getframe_bridge_force_plain_declined` | aborted 20, guard_failures 4114 | aborted 0, guard_failures 201 |
| `getframe_bridge_force_after_store_declined` | aborted 20, guard_failures 4114 | aborted 0, guard_failures 201 |
| `getframe_inline_subwalk_multiframe` | aborted 15, compiled 0 | aborted 1, compiled 2 |

Six fixtures whose coverage rode on the removed force take it from `sys._getframe(0).f_locals` instead — the call folds and the getset forces the portal on its own — and each reproduces its recorded counters **exactly** on all three backends. `f_locals` is the only frame getset that forces: swapping the read for `f_lasti` or `f_lineno` leaves `loops_aborted` at 0, which corrects a claim the fold's own doc carried. The three remaining fixtures keep their shape and re-record the duplicate escape that is gone.

### What this does *not* do

It does not extend the fold to constant depth ≥ 1, which is what the Codex review item asked for. Reading `_do_jit_force_virtual` first changed the framing: upstream residualizes each `f_backref()` hop too (`OS_JIT_FORCE_VIRTUAL`), so the gap was never that pyre residualized — it was that pyre's residual *forced*. What remains is that pyre emits one opaque `bh_call_fn(_getframe, …)` where upstream emits the getfield/`hide()`/force-virtual chain; that is now a heap-cache and call-overhead question rather than a lost-loop one, and it is filed rather than done here.

`sys._getframemodulename` keeps the same pre-walk force. It takes no force on the frame it *ends* at and then reads that frame's `w_globals`, so it is a separate question and is left alone.

## Verification notes

- The branch was rebased onto `main` earlier in the session and the LLBC artefacts were stale at that HEAD; every measurement here is after `scripts/extract-llbc.py pyre-object pyre-interpreter pyre-jit`.
- A control run at HEAD (fix reverted, fixtures untouched) reproduces every committed baseline exactly, so the nine moved fixtures are attributable to this change and not to the rebase.
- One fixture, `cranelift/synth/inline_freevar_after_mayforce`, reports `jit-stats unstable (not gated)` — check.py's own repeat-to-repeat detector, `guard_failures` 466 vs 471, unrelated to these commits.

## Summary by CodeRabbit

- **New Features**
  - Added support for `sys.audit` and `sys.addaudithook`, including hook registration, event dispatch, validation, and error handling.
  - Improved `sys._getframe` behavior, including frame-local access and audit-hook interactions.
  - Added portal-position tracing and expanded diagnostic counters.

- **Bug Fixes**
  - Improved handling of frame escapes, retired traces, inline tracing, and oversized traces.

- **Tests**
  - Added coverage for audit hooks, frame auditing, recursive drivers, and updated JIT execution statistics.


youknowone:
Codex P1 on `residual_call.rs:1356` (re-arm `mark_escape_flush_undo_pending` in the locals-only escape-flush arm) is answered in #1132, and the behaviour is deliberately left as-is.

That arm was removed by `jit: drop the locals-region leg's deferred escape-flush undo arm` on measurement, not on style: the `value-stack underflow` it had been added for came from an A/B that ran on stale LLBC artefacts. Re-running the arm that reproduced, on artefacts that pass `PYRE_LLBC_STRICT=1`, gives 0/10 panics on cranelift, with `pyre/check.py --synthetic-only` identical with and without the call (dynasm 1 failed / 390 passed either way) and no recorded counter or fixture output moving. The detector is not blind — deleting the `if !flushed` test from `LiveLastInstrGuard::drop` still makes `getframe_caller_resume_coord_two_call_sites` print the wrong answer 3/3.

What the review did surface is real, though: the comment below that arm still said "the undo stays armed so the legacy replay re-enters the pre-flush frame", and nothing consumes the capture from there. #1132 corrects the comment to state what actually happens — the committed-pc walk-end leg is gated on a pc this arm never sets, the deferred leg on a flag it never arms, so the capture sits until `capture_escape_flush_undo` supersedes it or the walk-start reset drops it.

The other findings from both reviewers are fixed in #1132. Note also that this PR was squash-merged at `d2f4ce1522e`, which predates two commits of the branch; those are the first two commits of #1132.

— *commented by Claude*

PR #1117, opened 2026-08-08
jit: seed the cut-trace escape closure from post-cut guard snapshots

youknowone:
`main`'s CPython gate has been red on exactly one module since 2026-08-07:

```
── REGRESSIONS (1) ──
  - test.test_re: PASS -> FAIL  rc=1
```

```
File "lib-python/3/re/_parser.py", line 460, in _parse_sub
  itemsappend(_parse(source, state, verbose, nested + 1, not nested and not items))
TypeError: _parse() missing 1 required positional argument: 'nested'
```

The message says arity; the defect is a **value**. `PYRE_BH_NULL_ARG=1` names it in
one run — `[bh-null-arg] arg[3]/5 NULL … at <frame _parser.py, line 460, code
_parse_sub>`. `args[3]` is `nested + 1`; a NULL in the argument vector is dropped
by `bh_call_fn_impl`'s `reload_args`, so five arguments arrive as four and the
binder reports the *name* as missing.

## Cause

`TreeLoop::cut_trace_from_with_consts` seeds its escaped-ref closure from
**post-cut operation arguments only**. A snapshot is not `fail_args` — `fail_args`
are attached after optimization by `store_final_boxes_in_guard`, so the cut
namespace owes them nothing, but a snapshot is captured during tracing and is the
only description a guard has of the frame it resumes into.

`nested + 1` is computed once before the cut and stays live on the Python operand
stack across the merge point, while every post-cut operand reads the cached
pre-cut box. So no post-cut *argument* names it, the seeding loop never sees it,
the snapshot remap's `else` arm rewrites it to `OpRef::NONE`, `_number_boxes`
emits `NULLREF`, and `store_live_frame_array_slot` writes NULL into the resumed
frame's operand slot.

Measured at the drop site, **every** dropped ref was snapshot-only
(`used_by_postcut_op=false`, 632 vable / 221 frame / 58 vref).

Upstream cannot hit this: `CutTrace` (`opencoder.py`) is a view over the same
trace and cuts at a jitdriver merge point where the whole live set is already
`inputargs` — a snapshot naming a position before `start` would trip
`TraceIterator._get`'s assert. Pyre cuts at a position whose live set is wider
than `original_boxes` (a Python operand stack lives in the virtualizable), so it
needs the prefix re-emission it already had, for op args only.

## Fix

Seed the closure from the vable, vref and frame sections of every post-cut
guard's snapshot as well.

Prefix re-emission **re-executes** the definition and the BFS pulls the whole
transitive cone, so a root is admitted only when its entire pre-cut dependency
cone is `has_no_side_effect()` — the condition is on the cone, not the root: an
allocation whose field feeds from a residual call would re-issue that call.
Measured seeds across a whole `test_re` run are exactly two: `NewWithVtable`
(admitted — resume already hands a virtual a fresh object, and the boxed operands
this recovers are compared by value) and `CallMayForceR` (**refused** — an extra
observable call is worse than the NULL slot it would repair).

## That refusal is a documented gap, and it was measured before being left open

A temporary probe at the refusal site plus the pre-existing `PYRE_BH_NULL_ARG`,
over **all 109 gated modules** (108 ran to completion; `test_runpy` truncated at
the gate's own 300 s, both counters 0 up to the cut):

| | |
|---|---|
| `bh_null` (**harm**) | **0 in every module** |
| `cut_refused` (**occurrence**) | `test.test_re` only, 3–4 per run, all `CallMayForceR` |
| that module | green with this patch |

Occurrence without observable harm, so this patch adds no machinery for it. The
shape that would close it without re-execution already exists in the same
function — phase 4 carries an `orig_inputarg_escaped` ref as an **extra inputarg**
instead of re-emitting it — and is worth extending when harm is actually observed.

## Verification

* `test.test_re` — **303 errors → 0**, `OK (skipped=13)`, rc=0. `PYRE_NO_JIT=1` was
  already `OK`, i.e. JIT-only.
* `pyre/check.py --backend dynasm` — 410/410 at `59dfaf3dbe3`. At the current base
  it is 409/410, and the one red is **not this patch**: `synth/pypy_type_surface`
  reports `bridges_compiled 102 -> 5, guard_failures 20497 -> 1011`. #1086
  (`e5eff816849`) rewrote 1063 `.jitstats` files and recorded **the storming state**
  for this fixture over a `main` where the type-name fold is fixed — the observed
  5 / 1011 (loops 11) is the correct post-fix reading. Attributed by a same-tree
  patch-on/off A/B: the base build without this commit fails with the byte-identical
  message. Not re-recorded here — re-recording a regression as a baseline is what
  produced this red.
* `cargo test -p majit-metainterp cut_trace_from` — 7 passed. The two new tests
  guard one half each: `test_cut_trace_from_reemits_ref_held_only_by_a_snapshot`
  goes red when the seed is disabled, and
  `test_cut_trace_from_leaves_side_effecting_snapshot_ref_unmapped` goes red when
  the cone filter is removed.
* `test_strftime`, an intermittent `_parse_sub` `IndexError` recorded at 4/25 —
  **0/25** here (and 0/25 independently on an earlier base). Strong, not proof.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trace cutting to correctly handle references captured in guard snapshots.
  * Preserved safe snapshot references while preventing side-effecting references from being reused.
  * Fixed handling of guard failure arguments to avoid exposing invalid references.
* **Tests**
  * Added coverage for snapshot-only reference restoration and side-effecting call safeguards.

### DIFF
--- a/.github/workflows/pyre-ci.yml
+++ b/.github/workflows/pyre-ci.yml
@@ -442,6 +442,15 @@ jobs:
       env:
         PYRE_CHECK_PYTHON3: ${{ steps.cpython.outputs.python-path }}
       run: ${{ steps.cpython.outputs.python-path }} pyre/extra_tests/parity_tests/run.py
+    - name: Run the vendored root extra_tests
+      # PyPy's own `extra_tests/` tree, run in place through a driver that
+      # supplies the pytest names those files use. Only `run.py`'s ENABLED list
+      # runs; the rest of the tree needs pytest surface, third-party packages,
+      # or PyPy modules that are not here. Windows selects nothing today and
+      # the runner exits 0 saying so.
+      env:
+        PYRE_CHECK_PYTHON3: ${{ steps.cpython.outputs.python-path }}
+      run: ${{ steps.cpython.outputs.python-path }} pyre/extra_tests/upstream/run.py
     - name: JIT structural stats vs committed baseline (Linux, informational)
       # Diff each benchmark's [jit-stats] counters against the committed
       # pyre/bench/<name>.<backend>.jitstats baselines. The counters are decided
--- a/majit/majit-metainterp/src/history.rs
+++ b/majit/majit-metainterp/src/history.rs
@@ -767,12 +767,94 @@ impl TreeLoop {
 
         // Seed with refs used by post-cut ops (args only, not fail_args).
         // RPython CutTrace parity: pre-cut refs in fail_args map to
-        // OpRef::NONE (resume data handles materialization). Only regular
-        // op args seed escaped refs for prefix re-emission.
+        // OpRef::NONE (resume data handles materialization) — fail_args are
+        // attached after optimization by `store_final_boxes_in_guard`, so the
+        // cut namespace owes them nothing. Only regular op args seed escaped
+        // refs for prefix re-emission.
         for op in cut_ops {
             for arg in op.getarglist().iter() {
                 if is_pre_cut_ref(&arg.to_opref()) && escaped_set.insert(arg.to_opref()) {
                     queue.push_back(arg.to_opref());
+                }
+            }
+        }
+
+        // A snapshot is NOT fail_args: it is captured during tracing and is the
+        // only description a guard has of the frame state it resumes into. Some
+        // pre-cut boxes reach the cut region through a snapshot slot alone —
+        // a pure loop-invariant value (`nested + 1` computed once before the
+        // cut) stays live on the Python operand stack across the merge point
+        // while every post-cut operand reads the cached pre-cut box, so no
+        // post-cut ARGUMENT names it and the loop above never sees it.
+        //
+        // Upstream never has to re-emit any of this: `CutTrace`
+        // (`opencoder.py`) is a view over the same trace, and the cut is a
+        // jitdriver merge point where the whole live set is already
+        // `inputargs` — a snapshot referencing a position before `start` would
+        // trip `TraceIterator._get`'s assert. Pyre cuts at a position whose
+        // live set is wider than `original_boxes`, so a snapshot-only
+        // reference has to seed the same prefix re-emission the operand path
+        // uses. Left unseeded, the remap below rewrote the slot to
+        // `OpRef::NONE`, `_number_boxes` emitted `NULLREF`, and the resumed
+        // frame's operand slot was written NULL — the argument of a live call
+        // arriving unbound.
+        //
+        // Prefix re-emission RE-EXECUTES the definition at the cut point, so it
+        // reconstructs faithfully only where nothing it drags along carries a
+        // side effect. The BFS below pulls in the whole transitive definition
+        // cone, so the condition is on the CONE, not the root: an allocation
+        // whose field feeds from a residual call would re-issue that call. Walk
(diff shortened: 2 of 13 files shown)


youknowone:
CI caught a regression this patch introduces, so I have moved it to draft.

The CPython gate goes `FAIL 1 -> FAIL 0` — `test.test_re` is fixed — but two modules that passed now crash:

```
test.test_collections: PASS -> CRASH  majit/majit-metainterp/src/compile.rs:2152
  compile.py:458 assert i == len(inputargs) failed (18 != 19)
test.test_urlparse:    PASS -> CRASH  same site, (21 != 22)
```

Both reproduce locally and deterministically (`rc=101`), and the base build without this commit does not crash.

`patch_new_loop_to_load_virtualizable_fields` requires `inputargs` to be exactly `red args + vable static fields + array items`, and asserts it. This patch widens the escaped-ref set, so phase 4's fallback — an escaped original inputarg with no pool constant becomes an **extra inputarg** — appends one arg the vinfo expansion cannot account for. A snapshot's frame boxes are the frame's locals, i.e. original inputargs, so the fallback that was rare under op-argument seeding now fires often.

Not merging until the seed is restricted to roots reconstructible without changing the entry contract.

— *commented by Claude*

youknowone:
Fixed in `ebf5f096f38`.

`patch_new_loop_to_load_virtualizable_fields` requires `inputargs` to be exactly the red args followed by the virtualizable's static fields and array items, and asserts it (`compile.py:458`). Phase 4 has two arms for an escaped **original** inputarg: remap to its pool constant, or append it as an extra inputarg — the second widens that list. A snapshot's frame boxes are the frame's locals, i.e. original inputargs, so seeding from snapshots reaches them constantly where operand seeding reached them rarely, and a latent fallback became the common path.

The cone gate now refuses a root whose cone reaches an original inputarg with no pool constant, testing **the same `inputarg_consts` lookup phase 4 branches on** so the two cannot disagree. A refused slot keeps the `OpRef::NONE` it had, exactly like the `CallMayForceR` case. `is_pre_cut_ref` already excludes `original_boxes`, so this only bites inputargs that are neither original boxes nor pool constants.

Verified on this base:

| | |
|---|---|
| CPython gate (the job that was red) | `PASS 108 / FAIL 0 / CRASH 0`, rc=0 |
| `test_re` / `test_collections` / `test_urlparse` | all rc=0 |
| `cargo test -p majit-metainterp cut_trace_from` | 8 passed |
| `test_strftime` | 0/25 |
| `check.py --backend dynasm` | 409/410 |

The new test is non-vacuous: with the gate removed it fails on `cut.inputargs.len()` — asserting the *length* is what discriminates, since the snapshot slot is `NONE` either way.

The one `check.py` red is not from this branch. `synth/pypy_type_surface` reports `bridges_compiled 102 -> 5, guard_failures 20497 -> 1011`, i.e. **recorded → observed**, and the observed 5 / 1011 (loops 11) is the correct post-fix reading; #1086 (`e5eff816849`) rewrote 1063 `.jitstats` files and recorded the storming state for this fixture over a `main` where the type-name fold is already fixed. Attributed by a same-tree patch-on/off A/B: the base build without these commits fails with the byte-identical message. Not re-recorded here.

— *commented by Claude*

PR #1122, opened 2026-08-09
jit: re-emit a cut-trace ref held only by a guard snapshot, or decline the cut

youknowone:
A cut trace could drop a value that only a post-cut guard's snapshot named, writing NULL into the resumed frame where a live call argument belonged. This fixes that, and closes one of the two holes the fix exposed.

## The defect

`cut_trace_from_with_consts` seeded its escape closure from post-cut **operand arguments** only. A pre-cut ref that no post-cut op names as an argument, but that a post-cut guard's snapshot still holds, was therefore never marked escaped. Phase 1's remap rewrote that slot to `OpRef::NONE`, `_number_boxes` encoded `NULLREF`, and the resumed frame's operand slot was written NULL — the argument of a live call arriving unbound.

Upstream never hits this. `CutTrace` (`opencoder.py`) is a *view* over the same trace, and upstream cuts at a jitdriver merge point where the whole live set is already `inputargs`. Pyre materializes the cut eagerly and cuts where the live set is *wider* than `original_boxes`, because the Python operand stack lives in the virtualizable rather than in the reds.

## What the commits do

1. **Seed the escape closure from post-cut guard snapshots** — the snapshot-only refs now join the same prefix re-emission the operand path already used.
2. **Cancel compilation when the cut cannot represent such a ref.** Falling back to the uncut trace is *not* available: once a merge point is live, the loop token, entry contract and optimizer seed downstream are all built for the cut namespace. Handing them the uncut trace fails `test_re` outright — measured 3/3 failing vs 0/3 on a same-binary toggle. Cancelling runs the loop in the interpreter, which is correct. Upstream's `trace.cut_trace_from` (`compile.py`) is total, but the cancellation lands on the outcome the `except InvalidLoop` arm just below it already defines.
3. **Decline a branch-guard kept slot that has no per-slot source** (`DispatchError::BranchGuardKeptSlotUnsourced`) — see the caveats below.
4. **Restrict re-emission to always-pure definition cones.**

## Commit 4 fixes a soundness bug commits 1–3 introduced

Commit 1 admitted any `has_no_side_effect()` cone for re-emission. That predicate spans 37 opcodes `is_always_pure()` rejects, and re-emission *re-executes* the definition at the cut point, so admitting them is wrong three separate ways:

- **Loads** (`GcLoad*`, `Getfield*`, `Getarrayitem*`, `RawLoad*`, `ThreadlocalrefGet`, …) cause no effect but *observe* one. A pre-cut store between the load and the cut is itself side-effecting and therefore never joins the cone, so replaying the load returns the **newer** value, not the one the snapshot named.
- **Allocations** (`New*`) take their contents from separate `SetfieldGc` stores, likewise outside the argument cone (`NewWithVtable` is built as `Op::with_descr(…, &[], sd)` — an empty arg list). Replaying the allocation alone yields a **blank object**.
- **`ForceToken` / `VirtualRef*`** mint identity; a replay mints a different one.

Narrowed to `is_always_pure()`. This was caught by the Codex parity review, not by the test suite — `test_re` passed both before and after.

**Measured**: on `test_re`, 4 cut attempts, 1 decline, root `NewWithVtable` — i.e. the old predicate's one admitted case in this corpus was re-emitting a blank allocation into a resumed frame's snapshot slot.

The pre-existing decline test uses `CallMayForceR`, which *both* predicates reject, so it could not distinguish them. The new test covers `GetfieldGcR` / `GcLoadR` / `NewWithVtable` and first asserts each opcode actually sits between the two predicates, so it cannot pass for the wrong reason.

## Caveats — please read before approving

**The kept-slot decline (commit 3) closes only half of its stated hole.** `collect_outer_active_boxes` prefers the vstack mirror, then `kept_recovered`, then a stale merge-colour read of `regs_r[color]`. The new decline fires only when the mirror entry is `OpRef::NONE`. It deliberately does **not** fire when the mirror holds a NULL const-ptr.

I instrumented rather than assumed. A census over `test_re` + `test_strftime` + 393 synthetic benchmarks produced **3 hits, all with `recovered=false`, all NULL const-ptr, all inside one *passing* synth**. So the gate's stated justification ("the recovery covered it") is false, but the corpus demonstrates no wrong behaviour. Widening the decline to cover the NULL case **regressed** `synth/list_append_write_barrier_gc` from loops 12 / bridges 5 / aborts 1 to loops 11 / bridges 3 / aborts 4, so it is left narrow and documented in place.

The root cause is that pyre's mirror NULL is overloaded across ≥8 writers, meaning both "live NULL" (the `PUSH_NULL` / `self_or_null` sentinel) and "unset". Upstream has no such ambiguity — `get_list_of_active_boxes` (`pyjitpl.py:225`) reads `registers_r[index]` directly and the register bank *is* the source. Unifying those semantics is filed as follow-up work; it is a wider change than this diff should carry.

**`BranchGuardKeptSlotUnsourced` fires 0 times in the corpus.** It is a guard against a state the census never reached, so it is exercised by construction and by unit tests, not by execution. Its `TraceAction` mapping is deliberately listed next to its sibling kept-stack decline rather than retiring the location, because the mirror shape and recovery set behind it are per-walk, not a static property of the coordinate.

## Verification

Run on base `eba36d13ed9` with the identical tree (the branch has since been rebased onto current `main`; the four commits' own diffs are unchanged):

- `cargo test --workspace --no-default-features --features dynasm` — **7533 passed / 0 failed**
- `test_re` — 165 tests OK, 0 failures
- `test_strftime` — 4 tests OK, 0 failures
- `synth/list_append_write_barrier_gc` — loops 12 / bridges 5 / aborts 1 / 1345, matching baseline exactly
- `cargo fmt --all --check` — clean

**`check.py` has not been re-run on the current base.** Its run was killed mid-flight when the branch was rebased under it, which changed `pyre-interpreter` sources and invalidated both the build inputs and the LLBC. It was fully green (dynasm 410/410, cranelift 410/410, wasm 406/406) on the pre-rebase tree with the same four commits. I am re-running it against this base and will post the result here.

— *commented by Claude*

## Summary by CodeRabbit

- **New Features**
  - Added incremental garbage-collection controls, hooks, and collection statistics.
  - Added heap dumps, type metadata, root/referent inspection, memory usage, and memory-pressure APIs.
  - Added assembler/JIT memory usage statistics across supported backends.
  - Enabled managed interpreter GC by default, with an environment-variable rollback option.
  - Added `__pypy__.add_memory_pressure()` support.

- **Bug Fixes**
  - Improved frame reallocation barriers, pinning validation, finalizer handling, and GC memory accounting.

- **Tests**
  - Added coverage for GC hooks, heap dumps, collection transitions, statistics, and memory accounting.


youknowone:
### Note for reviewers: this predicate was previously measured as unaffordable — that verdict does not carry over

Codex raised the same `has_no_side_effect()` → `is_always_pure()` finding on #1117. It was measured then and rejected: tightening the predicate admitted 0 of 7 censused seeds and `test_re` went `PASS -> FAIL` with **303 errors**, byte-identical to the pre-fix signature.

That measurement was correct, and it does not apply here, because the operative variable was never the predicate — it was **what a refusal does**:

| | refused root ⇒ | `test_re` |
|---|---|---|
| #1117 | keep the `OpRef::NONE` mapping and compile anyway | 303 errors |
| this PR | **cancel the compilation** | OK (165 tests) |

Keeping `NONE` *is* the original defect — a NULL written into the resumed frame's operand slot — so tightening the admit rule there simply routed more cases back into the bug. The tell is that the failure reproduced the original signature exactly, which indicates the refusal path leads into the defect rather than that the refusal is too strict.

Commit 2 (cancel-on-decline) is therefore the enabler for commit 4: cancelling runs the loop in the interpreter, which is correct, and that is what makes the sound predicate affordable.

One consequence worth noting: the #1117 census found **5 of 7** seeds were in `vable_boxes` rather than frame slots, and two of the three admitted seeds were re-deriving a *virtualizable array* slot — which upstream forbids outright (`consume_virtualizable_boxes` reads slot zero as the virtualizable itself and sizes the rest against it). Under `is_always_pure()` every one of those 7 candidates is refused, so that re-derivation no longer occurs at all.

— *commented by Claude*

youknowone:
### `check.py` is RED on this branch — converting to draft

Run on `dba0aa26c48`, base `48f800ec16f`. The run is valid: `git rev-parse HEAD` bookends the run and START == END, so nothing moved underneath it.

```
FAIL ×3 backends  synth/nested_break_not_hot
                  bridges_compiled 4 -> 0, guard_failures 1001 -> 5480
FAIL dynasm cpython-suite
  test.test_collections PASS -> CRASH  majit/majit-metainterp/src/compile.rs:2152
                                       compile.py:458 assert i == len(inputargs) failed (16 != 17)
  test.test_urlparse    PASS -> CRASH  ... (21 != 22)

dynasm 2 failed / 409 passed · cranelift 1 / 409 · wasm 1 / 405
```

Reproduced directly: both `rc=101`, and `PYRE_NO_JIT=1` gives `rc=0` — JIT defects, not base flakes.

**The crash is a known regression of this exact seed, and its fix is missing from this branch.** Seeding the escape closure from guard snapshots reaches *original inputargs* constantly, because a snapshot's frame boxes **are** the frame's locals. `cut_trace_from`'s phase 4 then takes its "append as an extra inputarg" arm, which widens `inputargs` past what `patch_new_loop_to_load_virtualizable_fields` asserts (`compile.py:458`: inputargs is exactly red args + static fields + array items).

The repair is to refuse a cone that reaches an original inputarg with no pool constant, testing **the same predicate phase 4 branches on** (`inputarg_consts.get(r.raw() as usize)`) so the gate and the arm cannot disagree. Neither that gate nor its test is present here.

**Not yet established:** whether the crash comes from commit 4, from the rebase onto the newer base, or was already present in commits 1–3. The earlier green run predates *both* commit 4 and the rebase, so two variables moved at once. Tightening the cone predicate only makes the seed decline more, which cannot widen `inputargs` — so I expect commits 1–3 to already carry it, but that is reasoning, not measurement, and I am measuring it rather than asserting it.

The `nested_break_not_hot` jit-stats regression (`bridges_compiled 4 -> 0`) is a separate, real coverage loss and is being investigated on its own.

Both are being fixed before this leaves draft.

— *commented by Claude*

youknowone:
### The cone gate treats a pre-cut original inputarg as an unconditional leaf, and that premise is measurably false

Context: #1121 carried an independent fix for the same defect. Its cut-trace commits are now **withdrawn** in favour of this PR — `is_always_pure()` is the correct predicate and cancel-on-decline is the correct refusal, and #1121 had neither. One finding from that work does not carry over, so it is reported here rather than lost.

**The claim.** In `snapshot_cone_is_reemittable`:

```rust
if r.raw() < num_original_inputargs || !seen.insert(r) {
    continue;
}
```

justified above by

> A pre-cut ORIGINAL inputarg has no definition to re-execute — phase 4 maps it to its pool constant — so it is a leaf.

Phase 4 has a **second arm**. When `inputarg_consts.get(r.raw())` is `None` it appends the value as an extra inputarg (`history.rs:955-970`, logging `appended as extra inputarg #N`). `patch_new_loop_to_load_virtualizable_fields` then requires the run after the red args to be exactly the virtualizable's expansion and asserts `i == len(inputargs)` (`compile.rs:2152`, `compile.py:458`). One appended arg fails that assert.

**Reproduction against this PR's own logic.** I transplanted this PR's leaf handling *and* its `is_always_pure()` predicate into a tree carrying the snapshot seed, then ran a test whose seed root is `IntAdd(iarg(1), iarg(0))` — always-pure, so it passes the stricter gate — named only by a post-cut guard's snapshot, with 2 original inputargs and 1 original box:

```
test_cut_trace_from_declines_a_snapshot_root_needing_an_extra_inputarg ... FAILED
assertion `left == right` failed: the cut grew an inputarg past the virtualizable expansion
  left: 2
 right: 1
```

So the widening is reachable under `is_always_pure()`, not only under the weaker predicate.

**Caveat on the field measurement, stated precisely.** I also observed this as a real crash — `test.test_urlparse` `PASS -> CRASH`, `assert i == len(inputargs) failed (21 != 22)`, with exactly one `[jit][cut-escape] … appended as extra inputarg #21` under `MAJIT_LOG=1`. That run used `has_no_side_effect()`. I have **not** measured whether that particular root survives `is_always_pure()`, so treat the unit test above as the evidence about this PR and the crash as the motivation. This PR's CPython gate is the authority.

**Suggested guard** — test the same predicate phase 4 branches on, so the two cannot disagree:

```rust
if r.raw() < num_original_inputargs {
    if inputarg_consts.get(r.raw() as usize).is_none() {
        return false;
    }
    continue;
}
if !seen.insert(r) {
    continue;
}
```

Here `false` reaches this PR's `return None`, so such a cut is cancelled rather than compiled with a widened entry contract — the same disposition the impure-cone case already gets. `is_pre_cut_ref` already excludes `original_boxes`, so this only bites inputargs that are neither original boxes nor pool constants.

Asserting the **length** is what discriminates: the snapshot slot is `NONE` either way, so a test that only checks the slot passes with the bug present.

The withdrawn commit with this guard and its test is `773ba81a070` (patch preserved locally; happy to hand it over in any form that is convenient).

— *commented by Claude*

youknowone:
### Pushed: the entry-contract crash is fixed; one jit-stats regression remains, so this stays draft

`4233d6b07f2` — *jit: refuse a snapshot seed whose cone needs an undeliverable inputarg*. The branch is also rebased onto current `main` (`range-diff` pairs all four earlier commits `=`, i.e. content-identical, base moved).

**Fixed.** A snapshot's frame boxes are the frame's locals, i.e. *original inputargs*, so seeding the escape closure from guard snapshots reaches phase 4's "append as an extra inputarg" arm constantly where operand seeding reached it rarely. That widens `inputargs` past what `patch_new_loop_to_load_virtualizable_fields` asserts. The gate treats an original inputarg as a re-emittable leaf only when phase 4 can map it to a pool constant, testing the same `inputarg_consts.get(r.raw() as usize)` expression phase 4 branches on so the two cannot disagree.

The test covers both arms — declined without a pool constant, produced *and* `inputargs.len() == original_boxes.len()` with one. The second arm is what a gate that simply declined everything would fail. Proven non-vacuous: with the gate disabled it fails as `the cut invented an inputarg: len=Some(2), original_boxes=1`.

`test_collections` rc=0, `test_urlparse` rc=0, `test_re` 165 OK, `test_strftime` 4 OK, cut-trace unit tests 9/9.

**Still red, and I do not think it should be blessed by re-recording:** `synth/nested_break_not_hot`, `bridges_compiled 4 -> 0`, `guard_failures 1001 -> 5480` (`loops_compiled` unchanged at 2).

Diagnosed rather than guessed. With `MAJIT_LOG=1 PYRE_FBW_DEBUG_ABORT=1`: **26 cut-declines, every one a `NewWithVtable` root**; `KEPT-SLOT-UNSOURCED` fires 0 times and the extra-inputarg arm 0 times — so neither the kept-slot decline nor the new gate is involved. It is the `is_always_pure()` tightening.

The decisive detail is what those allocations look like:

```
v27 = NewWithVtable(W_IntObject)
SetfieldGc(v27, v17)  W_IntObject.intval   is_immutable: true
```

Re-emitting `v27` alone yields `intval = 0`. So the 4 bridges the old predicate bought were standing on blank `W_IntObject`s — that part is not a regression to mourn.

But the baseline `bridges_compiled=4` was recorded on `main` (`e5eff816849`, #1086), and **`main` has no snapshot seeding at all**. So those bridges do not depend on the unsound admission: on `main` the slot simply becomes `OpRef::NONE` and nothing in this benchmark reads it. The regression is therefore ours, and it is that *cancelling the whole compilation is coarser than the problem* — the value only matters if the resumed frame reads that slot, and here it does not.

The sound repair that recovers them is to replay the initializing stores along with the allocation: every field here is `is_immutable: true`, and pyre's numeric identity is value-based, so `New` + its immutable initializing `SetfieldGc`s reconstructs an equal object. That needs the stores collected, all fields proven immutable, each store's value cone proven re-emittable, and escape analysis — this trace really does contain `SetfieldGc(v41, v42)`, an object stored into another. That last part is exactly where the previously-flagged unsound version came from, so I am not bolting it on.

Leaving this in draft until that is settled.

*Note for whoever picks this up:* #1117 seeds the same closure and carries an equivalent entry-contract gate, but keeps the `has_no_side_effect()` cone predicate and the keep-`NONE`-on-refusal behaviour. Neither PR is merged and `main` has no seeding yet, so the two overlap in the same function.

— *commented by Claude*

youknowone:
## `check.py` finds a bridge regression this design owns

I adopted this PR's three commits onto `fib_recursive` (the seed commit is
byte-identical between the two branches, so only `cancel-on-decline`, the
branch-guard kept-slot decline, and `is_always_pure()` came over) and ran the
gates. Two of three are green, the third is not:

| gate | result |
|---|---|
| `pyre/cpython_tests/run.py --backend dynasm --baseline …` (109 modules) | PASS 108 / FAIL 0 / CRASH 0, no regressions |
| `cargo test --workspace` | 7535 passed, 0 failing suites |
| `pyre/check.py --backend dynasm` | **410 passed, 1 FAILED** |

```
FAIL dynasm synth/nested_break_not_hot  jit-stats change —
  regressed: bridges_compiled 4 -> 0, guard_failures 1001 -> 5480
  (observed loops_compiled=2 bridges_compiled=0)
```

Reproducible 3/3 with identical numbers, so it is not the known `check.py`
non-determinism.

### Attribution

One build carried both arms — an env toggle on an extra guard my branch adds,
plus an env-gated decline-reason log:

| arm | cone predicate | `bridges_compiled` |
|---|---|---|
| A | this PR's commits + my extra guard | 4 → 0 |
| B | `PYRE_CUT_NO_INPUTARG_GUARD=1`, leaving **this PR's predicate alone** | 4 → 0 |

Decline census over the fixture: **26 declines, every one
`impure NewWithVtable`**, none from my guard. So the loss is
`is_always_pure()` refusing `New*` combined with cancel-on-decline. Arm B
reproduces this PR's predicate inside my own build; I did not build this branch,
so read it as "the predicate behaves this way", not as a run of this branch.

The fixture is `nested_break_not_hot`, whose subject is operand-stack
reconstruction when an inner `break` pops the inner iterator and control resumes
at the outer `FOR_ITER` — so the declining values are live, already-initialised
objects on the operand stack.

~~I tried to recover it soundly: admit a `New*` root only when no pre-cut op names
it (no store ran, so the blank object's defaults are the recorded contents; no
second reference exists, so the fresh identity is uncomparable). It admits
**nothing** — all 26 allocations are touched, and the numbers stayed
byte-identical — so I dropped it rather than ship inert machinery.~~

> **Corrected after posting.** The struck paragraph is wrong: "no pre-cut op
> names it" was *my* precondition, and its 0-of-26 admission rate indicts the
> precondition, not the population. Replaying a `New*` root **together with the
> stores that filled it** does recover the fixture — `check.py --backend dynasm`
> returns to 411/411. See
> https://github.com/youknowone/pyre/pull/1122#issuecomment-5231789908.

Note that re-recording the `.jitstats` would bless the regression, so that is not
the answer here.

### Separately: the cone walk skips original inputargs unchecked

```rust
if r.raw() < num_original_inputargs || !seen.insert(r) { continue; }
```

An original inputarg that has no pool constant is not a benign leaf: phase 4's
non-constant arm appends it as an extra inputarg, and
`patch_new_loop_to_load_virtualizable_fields` ends with
`assert i == len(inputargs)` (`compile.py:458`), requiring exactly red args +
static fields + array items. `is_always_pure()` does not cover this — the root
can be a plain `IntAdd` whose cone is replay-safe while the *entry contract*
still cannot represent the value.

This is not hypothetical: PR #1117 hit it as `test.test_collections` and
`test.test_urlparse` `PASS -> CRASH` with
`assert i == len(inputargs) failed (18 != 19)`. Transplanting the cone-walk shape
above into my tree fails the regression test for it with "the cut invented an
inputarg the virtualizable expansion cannot account for" (2 inputargs for 1
original box). The 109-module suite is where those two crashes surface, and this
PR's recorded verification (`test_re`, `test_strftime`, `cargo test --workspace`)
does not include it.

The guard is one branch in the same closure — test the predicate phase 4 branches
on (`inputarg_consts.get(raw)`) and decline when the constant is absent.

— *commented by Claude*

youknowone:
## Correction: the sound recovery is *not* inert — replay the allocation together with its stores

This corrects the last claim of my previous comment
(https://github.com/youknowone/pyre/pull/1122#issuecomment-5231581491), which
said the sound recovery "admits **nothing** … so I dropped it rather than ship
inert machinery". That verdict was a property of *my precondition*, not of the
population — a 0-of-26 admission rate indicts the predicate before it indicts
the subjects, and I read it the wrong way round.

The census that settled it: across all 26 declining `NewWithVtable` roots in
`nested_break_not_hot`, the pre-cut ops naming the allocation were
`[SetfieldGc, GuardClass]` and nothing else. The population is uniform and
nearly trivial. "Untouched" was the wrong precondition; **"filled, and not
published"** is the right one.

### The rule

Admit a `New*` **root** when every pre-cut op that names it is either

* a **guard** (reads it, writes nothing), or
* a `Setfield*` / `Setarrayitem*` / `Setinteriorfield*` whose **`arg(0)`** is the
  allocation — i.e. a store *into* it.

Decline as soon as the allocation appears anywhere else: as a store's *value*
(published into some other object), as an argument to a call, or as an argument
of its own filling store (self-reference). Non-root cone nodes keep
`is_always_pure()` unchanged — the widening is the seed only, and it composes
with, rather than replaces, this PR's predicate.

The admitted stores ride along in the escape set. `op_escaped` is
`sort_by_key(|r| r.raw())`, so they re-emit *after* the allocation in program
order, and their own value operands are pushed back onto the cone stack, so they
are held to the same predicate as any other cone node.

Soundness: at bridge entry the allocation is re-executed (fresh identity) and
then re-filled by exactly the stores that produced its recorded contents, so the
replayed object's fields equal the recorded object's fields. Because no admitted
op publishes it, nothing else holds a reference to the pre-cut object — no
identity comparison and no aliasing write can observe the substitution. The
earlier `is_always_pure()` failure mode (re-executing `NewWithVtable` alone and
getting default fields) is exactly what the store replay closes.

### Numbers

Built on `48f800ec16f` + this PR's three commits + the inputarg-contract guard
from my previous comment + the above:

| gate | with this PR's predicate alone | with the store replay |
|---|---|---|
| `pyre/check.py --backend dynasm` | 410 passed, **1 FAILED** (`nested_break_not_hot`, `bridges_compiled 4 -> 0`) | **ALL PASSED 411 / 411** |
| `pyre/cpython_tests/run.py --backend dynasm` (109 modules) | PASS 108 / FAIL 0 / CRASH 0 | PASS 108 / FAIL 0 / CRASH 0 |
| `cargo test --workspace` | 7535 passed, 0 failing suites | 7537 passed, 0 failing suites |

No `.jitstats` was re-recorded.

### Non-vacuity of the new tests

The first version of the escape test passed even with the
`other.arg(0) == alloc` check neutralised, because the `SetfieldGc(holder,
alloc)` fixture trips the self-reference check first. The committed test also
carries a `CallMayForce(alloc)` shape — neither a guard nor a store, and with no
operand past the receiver — so only the store-shape arm can reject it.
Neutralising that arm now fails with *"an allocation published by a call was
replayed"*.

The commits are on `fib_recursive` (`jit: replay a cut-trace allocation together
with the stores that filled it`, plus `jit: refuse a snapshot seed whose cone
needs an undeliverable inputarg`) and are cherry-pickable onto this branch —
they sit on top of this PR's three commits unmodified.

— *commented by Claude*

youknowone:
Rebased and repaired. Three changes since the last push.

**The branch did not compile.** main's `1a63e36e37a` (list.pop() fold) added a
`collect_outer_active_boxes(...)` call site with 13 arguments; this branch's
kept-slot commit added a 14th parameter. Different files, no overlapping hunks,
so the rebase applied both without a conflict and the result was red for a build
failure — not for anything this PR is about. `origin/gc-decouple` was broken the
same way. Fixed at the `orthodox_list_pop_commit` site, which is a pre-call
snapshot and so reports no unrecovered kept slot.

**The allocation-replay fix is back.** A rebase had dropped it: `is_always_pure()`
alone refuses every `New*` root, and on `synth/nested_break_not_hot` that fired
26 times, taking `bridges_compiled` 4 → 0 and `guard_failures` 1001 → 5480. The
admission rule (replay the allocation together with the stores that filled it;
refuse a store into another object or a call that receives it, both of which
publish the identity) is taken from `a2111b9c41b` + `d3836814c1a` on the
`fib_recursive` branch, which developed it on top of an equivalent commit series.

**One hypothesis refuted and recorded in-code.** Upstream reads the operand stack
out of a single list — `create_top_snapshot(frame, vable_boxes, ...)`
(`opencoder.py:767`) and `live_arg_boxes += self.virtualizable_boxes`
(`pyjitpl.py:2987`) are the same object — which argues for letting the
virtualizable shadow answer for an operand-stack slot even when it reads NULL,
so a post-header snapshot can never leave the merge point's published set.
Measured: it is wrong. `nested_break_not_hot` then raises
`TypeError: comparison on null operand` in `plain_break`, because the traced arm
has already popped a slot the guard's resume target still reads. The comment at
the site records this so it is not retried.

The residue is a real divergence, filed rather than patched: pyre's branch guards
resume at the not-taken **Python bytecode** (`resume_pc = other_target`), while
`opimpl_goto_if_not` uses `resumepc=orgpc` and re-executes the jitcode opcode, so
upstream never reconstructs a popped stack. Closing that needs a blackhole that
re-enters mid-opcode.

## Verification at a29b6d93414

- `cargo test --workspace --no-default-features --features dynasm` — **7085 passed / 1 failed**
- `test_re` 165 OK, `test_strftime` 4 OK, `cargo fmt --all --check` clean
- jitstats sweep over **all 403 committed dynasm baselines**: **zero** differences,
  `nested_break_not_hot` back to 4 / 1001 / 2 / 0 exactly

The one failing test is `jitcode_runtime::tests::list_pop_end_jitcode_resolves_charon_body`
("build-time pipeline must contain the charon `w_list_pop_end_inner` jitcode").
That jitcode arrives with main's `1a63e36e37a` and this run built with
`PYRE_LLBC_STRICT=0` against stale LLBC artefacts, so it is absent locally. It is
not a code failure and CI's fresh extraction should carry it — but I have not
verified that here, so calling it out rather than calling the suite green.

`check.py` has not been re-run locally for the same reason.

— *commented by Claude*

youknowone:
## Rebased onto `f705ef075e1`, plus the review round and the two CPython-suite reds

### Rebase

The branch is now linear on top of `f705ef075e1`, which carries `#1147`
(`type_methods`: read the locale for the `n` presentation type). That is the
commit for `test.test_format::test_locale`, the one red the macOS
`check.py` and `CPython suite (gate)` legs carried **on main as well** — main's
own macOS run at `863e64529ad` failed on that and nothing else. So it should
leave both gates with only this branch's own findings.

### `test.test_str::test_raiseMemError` — fixed

`sys.getsizeof` charged the collector's two-word pre-header on the question
"does the collector own this instance". This branch moved interpreter
allocations into the collector's heap, so every runtime-built leaf value
gained 16 bytes while the same value folded into a code constant did not:
`'a'*42` measured **83 folded / 99 at run time**, `int` 44 against CPython's
28, `float` 40 against 24, `object()` 32 against 16.

The pre-header is a property of the *type*, not of the heap the instance
happened to land in — `_PyType_PreHeaderSize` is
`_PyType_IS_GC(tp) * sizeof(PyGC_Head)` plus the managed-dict term. So
`W_TypeObject` grows `flag_have_gc`, cleared at startup for the leaf value
types (`int`/`long`/`bool`/`float`/`complex`/`str`/`bytes`/`bytearray`/
`NoneType`/`NotImplementedType`/`ellipsis`/`range`/`object`), and `get_sizeof`
reads it. Heap types keep the default set, because `type_new` sets the flag
unconditionally.

Measured against CPython 3.14 on the same script, every subject now agrees
exactly:

| | folded str | runtime str | int | small int | bytes | float | `object()` | `[]` | `[1]` | `()` | `{}` | `C()` | `S()` |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| pyre | 83 | 83 | 36 | 28 | 37 | 24 | 16 | 56 | 64 | 48 | 64 | 48 | 40 |
| CPython | 83 | 83 | 36 | 28 | 37 | 24 | 16 | 56 | 64 | 48 | 64 | 48 | 40 |

`cpython_tests/run.py --filter test_str` is back to 7 PASS / 0 FAIL.

### `test.test_descr::test_slots` — baselined, with the reason recorded

The failing assertion is the lookup-leak check at `test_descr.py:1283-1288`:
it brackets a loop with `len(gc.get_objects())` and requires the count not to
move. pyre's `gc.get_objects()` enumerates every app-level object the
collector holds — `referents.py:117-126` filters only through
`try_cast_gcref_to_w_root`, so ints are in — while CPython's returns only
GC-tracked containers. Binding a fresh int inside the loop therefore moves the
count by construction.

Controls: an empty loop gives the same delta, back-to-back calls differ by one,
five calls climb by one each, `PYRE_GC_INTERP=0` is flat, and PyPy behaves the
same. It is the upstream-faithful answer to a test CPython's own `get_objects`
contract lets pass, so it is recorded as `FAIL` in `baseline.json` with that
reason in the entry rather than worked around.

### Review round

**Fixed**

* `majit-backend-wasm/src/lib.rs` — record the compiled block *after* the
  host-rejection decline, matching `compile_bridge`.
* `majit-gc/src/lib.rs` — **both** `collect_step` fallbacks. `is_done__states`
  (`rgc.py:52-60`) needs `oldstate != 0 && newstate == 0`, so `(SCANNING,
  SCANNING)` never completes. Both now report `_encode_states(1, 0)` —
  `MARKING → SCANNING` — which is literally what `rgc.py:20-31` returns, rather
  than the `SWEEPING → SCANNING` the review suggested.
* `pyjitpl.rs` — `retire_speculative_cut_key` extracted, with
  `only_a_compiled_outcome_keeps_the_cut_ownership_marker` driving
  `compile_loop` to `Cancelled`/`Aborted` and asserting the key is gone.
* `executioncontext.rs` — `debug_assert_eq!` that a nonperiodic action
  registers once.
* `gc/mod.rs:932` — `w_int_new(total_gc_time)` is built and pinned before the
  stats slot is read.
* `gc_interp.rs` — `would_collect()` now asks `gc.isenabled()`. The safepoint's
  collection is `do_collect_oldgen_nonmoving`, which runs a full cycle whatever
  the flag says, so without this read a `gc.disable()` region was still
  collected; `incminimark.py:831-832` is the upstream early return, and only an
  explicit `gc.collect()` carries `force_enabled` past it. Measured with the
  app-level hooks as the counter, 4000 churn iterations either side:
  `disabled: {}` — no minor, no collect_step, no collect — against
  `enabled: {on_gc_collect_step: 2, on_gc_minor: 1, on_gc_collect: 2}`.
* `gc/mod.rs` getattribute — the storage-slot rule is now one shared
  `is_hidden_stat_slot`, and it stops hiding dunders. `W_GcCollectStepStats` is
  an ordinary `TypeDef` object upstream, so `__class__`/`__repr__` answer as on
  any other; `__dict__` stays hidden because these objects have none.
* `collector.rs` — `do_get_referents`' walk terminator re-spelled the app-level
  predicate inline and dropped the hide bit, so an internal node with a
  Python-object prefix was *reported* instead of being looked through. It now
  calls `is_app_level_object_ref`, which also gained the dummy rejection that
  opens `try_cast_gcref_to_w_root` (`referents.py:18`). Covered by
  `get_referents_looks_through_a_hidden_or_dummy_node`, checked non-vacuous:
  with the two rejections neutralised the test reports the hidden node instead
  of the leaf behind it.

**Won't fix, with the upstream line**

* `gc/hook.rs:73` — re-check `w_callable` for `None` before calling. Upstream
  does not: `hook.py:171` is a bare
  `self.space.call_function(self.w_callable, w_stats)`. It is safe there
  because the GC only reaches the hook through `is_gc_minor_enabled()`
  (`hook.py:28`), which reads the same flag the setter writes. pyre has the
  same gate.
* `gc/mod.rs:74-85 wrap_rooted` — root `target` across the allocation.
  `malloc_typed_managed` (`lltype.rs:303-327`) allocates through
  `try_gc_alloc`, the **non**-collecting entry point, so nothing can move
  between the `shadow_stack_get` read and the store. The old-gen fallback that
  path can take already applies `try_gc_write_barrier_managed` on the line
  below.

### Still open

`synth/str_fstring` dynasm reports `guard_failures` 659 on macOS where
ubuntu-24.04 and windows-latest both report 658, on all three runs of this
branch — the same knife edge `check.py:518-530` documents for
`recursive_call_frame_relocation` (638 preamble / 637 body / 636 no crossing).
Deciding between a `.darwin.jitstats` overlay and a real fix needs the value
read back from this host first, so it is not re-recorded in this push.

— *commented by Claude*

PR #1124, opened 2026-08-09
majit: add a StaticLength arm to the RangeTo slice-index fold

youknowone:
## What this changes

Adds a `StaticLength` arm to the RangeTo slice-index fold, beside the existing `MinusOne` arm. A residual `&receiver[..end]` is rewritten only when the receiver's length is fixed by `__array_repeat`'s `ConstInt` count, a comparison dominating the site proves the bound on the proving edge, and `end` matches the comparison operand as the same `ArrayLen` value.

## Census result

Three-stream census A/B at the PR head `c78b1c6ecd4` (arm = HEAD, base = HEAD~1 for the two touched files), both arms real builds, per-arm LLBC extraction, HEAD stable across both:

| stream | base | arm | newly good |
|---|---|---|---|
| phaseA | 1677 | 1677 | 0 |
| phaseB | 11 | 10 | **1 — `pyre_interpreter::display::<Impl>::push_onto`** |
| skip | 1678 | 1678 | 0 |

(unique subjects, not raw lines) — `NEWLY BAD 0/0/0`, VERDICT CLEAN.

**Determinism control:** the same code was censused twice back to back; the subject sets were identical across all three streams (0 movement). So the single-subject delta is a real effect of this change, not run-to-run variation.

Please still calibrate expectations: this clears **one** rtyper subject. It is not a broad coverage win.

## The one real corpus site declines, and that is deliberate

`call_function_impl_result` in `pyre/pyre-interpreter/src/call.rs` is **not** rewritten. That site genuinely is in bounds — its `args: &[PyObjectRef]` is a shared reference and so cannot change length between the two `ArrayLen` reads — but the frontend has no shared-reference/immutability notion with which to prove it, so the fold correctly refuses. The anchor test is named `call_function_impl_result_declines_residual_array_index` and its doc comment says exactly this.

## Why the rule is shaped the way it is

An earlier revision of this fold rewrote that site. A three-lens adversarial review then proved **by execution** that it did so only through two unsound admissions:

1. The stability scan window ran `[proving_edge .. site_op)`, but the interval that must be free of length-changing operations is `[earlier ArrayLen .. later ArrayLen]`. In any acyclic diamond the comparison block is reachable backward from the site but never forward from the proving edge, so a mutation between the compared length read and the branch was never scanned. Moving a byte-identical mutating call from the site block into the entry block flipped decline → admit.
2. `may_reference_base` returned early for any operation that did not *syntactically* name the base, so a mutation reached through a second handle — an owner object, or a base published by an allowlisted `ArrayWrite` and reloaded — was invisible even inside the window.

Both emit `getslice(buf, 0, 9)` on a length-8 array, and `ll_listslice_startstop` **clamps** rather than panicking, so the failure mode is a silent wrong-length slice rather than a crash.

This PR fixes both. The window now spans the two `ArrayLen` definitions ordered by dominance (unorderable reads decline); inside it, a call outside the read-only allowlist declines regardless of its operands; and the base declines if it was published into a `FieldWrite`/`ArrayWrite` value slot or passed to a non-allowlisted call before the later read.

## A latent miscompile fix to code already on main

`range_feeds_only_index` now requires **exactly one** `end` `FieldWrite` on the range value before either arm may substitute the captured operand.

This tightens the **`MinusOne` arm already on main**, whose guard is

```rust
let is_construction_write =
    matches!(&op.kind, OpKind::FieldWrite { base, .. } if base == range_result);
```

— keyed only on the base, with no field-name check and no uniqueness count. A later `range.end = m` therefore passes the consumer gate while the ctor-time value is the one planted, and the write becomes dead. Census-invisible, because it is a guard rather than a coverage lever, but real.

## Verification

- `cargo test --release -p majit-translate --lib` — 3167 passed, 0 failed, 33 ignored
- `front::slice_index` module — 31 passed, 0 failed
- anchor (`--ignored call_function_impl_result`) — 2 passed, 0 failed
- `cargo check -p majit-translate --all-targets`, `cargo fmt --check` — clean

Each of the four new negative regressions was verified by mutation: the guard was reverted, the test observed to fail, and the guard restored.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -24214,14 +24214,13 @@ mod tests {
         );
     }
 
-    /// Array slicing keeps the residual RangeTo path because a general stop has
-    /// no proof that `end <= slice.len()`.  Both halves are asserted — the
-    /// residual call is present AND no `__getslice_rangeto` marker was planted
-    /// — so a lowering change that drops the call for an unrelated reason is a
-    /// failure rather than a pass.
+    /// The real site is in bounds: its `args: &[PyObjectRef]` is shared and
+    /// therefore cannot change length between the two `ArrayLen` reads. The
+    /// frontend has no shared-reference/immutability notion with which to
+    /// prove that fact, however, so the sound static-length fold declines.
     #[test]
     #[ignore]
-    fn call_function_impl_result_keeps_residual_array_index() {
+    fn call_function_impl_result_declines_residual_array_index() {
         use crate::model::OpKind;
         let path = concat!(
             env!("CARGO_MANIFEST_DIR"),
@@ -24247,14 +24246,15 @@ mod tests {
                 })
                 .count()
         };
-        assert!(
-            calls_path(&["core", "array", "<Impl>", "index"]) >= 1,
-            "general RangeTo array index remains residual"
+        assert_eq!(
+            calls_path(&["core", "array", "<Impl>", "index"]),
+            1,
+            "unproved RangeTo array index remains residual"
         );
         assert_eq!(
             calls_path(&["__getslice_rangeto"]),
             0,
-            "no general RangeTo site is rewritten — the fold is declined"
+            "no RangeTo getslice call is planted without an immutability proof"
         );
     }
 }
--- a/majit/majit-translate/src/front/slice_index.rs
+++ b/majit/majit-translate/src/front/slice_index.rs
@@ -131,7 +131,8 @@ pub(crate) fn rewire_slice_index_rangefrom_sites(
 }
 
 /// Rewrite captured `RangeTo { end }` indexes through the synthetic-call
-/// channel. Only the MinusOne shape has a proven stop bound.
+/// channel. The stop must either be the receiver's `len - 1`, or be bounded
+/// by a dominating branch against the receiver's static array length.
 pub(crate) fn rewire_slice_index_rangeto_sites(
     graph: &mut FunctionGraph,
     sites: &[SliceIndexRangeToSite],
@@ -153,6 +154,7 @@ pub(crate) fn rewire_slice_index_rangeto_sites(
 enum SliceIndexBounds<'a> {
     RangeFrom { start: &'a Variable },
     MinusOne { end: &'a Variable },
+    StaticLength { end: &'a Variable },
 }
 
 fn rewire_one_slice_index_rangeto_site(
@@ -174,6 +176,8 @@ fn rewire_one_slice_index_rangeto_site(
         .ok_or_else(|| format!("{}: no RangeTo index consumer", graph.name))?;
     let bounds = if minus_one_end_matches(graph, &site.end, &slice) {
         SliceIndexBounds::MinusOne { end: &site.end }
+    } else if rangeto_static_length_bound_matches(graph, &site.end, &slice, &range) {
+        SliceIndexBounds::StaticLength { end: &site.end }
     } else {
         return Err(format!(
             "{}: RangeTo stop has no proof that end <= slice length — declining",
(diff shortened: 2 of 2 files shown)


youknowone:
**Correction to the original PR description.**

The description as first posted stated the census yield of this PR was **zero**. That figure was measured at `eba36d13ed9` and was correct there, but it is **not** correct for the code as it will land. The branch was subsequently rebased onto commits that touch `majit-translate` (including `front/mir.rs`), and re-measuring at the PR head gives a different answer:

| stream | base | arm | newly good |
|---|---|---|---|
| phaseA | 1677 | 1677 | 0 |
| phaseB | 11 | 10 | **1 — `pyre_interpreter::display::<Impl>::push_onto`** |
| skip | 1678 | 1678 | 0 |

`NEWLY BAD 0/0/0`, VERDICT CLEAN. Unique subjects, not raw lines.

I re-ran the same code twice as a determinism control before believing the delta: the subject sets were identical across all three streams, so this is a real effect of the change rather than run-to-run variation.

The description has been updated. The earlier zero was a stale measurement on my part, not a property of the change.

— *commented by Claude*

youknowone:
## CI triage — none of the three red checks is caused by this PR

Control used throughout is the **parent-sha run** (`0c3c4777ba9`, run `31303981088`), not the merge base.

| check | parent `0c3c4777ba9` | this PR `c78b1c6ecd4` | verdict |
|---|---|---|---|
| CPython suite (gate) | fail | fail | pre-existing |
| pyre/check.py (macos-latest) | fail | fail | pre-existing |
| pyre/check.py (ubuntu-24.04) | **pass** | **fail** | false red — see below |

### CPython suite (gate) and check.py (macOS)

Byte-for-byte the same failures on both commits:

```
test.test_pickletools: PASS -> FAIL  rc=1  FAILED (errors=1,   skipped=13)  ERROR: test_ints
test.test_re:          PASS -> FAIL  rc=1  FAILED (errors=303, skipped=13)
```

macOS `check.py` additionally reports `dynasm 1 failed, 410 passed` on both. Identical error and skip counts on parent and head.

### check.py (ubuntu) — a perf gate whose denominator halved

This one did flip from pass to fail, so it deserved a real look. The failing rows are a single fixture out of 410, on both backends at once:

```
FAIL dynasm    synth/mapdict_frozen_unboxing_fold  exec 1.30s > pypy 0.02s  ratio 64.9x > gate 63x
FAIL cranelift synth/mapdict_frozen_unboxing_fold  exec 1.52s > pypy 0.02s  ratio 70.8x > gate 63x
```

Comparing the absolute columns against the parent run:

| | cpython | **pypy** | dynasm | cranelift | third |
|---|---|---|---|---|---|
| parent (passed) | 0.24s | **0.06s** | 1.38s → 28.5x | 1.57s → 32.5x | 2.35s → 50.4x |
| this PR (failed) | 0.22s | **0.03s** | 1.37s → 68.8x | 1.60s → 80.6x | 2.26s → 117.3x |

**pyre's own execution times did not regress** — dynasm 1.38s → 1.37s, the third column 2.35s → 2.26s, cranelift 1.57s → 1.60s (+1.9%, within noise). What moved is the **pypy denominator, 0.06s → 0.03s**, and all three ratios rose by the same ~2.2× factor as a result.

Three ratios moving together by an identical factor while the numerators stay flat is a denominator move, not a codegen change — a codegen regression cannot lift every backend's ratio by the same multiple while leaving every backend's absolute time unchanged. At 0.02–0.03s the pypy measurement is at its resolution floor, where a few milliseconds swings the ratio past a 63x gate that this row was already sitting under by a small margin.

This is the known denominator-collapse class in the synth perf gate, not a property of this change.

— *commented by Claude*

youknowone:
Follow-up on the parity review: #1133.

**§2 (`resolve_block_alias` dropped non-variable incoming arguments)** — real, and fixed there. `filter_map` discarded a non-variable incoming argument instead of treating it as a disagreement, so `[Variable(v), Constant(7)]` resolved to `v`'s root. Collecting into `Option<Vec<Variable>>` short-circuits on the first one. Links out of blocks unreachable from `startblock` cannot execute and are excluded from the walk rather than allowed to veto a resolution.

**§1 (the `end` write is matched as `field.name == "end"`)** — the code reading is right, but this PR did not narrow a rewrite the older code performed. `is_construction_write` has never keyed on the field name and still does not; the `end_writes` gate only requires the write to be unique and named. This PR also added `resolve_block_alias`, which widened which sites are recognised at all — so accepting `__pos_0` would admit sites only the new alias resolution makes visible, not restore an earlier rewrite. It was implemented and measured, cleared no additional prepass subject, and is not included; a test records the decline with that reasoning.

**§3 (dominance/reachability sets vs. the `decompose_slice_args()` contract)** — not addressed; it is an objection to the fold's design rather than to these two defects.

One measurement note for anyone reading the census numbers in this PR's history: `pyre_interpreter::display::<Impl>::push_onto` is **not stable** in the three-stream prepass census — on a byte-identical tree it scored phaseB 15 once and 14 eight times across independent runs. Single-observation 1-subject deltas on that row are not evidence in either direction.

— *commented by Claude*

PR #1127, opened 2026-08-09
jit: coerce a closing JUMP to the label arity only when it targets that label

youknowone:
`assemble_peeled_trace_with_jump_args` applied `compile.py:334`'s equal-arity rule
to every close whose `jump_to_self` flag was set. That flag records that the
closing JUMP was **redirected onto some existing target token**, not that it
targets the local label, so a cross-token JUMP was truncated to the local label's
arity — 8 args to 7, dropping a trailing `ConstInt(1)` that pypy spends on a
literal in the same slot.

This renames the flag to `jump_was_redirected` and compares the JUMP's target
descr against the local label's before coercing. A JUMP carrying another token's
virtual-state arity (`unroll.py:346-357`) now passes through unchanged;
`compile.py:334` asserts equality for the same-label close only.

Also corrects the comment at the `jump_to_existing_trace` discard site, which
attributed the fallback to cranelift's one-function-per-trace layout. That was
measured false: admitting the external close breaks dynasm too, which relocates
every LABEL and refuses a JUMP below the first page, so the branch does name a
live address.

## Reachability

The change is a **no-op on today's corpus, measured rather than argued**. Every
bench script (394 synth + 23 top-level = 417) was swept with the release binary
under `MAJIT_LOG=1`, tallying both sides of the new predicate:

| `redirected` / `targets_local` | occurrences |
|---|---|
| true / true | 543 |
| **true / false** (the only branch this changes) | **0** |
| false / true | 0 |
| false / false | 18 |

Every redirected close targets its local label, because the discard at
`jump_to_existing_trace` removes an external close before assembly ever sees it.
The fix becomes load-bearing when that discard is lifted.

## Fixture

`pyre/bench/synth/retrace_outer_loop_type_flip.py` — a nested `while` whose
accumulator flips `int -> float` in the **outer** loop, so the retrace grown for
it closes onto the token the inner loop left behind. Recorded at
`loops_compiled=1 loops_aborted=2 retraces_compiled=0 guard_failures=590`,
identical on all three backends. That is pyre's current *declined* state, and it
is the acceptance test for whatever admits the close.

## Verification

- `cargo test -p majit-metainterp --no-default-features --features dynasm` — 1457 passed
- `cargo fmt` clean
- `pyre/check.py --backend dynasm` at base `1a63e36e37a` — full bench corpus (411) with zero failures
- Earlier at base `b925c3e5ead`: dynasm 410/411, cranelift 410/411, wasm 403/407, where all 6 failures were the #1086 baseline reds since restored by #1104

New regression test `test_assemble_peeled_trace_keeps_external_target_jump_arity`
pins the pass-through; the existing
`test_assemble_peeled_trace_passes_through_resolved_body_inputs` was corrected to
model the same-label close it was actually exercising.

— *authored by Claude*

### DIFF
--- a/majit/majit-metainterp/src/optimizeopt/unroll.rs
+++ b/majit/majit-metainterp/src/optimizeopt/unroll.rs
@@ -1551,7 +1551,7 @@ impl UnrollOptimizer {
         }
         let mut body_ops = p2_ops;
         let mut redirected_tail_ops = Vec::new();
-        let jump_to_self = {
+        let jump_was_redirected = {
             let body_jump_args: Vec<OpRef> = body_terminal_op
                 .as_ref()
                 .map(|jump| jump.getarglist().iter().map(|a| a.to_opref()).collect())
@@ -1779,12 +1779,10 @@ impl UnrollOptimizer {
                 // Only take jump_ctx ops if we don't already have
                 // a self-loop Jump from the retrace path.
                 redirected_tail_ops = std::mem::take(&mut jump_ctx.new_operations);
-                // Check if the redirected Jump targets the current body token
-                // (last in target_tokens) or an external token from a previous
-                // compilation.  The Cranelift backend compiles each trace as a
-                // single function — cross-function jumps to external target
-                // tokens are not supported.  Discard the redirected tail and
-                // restore the body's original self-loop Jump instead.
+                // Check whether the redirected Jump targets the current body
+                // token (last in target_tokens) or an external token from a
+                // previous compilation. An external close is discarded and the
+                // body's original self-loop Jump restored instead.
                 let current_body_descr_idx = self
                     .target_tokens
                     .last()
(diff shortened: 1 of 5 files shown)


youknowone:
Rebased onto `23aaac5c09e`. The commit's patch is byte-identical to the reviewed one.

**Where the verification stands.** The full dynasm bench corpus (411 benches, zero
failures) and the 417-script reachability sweep were run at base `1a63e36e37a`.
The branch has since been rebased twice by its author; `241a97fa4bc` (#1124)
touches `majit-translate`, which invalidates `pyre-jit.ullbc`, so the local
binaries no longer describe this base. CI is the authority for `23aaac5c09e`.
Nothing in either base bump touches `optimizeopt/`, the dynasm backend, or
`majit-ir`, so the reachability result — no corpus bench takes the branch this
change alters — is not expected to move.

**Follow-up, not in this PR.** The `jump_to_existing_trace` discard this PR only
re-comments is still there, and the root cause of why it cannot simply be lifted
is now pinned: upstream's foreign close never crosses a `JitCellToken` —
`compile.py:287-289` resets `jitcell_token.target_tokens` to `[start_descr]` and
`start_descr.virtual_state is None`, so `unroll.py:177-179` skips it and the close
in `compile_loop` is structurally always a self-close. A foreign token is
reachable only via `compile.py:355` (`compile_retrace` →
`get_procedure_token(greenkey)`, which filters `invalidated` at
`warmstate.py:191-196`) and via `optimize_bridge`, and the result is installed as
a bridge under that same token (`compile.py:797-811`).

pyre instead seeds `unroll_opt.target_tokens` in **both** `compile_loop` and
`compile_retrace` from a green-key side table that survives recompiles, so the
matched token can belong to a previous compilation's `ExecutableBuffer`; dynasm
bakes its `ll_loop_code()` as an absolute `br`, and the blanket
`set_original_jitcell_token_number` re-stamp makes the port of `compile.py:197`
compare equal and skip the `record_jump_to` keepalive, after which `memmgr` is
free to evict the target.

That predicts the split seen when the discard was lifted: the five benches that
broke are recompile-at-the-same-key shapes that cannot reach `compile_retrace` at
all (`retrace_limit` defaults to 0, `rpython/rlib/jit.py:595`, and none of them
calls `pypyjit.set_param`), while `synth/retrace_outer_loop_type_flip` sets
`retrace_limit=5` and needs the close only in the retrace-attach path, bound to
the same CLT that owns the target LABEL. The fix is to make the discard test the
real invariant — same artifact, and target live — rather than "is this the body
token I just minted".

— *commented by Claude*

youknowone:
**CI triage — the `pyre/check.py (macos-latest)` red is inherited.**

Both failures are in the `cpython-suite` phase; the bench corpus phase is clean:

- `test.test_pickletools` — `errors=1`, `test_ints`
- `test.test_re` — `errors=303`

main's own run at this PR's exact base `23aaac5c09e`
([run 31316170676](https://github.com/youknowone/pyre/actions/runs/31316170676))
fails the same job with the **same two tests and the same counts**, and its
`CPython suite (gate)` and all three `pyre/check.py` jobs are red there too. So
this branch adds nothing: `test_re` is the documented JIT-only lost-argument bug
and `test_pickletools` came in with #1087.

— *commented by Claude*

PR #1132, opened 2026-08-09
sys: root the audit emit's livevars, and land the two _getframe commits #1112 merged without

youknowone:
Follow-up to #1112, rebased onto current main.

#1112 was squash-merged at `d2f4ce1522e`, which predates part of the branch, so
one commit here is work that PR was supposed to carry and did not. The rest
answer the review comments the merged PR collected.

### Not in main despite #1112

* **`sys: force the frame `_getframemodulename` reads instead of the stack top`** —
  main still forces `topframeref` at the walk START and reads `w_globals` off the
  frame the walk ENDS at without forcing it. `w_globals` is one of the six fields
  `interp_jit.py:25-30` declares virtualizable, so the force belongs at the
  consumer; forcing the walk escapes the traced virtualizable and
  `vable_after_residual_call` aborts with ABORT_ESCAPE. Measured on the
  `getframe_*` corpus: vable-escape aborts 5 → 0 for depth ≥ 1.

The branch's second unmerged commit — unwrap and rebuild `sys.audit`'s event
rather than forwarding the caller's object — has been **dropped**: #1113 landed
the same round trip meanwhile (`str_utf8_w` + `w_str_new`). See the note at the
bottom for the one thing that commit did differently.

### Review answers

* **`sys: root the audit emit's livevars ...`** — Codex P1 (reload pinned audit
  objects after collection points) and CodeRabbit's two `vm.rs` findings. Every
  value the audit path needs after an allocation now comes back through its
  shadow-stack slot; `trigger_audit_events` publishes the event, the arguments
  and the hook set in one phase and normalizes once, rather than `pin_root` per
  value (`gc_roots::pin_roots` documents why the per-value form reopens the
  window it exists to close). `audit` and `sys_audit` root their arguments in
  front of the event wrap, `getframe` roots the frame it returns across the
  emit, and `sys_addaudithook` pins before the `sys.addaudithook` event rather
  than after. The commit also moves the `__cantrace__` lookup to
  `findattr_result`: `space.findattr` (`baseobjspace.py:881-888`) answers `None`
  for any non-async error, where the bare `findattr` panicked.
* **`jit: bump the inline-subwalk close counter below the driver lookup`** —
  CodeRabbit on `state.rs`.
* **`bench: note which audit-hook arms are unreachable ...`** — CodeRabbit's three
  fixture comments. The requested extra coverage for the propagating arm of the
  refusal predicate is **not** added, and the header now records why it cannot
  be: a hook is installed for the life of the interpreter and the first one that
  raises masks every hook behind it, and upstream's own facility for clearing the
  set between cases (`__pypy__._testing_clear_audithooks`,
  `interp_magic.py:292`) refuses to run once translated.
* **`bench: retarget the audit-hook note at `error_is_exception``** — the same
  header cited `error_is_runtime_error` and `RuntimeError`, both of which #1126
  replaced (`Exception`, renamed predicate).

### One review finding deliberately not acted on

Codex P1 on `residual_call.rs:1356` asks for `mark_escape_flush_undo_pending()`
back in the locals-only escape-flush arm. That arm was removed by
`jit: drop the locals-region leg's deferred escape-flush undo arm` on measurement:
the `value-stack underflow` it was added for came from an A/B run on stale LLBC,
and re-running the reproducing arm on artefacts that pass `PYRE_LLBC_STRICT=1`
gives 0/10 panics on cranelift with the whole synthetic suite and every recorded
counter unmoved. Restoring the call would revert an executed result on a static
argument, so this PR fixes the **comment** instead — it still claimed the capture
was consumed by a legacy replay, which is what stopped being true.

### One open question, left as main has it

`sys_audit` unwraps with `str_utf8_w`, and #1113's comment states that choice is
deliberate: "an event name holding a lone surrogate is a `UnicodeEncodeError` at
the call". The dropped commit used the surrogate-tolerant read instead. Measured:

```
pypy3 7.3.22    sys.audit("\ud800bad")  → accepted, the hook receives '\ud800bad'
CPython 3.14.6  sys.audit("\ud800bad")  → UnicodeEncodeError
```

Upstream's spec is `@unwrap_spec(event="text")`, i.e. `space.text_w`, and the
PyPy oracle accepts the name. This PR does **not** change it — reversing a
documented, deliberate decision is not something a conflict resolution should
do silently. Raising it here so the call is made on the evidence rather than by
whichever branch merged last.

### Verification

`PYRE_LLBC_STRICT=1 python3 pyre/check.py --synthetic-only`:

```
FAILED: dynasm 1 failed, 394 passed
FAILED: cranelift 1 failed, 394 passed
ALL PASSED: wasm 394/394
```

The single failure is `synth/list_pop_append`, an inherited base red: the CI run
for #1119's own merge commit (`1a63e36e37a`, run `31313268003`, `headSha` equal)
already reads `FAIL dynasm 59.8x > gate 22x` / `FAIL cranelift 68.9x > gate 22x`
on the fixture that same commit introduced the 22x ceiling for, and every main
run since carries it.

Output parity against `pypy3` re-checked for `sys_audit_hooks`,
`getframe_audit_event`, `getframe_inline_subwalk_multiframe` and
`getframe_while_escaping_read_frame_identity`. `cargo fmt --check` clean.

### DIFF
--- a/pyre/bench/synth/getframe_inline_subwalk_multiframe.py
+++ b/pyre/bench/synth/getframe_inline_subwalk_multiframe.py
@@ -54,7 +54,7 @@ import sys
 
 
 def leaf(x):
-    sys._getframe(0).f_locals
+    sys._getframe(0).f_locals  # noqa: B018 — the read itself is the force under test
     return sys._getframe(2).f_locals["base"] + x
 
 
--- a/pyre/bench/synth/getframe_while_escaping_read_frame_identity.py
+++ b/pyre/bench/synth/getframe_while_escaping_read_frame_identity.py
@@ -34,8 +34,8 @@
 # Both reads still escape — `_gf()` names `leaf_a`'s published frame and
 # `_gf(1)` names `part_b`'s portal — but only once per call now that
 # `sys._getframe` forces the frame it RETURNS and not also the top of the stack.
-# The multi-frame adopts this file exists for are unmoved at 10; what the
-# duplicate escape carried was the single-frame count, 5 -> 0, alongside
+# The multi-frame adoption count this file exists for is unmoved at 10.  What
+# the duplicate escape carried was the single-frame count, 5 -> 0, alongside
 # `part_a`'s loop compiling.
 import sys
 
--- a/pyre/bench/synth/sys_audit_hooks.py
+++ b/pyre/bench/synth/sys_audit_hooks.py
@@ -10,16 +10,18 @@
 #   * `sys.audit` is free when no hook is installed, so the count a hook sees
 #     starts at the first `addaudithook`, not at interpreter start;
 #   * installing a hook emits `sys.addaudithook` to the hooks already there,
-#     and a `RuntimeError` out of that event means those hooks REFUSED the new
-#     one: it is dropped and the refusal does not propagate.  Any other
-#     exception does propagate -- not exercised here, and it cannot be: a hook
-#     is installed for the life of the interpreter and the first one that
-#     raises masks every hook behind it, so reaching the propagating branch
+#     and an `Exception` out of that event means those hooks REFUSED the new
+#     one: it is dropped and the refusal does not propagate.  A `BaseException`
+#     outside `Exception` does propagate -- not exercised here, and it cannot
+#     be: a hook is installed for the life of the interpreter and the first one
+#     that raises masks every hook behind it, so reaching the propagating branch
 #     needs the hook set cleared between cases.  Upstream's own facility for
 #     that (`__pypy__._testing_clear_audithooks`, `interp_magic.py:292`) refuses
 #     to run once translated, so no app-level program on a built interpreter
-#     can get there.  Same for the RuntimeError-SUBCLASS arm of
-#     `error_is_runtime_error`.
+#     can get there.  The refusal below raises from app code, so it carries a
+#     real exception object and takes `error_is_exception`'s isinstance arm;
+#     the `PyErrorKind` fallback behind it answers only for an error that never
+#     materialised one, which nothing app-level can hand this event.
 #   * `__cantrace__` on a hook is honoured (the flag exists so a tracing hook
 #     can opt back in); nothing here can observe the tracing state, so only the
 #     attribute lookup path is exercised.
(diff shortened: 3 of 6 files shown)


youknowone:
CI attribution for this PR — every failing row is inherited from main or a measurement artifact, verified against `main @ eba122397df` (run [31346388732](https://github.com/youknowone/pyre/actions/runs/31346388732)), the same sha this branch sits on.

| job | main @ eba122397df | this PR | verdict |
|---|---|---|---|
| `pyre/check.py (ubuntu)` | `FAIL wasm synth/unary_negative jit-stats change — improved: guard_failures 13 -> 2` | byte-identical row | inherited |
| `pyre/check.py (windows)` | synth 414/414 pass; job fails on posix parity (`access() requires 2 arguments`, `utime(bytearray)`, `No module named 'posix'`) | identical | inherited |
| `pyre/check.py (macos)` | pass | `synth/list_setslice` 0.1x < floor 0.2x | measurement artifact, see below |
| `cargo test (ubuntu)` | pass | passed on re-run (was a `memory allocation of 8 bytes failed` SIGABRT in `pyre-object --lib`, a crate this branch does not touch) | runner flake |
| `CPython suite (gate)` | pass | pass | — |

**`unary_negative`** — the recorded baselines are `dynasm=2`, `cranelift=2`, `wasm=13`; the new reading of 2 makes wasm agree with the natives. It reproduces on main itself, so the re-record belongs with whichever change moved it, not here. A local A/B pins the window: the same branch code read `ALL PASSED: wasm 394/394` on base `1b31a5b3060` and produces this row on `eba122397df`, same host — the only variable is main.

**`list_setslice`** — read the two runs' own startup lines:

```
main  macOS   cpython=0.016s pypy=0.017s dynasm=0.056s cranelift=0.053s   -> 0.4x / 0.5x, pass
this  macOS   cpython=0.020s pypy=0.021s dynasm=0.080s cranelift=0.070s   -> 0.1x / 0.1x, fail
```

The whole host was ~20% slower and dynasm startup went 0.056s -> 0.080s, while the fixture's total stayed 0.07-0.08s on both. `exec = total - startup` therefore goes negative and clamps to the 0.01s floor, which is the entire 0.1x. Locally the same fixture reads 0.5-0.6x, 3/3. This is the floor side of the denominator problem #1121 buffered on the ceiling side.

No jitstats or ceiling has been re-recorded in this PR.

— *commented by Claude*

PR #1138, opened 2026-08-10
jit: a dead-var link-arg trim scoped on the wrong reachability, a shared-receiver locals_w_mut!, bridge opt-fuel, and 91 stale CPython-suite baseline entries

youknowone:
Three independent changes that came out of chasing a bridge-path wrong-code
failure. The failure itself turned out to be #1130, already merged — this
branch is rebased onto it and carries only what the hunt produced on its own
merits.

## `prune_dead_phis` trimmed a link the way its own reachability walk saw it

`prune_dead_phis` (`majit/majit-translate/src/model.rs`, the port of
`simplify.py:425-479 transform_dead_op_vars`) drops a dead `block.inputargs[i]`
from every **reachable** block in Step 7, and trimmed the matching
`Link.args[i]` in Step 6 only for links whose **source** block was reachable.

Those two scopes are not the same set. A block the reachability walk excludes
can still name a reachable block as its link target — an orphan
`eliminate_empty_blocks` bypassed, or one of the merge blocks `jtransform`
leaves whose inputargs are phi targets with no backing `OpKind::Input`. Its
link kept the full argument list while the target lost a column.

Nothing asserts `len(link.args) == len(link.target.inputargs)`, so the damage
surfaces one pass later. `remove_duplicate_inputargs` reads each phi column by
index across every incoming link, so the untrimmed link contributes its value
one slot over; the union-find then merges two variables that hold different
values, and the resulting rename is applied to every block in the graph.

Keying the trim on the **target**'s reachability closes the gap. Reachability
is closed under exits, so this only ever adds links to the trim — it can never
drop one that was previously trimmed.

Upstream has no equivalent gap: `graph.iterblocks()` *is* its block list, so an
unreachable block is not in `blocks` and there is no link for it to skip.

A regression test builds the shape directly — entry and an orphan both jumping
to a merge whose inputargs are `[dead, live]` — and fails on the previous code
with the surviving column reading back as the orphan's value.

## `locals_w_mut!` accepted a shared receiver

`locals_cells_stack_w` is a raw pointer, so it reads fine through a shared
`&PyFrame`. The macro that replaced the old `&mut self` accessor therefore let
a safe caller holding a shared frame mint `&mut` to the array. `&mut *` on the
receiver restores the requirement the accessor had.

The result's lifetime stays unconstrained — `&mut *ptr` has none to bind — so
overlapping calls still alias invisibly to the borrow checker. Binding it needs
a function signature, which is the accessor form whose separate graph defeats
the getfield/setfield pairing the macro exists for; the doc comment says so
rather than leaving the remaining hole unmarked.

## `MAJIT_MAX_BRIDGES` — opt-fuel over bridge compilation

`MAJIT_NO_BRIDGE` is all-or-nothing. It can attribute a wrong value to the
bridge path but not to a bridge. `MAJIT_MAX_BRIDGES=N` compiles the first N and
declines the rest, which turns that attribution into a bisection over N: the
boundary names the compilation that first produces the wrong value.

The counter is consumed last in the `should_bridge` chain, so it only advances
on a guard failure the other conditions already admitted and the index counts
bridges actually taken. `MAJIT_BRIDGE_FUEL_LOG=1` prints each one, which is what
pairs an index with the `@@@GUARD` line `MAJIT_GUARDLOG` emits beside it.

On the failure that motivated it the bisection was 6 runs of a few seconds each
against a repro that had cost a full rebuild per arm, and it landed on a single
bridge index with a stable 3/3 boundary on both sides.

## The suite baseline recorded 91 working modules as broken

Separate from the three changes above, and found while asking what the green
gate actually covers.

`pyre/cpython_tests/run.py` runs only the modules `baseline.json` records as
PASS — everything else is deselected and carries no signal. That makes the file
stale in one direction: a module that starts passing stays outside the gate
until someone runs the non-PASS set and records it. Nothing does that
automatically.

Running the full 434 on dynasm turns up 94 such modules. 91 reproduce as PASS in
a second run at lower concurrency and are promoted here — **63 IMPORTERROR, 16
CRASH, 9 FAIL, 3 TIMEOUT**. The CRASH set alone includes `test_dict`,
`test_tuple`, `test_range`, `test_slice`, `test_decimal`, `test_baseexception`,
`test_userdict`, `test_userlist`, `test_weakset`, `test_queue`,
`test_configparser` and `test_format`.

Three more (`test_embed`, `test_frozen`, `test_tools`) pass under `--full` but
stay put: they are `KNOWN_SKIPS`, which is a policy about what the suite should
never gate on rather than a record of what works. `--full` also reports
`test_c_locale_coercion` as `PASS -> FAIL`; that is a `KNOWN_SKIPS` entry whose
stated reason *is* the failure it produces (it asserts a child's stderr is empty
while `MAJIT_STATS=1` writes a `[jit-stats]` line to every process), so the gate
lane never runs it and its stale PASS is inert.

The gate now selects **201 modules and runs 200 in 158s**, against 109 in 170s
before — the coverage roughly doubles and the wall clock does not move.

What remains genuinely broken on dynasm, for the record: 137 FAIL, 26 TIMEOUT,
10 IMPORTERROR, and 4 CRASH (`test_mmap`, `test_resource`, `test_statistics`,
`test_weakref`).

## Gates

All on `729bcb51c9e`, which did not move across the runs; working tree clean;
`build/llbc/*.ullbc` re-extracted for all three crates first, and no run
reported `LLBC STALE`.

| gate | result |
|---|---|
| `cargo test --all --no-default-features --features dynasm` | exit 0, 103 `test result` lines |
| `pyre/check.py --backend dynasm` | **ALL PASSED 415/415**, incl. vendored CPython suite 109 modules |
| `pyre/check.py --backend cranelift` | **ALL PASSED 414/414** |
| `pyre/extra_tests/parity_tests/run.py` | `all parity tests pass` |

`test.test_pickletools` — red on the previous base and the reason this branch
went looking — passes here both with bridges on and under `MAJIT_NO_BRIDGE=1`,
which is #1130 doing its job; the baseline already recorded it as `PASS`, so it
was a real regression rather than a stale entry.

### DIFF
--- a/majit/majit-metainterp/src/jitdriver.rs
+++ b/majit/majit-metainterp/src/jitdriver.rs
@@ -466,6 +466,31 @@ pub fn no_bridge_enabled() -> bool {
     static FLAG: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
     *FLAG.get_or_init(|| std::env::var_os("MAJIT_NO_BRIDGE").is_some())
 }
+/// `MAJIT_MAX_BRIDGES=N` (diagnostic): allow the first N bridge compilations
+/// and behave as `MAJIT_NO_BRIDGE` from then on.  Bisecting N names the bridge
+/// whose compilation first produces a wrong value, at seconds per run rather
+/// than a rebuild per arm.  Consumes fuel only when the rest of `should_bridge`
+/// already held, so the count is bridges actually taken — place it last in the
+/// `&&` chain.  `MAJIT_BRIDGE_FUEL_LOG` reports each one taken.
+fn bridge_fuel_take() -> bool {
+    static LIMIT: std::sync::OnceLock<Option<u64>> = std::sync::OnceLock::new();
+    let Some(limit) = *LIMIT.get_or_init(|| {
+        std::env::var("MAJIT_MAX_BRIDGES")
+            .ok()
+            .and_then(|v| v.parse().ok())
+    }) else {
+        return true;
+    };
+    static USED: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+    let n = USED.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+    if n >= limit {
+        return false;
+    }
+    if std::env::var_os("MAJIT_BRIDGE_FUEL_LOG").is_some() {
+        eprintln!("@@@FUEL bridge #{n}");
+    }
+    true
+}
 fn guardlog_enabled() -> bool {
     static FLAG: std::sync::OnceLock<bool> = std::sync::OnceLock::new();
     *FLAG.get_or_init(|| std::env::var_os("MAJIT_GUARDLOG").is_some())
@@ -4413,7 +4438,8 @@ impl<S: JitState> JitDriver<S> {
             // pending-field prologue (resume.py:993-1007).
             let should_bridge = must_compile
                 && !majit_metainterp::MetaInterp::<S::Meta>::stack_almost_full()
-                && !no_bridge_enabled();
+                && !no_bridge_enabled()
+                && bridge_fuel_take();
 
             // compile.py:710 recovery_layout header_pc parity:
             // guard resume_pc comes from the guard's recovery metadata.
@@ -5715,7 +5741,8 @@ impl<S: JitState> JitDriver<S> {
         // every bridge.
         let should_bridge = must_compile
             && !majit_metainterp::MetaInterp::<S::Meta>::stack_almost_full()
-            && !no_bridge_enabled();
+            && !no_bridge_enabled()
+            && bridge_fuel_take();
 
         // Same `@@@GUARD` line the sibling loops emit. Without it this loop —
         // the one pyre reaches — had no per-guard-failure trace at all, so
@@ -6860,7 +6887,8 @@ impl<S: JitState> JitDriver<S> {
             // resume defects from the blackhole path.
             let should_bridge = must_compile
                 && !majit_metainterp::MetaInterp::<S::Meta>::stack_almost_full()
-                && !no_bridge_enabled();
+                && !no_bridge_enabled()
+                && bridge_fuel_take();
 
             // compile.py:710 recovery_layout header_pc parity:
             // guard resume_pc comes from the guard's recovery metadata.
--- a/majit/majit-translate/src/model.rs
+++ b/majit/majit-translate/src/model.rs
@@ -4053,13 +4053,30 @@ pub fn prune_dead_phis(graph: &mut FunctionGraph) {
     // Step 6: trim Link.args at indices whose target inputarg is dead.
     // `simplify.py:512-516`.  Walk in reverse so removals don't shift
     // surviving indices.
+    //
+    // The scope is the *target*'s reachability, not the source's.  Step 7
+    // trims the inputargs of every reachable block, and a block the
+    // reachability walk excludes — an orphan `eliminate_empty_blocks`
+    // bypassed, or one of the merge blocks jtransform leaves whose
+    // inputargs are phi targets rather than parameters — can still name a
+    // reachable block as its link target.  Skipping that link because its
+    // own block is unreachable breaks `len(link.args) ==
+    // len(link.target.inputargs)` with no diagnostic:
+    // `remove_duplicate_inputargs` reads each column by index across every
+    // incoming link, so the untrimmed one contributes the value one slot
+    // over, and the union-find merges two variables that are not the same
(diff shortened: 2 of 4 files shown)


youknowone:
## `main` is red on `CPython suite (gate)` since this merge

The baseline refresh in this PR promoted `test.test_format` from `CRASH` to `PASS`, and the gate now runs it and fails.

**This commit's own run** — [run 31370182942](https://github.com/youknowone/pyre/actions/runs/31370182942), job `93400431577`, `36666ef933c`, no other branch involved:

```
200 to run, 22 skipped, 212 not gated (non-PASS), timeout=300s
── REGRESSIONS (1) ──
  - test.test_format: PASS -> FAIL  rc=1 FAILED (failures=1, skipped=2) | FAIL: test_locale (__main__.FormatTest.test_locale)
```

**The parent-SHA control** — [run 31353530450](https://github.com/youknowone/pyre/actions/runs/31353530450) at `0768a7cd41f`, green:

```
109 to run, 22 skipped, 303 not gated (non-PASS)
  PASS 109   FAIL 0
```

`test.test_format` was `CRASH` in the baseline there, so the gate never ran it. The 91-entry refresh moved it into the gated set (`109 → 200 to run`), which is what surfaced the failure. Every open PR inherits this through its merge ref — e.g. #1141, whose only red is this exact line.

### The failure is real, and JIT-independent

```
AssertionError: ',' not found in '123456789'
    lib-python/3/test/test_format.py:440  self.assertIn(sep, text)
```

Reproduced locally on macOS, `LANG=en_US.UTF-8`, and it reproduces with the JIT off:

```
$ PYRE_NO_JIT=1 ./target/release/pyre-dynasm -c "
import locale; locale.setlocale(locale.LC_ALL,'')
print(repr(format(123456789,'n')), repr(format(1234.5,'n')))"
'123456789' '1234.5'          # CPython: '123,456,789' '1,234.5'
```

The locale data and the grouping machinery both exist — only the `'n'` path never consults the locale:

```
locale.localeconv()        -> grouping [3, 0], thousands_sep ','   # data is there
'{:,d}'.format(123456789)  -> '123,456,789'                        # insertion works
'{:n}'.format(123456789)   -> '123456789'                          # locale never read
```

### Where

`pyre/pyre-interpreter/src/type_methods.rs:2452`

```rust
if let Some(separator) = p.grouping {
    magnitude = separate_integer_digits(magnitude, interval, separator, displayed_digits);
}
```

`'n'` carries no `,`/`_` separator, so `p.grouping` is `None` and this branch is never taken. `interval` is a per-`ty` constant in the radix table (`'n' => (10, 3, false, "")`, `:2386`), so there is no channel for a locale grouping vector either.

Upstream is shaped differently: `newformat.py:642-658` `_get_locale(tp)` branches three ways and for `'n'` reads `rlocale.numeric_formatting()` (`:643-644`); insertion is then a **single** routine `_group_digits` (`:738-778`) shared by `,`/`_` and `'n'`, driven by a grouping vector that is `"\3"`/`"\4"` for the literal separators and the locale vector for `'n'`. pyre has only `separate_integer_digits` (`:2332`, RustPython-shaped), so the locale channel is absent by construction.

### Notes for whoever picks up the fix

- **Float `'n'` is not a one-line change.** `format_finite_float` delegates render+group+pad end-to-end to `FormatSpec::format_float` (`:3211-3216`); upstream instead renders unpadded `'g'`, splits with `_parse_number` (`:984-991`), groups, substitutes `_loc_dec` in `_fill_number` (`:798-813`), then pads.
- **`format(1234,'012n')` must be `'0,001,234'`**, which comes from `_group_digits`'s `min_width` argument (`:759`). A render-then-`pad_to_width` shortcut gives `'0000001,234'` and **this test would not catch it** — `test_locale` only asserts `sep in text` and `text.replace(sep,'') == '123456789'`.
- **`rustpython_host_env::locale::localeconv_data().grouping` cannot be used as-is** for the grouping vector: its `copy_grouping` stops at `CHAR_MAX` *and drops it*, collapsing `"\3\x7f"` (stop) and `"\3"` (repeat) to the same value. `interp_locale.rs:225-235` already does the raw NUL-terminated walk — minus the trailing `0` it appends at `:220-222`, which is `_w_copy_grouping`'s app-level fixup and must not reach the formatter.
- Windows/wasm/sandbox each need an arm; note pyre's `localeconv` is already `unix`-gated (`interp_locale.rs:200`) while `host_env` supports `any(unix, windows)`, and the sandbox build installs raising stubs — `format()` must not start raising, which diverges from upstream's `sandboxsafe=True` (`rlocale.py:167`).

Until it is fixed, the alternative is to put `test.test_format` back to its pre-refresh state so the gate reflects reality.

— *commented by Claude*

PR #1141, opened 2026-08-10
majit: resolve a boxing cluster's header pointers across the block boundary

youknowone:
`fuse_boxing_alloc` — pyre's port of `rewrite_op_malloc` — never fired anywhere
in the tree. This makes it fire at seven sites, and guards the miscompile that
firing it exposes.

## Why it never fired

The pass resolves `ob_header.ob_type` through `const_ref_addr`, which gave up
whenever no operation produced the value. A call ends a block, so a boxing
cluster whose header stores sit before one carries the type pointer into the
next block as a `Block.inputargs` phi while its `ConstRefAddr` producer stays
in the predecessor. The lookup found no producer and reported the vtable
unresolved.

The in-tree comment recorded that outcome and attributed it to an empty
`HostStaticAddrs.pytypes`. That attribution is wrong — the table is populated
in the build-script pipeline, and the addresses resolve as soon as the walk
crosses the block boundary. The comment is corrected here.

The fix follows the links when the value is a phi, requires **every**
predecessor to agree, and treats a non-variable link argument as a
disagreement. The cast walk and the phi walk are now shared by both resolvers.

Census: **0 → 7** fusions — `w_int_box_slow`, `w_int_new_unique`,
`w_complex_new`, `w_object_mutable_cell_new`, `w_int_mutable_cell_new`,
`w_dict_new_unmanaged_side_table_value`, `w_module_new_aliasing_dict`.

## What firing it exposes

The fusion discards the whole `ob_header` subtree and the runtime
re-synthesises **both** header words from the kept vtable
(`materialize_virtual_object`, `w_class = get_instantiate(vtable)`). That
substitution is only valid where the cluster's own `w_class` is
`get_instantiate(&T)` for the very `&T` its `ob_type` names.

`w_bytes_subclass_from_bytes` stores a base `ob_type` beside a `w_class` read
back off the shadow stack. Re-synthesising it would answer the **base** type
for `type(B(b'x'))`. This requires the two to agree and declines otherwise,
pinned by a test with a control row that fuses.

## The deviation, named

Dropping the whole header subtree is itself the deviation from upstream.
`jtransform.py` suppresses exactly one field, by the literal name `typeptr`
(908-911, 952-954), and `heaptracker.py:66-69` skips that name while
**recursing** into every other field of a nested header struct. Upstream, a
second header word is an ordinary traced field — as the Python-level class is
for user subclasses (`mapdict.py:751-752`).

The faithful shape is to keep `ob_type` vtable-derived and re-emit `w_class` as
an ordinary payload store, which would also fuse the subclass constructors this
guard declines. That needs per-group header `w_class` fielddescrs
(`function_header_w_class_descr` is the pattern) — the shared `w_class_descr()`
has `index_in_parent == 0` and would be dropped silently by the materialize
replay — and `virtualize.rs:944-958` names the read/write descr-identity split
as the prerequisite. Guarding first; the path is recorded in the commit
message.

## No performance claim

`bench/synth/list_pop_append` reads the same with and without this change, and
a pre-registered negative control — which also reverts the fusion, i.e. the
exact shape whose regression motivated the `dont_look_inside` boundary — does
**not** reproduce that regression. A negative control that fails to differ
voids every arm, so the bench no longer discriminates this mechanism and
nothing was decided from it. In particular the peer-merged boundary was left
untouched. This is stated in the commit message and in the `intobject.rs`
comment.

## Verification

- `pyre/check.py` (full): dynasm **415/415**, cranelift **414/414**, wasm
  **410/410**, exit 0, no jitstats drift, CPython gate passed
- `cargo test -p majit-translate -p pyre-jit-trace`: 21 suites ok

Four single-block fixtures modelled `ob_type` alone, or fed `w_class` the type
pointer instead of the instantiate slot; they now build the header the
constructors actually build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object boxing optimization to resolve type information across control-flow paths.
  * Prevented unsafe allocation fusion when class information does not match the object type.
  * Removed unnecessary initialization reads when boxing fusion succeeds.

* **Documentation**
  * Clarified virtual-object materialization and integer-boxing behavior, including vtable resolution and allocation boundaries.

* **Tests**
  * Expanded coverage for matching and mismatched class sources, complete object headers, and cleanup of unused initialization data.

### DIFF
--- a/majit/majit-translate/src/model.rs
+++ b/majit/majit-translate/src/model.rs
@@ -2980,7 +2980,27 @@ pub fn fuse_boxing_alloc(
                 _ => None,
             })
     }
-    fn const_ref_addr(graph: &FunctionGraph, var: &Variable, depth: u32) -> Option<i64> {
+    /// Resolve `var` to the constant address `terminal` reads off its producer.
+    ///
+    /// The walk itself is shared by every header spelling: it steps through
+    /// `__pyre_cast_instance[<root>]` pointer reinterprets, and through the
+    /// links when `var` is a `Block.inputargs` phi rather than an op result.
+    /// A boxing cluster whose header stores sit before a call is exactly that
+    /// shape — each call ends a block, so the header value crosses the
+    /// boundary as a link arg while its producer stays in the predecessor.
+    /// Every predecessor must agree: a phi merging two different pointers is
+    /// not a constant and must keep the cluster unfused.  `depth` bounds the
+    /// walk, which is what keeps a loop-carried phi (whose own link arg is
+    /// itself) from recursing forever.
+    fn resolve_addr<F>(
+        graph: &FunctionGraph,
+        var: &Variable,
+        depth: u32,
+        terminal: &F,
+    ) -> Option<i64>
+    where
+        F: Fn(&FunctionGraph, &OpKind, u32) -> Option<i64>,
+    {
         if depth == 0 {
             return None;
         }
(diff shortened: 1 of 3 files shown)


youknowone:
Both reds on this PR — `CPython suite (gate)` and `pyre/check.py (macos-latest)` — are the same single inherited failure, not caused by this commit. check.py's `cpython-suite` row is the same gate (it is `skip`ped on ubuntu, which is why only the macOS host is red); everything else in that run is green: `dynasm 1 failed, 414 passed`.

```
test.test_format: PASS -> FAIL | FAIL: test_locale (__main__.FormatTest.test_locale)
```

The identical failure occurs on `main` at `36666ef933c` ([run 31370182942](https://github.com/youknowone/pyre/actions/runs/31370182942), job `93400431577`), which contains none of this branch. #1138's baseline refresh moved `test.test_format` from `CRASH` to `PASS`, and every open PR picks that up through its merge ref.

It is also JIT-independent — `PYRE_NO_JIT=1` reproduces it — whereas this PR changes only a JIT lowering pass. Full RCA and the fix location: https://github.com/youknowone/pyre/pull/1138#issuecomment-5240844846

— *commented by Claude*

PR #1190, opened 2026-08-13
pip: support installed venvs and build isolation

youknowone:
## Summary

- make packaged pyre discover its staged stdlib and support venv, sysconfig, import, mmap, and typing behavior required by pip
- implement the SSL surface with rustls and the zlib streaming surface without introducing an OpenSSL dependency
- preserve per-frame red-frame identity across CALL_ASSEMBLER and blackhole resume so pip build isolation does not corrupt callee globals or class cells
- stage the stdlib in release builds and ignore installation paths written by pip
- pin RustPython dependencies to upstream main merge commit 212c0d0b154b45565b7f27fcb116abc6299608ca, which contains the merged compatibility and future-annotation fixes

## Why

pip reaches substantially more of the interpreter than wheel-only imports: HTTPS, compression, mmap, venv/sysconfig paths, isolated subprocesses, and class construction inside the JIT. The remaining sdist failure was a frame-identity collapse during CALL_ASSEMBLER blackhole continuation: stale green state overwrote the live red callee frame. This change keeps activation state on the live frame, matching the one-red-frame portal contract.

## Validation

- cargo check --features dynasm
- cargo test --features dynasm
- pyre -m test test_ssl test_zlib test_ensurepip test_super test_unittest: 1,434 run, 52 skipped, all successful
- future annotations smoke test against the merged RustPython main revision
- installed Homebrew-style venv: pip sdist build isolation, wheel build, install, import, and uninstall
- dynasm pre-merge benchmark suite: 17/17 passed
- rtyper prepass census: phase A 1499, phase B 4
- cargo fmt --all -- --check
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added bundled standard-library and pip assets to release packages.
  * Improved Python 3.14 compatibility, including implementation metadata, ABI headers, import paths, and slice constants.
  * Added enhanced TLS certificate-chain handling, verification, cipher reporting, and channel binding.
  * Expanded `mmap` support with safer resource handling, resizing, validation, and platform features.
  * Improved zlib streaming, copying, dictionaries, compression options, and format handling.

* **Bug Fixes**
  * Improved socket timeout behavior and JIT execution recovery.
  * Added more reliable virtual-environment and standard-library discovery.


youknowone:
### Re: Codex parity review §2

Both §2 items are the same deliberate divergence, and it is the point of the
change rather than an oversight — `stdlib_at_prefix` and `scripts/stage-stdlib.py`
each carry a comment stating it.

`lib_pypy` is mostly PyPy's cffi / pure-Python shims for modules pyre implements
natively or not at all (`_testcapi`, `_md5`, `_sha*`, `_dbm`, `_sqlite3`, …). The
CPython suite branches on whether `import X` succeeds, so putting those on
`sys.path` turns a clean skip into a run against a module that cannot answer.
Measured on this branch's own gate runs: with `lib_pypy` on `sys.path` the gate
reported **16** regressions (`_md5` / `_hashlib` / `_dbm` / `_testcapi` shims);
without it, **6** — five of which reproduce on `main` at this branch's merge base,
and the sixth (`test_range`) also reproduced on an unrelated branch at the same
base, so none of the six are contributed by this branch.

The one module the pip work actually needed from `lib_pypy` — `_sysconfigdata.py`
— was moved into `lib-python/3` instead, so the supported surface stays
`lib-python/3` plus the builtin modules, in both the source tree and the staged
release tree.

§3's zlib message mismatches are left as-is: under the project's spec/implementation
split the wording follows CPython 3.14, which is what those messages already spell.

— *commented by Claude*

PR #1182, opened 2026-08-12
jit: speed up exception construction, traceback, and property getters

youknowone:
## Summary

- allocate traceback nodes emitted by compiled code in the nursery while retaining stable allocation for Rust callers
- specialize SystemExit construction, including its code field and all supported argument shapes
- trace straight-line raising property getters through their exception-constructor call instead of executing an opaque residual every iteration
- resize the property exception fixture so its PyPy baseline clears the timing floor and its 30x CI ceiling is enforced

## Property getter result

The PyPy oracle reports one loop, zero bridges, and zero forcings. Pyre now reports one loop, zero bridges, zero aborts, and one initial guard failure. The focused check.py run passes at 12.2x PyPy; the prior CI result was an informational lower bound of roughly 31-37x because the PyPy measurement was clamped.

## Verification

- cargo check --features dynasm
- cargo test --features dynasm
- cargo fmt --all -- --check
- focused check.py synthetic run for property_getattr_exceptions: PASS
- property_protocol_hot and property_custom_hook_decline: correct output, no JIT aborts
- all eight primary benchmarks completed successfully

The local check and test used the LLBC fingerprint bypass only because unrelated, unstaged interpreter edits changed after LLBC extraction. The final commit in this PR is translator/fixture-only; clean CI will perform the normal fingerprint checks.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved generator function names and qualified names, including non-ASCII names.
  - Improved exception handling for `SystemExit`, traceback chains, and nested property getter calls.
  - Preserved module name identity, including names containing lone surrogate characters.
  - Improved garbage-collection safety and validation for large allocations.
  - Ensured traceback data remains available after garbage collection.

- **Tests & Benchmarks**
  - Added regression coverage for generator names, exit-code argument shapes, traceback retention, and module names.
  - Updated benchmark expectations and baseline documentation.

PR #1194, opened 2026-08-13
jit: count eval-breaker back-edge poll failures separately

youknowone:
`close_loop_args_at` records the eval-breaker poll as a real guard at every
loop close, so a major collection arming the word failed that guard and
`record_guard_failure_event` counted it in `guard_failures` like any other.
That total mixed deoptimizations the compiled code caused with the timing of
collections, and moved between hosts while `loops_compiled` and
`bridges_compiled` stood still.

## What changed

- `is_back_edge_poll_guard` walks back from a guard's condition to a
  `RawLoadI` of the published eval-breaker address. It anchors on that load
  rather than on the whole opcode shape, because the load is the one link that
  cannot be dropped (it must stay non-pure or CSE forwards the preamble's
  guarded zero into the body and deletes the body's poll).
- `FailDescr::is_back_edge_poll` / `set_back_edge_poll`, owned by
  `ResumeGuardDescr` and `ResumeGuardCopiedDescr`, following the per-emission
  scoping `source_op_index` already uses.
- `store_final_boxes_in_guard` stamps it — the single point where a guard's
  condition chain and its descr are both in hand. Both emit paths reach it, and
  every re-emission (unroll preamble and body, bridge back edges) mints its
  descr there, so no backend change is needed.
- `record_guard_failure_event` splits only the tally. The census,
  `warm_state.log_guard_failure` and the hook still see every failure, so
  bridge thresholds are untouched and `bridges_compiled` moved nowhere.
- `check.py` keeps `back_edge_polls` out of `JITSTATS_SNAPSHOT_FIELDS`.
  Recording it would relocate the sensitivity onto a new key instead of
  removing it. pyrex, pyre-wasm and pyre-wasm-runner report it.

## Measurements

Across all 404 dynasm synthetic fixtures, comparing each fixture's
`guard_failures` at check.py's pinned `PYPY_GC_MIN=256MB` against the same
fixture at 8GB, where no major collection happens:

| | before | after |
|---|---|---|
| fixtures whose count differs between the two settings | 11 | 0 |
| poll failures folded into `guard_failures` | 21 | 0 |

Baselines re-recorded: 11 on dynasm, 11 on cranelift, 9 on wasm. Every new
value was independently produced by running the fixture at 8GB before the
change. `loops_compiled` and `bridges_compiled` are unchanged everywhere.

`pyre_env`'s note said the threshold pin pushes past every fixture's working
set; it was generalized from two fixtures and those 11 refute it, so the note
is corrected to what was measured.

## Follow-up commit

`str_fstring`'s cranelift baseline had been re-recorded 659 -> 658 by
subtracting dynasm's delta instead of measuring cranelift, which carried two
poll failures where dynasm carried one. Corrected to 657, the measured value.

## Stability of the result

`str_fstring` at check.py's pinned environment: **20/20 runs report 657 on
each backend**. Sweeping nursery (2/4/8/16MB) x major threshold
(128MB/256MB/1GB/8GB), 32 cells: 31 report 657 while `back_edge_polls` moves
0 -> 1 -> 3. The one exception is cranelift at nursery=16MB with a 128MB
threshold, which reports 658 deterministically (12/12; dynasm reports 657
12/12 at the same setting) — under four times as many collections, cranelift's
code takes one extra genuine deopt. So the poll contribution is removed
entirely, and what remains is a real consequence of collecting, not a counting
artefact.

## Summary by CodeRabbit

- **New Features**
  - Added separate JIT statistics for eval-breaker back-edge polls.
  - Exposed back-edge poll counts in runtime, WebAssembly, and diagnostic summaries.
  - Improved tracking of polling events during compiled execution.

- **Bug Fixes**
  - Corrected garbage-collection analysis for several allocation operations.

- **Tests**
  - Added coverage for poll detection and allocation analysis.
  - Updated benchmark statistics to reflect revised guard-failure counts.

- **Documentation**
  - Clarified trace labels, runtime counters, and diagnostic threshold behavior.


youknowone:
Rebased onto current `main` and pushed three more commits. Summary of this round.

### Is 657 the right value for `str_fstring`?

Yes, on three independent grounds:

* **Ground truth.** At `PYPY_GC_MIN=8GB` no major collection happens at all (`back_edge_polls=0`), so the schedule term cannot contribute — cranelift reads **657**.
* **The pinned config.** At check.py's pin, dynasm and cranelift both read 657, deterministic 3/3 each.
* **A sweep.** Over a nursery x `PYPY_GC_MIN` grid, 13 of 14 cells read 657.

`main` currently records 658 for `str_fstring.cranelift`; that is the pre-split value, which includes the poll. The correction to 657 is exactly what separating the counter produces.

### The one outlier, traced

One cell (cranelift, nursery=16MB) reads 658. Tracing it needed a diagnostic that did not exist: `compile_bridge` never called `log_optimized_trace`, so only the three loop paths were dumped, while bridges still consumed trace ids from the same counter. That left gaps in the dump sequence, so a dumped trace could not be lined up with the `trace=` field of a guard-failure log. Added in `jit: log optimized bridge traces under MAJIT_LOG_OPT`.

With bridges dumped, the outlier is one guard: trace 6 index 5, `GuardFalse(IntIsTrue(GetfieldGcI(v462)))` reading `W_IntObject.intval` off a `CallMayForceR` result, failing 52 times instead of 51.

It is **not** a missed poll — the poll chain is `IntIsTrue(IntAnd(RawLoadI(&EVAL_BREAKER_WORD)))`, and this guard fails 51 times in runs where `back_edge_polls=0`, which no stamped guard can do. So the stamping has no bridge gap.

What it is: the two runs' `(trace, fail)` censuses are otherwise identical, **including which three guards the polls land on**. The discriminator is which *iteration* a bailout lands on — an eval-breaker bailout returns control to the interpreter, so a poll firing one iteration earlier or later moves a single iteration across the compiled/interpreted boundary, and a downstream data guard is then executed once more. It is sporadic rather than systematic: at nursery=16MB, `PYPY_GC_MIN=32MB` with **12** polls reads 657 while `64MB` with **6** polls reads 658 — non-monotone in poll count. Every zero-poll row reads 51, and all four rows at the pinned 4MB nursery read 657.

The consequence worth stating: only the first-order term can be made schedule-invariant, and that is what this PR removes. A bailout perturbing the compiled/interpreted split is inherent to counting deopts; no re-tally can remove it.

### Three more wasm baselines

The full gate surfaced three wasm fixtures whose `guard_failures` fall by exactly one with `loops_compiled`/`bridges_compiled` frozen. Each measures `back_edge_polls=1`, so the drop is precisely the poll term: `inline_gate_operand_provenance` 5->4, `math_isqrt_compare_bridge_resume` 403->402, `recursive_call_frame_relocation` 638->637. Edited by hand rather than via `--snapshot`, which rewrites the whole corpus.

### Gate

`pyre/check.py` locally: dynasm 425/425, cranelift 425/425, wasm 418/418.

— *commented by Claude*

youknowone:
### Re: "Verify that shared/copied guard descrs inherit the back-edge-poll classification"

Checked, and the conclusion is that a copied descr **should not** inherit it — reading through `prev` would introduce a misclassification rather than fix one. Two independent reasons:

**A poll guard can never be the sharer.** Both emit paths gate sharing on the same test — `!op.has_descr() && op.rd_resume_position.get() < 0` (`optimizeopt/optimizer.rs` `emit_guard_operation`, mirrored in `optimizeopt/mod.rs`). The poll is emitted by `close_loop_args_at` via `generate_guard`, which goes through `generate_guard_core` → `capture_resumedata(resume_pc)`; it has to, since the guard's whole purpose is to bail out at the back edge. So it always carries a resume position and can never satisfy `rd_resume_position < 0`. It always lands in `store_final_boxes_in_guard`, which is where the stamp runs.

**When a poll guard is the donor, the sharer is not a poll.** Sharing applies to descrless optimizer-created follow-up guards (`GUARD_NO_EXCEPTION` after `CALL_MAY_FORCE`, per the comment at `optimizer.py:661-664`). Such a guard has a different condition chain, so `false` is the correct answer for it, and chasing `prev` would stamp it as a poll whenever its donor happened to be one.

This is the same per-emission rule the neighbouring `force_token_slots` field already documents: "Owned per copied descr so each emission's GC-root classification stays distinct — no sharing through `prev`."

Corroborated by measurement: in `str_fstring` the three guards carrying `back_edge_poll: true` in the optimized dumps are exactly the three whose failures a no-collection/pin/heavy three-point subtraction identifies as polls, and `guard_failures` reconciles with `back_edge_polls` on every run.

Added the reasoning as a doc comment on the accessor in `majit: document why a copied descr does not inherit back_edge_poll` so the question does not have to be re-derived.

— *commented by Claude*

youknowone:
Rebased onto `origin/main` (`3233b3cb42f`) and force-pushed.

### The 657/658 "flake" was not a flake

`str_fstring` read `guard_failures=657` standalone (20/20) but `658` under `check.py`, both deterministic. The decision variable is **the child's stderr destination**, not stdout:

| stdout | stderr | guard_failures |
|---|---|---|
| FILE | FILE | 658 (5/5) |
| /dev/null | FILE | 658 (5/5) |
| FILE | PIPE | 657 (5/5) |

`MAJIT_STATS=1` and `PYRE_DESCR_SPELLING_GATE=1` write multi-KB lines to stderr *during* the run; pipe vs regular file changes the write path's buffering and allocation, which moves where old-gen use crosses the major-collection threshold, which flips which copy of the peeled back-edge poll catches the armed bit. Per the peel note in `check.py`, the preamble copy costs two bailouts to the body copy's one, because resuming from the preamble re-enters at the loop head and re-fails its `FOR_ITER` guard — a real deopt, so the poll/deopt split cannot remove it.

`_run_timed_unix` hands the child `tempfile.TemporaryFile()` for both streams (pipes would deadlock against `wait4`), so the gate always sees the FILE reading. Measuring a gate number with stderr piped into `rg` measures something else.

### Platform overlays

`_jitstats_baseline_path` prefers `<name>.<backend>.<platform>[.github-actions].jitstats` over the generic file. The synthetic corpus contains exactly four such overlays and **all four are `str_fstring`'s**, so this fixture is the one place the preference bites — and the counter-split commit had updated only the generic files. All six `str_fstring` baselines now drop by the one `back_edge_poll` each records; `dynasm.darwin` is measured directly and confirms the rule at 658.

### Bridge dumps

`compile_bridge` did not call `log_optimized_trace`, yet bridges draw trace ids from the same counter loops draw from. The gaps made a dumped trace impossible to line up with the `trace=` field of a guard-failure log. Bridges are dumped now.

Note also that the runtime `@@@GUARD` log prints `fail_index_per_trace`, which is assigned by the backend assembler and is **not** unique across traces — a census must key on `(trace, fail)`.

### Pre-existing, not from this branch

`exception_traceback_loop_forms.wasm.jitstats` records `guard_failures=810` while the fixture now runs 811. #1182 set all three backends to 810; b0f34c0af3e (#1188) raised `cranelift` and `dynasm` back to 811 and left the wasm file untouched. This branch is byte-identical to `main` on all three baselines, and the fixture measures `back_edge_polls=0`, so the counter split provably cannot move it.

Gates on the pre-rebase tree: dynasm 425/425, cranelift 425/425.

— *commented by Claude*

PR #1197, opened 2026-08-13
majit: keep a boxing cluster's w_class as its own store

youknowone:
> **This PR's contents changed.** It previously carried `module: store the module
> name as the wrapped object`; that change is now on `main` (it landed through the
> #1187 review follow-ups), so the branch rebased down to the single commit below.

`fuse_boxing_alloc` dropped both header stores and let the `NewWithVtable`'s
vtable stamp `ob_type` and `w_class`, so it could only fire where the cluster's
`w_class` was `get_instantiate(&T)` for the very `&T` its `ob_type` named.
Anything else declined and left `malloc_typed` residual, which the rtyper
rejects fail-closed; the whole graph then census-Skips to the legacy walker.

## Change

`jtransform.py:1040-1045 rewrite_op_malloc` emits `new_with_vtable` carrying
only the sizedescr and leaves every other field to its own `setfield`. Do the
same with `w_class`: lower the cluster and re-emit the store as a `FieldWrite`
on the new object whenever the vtable does not stand for it.

Predecessors that disagree — one folding into the vtable and another not, or two
naming different classes — still decline, as they already did for a disagreeing
`ob_type` (`malloc.py:176-186`).

`prune_dead_boxing_remnants` needs no change: its liveness gate already keeps a
producer the flow reaches, which the re-emitted store now makes the `w_class`
feed.

## Measurement

Both arms run on identical LLBC — only `majit-translate/src/model.rs` differs
(`PYRE_LLBC_STRICT=0` demotes the repo-wide #1164 digest, which a translator-only
swap necessarily trips).

| | `main` | this PR |
|---|---|---|
| `survived fuse_boxing_alloc unfused` | 36 lines / 18 graphs | 4 / 4 |
| `PREPASS phaseA fail` | 1497 | 1487 |

`w_long_from_raw` pairs `&LONG_TYPE` with `get_instantiate(&INT_TYPE)` so that
`type(1 << 100) is int`, and is the root cited by **14 of those 18**. The other 4
are `w_generator_or_coroutine_new` and its three callers, which decline for the
other reason — three `ob_type` constants merge into one malloc, so no single
vtable stands for the cluster — and are exactly the 4 that survive.

**Clearing the reject is not the same as lifting the graph.** Of the 18: 9 now
annotate, 4 keep that reject, and 5 fail on a *different* error it was standing
in front of (`compute_at_fixpoint`, on `unary_negative_value` /
`unary_invert_value` under both spellings and on `PositionCookie::to_object`).

## The −10 is not uniform: 14 fixed, 4 newly failing

Diffing the failing-graph *sets* rather than the totals decomposes the −10 into
14 fixed and 4 regressed. (That 14 is a different set from the 14 above: the 9
that now annotate, plus 5 that had been failing only through a cascade from them
— `__pyre_wrap_random`, `w_dict_proxy_new`, `w_float_new`, `w_range_compute_item`,
`w_range_new`.)

The 4 that regress are `PyFrame::getorcreate_debug_data` and its three callers:

```
UnionError in mergeinputargs: longobject::W_LongObject ∪ pyframe::FrameDebugData
In <FunctionGraph of write>
```

`core::ptr::write` is lifted as **one** graph shared by every instantiation
(`Lowering::new` takes the declaration's `fd.generics`; the instantiation lives
at the call site), and the raw write survives fusion by design —
`sink_fused_boxing_aggregates_at_raw_writes`' test asserts "the raw write must
survive the rewrite". Until now only `FrameDebugData` reached that graph;
`w_long_from_raw` lifting means `W_LongObject` reaches it too, and slot 0 unions
two instances with no common base.

This is a latent landmine rather than a defect in this rewrite: **any** change
that lets `w_long_from_raw` lift arms it, so there is no narrower variant of
this patch that avoids it. The upstream shape is per-instantiation
specialization — `Specializer::Argtype` (`specialize:argtype(N)`) is already
implemented — but nothing in production ever sets `annspecialcase` on an
LLBC-lifted function, so wiring it is a front-end change of its own and is left
as follow-up. The 4 graphs fall back to the legacy walker, the same fail-closed
path the reject already had them on the edge of.

Widening the union instead would be the wrong repair: the analogous
`List ∪ Inst(classdef-less)` arm lifted 5 graphs and was reverted because phaseB
silently accepted the resulting erasure.

## Verification

- `cargo test -p majit-translate --release --lib` — 3190 passed, 0 failed.
- `python3 pyre/check.py --backend dynasm` — 425/425; `--backend cranelift` — 425/425.
- Semantic probes on the lifted graphs: `type(1 << 100) is int`, hot-loop `-x` /
  `~x` arithmetic, and `class B(int)` subclass identity all correct.

### DIFF
--- a/majit/majit-translate/src/model.rs
+++ b/majit/majit-translate/src/model.rs
@@ -2906,9 +2906,12 @@ pub fn fuse_boxing_alloc(
     // Derive a boxing struct's scalar payload fields from its registered field
     // layout, in struct-declaration order, skipping the `ob_header` (PyObject
     // base): the header's type pointer is captured separately into
-    // `NewWithVtable.vtable` (see `resolve_vtable_addr`) and the runtime stamps
+    // `NewWithVtable.vtable` (see `resolve_header_plan`) and the runtime stamps
     // `ob_type` / `w_class` from it, so only the non-header setfield(s) are
     // re-emitted (oracle: `box_trace.rs trace_box_float` / `trace_box_int`).
+    // The one header field that survives as its own store is a `w_class` the
+    // vtable cannot stand for, which `resolve_header_plan` hands back as a
+    // payload.
     // This mirrors the type-generic malloc lowering in the front end
     // (`rpython/rtyper/rbuiltin.py:349-385` `rtype_malloc`; `rclass.py` reads
     // the exact lltype's field layout); `struct_field_attrs` is the
@@ -2969,7 +2972,7 @@ pub fn fuse_boxing_alloc(
     // `value.ob_header.ob_type`; it must match `w_class`, since values may carry
     // a subclass there.
     fn store_value(graph: &FunctionGraph, base: &Variable, field_name: &str) -> Option<Variable> {
-        let (_, value) = unique_store(graph, base, field_name)?;
+        let (_, value, _) = unique_store(graph, base, field_name)?;
         value.as_variable().cloned()
     }
     /// The unique graph-wide store to `base.field_name`, or `None` if absent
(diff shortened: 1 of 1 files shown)


youknowone:
Rebased onto current `main`, which changes what this PR contains.

The `module: store the module name as the wrapped object` commit is no longer
here — an equivalent `w_name: PyObjectRef` port landed on `main` through the
#1187 review follow-ups, so the rebase dropped it. I diffed my version against
`main`'s across all five touched files before letting it go: the structure
matches and `main` additionally carries a
`module_init_retains_a_surrogate_name_object` test, so nothing was lost.

What remains is the single majit commit, and the description above has been
rewritten for it. One thing worth reviewer attention: the census improvement is
**14 graphs fixed against 4 newly failing**, not a uniform gain. The 4 are
diagnosed in the description — `core::ptr::write` is lifted as one graph for all
instantiations, so the first change that lets a second type reach it hits a
`UnionError`. That is reachable by any patch that makes `w_long_from_raw` lift,
not something this one can dodge.

— *commented by Claude*

PR #1211, opened 2026-08-14
marshal: wrapped code constants; _typing: an absent type-parameter default; PR 1174 review follow-ups

youknowone:
Follow-ups to the #1174 review, plus the two findings that review left blocked.

## `marshal`: wrapped code constants

`MarshalBag`'s blanket impl over `ConstantBag` answered `BadType` for
`make_list` / `make_set` / `make_dict`, so a `code` object whose `co_consts`
held a list, set or dict round-tripped as an error rather than as the object.

rustpython is repinned to `fd7d107c7` (RustPython/RustPython#8516), whose reader
decodes a code object's fields through the runtime bag instead of requiring
every `co_consts` entry to fit the compiler constant enum. `PyreMarshalBag`
implements the five hooks it needs: `code_constant_from_value` records a
`ConstantData::None` shape placeholder for a value the enum cannot describe,
and `make_code_with_constants` stores the decoded objects into `co_consts_w`.
`w_code_fill_consts_from_tuple` splits into a slice-taking
`w_code_fill_const_slots` so the reader, which holds the values but no tuple,
shares the path `code.replace(co_consts=...)` already used.

Round trip of `co_consts=([1, 2], {"a": 3}, {4, 5}, 6, "s", None)`, a nested
code object, and shared-constant identity all match CPython 3.14 and pypy3.

## `_typing`: an absent default is not `NoDefault`

`_Py_make_typevar` / `_Py_make_paramspec` / `_Py_make_typevartuple` leave
`default_value` NULL, while `typevar_new_impl` stores the `NoDefault` sentinel
its signature defaults to. `typevar_evaluate_default` answers `None` for the
first and a constant evaluator for the second, so a compiler-created type
parameter reported `<constevaluator NoDefault>` where `None` is correct.

`_MISSING`, which already stands for the NULL `bound` and `constraints` slots,
now stands for the NULL `default_value` too. `ParamSpec._make` and
`TypeVarTuple._make` join `TypeVar._make`, and the three intrinsics construct
through them. `__default__`, `evaluate_default` and `has_default` follow
`typevar_default`, `typevar_evaluate_default` and `typevar_has_default_impl`.

## Review follow-ups

* `_typing._index` narrows an `__index__` result with `int.__index__` instead
  of `int()`, so an int subclass keeps its stored value.
* `descr_function_new` binds its six constructor parameters through
  `bind_builtin_kwargs`, which reports the arity, the by-name-and-position
  duplicate and the missing-required cases.
* `marshal.load` converts a `readinto` count through `getindex_w`.
* `WriterRefs` indexes its entries by `gc_identity_hash`, replacing the linear
  scan over every written object.
* `UnionBuilder::finish` reads the member list and the parameters tuple out of
  the shadow stack after the frozenset and tuple allocations, and takes the
  parameters slot rather than a by-value pointer.
* `inline_call` builds its `LoopBearingCalleeInlineUnsupported` decline through
  `DispatchError::callee_inline_unsupported`, so `PYRE_LB_SITE` records the site.
* `w_list_new_with_strategy` debug-asserts that the items support the requested
  strategy before `build_list_storage` unboxes them.
* `Cpu` drops `#[derive(Debug)]`: the derived formatter requires
  `core::marker::FnPtr` for every helper field, which Charon does not model, so
  extraction emitted two trait-resolution warnings per function pointer.

## Note on the branch name

The remote already has an unrelated `builtins` branch owned by another
worktree, so this line is pushed as `marshal-consts`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1077,7 +1077,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3154,7 +3154,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -3216,7 +3216,7 @@ dependencies = [
 [[package]]
 name = "rustpython-codegen"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=d64cc2cff5ee611862ca9d2a27ccbd9a6f740e5d#d64cc2cff5ee611862ca9d2a27ccbd9a6f740e5d"
+source = "git+https://github.com/RustPython/RustPython.git?rev=2690c16fdceedf1c248a64a360de90e3ffac3dfc#2690c16fdceedf1c248a64a360de90e3ffac3dfc"
 dependencies = [
  "bitflags 2.13.0",
  "indexmap",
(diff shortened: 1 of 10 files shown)


youknowone:
Superseded by #1212, which carries the same four commits on `builtins`. — *commented by Claude*

PR #1213, opened 2026-08-14
majit: array element write-sets for residual calls, per-jitdriver enable_opts, and resume-path fixes

youknowone:
Twelve commits on top of `main`, in three groups.

## Array element write-sets for residual calls

`residual_writes` could not express a write to an array element, so a helper
that stores into one had no way to declare it. This adds the vocabulary and
the emit side, then closes two gaps in how the write-set is recovered:

- `8ac021ff56f` emit-side `setarrayitem_gc_i` builders
- `4e58ff3f623` the `array_fields` vocabulary for non-virtualizable element access
- `a07f9e58deb` recover array write-sets by raw-set identity when `compute_bitstrings` has not run
- `c008f77bdb7` key the fallback on shape rather than pointer identity
- `acd5157e74d` the `field[]` element write-set in `residual_writes`

New tests: `array_write_set_invalidation.rs`, `jit_interp_residual_array_write.rs`.

## Per-jitdriver `enable_opts`

`unroll` was on for every driver. `warmspot.py:73` makes it per-driver
upstream, so the knob is exposed and unrolling reads it:

- `3a525f5f6b3` gate unrolling on the jitdriver's `enable_opts`
- `4b075433008` expose `set_param_enable_opts` on `JitDriver`

## Resume-path fixes

- `a33c8b8e3d8` scope the inline-frame identity-slot snapshot trim to `split_dispatch`
- `948e34df157` seed a walk re-entered after a declined close at the pc that declined
- `54e723c8383` route both close re-entries through one resume publisher
- `ab5199cb1ee` align nursery allocation with upstream regalloc
- `dd927e894bb` describe what `opt_time` measures

## Verification

`pyre/check.py` was still running when this PR was opened; its result is
reported in a follow-up comment rather than asserted here.

The aheui consumer of the `enable_opts` change lives in a separate repository
(`youknowone/aheui-rust`) and is not part of this PR.

https://claude.ai/code/session_011NLDxh1mE6nALnJFaCZNoJ

### DIFF
--- a/majit/majit-backend-dynasm/src/aarch64/assembler.rs
+++ b/majit/majit-backend-dynasm/src/aarch64/assembler.rs
@@ -3387,9 +3387,11 @@ impl<'a> AssemblerARM64<'a> {
                     let rv = r.value;
                     dynasm!(self.mc ; .arch aarch64 ; mov X(rv), x0);
                 }
-                if !op.pos.get().is_none() {
-                    self.store_rax_to_result(op.pos.get());
-                }
+                // aarch64/regalloc.py:964-966 keeps malloc's result in the
+                // selected result register (x0); it does not eagerly mirror
+                // every result into its jitframe slot.  Regalloc emits the
+                // eventual spill when a guard or register-pressure boundary
+                // actually needs one.
             }
             // aarch64/assembler.py:715 malloc_cond_varsize_frame
             OpCode::CallMallocNurseryVarsizeFrame => {
@@ -6546,9 +6548,11 @@ impl<'a> AssemblerARM64<'a> {
             return;
         }
 
-        // x0 = nursery_free, x1 = new_free, x2 = scratch, x3 = nursery_top
-        self.emit_mov_imm64(2, nf_addr as i64);
-        dynasm!(self.mc ; .arch aarch64 ; ldr x0, [x2]);
+        // aarch64/assembler.py:682-708 uses only x0/x1 plus the reserved IP
+        // scratch registers.  Keep x2..x13 available to regalloc on the fast
+        // path; the collecting slow path spills/restores them below.
+        self.emit_mov_imm64(16, nf_addr as i64);
+        dynasm!(self.mc ; .arch aarch64 ; ldr x0, [x16]);
 
         if total_size < 4096 {
             let ts = total_size as u32;
@@ -6558,18 +6562,18 @@ impl<'a> AssemblerARM64<'a> {
             dynasm!(self.mc ; .arch aarch64 ; add x1, x0, x1);
         }
 
-        self.emit_mov_imm64(3, nt_addr as i64);
-        dynasm!(self.mc ; .arch aarch64 ; ldr x3, [x3]);
-        dynasm!(self.mc ; .arch aarch64 ; cmp x1, x3);
+        self.emit_mov_imm64(17, nt_addr as i64);
+        dynasm!(self.mc ; .arch aarch64 ; ldr x17, [x17]);
+        dynasm!(self.mc ; .arch aarch64 ; cmp x1, x17);
 
         let slow_path = self.mc.new_dynamic_label();
         let done = self.mc.new_dynamic_label();
         dynasm!(self.mc ; .arch aarch64 ; b.hi =>slow_path);
 
-        // Fast path: bump nursery_free, zero header, return payload ptr
-        self.emit_mov_imm64(2, nf_addr as i64);
+        // Fast path: x16 still holds &nursery_free; bump it, zero the header,
+        // and return the payload pointer.
         dynasm!(self.mc ; .arch aarch64
-            ; str x1, [x2]       // *nursery_free = new_free
+            ; str x1, [x16]      // *nursery_free = new_free
             ; str xzr, [x0]      // zero GcHeader
         );
         let hs = gc_hdr as u32;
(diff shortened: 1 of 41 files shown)


youknowone:
## Review findings

All three were real. Fixed in `51115e1d360`, with a regression test for the one that had none.

**P1 — unsigned array element loads.** Both the read and the write site passed `is_item_signed: true` unconditionally, so an element declared `u8` holding `0x80` reached the trace as `-128` where the concrete Rust read yields `128`. `descr.py:240-254 get_type_flag` takes signedness from the declared type; both sites now derive it as `(<Elem>::MIN as i128) < 0`, which resolves through a type alias and fails to compile for a non-integer element type rather than loading it as a signed word. New test `jit_interp_array_element_signedness.rs` pins `u8` and `i64`; reverting the derivation reddens the `u8` case with `interned [(1, true)]`.

**P1 — array assignment operand order.** Correct: Rust evaluates the assigned value before the assignee place, and both the annotated Rust and the jitcode lowering bound the base and index first. `lower_array_field_base` is now split into `match_array_field_base`, which rejects without emitting anything, and `emit_array_field_base`, so the write can finish matching the shape and then emit the RHS, the base and the index in that order. The read keeps its order — a place expression does evaluate its base before its index.

**P2 — empty `enable_opts`.** Correct, and it inverted the caller's request rather than merely widening it: `warmstate.py:284` substitutes the full list for `'all'` (and for `None`, which this `&str` surface cannot express) and for nothing else, so `""` splits into no names and selects nothing. Treating it as `'all'` meant a driver asking for no optimizations kept unrolling and every other pass. Unit test added.

One further divergence in the same function is **not** fixed here: upstream raises `ValueError` on an unknown pass name (`warmstate.py:288-289`) and this port silently ignores it. Fixing it means giving the `&str` surface a fallible signature, which reaches the `sys` parameter path in `pyre-jit`; that is a separate change.

## `pyre/check.py`

RC=1. **dynasm 425/425 passed.** Three failures, none of which I have re-recorded:

- `cranelift synth/str_fstring` — `guard_failures 658 -> 659`. Deterministic here: 659 on 2/2 repeats and on 4 different env lengths, so it is not a per-run lottery today. The baseline it grades against is the `darwin` overlay, last written by `b0f34c0af3e` on `main`; this branch does not touch it, and the generic (non-darwin) baseline already reads 659. This counter is also a known band artifact — at check.py's pinned `PYPY_GC_MIN=256MB` the fixture sits on a major-collection threshold, so the overlay records a phase rather than a property. **Unresolved:** whether this branch moved it. Only a control build of `origin/main` on this host answers that, and I have not run one — the resume-path commits here are a plausible cause, so I am not claiming it is pre-existing.
- `wasm fannkuch` — 3.3x vs a 3x wall-clock gate.
- `wasm synth/global_quasiimmut_invalidation` — 3.3x vs the same gate.

Both wasm failures are wall-clock ratios measured on a shared host at load average 19–36, where this repository's timing gates are not readable. They need a re-run on a quiet host before they mean anything.

— *commented by Claude*

youknowone:
## Second review round — all three findings real, fixed in `a2cd03ef929`

**P1 — residual write-set signedness (`lower_stmt.rs`).** Correct, and it is a regression the previous round introduced: `writes_array_descr_by_shape` keys on `(base_size, item_size, item_type, is_item_signed)`, so once the read derived signedness and the write-set kept its literal `true`, a `u8`-element `residual_writes` declaration minted a shape no read ever produces and stopped invalidating the load it names. The write set now derives it exactly as the read and write do. New macro-level fixture `unsigned_elements::a_u8_declaration_reaches_the_unsigned_read_and_not_a_signed_one` in `jit_interp_residual_array_write.rs`; restoring the literal reddens it.

**P1 — GC-managed descr for a header-less buffer (`lower_vable.rs`).** Correct. `is_gc_managed` threads from `CanonicalBhDescr::Array` through `DispatchArrayDescrKey` to the runtime descr, and `ArrayPtrInfo::make_guards` (`optimizeopt/info.rs:668`) gates `GUARD_GC_TYPE` on it, reading a type-id word at `ptr - GcHeader::SIZE`. Both element sites now intern through the existing `add_raw_int_array_descr_signed` — same shape, `is_gc_managed = false` — and `residual_write_effect_info` mints its write descr to match. The `&[u8]` `env` array keeps `add_gc_int_array_descr`: its base is a green constant, so the optimizer gives it `PtrInfo::Constant` and never reaches this arm. New test `an_array_field_element_descr_is_not_gc_managed`; reverting reddens it with `interned [(1, false, true)]`.

**P2 — no field-type check on `array_fields`.** Correct — `int_fields` has one in `field_scalar_tokens` and the array vocabulary had none. `emit_array_field_base` now emits `const _: fn(&Struct) -> Elem = |__s| unsafe { *__s.field };`, which names the *pointee* so it accepts `*const T` as well as `*mut T`. Verified: changing a fixture to `ByteStack::data => u16` over a `*mut u8` field now fails to compile with `E0308` instead of silently taking its stride from the declaration.

### Not changed here

The parity review's §2/§3 make a deeper point than the descriptor flag: upstream lowers native C-array pointers through `raw_load_*`/`raw_store_*`, not GC-array operations. Switching `array_fields` to those ops is a different change — `raw_load_i` takes a byte-offset effective address rather than a scaled index, and the `residual_writes` element write-set machinery is built on array descrs. Marking the descriptor raw removes the unsound guard, which is what the inline finding asked for; the op-family divergence is separate and still open. §3's pre-existing `&[u8]` case is untouched for the reason above.

`cargo test -p majit-metainterp --features dynasm` is green (1514 + all integration binaries). `cargo check --workspace --all-targets` and `cargo fmt --check` are clean. The `pyre/check.py` status is unchanged from the previous comment and I have not re-run it on this commit.

— *commented by Claude*

PR #1214, opened 2026-08-14
jit: fold census + kill switch, symbolic-fnaddr naming, fat-pointer publication removal, and two crashes it exposed

youknowone:
Nineteen commits on the walker-descent line: diagnostics that made the
remaining blockers measurable, the fnaddr-table corrections that came out of
that, and two crash fixes the new instrumentation exposed.

## The two crashes

Both were found by suppressing a hand-written trace fold, which routes the case
to the generic residual. A crash there is not the suppression's bug — it is a
pre-existing unsoundness the fold was hiding.

**`interpreter: root the args and the exception in exc_constructor!`** — the
macro-generated public exception constructors held two live values across two
allocations with no shadow-stack roots. `args` is the caller's raw slice and
`args.to_vec()` ran *after* `w_exception_new_empty`; a collection there
relocates the young arguments and updates the caller's rooted slots but not the
slice, so the copy carried pre-move addresses into the fresh args list. That
produces a capacity-1 items block whose `items[0]` points at a forwarded
object, which the collector later reads as a header. `exc` was separately
unrooted across the arg-list build — a distinct hazard, since `exc` is
non-moving oldgen and can be swept rather than moved. `w_exception_new` and
`w_exception_new_wtf8` already root the value they hold; the macro did not.

Measured with both binaries built at the same base and differing only by this
file: `PYRE_FBW_NO_SPECIALIZE=binary_op_int` on
`pyre/bench/synth/subscr_user_getitem_inline.py` gives **20/20 SIGABRT before,
0/20 after**, and the control's panic names the object the diagnosis pointed at
(`holder_type_id=9` = `ItemsBlock`, `holder_offset=8` = `items[0]`).

**`interpreter: guard the null callable in classify_callable's TypeError
path`** — the not-callable arm read `(*(*callable).ob_type).name` through
`unwrap_or_else`, dereferencing a null `callable`. `typedef::type` returns
`None` both for a null pointer and for a non-null unregistered type, so the
`None` arm cannot tell them apart. Branching on `is_null()` turns an
empty-stderr segfault into a traceback that names the failing expression. The
operand-window defect behind it is diagnosed but not fixed here.

## Diagnostics

`jit: count and optionally suppress the walker specializations` adds
`PYRE_FBW_SPEC_CENSUS` and `PYRE_FBW_NO_SPECIALIZE=<labels|all>` over the 56
fold rows. Silent when unset. With all folds suppressed the corpus still runs
402/408, so the folds are not load-bearing for the overwhelming majority — but
suppression *raises* the abort count (79 → 152) because a fold blocks the
descent from being attempted rather than replacing it.

`jit: record what each symbolic fnaddr stands for and carry it to the abort`
makes the abort name its blocker instead of a raw hash, which is what turned
the blocker census from hashes into a work list.

## fnaddr table

`jit: stop publishing fat-pointer-argument helpers in the trace fnaddr table` —
a published helper whose argument is a two-word fat pointer is uncallable
through the one-register-per-slot residual ABI; one word is passed and two are
read. The symbolic abort was the only thing preventing those calls, so any
change that widens descent arms them.

The `w_none` / `shadow_stack_cell` / shadow-stack-bracket publications and the
`abs` gateway go the other way, adding addresses that are ABI-clean.

## Not wired

`jit: add a ConstStr op kind and an unwired str_const_fold pass` lands the fold
for synthetic `__str_const` calls with no caller. Wiring it without a one-word
string representation turns a string constant into a traced value no
string-taking consumer can read; that was observed as a silent wrong answer in
`pickle._dumps`. It stays dormant deliberately.

## Verification

408/408 `pyre/bench/synth` fixtures exit 0; executed descent aborts 79
occurrences / 14 builtins, unchanged from the base. The corpus gate was last
run before this branch was rebased onto the current `main`; it is being re-run
at the new base and I will post the result.

## Summary by CodeRabbit

* **New Features**
  * Improved handling and preservation of string constants during translation.
  * Added optional JIT specialization census reporting and diagnostics.
  * Expanded diagnostics for unsupported calls and translation failures.
  * Added symbolic function-address reporting for pipeline analysis.

* **Bug Fixes**
  * Improved `abs()` compatibility and error handling.
  * Corrected optional size handling across I/O operations.
  * Fixed zero-sized structure field emission.
  * Improved callable error messages and runtime memory safety.
  * Preserved updated values when restoring JIT execution state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2056,6 +2056,7 @@ dependencies = [
  "quote",
  "rustpython-compiler",
  "rustpython-compiler-core",
+ "rustpython-wtf8",
  "serde",
  "serde_json",
  "syn",
--- a/majit/majit-translate/Cargo.toml
+++ b/majit/majit-translate/Cargo.toml
@@ -42,6 +42,7 @@ libc = { workspace = true }
 # Critically this is NOT a dependency on pyre: rustpython-compiler-core is
 # external to both majit and pyre. The `rpython/` ⊥ `pypy/` boundary holds.
 rustpython-compiler-core = { workspace = true }
+rustpython-wtf8 = { workspace = true }
 # `md-5` mirrors upstream `hashlib.md5(...).hexdigest()` used by
 # `rpython/translator/backendopt/stat.py:48` to fingerprint each
 # graph's `co_code` when `print_statistics(save_per_graph_details=...)`
--- a/majit/majit-translate/src/codegen.rs
+++ b/majit/majit-translate/src/codegen.rs
@@ -171,6 +171,7 @@ mod tests {
                 main_jitcode_index: 0,
             }],
             jitcodes: vec![main_jitcode],
+            symbolic_fnaddr_paths: Vec::new(),
             jitcodes_by_path: indexmap::IndexMap::new(),
             indirectcalltarget_indices: Vec::new(),
             insns: indexmap::IndexMap::new(),
--- a/majit/majit-translate/src/codewriter/assembler.rs
+++ b/majit/majit-translate/src/codewriter/assembler.rs
@@ -1495,6 +1495,23 @@ impl Assembler {
                 let opnum = self.get_opnum(&key);
                 state.code[startposition] = opnum;
             }
+            OpKind::ConstStr(bytes) => {
+                let p = crate::translator::rtyper::lltypesystem::rstr::const_str_cache_llstr(bytes)
+                    .expect("prebuilt STR constant must materialize");
+                let const_value = crate::flowspace::model::ConstValue::LLPtr(Box::new(p));
+                let idx = self.emit_const_r(&const_value, state);
+                state.code.push(idx);
+                argcodes.push('r');
+                if let Some(result) = op.result.as_ref() {
+                    argcodes.push('>');
+                    let (reg, kc) = self.lookup_reg_with_kind_var(result, regallocs);
+                    argcodes.push(kc);
+                    state.code.push(reg);
+                }
+                let key = format!("ref_copy/{argcodes}");
+                let opnum = self.get_opnum(&key);
+                state.code[startposition] = opnum;
+            }
             OpKind::ConstRefNull => {
                 let const_value = crate::flowspace::model::ConstValue::LLAddress(
                     crate::translator::rtyper::lltypesystem::lltype::_address::Null,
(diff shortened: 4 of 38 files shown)


youknowone:
Re-gate at the rebased base (`30932b50068`, on `origin/main` `08f1b397b4d`), as promised in the description.

The rebase pulled in `majit/majit-trace/src/counter.rs`, `pyre/check.py` and three jitstats baselines that the original gate had not compiled, so the earlier result did not carry over and the corpus was re-run from a fresh `extract-llbc` and release build.

```
extract-llbc  RC=0
release build RC=0
fixture run status:   408  exit=0
executed descent aborts — 79 occurrences, 14 builtins
  len         28 occurrences / 17 fixtures   __str_const::__len__
  isinstance  22 occurrences /  8 fixtures   pyre_interpreter::baseobjspace::isinstance
```

Identical to the pre-rebase result, so nothing in this branch interacts with what the rebase brought in.

— *commented by Claude*

youknowone:
Pushed `3d93f070fec` — the root cause behind the null-callable guard earlier in this branch.

`vstack_boxes` has two producers, the executed `setarrayitem_vable_r` store and the pc-derived reconstruction in `reconcile_vstack_at_boundary`, with no precedence rule between them. Upstream never needs one: `_opimpl_setarrayitem_vable` (`pyjitpl.py:1245`) is the only writer of `virtualizable_boxes` and indexes off the descr, and a block transition is `self.pc = target` (`pyjitpl.py:507`) with nothing saved or reconciled. The reconstruction is a consequence of walking flattened jitcode, so there is no upstream code to port here — only the principle that an execution-derived store is authoritative and a pc-derived one is not.

The out-of-order region restores its snapshot when the walk returns to the same py_pc at the same depth, and that coordinate does not establish the premise the restore states ("no Python opcode retired in between") — the arming boundary can be reported mid-opcode, so the snapshot can hold a state no opcode ever left behind.

Two designs were tried against the evidence and rejected before this one:

- *detect the NULL and reorder the operands* — the slot is not unwritten, it is overwritten, so this papers over a mirror that is confidently wrong;
- *if the mirror changed since the save, keep the current boxes* — the predicate is vacuously true, because the arming boundary always runs `ShadowReseed`, which clears and resizes the whole vector. It would disable the restore in every region, and three windows in the passing run show the restore legitimately recovering slots the hole-fill declined — including the `leaf(i)` result that is `seen.add`'s own argument.

Hence per-slot merge rather than either side whole: one window can hold both an executed store and a slot the reseed dropped.

```
PYRE_FBW_NO_SPECIALIZE=load_bound_method_attr … getframe_inlined_callee_own_frame.py
  before  5/5 exit 1   TypeError: <NULL> object is not callable
  after   5/5 exit 0   199990000 [('leaf', 'main')]   (stderr empty)
no selector (control)   3/3 exit 0, byte-identical stdout
408 corpus              408 exit=0, aborts 79 occurrences / 14 builtins — zero movement
```

Caveat worth stating: there is no unit-test coverage of the restore path — every `vstack_reorder` occurrence in `jitcode_dispatch/tests.rs` is a `None`/`u32::MAX` initializer — so this change is graded by the corpus alone.

— *commented by Claude*

PR #1218, opened 2026-08-14
check.py: raise the major-collection threshold pin past str_fstring

youknowone:
Three `str_fstring` overlays held what looked like a host disagreement. It was
not one, and the fix is a pin rather than a per-platform baseline.

## The split was not a platform property

`pyre_env` pins `PYPY_GC_MIN` so a fixture's collection schedule is a property
of the tree rather than of the machine. The criterion its comment states is
"past **every** fixture's working set", but 256MB was measured against the two
fixtures that motivated it, and `str_fstring` — six hot loops merged into one
run — sits above it.

At 256MB it reads `guard_failures` 658 or 659 as a function of nothing but how
much the process allocated before its loops got hot. A perturbation with **no
semantic content** moves it, on one machine:

| environment | `guard_failures` |
|---|---|
| 9 variables (`env -i` + the pins) | 658 |
| 49 variables | **659** |
| 69 / 109 / 209 variables | 658 |

Non-monotonic in the count, so it is a knife edge and not a size threshold;
padding a *single* variable to 16000 characters moves nothing. That is the
preamble-vs-body crossing `pyre_env` already describes, and it is exactly what
the overlays were holding — each runner has its own startup allocation volume,
so each landed on its own side.

## 512MB removes the event instead of relocating it

`str_fstring` then reads 657 at every nursery from 4MB to 32MB **and** under
every perturbation of that volume — the same double invariance that already
stood for `recursive_call_frame_relocation` and `closure_per_call`.

Eight other fixtures move with it, none a knife edge at either value: seven lose
a crossing (`arith_int_bool` 2214 → 2212, `build_set_hashability` 4 → 2,
`bytes_split_whitespace_maxsplit` 3 → 2, `delete_negative_open_slice_hot`
1405 → 1404, `instance_dict_reassign` 2 → 1, `newslice_step_hot` 5 → 3,
`unpack_ex_hot` 3 → 2) and `inlined_helper_mutation` 3 → 4 relocates one. **No
`loops_compiled` or `bridges_compiled` moves anywhere in the suite**, which is
the signature that only the collection schedule changed.

## The wasm guest converges too

The old comment left this open: at 256MB the guest still crossed where the
native backends did not, holding `recursive_call_frame_relocation` 638 and
`closure_per_call` 415 against their 636 and 414. Both now read the native
number, and `str_fstring` is a single 657 across all three backends.

## Cost

Paid by the one fixture that was over the old pin: `str_fstring` peak RSS
519MB → 618MB. Every other fixture is flat to within a megabyte
(`closure_per_call` 179 → 180MB, `recursive_call_frame_relocation` 159 → 160MB).
No fixture declares a `max-rss-mb` gate.

## Scope of the diff

28 baselines re-recorded, one line each. `--snapshot` additionally rewrote 1124
files with a missing-field top-up (`retraces_compiled=0` and friends); those are
reverted, because `_jit_stats_change` reads `old_fields.get(field, "0")` — an
absent baseline field already compares as 0, so the top-up changes no gate.

The three `str_fstring.*.darwin` / `.win32.github-actions` overlays are deleted
rather than re-recorded. No fixture carries an overlay now.

## Verification

`python3 pyre/check.py` (plain gate run, no `--snapshot`): dynasm 425/425,
cranelift 425/425, wasm 418/418. Locally that is darwin only.

**The prediction CI tests** is that ubuntu and windows also read 657, from a
single shared baseline. It rests on the crossing being gone rather than
relocated, and the pre-change evidence supports it: all three runners read
either 658 or 659 — inside the ±1 knife-edge band, never a structurally
different number — with `loops_compiled` 6 and `bridges_compiled` 3 identical
everywhere. The falsifier is a runner reading 657 ± anything other than 0; that
would mean the fixture's count is sensitive to the per-target lifted set rather
than to the collection schedule, and the diagnosis above would be wrong.

## Second commit: the wasm overflow check against a constant

The three wasm perf-ratio failures on ubuntu (`raise_catch` 3.4x,
`synth/global_quasiimmut_invalidation` 5.3x,
`synth/short_circuit_value_kept_stack` 3.2x, gate 3.0x) are inherited — `main`
reads 3.2x / 5.1x / 3.4x on the same three.

Two candidate causes were measured and dropped. Compile latency: `compile_ms` is
2.3–15.0ms against execution times of hundreds of ms. Allocation: the wasm
backend really does skip the GC rewrite pass that dynasm and cranelift run
(`runner.rs:1839`, `compiler.rs:8586`), so upstream's `gen_malloc_nursery`
fusion (`rewrite.py:879`) never applies and each `New` inlines its own bounds
check whose slow path reloads every live Ref — but **both failing synth benches
report `gc_minors=0`**, so they allocate nothing and that is not their
mechanism. Recorded here because the missing fusion is real and still worth
fixing; it just is not this.

What did hold is the overflow check. `INT_ADD_OVF` / `INT_SUB_OVF` expanded to
the sign-comparison form `((a ^ r) & (b ^ r)) >>s 63` for every operand pair —
about ten instructions where x86 uses one `jo`. Both benches run ~8 add/sub per
iteration, nearly all against a constant, and against a constant the predicate
is a single signed comparison against a bound folded at codegen time, which also
drops the dependency on the result.

Measured per-iteration cost (slope across two `N`, so fixed cost is subtracted):

| bench | wasm before | wasm after |
|---|---|---|
| `global_quasiimmut_invalidation` | 20.49 ns/it | 18.11 ns/it |
| `short_circuit_value_kept_stack` | 15.84 ns/it | 13.05 ns/it |

`ovf_const_bound` is checked against `checked_add`/`checked_sub` over the
extremes, and the emitted traces are executed for the same cases — including
`a + i64::MIN` and `a - i64::MIN`, where the bound is closest to overflowing
itself. All 425 synth fixtures keep their counters on every backend, which is
the evidence the change is semantics-neutral rather than merely green.

**It was not enough, and CI says why.** After this change ubuntu reads
`raise_catch` 3.4 → 3.3x, `global_quasiimmut_invalidation` 5.3 → 5.2x,
`short_circuit_value_kept_stack` 3.2 → **3.4x** — net zero, i.e. noise. The
change is kept because it is unambiguously fewer instructions and is worth
−12%/−18% on aarch64, but it does not move this gate.

The exec times say the gate is not about guest code length at all:

| | darwin (aarch64) | ubuntu (x86_64) |
|---|---|---|
| dynasm exec | 0.11s | 0.13s |
| wasm exec | 0.26s | **0.66s** |
| ratio | 2.4x | **5.2x** |

dynasm is within 18% across the two hosts; **wasm alone is 2.5x slower on
x86_64**. Two explanations were measured and killed: trace-module compilation is
`compile_ms` 11.8 / 18.2 / 2.2ms against 260ms+ runtimes (under 5%, and it sits
inside "exec" because it happens mid-run, so the startup subtraction never
touched it), and startup is measured live per host anyway (wasm 0.023s on ubuntu
vs 0.051s on darwin — faster there).

What is left is register pressure from fail-args live across guards. wasmtime
has ~16 GPRs on x86_64 with several reserved, against 31 on aarch64, and a
clause-count sweep already shows wasm's marginal cost per guard-carrying clause
growing superlinearly where dynasm's is flat — 3.37 → 4.30 → 6.11 ns against
1.10 → 1.40 → 1.52 ns — measured on aarch64, which has the budget to hide it.

That makes this gate **ungradeable on an aarch64 workstation**: a change can be
strictly fewer instructions, measure as a real win locally, and move ubuntu by
nothing. Targeting it means store-on-def for fail-args plus a static per-guard
fail-arg→home map (the role of `fail_locs`, `llsupport/assembler.py:24`), which
drops guard liveness to zero — a redesign of the spill path rather than an edit,
and one that needs an x86_64 host to grade.

## What this does not fix

`pyre/check.py (windows-latest)` has nine further divergences
(`mutate_then_raise_caught`, `exception_traceback_*`, `named_reraise_sibling_hot`,
…), all exception/raise paths, `loops_compiled` +1/+2 with dynasm ≡ cranelift
byte-identical. They are **not** this class: locally they are invariant to
nursery 1–64MB, to `PYPY_GC_MIN`, and to the perturbation above.

The live lead on the windows nine is that the front-end lifts a different
program per target — `all_descrs` reads 5456 on windows, 5489 on ubuntu and 5540
on macos (matching a local darwin build exactly), and CI extracts LLBC in three
separate `prepare-charon-llbc-{linux,macos,windows}` jobs, so the lifted set is
per-target by construction. That needs a census of the lifted set, not a GC knob.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file truncation behavior on Windows, including clearer error handling and refreshed file metadata.
  * Corrected integer overflow detection for constant-based addition and subtraction, including boundary and wrapping cases.
  * Improved consistency of runtime behavior across native and WebAssembly execution.
  * Avoided unsupported file-seekability comparisons in Windows-specific compatibility checks.

* **Performance**
  * Adjusted garbage-collection startup behavior to reduce unnecessary collection-boundary effects, with a potential increase in peak memory usage for some workloads.


youknowone:
## CI outcome: the prediction held on all three runners

The change predicted that all three platforms would read `str_fstring`
`guard_failures=657` from a single shared baseline, with the falsifier being any
runner reading something else. No runner did.

| `pyre/check.py` | main (`08f1b397b4d`) | this PR |
|---|---|---|
| macos-latest | dynasm 1 / cranelift 1 failed (`str_fstring`) | **SUCCESS — fully green** |
| ubuntu-24.04 | dynasm 1 / cranelift 1 failed (`str_fstring`), wasm 3 failed | dynasm **pass**, cranelift **pass**, wasm 3 failed |
| windows-latest | 10 failed per native backend | **9 failed** — `str_fstring` passes |

`str_fstring` is gone from every failure set, and no fixture carries an overlay
any more.

### The failures that remain are inherited, not introduced

- **ubuntu wasm ×3** — perf-ratio gates, not jit-stats: `raise_catch` 3.4x,
  `synth/global_quasiimmut_invalidation` 5.3x, `synth/short_circuit_value_kept_stack`
  3.2x against a 3x gate. main reads the same three at 3.2x / 5.1x / 3.4x, so
  this PR moves them within noise.
- **windows ×9** — `mutate_then_raise_caught`, `exception_traceback_*`,
  `named_reraise_sibling_hot`, … all exception/raise paths, `loops_compiled`
  +1/+2, dynasm ≡ cranelift byte-identical. Present on main. They are **not**
  the class this PR fixes: locally they are invariant to nursery 1–64MB, to
  `PYPY_GC_MIN`, and to the startup-allocation perturbation that moves
  `str_fstring`.
- **CPython suite (gate)** — red on main since 2026-08-13, unrelated to this PR.

### Note on the windows nine

The live lead is that the front-end lifts a different program per target:
`all_descrs` reads 5456 on windows, 5489 on ubuntu, 5540 on macos (matching a
local darwin build exactly), and CI extracts LLBC in three separate
`prepare-charon-llbc-{linux,macos,windows}` jobs — so the lifted set is
per-target by construction. Distance from the recording host tracks the
divergence count. That needs a census of the lifted set rather than a GC knob,
and is out of scope here.

— *commented by Claude*

youknowone:
@copilot resolve conflict

PR #1235, opened 2026-08-15
majit: lower discarded inline-helper results, and give the degraded-arm registry a denominator

youknowone:
Two gaps in the `#[jit_interp]` / `#[jit_inline]` macro surface, found by auditing what a consumer has to hand-write around.

## 1. `inline_int` / `inline_ref` / `inline_float` in statement position

Statement position and value position are lowered by two different functions. `lower_value.rs` has an arm for the three result-returning inline policies; `lower_stmt.rs` had one only for `inline_void`, so a helper registered `inline_int` and called for its effect reached the trailing `_ => return None`.

The two macro surfaces then diverge:

* `#[jit_inline]` reports `could not lower this helper into JitCode` and fails the build.
* `#[jit_interp]` degrades the enclosing dispatch arm to an abort stub, records it through `record_degraded_dispatch_arm`, and **builds clean** — every trace reaching that opcode aborts, once per threshold, forever, and the helper's stores never enter the trace.

`explicit_call_emits_post_live` already returns an answer for these three kinds, so the accounting said the policy was handled while the lowering did not handle it. The workaround is to bind the result to a `let _x` nothing reads, which makes the same source lower — so a machine that hit this looks like one that never did.

The new statement arm uses `alloc_reg()` for the destination the sub-jitcode's typed return needs, the way the discarded `ResidualInt` family below it already does; the rest is the value-position lowering.

## 2. A denominator for `degraded_dispatch_arms()`

`degraded_dispatch_arms()` is a numerator. An empty result reads as "no arm degraded" and as "no portal was ever built" at once, and only the first is a pass — the second is the same shape as the defect the registry exists to report, one level up. Every consumer gating on it therefore supplies its own proof that the portal was installed, and each supplies a different one.

`emit_dispatch_chain` now stages `record_dispatch_arm_census(state_type_name, n)` from the same admission test its emission loop uses, so an arm is counted exactly when a body is emitted for it. New API: `DispatchArmCensus`, `record_dispatch_arm_census`, `dispatch_arm_census`, and `assert_no_degraded_dispatch_arms(interp)`, which fails with one message when the portal was never installed and a different one when an arm degraded.

## Verification

* `jit_interp_discarded_inline_result.rs` — a `#[jit_interp]` machine with a discarding arm, a result-binding control arm, and a deliberately unlowerable arm as the registry's denominator. With the statement arm removed both subject assertions fail: `degraded=["OP_DISCARD_POP", "OP_ENCLOSED_BREAK"]`, and one helper splice instead of two.
* The splice count comes from the portal's sub-JitCode list, not from scanning bodies for `BC_INLINE_CALL`. Jitcode operands are bytes too, so a byte scan reports two inline calls even with the arm degraded to `[BC_ABORT]` — the first version of that assertion passed for exactly that reason.
* `jit_interp_inline_helper_typed_return.rs` — the `#[jit_inline]` half, where the same input previously failed the build.
* `jit_interp_ref_state_field.rs` — the success case for `assert_no_degraded_dispatch_arms`; a gate that panicked unconditionally would satisfy the two failure cases above.
* `cargo test -p majit-metainterp --features dynasm --no-fail-fast`: 29/29 binaries green. `cargo check --workspace --all-targets` RC=0. `cargo fmt --all --check` clean.

— *authored by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed discarded inline helper calls so they execute correctly in interpreted dispatch paths.
  * Improved validation for reference fields and pool-array declarations, including accurate item and length-field layouts.
  * Improved detection and reporting of degraded dispatch paths, unused declarations, and conflicting structure layouts.

* **Tests**
  * Added coverage for inline calls, stack mutations, field usage, pool-array layouts, dispatch tracking, and related diagnostics.

### DIFF
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/api.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/api.rs
@@ -531,9 +531,39 @@ pub(crate) fn generate_inline_helper_jitcode_with_calls(
     maybe_dump_liveness(&helper_name, &lowerer.op_metadata);
     let liveness_prebuild =
         liveness_prebuild_tokens(&lowerer.op_metadata, &lowerer.inline_liveness_prebuild);
+    // The body has lowered, so the consulted-key set is final. Reported per
+    // HELPER, not per machine: each `#[jit_inline]` carries its own
+    // `int_fields` / `ref_fields`, and a key one helper consults says nothing
+    // about the helper next to it.
+    //
+    // This surface is where the population lives. A `#[jit_interp]` machine
+    // declares a handful of keys; a consumer's helpers repeat theirs at every
+    // site, and a key that matches nothing at one site matches nothing at
+    // dozens.
+    let unconsulted_declarations = match inline_config.as_ref() {
+        Some(config) => {
+            let consulted = config.consulted_field_keys.borrow();
+            let mut keys: Vec<&String> = config
+                .int_fields
+                .keys()
+                .chain(config.ref_fields.keys())
+                .filter(|key| !consulted.contains(*key))
+                .collect();
+            keys.sort();
+            keys.dedup();
+            let records = keys.into_iter().map(|key| {
+                quote! {
+                    majit_metainterp::record_unconsulted_field_declaration(#helper_name, #key);
+                }
+            });
+            quote! { #(#records)* }
+        }
+        None => quote! {},
+    };
     let statements = lowerer.statements;
     Ok(Some(InlineHelperJitCode {
         body: quote! {
+            #unconsulted_declarations
             #(#statements)*
         },
         return_reg,
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
@@ -2498,6 +2498,22 @@ pub(super) fn lower_dispatch_chain(
         lowerer.emit_jump(&default_label);
     }
 
+    // The denominator for `record_degraded_dispatch_arm` below, staged from the
+    // same loop's own admission test so the two cannot drift: an arm is counted
+    // here exactly when the loop emits a body for it.  Without it an empty
+    // degraded registry reads as "nothing degraded" and as "no portal was
+    // built" at once, and only the first is a pass.
+    {
+        let census_interp = config.state_type_name.clone();
+        let census_arms = classified_arms
+            .iter()
+            .filter(|arm| !matches!(arm.pat, Pat::Wild(_)) && !is_lowercase_binding_pat(&arm.pat))
+            .count();
+        lowerer.emit_aux(quote::quote! {
+            majit_metainterp::record_dispatch_arm_census(#census_interp, #census_arms);
+        });
+    }
+
     for (arm_idx, arm) in classified_arms.iter().enumerate() {
         // `_` wildcard: skip here; handled by the default GOTO below.
         // All other patterns (including Pat::Ident like `OP_NOP`) are
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
@@ -3023,6 +3023,33 @@ pub(super) fn lower_dispatch_chain(
         }
     }
 
+    // Every arm has lowered, so the consulted-key set is final: report the
+    // declared `int_fields` / `ref_fields` keys no access site asked about.
+    //
+    // Reported rather than rejected, and the reason is the degraded arm. An
+    // arm that refused to lower never reached its field accesses, so a key
+    // used only there is unconsulted through no fault of the declaration —
+    // failing the build on it would turn one refusal into two. The consumer's
+    // gate reads this alongside `degraded_dispatch_arms` and can tell them
+    // apart; a compile error could not.
+    {
+        let interp = config.state_type_name.clone();
+        let consulted = config.consulted_field_keys.borrow();
+        let mut unconsulted: Vec<&String> = config
+            .int_fields
+            .keys()
+            .chain(config.ref_fields.keys())
+            .filter(|key| !consulted.contains(*key))
+            .collect();
+        unconsulted.sort();
+        unconsulted.dedup();
+        for key in unconsulted {
+            lowerer.emit_aux(quote::quote! {
+                majit_metainterp::record_unconsulted_field_declaration(#interp, #key);
+            });
+        }
+    }
+
     // After all arm guards, the default/exit path: unconditional GOTO.
     // default_label is bound at the typed-return emission site in
     // lower_dispatch_body (Task 1.7).
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/lower_stmt.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/lower_stmt.rs
@@ -1494,6 +1494,52 @@ impl<'c> Lowerer<'c> {
                     );
                     self.emit_op(OpMeta::live_marker(), post_live);
                 }
+                // A result-returning inline helper whose result the statement
+                // discards. The sub-jitcode still ends in a typed return
+                // opcode, so it needs a destination even though nothing reads
+                // it; `alloc_reg` mints one, the same way the discarded
+                // `ResidualInt` family below does. Everything else is the
+                // value-position lowering verbatim.
+                //
+                // Without this arm the call reaches `_ => return None` and the
+                // statement is dropped: `explicit_call_emits_post_live` already
+                // answers for these kinds, so the accounting says the policy is
+                // handled while the lowering says it is not, and the effect the
+                // helper performs leaves the trace with no diagnostic. The
+                // workaround is to bind the result to `let _x = ...`, which is
+                // a source change the declaration does not ask for.
+                crate::jit_interp::CallPolicyKind::InlineInt
+                | crate::jit_interp::CallPolicyKind::InlineRef
+                | crate::jit_interp::CallPolicyKind::InlineFloat => {
+                    let result_kind = binding_kind_for_inline_policy(kind)
+                        .expect("the arm's own patterns are the inline result policies");
+                    let throwaway_reg = self.alloc_reg();
+                    let builder_path = inline_builder_path(&call.func)?;
+                    let prebuild_path = inline_prebuild_path(&call.func)?;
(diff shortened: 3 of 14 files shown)


youknowone:
Added a third commit: `ref_fields` entries now carry a compile-time witness.

A `ref_fields = { Struct::field => Pointee }` entry is trusted twice and verified nowhere — the read becomes `getfield_gc_r` into the ref bank, and the binding's `struct_type` is what the next hop resolves `offset_of!` against. `field_scalar_tokens` now emits a witness for these keys, next to the one `int_fields` keys already got, admitting `*mut Pointee`, `*const Pointee` and `usize`.

Where it adds coverage was measured with the witness on and off, not assumed:

| surface | drifted pointee on a `*mut T` field |
|---|---|
| `#[jit_inline]` | already rejected — the concrete rewriter types the loaded value against the declaration, `E0308: expected .., found *mut Holder` |
| `#[jit_interp]` state field | **compiled clean**, faulted at run time. With the witness: `E0277: the trait bound *mut Holder: __MajitRefField is not satisfied` |

So the interp path is what this closes; on the inline path it is a second, earlier line.

The undeclared half is deliberately left open and the doc comment says so: a field named in no map reads into the Int bank, and stable Rust cannot assert that a type is *not* a pointer. Catching it needs a same-offset `is_ref` disagreement to reach `Assembler::register_struct_layout`, which today merges by offset alone and keeps whichever registration arrived first.

**Downstream check.** The one real consumer of these macros lives in a separate repository and pins majit by revision. It was built against this branch through a `[patch]` override: `cargo test -p aheui-jit` is fully green, including a `ref_fields` entry whose field is a `usize` carrier — which is why the witness admits `usize` rather than the two pointer spellings alone.

— *commented by Claude*

youknowone:
Two more, plus a correction to what the last one's doc claimed.

**Rebased onto `origin/main`** — the three earlier commits are the same patches (`8175b4e0`, `0cb8a70f`, `a2b8f4b4` by patch-id), now on top of #1212.

---

### `register_struct_layout` reports a registration that disagrees (`ad2286c9eb6`)

`ref_field_witness_tokens`' doc named this as the second line for a pointer field left out of one declaration's `ref_fields`. **That was wrong, and the code says so:** `struct_size_specs` is a field of `JitCodeBuilder`, and every jitcode calls `JitCodeBuilder::new()` — the portal, each `#[jit_inline]` helper, each sub-builder. Two *declarations* never land in one map. What does meet is two emit sites inside one jitcode, reaching the same member by two lowering paths. The doc now says that instead.

The check itself could not key on the offset alone. `an_ambiguous_offset_is_arbitrated_by_name_and_otherwise_declined` already pins that a flattened inline aggregate and its first leaf share an address — *"7 of 1714 submitted field specs share an offset with a sibling"* — so an offset-keyed conflict test is an instant false positive on the existing corpus. Keyed on `(offset, name)`, a hit splits three ways: the same field again (the common case, and what the dedup exists for), **Redescribed**, **DroppedSibling**. It reports; it does not arbitrate, so the merge is unchanged.

**The denominator is what makes the result readable, and my first one was wrong.** Counting *registrations* would have passed: a field at an offset nobody has registered yet is appended without comparing anything, so a corpus can submit thousands and never exercise the check. Counting *comparisons* — fields that arrived at an already-recorded offset — gives:

| corpus | comparisons ran | conflicts |
|---|---|---|
| 4 pyre benches (`nbody`, `fannkuch`, `inline_helper`, `nested_loop`) | **none** | 0 — vacuous |
| aheui `logo` | yes | 0 |

So a conflict count read off the pyre benches says nothing, and only the denominator reveals that. The check's live corpus is the macro path.

Falsification: with the recording disabled, 3 of the 4 new tests go red; the fourth is the negative control and correctly stays green.

---

### A `pool_arrays` declaration states its array's offset and length word (`3308129fdde`)

`add_ptr_array_descr` hard-coded `base_size = size_of::<usize>()` and `len_offset = Some(0)`, so the declaration silently required the consumer's struct to be `{ len, items.. }`. No `offset_of!` was emitted anywhere, so the assumption never met the layout — reordering that header, or dropping it, compiled clean on both sides and left the element read at `base + 8 + i*8`.

The grammar is now `<ref>.<items>[<len>] => <getter> [-> Elem]`. Both numbers are `offset_of!` on the declaration's own field names, plus two witnesses for what an offset cannot cover: that `items` holds pointer-width elements, and that `len` is the `usize` the lendescr reads it at.

`len_offset = None` is a supported shape rather than a degraded one — `ArrayPtrInfo.make_guards` skips the short preamble's `ARRAYLEN_GC` for a descr without a lendescr, so a plain fixed-size inline array declares no length. **The doc claiming a lendescr was required, and that the GC rewrite would panic without one, describes the state before that gate; it is removed rather than restated.**

Armed control on the real consumer: with `pools` moved ahead of the length word, the declared build stays byte-identical (996310 B, md5 `7fcdbfff0af449c4283c008e3ca317ce`, exit 42, jit == naive) and the hard-coded build **hangs with 0 bytes of output**.

The new test's machine puts its items behind a two-word pad. Its run assertion is gated on a compile counter *first*, and that is not decoration: the concrete fallback indexes the real field whatever the descr claims, so my first version of that test **passed on the deliberately broken build** — zero traces had compiled. With `can_enter_jit!` on the back edge it reads `left: 2180, right: 6000` under the old hard-code.

---

### Gates

- `cargo test -p majit-metainterp --features dynasm --no-fail-fast` — 31/31 binaries green
- `cargo test -p majit-macros --lib --features majit-metainterp/dynasm` — 116 passed
- `cargo check --workspace --all-targets` — clean; `cargo fmt --all -- --check` — clean
- `python3 ./pyre/check.py --backend dynasm` — **ALL PASSED, 435/435**
- aheui built and tested against this tree through a `[patch]` config; `logo` byte-identical, `jit == naive`

### Not done, and why

The audit's fourth item — a knob withholding a majit-fed root family so a consumer's GC root-walk oracle can be shown able to fail — **is dead**. The control already exists one layer down in the consumer (`AHEUI_GC_DROP_JIT_ROOTS`), and the surface the item named (`majit-gc/src/collector.rs`) is inert for that consumer anyway: it installs its own nursery allocator whose collect methods are empty, so MiniMark's root phases never manage those objects. A knob there would report "did not arm" while never having touched anything.

— *commented by Claude*

youknowone:
### A declared field key no access site consulted (`99d05d526a8`)

`int_fields` and `ref_fields` are consulted by `"StructType::field"`, built from the **declared** type of the base an access goes through. A key naming a struct nothing is typed as matches nothing — so it emits no descr width, and, because `field_scalar_tokens` emits its `const _: fn(&S) -> T` witness *only on a match*, no check of what it claims about the field either.

That second half is what makes this worth a mechanism rather than a tidy-up. **Measured on a real consumer: 36 entries declaring a `u32` field as `u8` compile clean; the same misdeclaration on a key that *is* consulted produces 13 `E0308`s.** An unconsulted entry is not merely dead — it is a false statement the build agrees to, and it is indistinguishable from a live one by reading the declaration.

Every consultation reaches the maps through `field_scalar_tokens`, so the consulted set is recorded in that one place and diffed against the declared keys once lowering finishes.

**Reported, not rejected** — deliberately unlike the `pool_arrays` base-name check sitting next to it. A degraded arm never reaches its field accesses, so a key used only in that arm is unconsulted through no fault of the declaration. Rejecting would turn one refusal into two. The gate's message carries the degraded arms alongside, so a reader can tell "stale" from "the only arm that used it refused to lower".

**Both surfaces, and they had to be done separately.** The first version recorded only from the `#[jit_interp]` portal. Pointed at the consumer, it reported exactly **one** key — while the 36 entries that motivated the whole check sat on the `#[jit_inline]` side, untouched. Same asymmetry these two macros produce on every input, reproduced inside the fix for it. With both surfaces recording, that consumer reports **61**, including two classes I did not predict:

- `stack_swap` does not consult its **own** `Stack::size` (swap never touches the count);
- `stack_push` / `stack_dup` / `queue_dup` declare `Node::next` and never consult it.

A helper reports under its own name, since each `#[jit_inline]` carries its own maps and a key one helper consults says nothing about the helper beside it.

### Gates

- `cargo test -p majit-metainterp --features dynasm --no-fail-fast` — 32/32 binaries green
- `cargo test -p majit-macros --lib --features majit-metainterp/dynasm` — 116 passed
- `cargo check --workspace --all-targets`, `cargo fmt --all -- --check` — clean
- `python3 ./pyre/check.py --backend dynasm` — **ALL PASSED, 435/435**
- consumer built against this tree through a `[patch]` config: output byte-identical (996310 B, md5 `7fcdbfff0af449c4283c008e3ca317ce`, exit 42) while the report names its 61 declarations

— *commented by Claude*

youknowone:
Two corrections to the comment above, and the list it refers to.

**The count is 60, not 61.** I miscounted the log lines.

**"Every consultation reaches the maps through `field_scalar_tokens`" is not true of `ref_fields`.** `RefFieldRewriter` reads it as well, to rewrite the concrete body's `x.field` into a deref of the raw-pointer carrier. Both walk the same source, so they agree except where the jitcode lowering stopped early and the rewriter did not — a degraded arm, which is the case the report already handles by reporting rather than rejecting. `8cf2937d573` says so in the comment; the claim was overbroad, the mechanism is unchanged.

The 60, on the consumer:

| class | count | what it is |
|---|---:|---|
| a foreign storage's `size` | 54 | `Port::size` at all 27 helpers; `Queue::size` at the 13 `ref(Stack)` helpers; `Stack::size` at the 14 `ref(Queue)` helpers |
| a helper's **own** `size` | 2 | `stack_swap`, `queue_swap` — swap exchanges two node values and never touches the count |
| `Node::next` | 3 | `stack_push`, `stack_dup`, `queue_dup` |
| `NodeJit::next` | 1 | on the `#[jit_interp]` machine |

The first class is the one the declarations were written for: a comment at each site says the three storages share the `head`/`size` prefix and the JIT reaches all of them through the `Stack` tag, "so the width is declared for every name the layout can be registered under". The layout is only ever registered under the tag the access base is *declared* as, so the other names never match.

The `Node::next` three are not what I expected and are worth stating separately, because they look load-bearing and are not. `stack_push` is

```rust
let old_head = stack.head;
let new_node = Node { value, next: old_head };
stack.head = new_node;
stack.size = stack.size + 1u32;
```

`next: old_head` is a struct-literal initialiser, which the `struct_allocs` lowering handles; there is no `.next` *access* for `ref_fields` to answer. Dropping the entry there leaves the consumer's output byte-identical (996310 B, md5 `7fcdbfff0af449c4283c008e3ca317ce`, exit 42), so it is a true positive rather than a rewriter-only use.

No deletions made — the report is the deliverable here, and what to do with the 60 is the consumer's call.

— *commented by Claude*

PR #1266, opened 2026-08-16
jit: widen the walker CALL_ASSEMBLER fold past four pyre-local declines

youknowone:
## What

A Python recursion deeper than the inline unroll bound, when it is called
from a loop body, never left the interpreter-residual path: one frame build
and one entry bridge per recursive call, for the whole run.

`try_walker_call_assembler_self_recursive`
(`pyre/pyre-jit-trace/src/jitcode_dispatch/inline_call.rs`) held four
declines with no counterpart in `_opimpl_recursive_call`
(`rpython/jit/metainterp/pyjitpl.py:1376-1436`). Each is replaced by the
narrower question it stood in for:

| was | now |
|---|---|
| callee must be self- or mutually recursive | a foreign callee is admitted when its body facts (`contains_raise` / `has_exception_table` / `has_abort_permanent`) all clear |
| caller's jitcode holds a `catch_exception` anywhere | the CALL's own `py_pc` sits inside a protected region, excluding 3.14's whole-body generator entry (`depth != 0 \|\| !lasti`) |
| `fbw_executed_body_residual` latch keyed on callee code identity | keyed on `selfrec_ca_fold_active()` scope, so a successful fold stops latching against later calls in the same walk |
| any loop-carried `InputArg` below the call | only a non-numeric one (a live iterator) |

The third gate's discriminator is what the slot holds: a numeric accumulator
(`total = (total + step(8, i)) % MOD`) keeps the fold worthwhile, a live
`FOR_ITER` iterator does not.

## Measurements

dynasm, same tree, interleaved, load 4:

| shape | before | after |
|---|---|---|
| `r(8)` x1M from a `while` | 1.21s | **0.06s** |
| `r(9)` x1M | 0.74s | **0.06s** |
| `recursive_call_frame_relocation` | 0.18s | 0.18s |
| `fib_recursive` | 0.28s | 0.28s |
| `generator_tree_recursion` | 0.15s | 0.15s |
| `foriter_call_resume_drops_iteration` | 0.18s | 0.18s |

`pyre/check.py` **436/436 on both backends**. Three jit-stats baselines
re-recorded downward (improvements):
`ca_bridge_multiframe_resume_double_call` 3437→3369, `recursion_memo_branch`
3554→3491, `recursive_forced_frame_kept_stack` 800→600.

## Second commit — the fixture

`bench/synth/recursion_past_unroll_bound_from_loop.py` holds the shape that
was uncovered: recursion two frames past `FBW_MAX_INLINE_RECURSION` with a
numeric loop-carried operand below the call. In `fib_recursive` and
`selfrec_bridge_nontail_promote` the recursion is itself the hot loop;
`recursive_call_frame_relocation` puts an iterator below the call instead.

## Third commit — comment corrections

The message of 85d16214f1a describes the foreign admission as requiring
`JC_DONT_TRACE_HERE`. It does not: the gate reads body facts only and never
consults `can_inline_callable`. Corrected in the third commit, along with two
other claims about upstream that the investigation below refuted.

## What was measured and rejected

Porting upstream's own admission condition onto this fold — it looks
obviously right and is not:

| configuration | result |
|---|---|
| uniform `!can_inline_callable` on every callee | 8 regressed benches on both backends, incl. `fib_recursive` |
| `!can_inline_callable \|\| count >= FBW_MAX_INLINE_RECURSION` | `fib_recursive` 0.28s → **2.45s** |

The reason, established afterwards: **the condition is already at upstream's
site.** `inline_call.rs:3050-3071` runs pyjitpl.py:1382-1416 in order —
`can_inline_callable`, then the `FBW_MAX_INLINE_RECURSION` bound, then the
`dont_trace_here` / `disable_noninlinable_function` side effect. This fold is
the port of what comes *after* :1416 (`assembler_call = True` →
`verify_green_args` → `do_recursive_call` → `do_residual_call(
assembler_call=True)`), and **upstream's emitter carries no admission
condition of its own**. Feeding `can_inline_callable` into the emitter applies
the gate twice, diverting callees upstream would have inlined onto
CALL_ASSEMBLER.

The `recursive_call` op site itself is not portable here. Upstream emits it
only from `guess_call_kind == 'recursive'` (`codewriter/call.py:116-120`), an
identity test `funcptr is jd.portal_runner_ptr` on a **graph constant** that
warmspot plants by rewriting the portal graph (`warmspot.py:1054-1071`).
pyre's jitcode comes from a bytecode codewriter where the callee is a runtime
stack value, so that predicate has no input; `indirect_call` never yields
`'recursive'` upstream (`call.py:137-139`), so there is no dynamic-callee
precedent to copy either. The op is fully ported in majit
(`recursive_portal_call!` → `exec_recursive_call_assembler` →
`bhimpl_recursive_call_*`); its only consumer is `majit/examples/tl`, a
`#[jit_interp]`-annotated Rust interpreter, which has the static portal call
upstream relies on.

https://claude.ai/code/session_01MLkGH6Ee8dMtvQqFYU8k5Q

## Summary by CodeRabbit

- **Improvements**
  - Improved handling of self-recursive calls, loop-carried values, protected exception paths, and eligible external calls during JIT compilation.
  - Improved traceback recording for compiled recursive calls, including runtime traceback hooks and clearer frame information.
  - Reduced unnecessary residual activity during optimized call folding.

- **Benchmarks**
  - Added coverage for recursion beyond the loop unrolling limit.
  - Updated benchmark statistics, showing fewer guard failures across several execution modes.


youknowone:
Pushed three more commits. The headline change is that the perf work turned out to be blocked on a **pre-existing traceback defect**, which is now fixed.

### `jit: emit the top-level frame's traceback node into the trace`

A trace finishing with `exit_frame_with_exception` recorded its root frame's traceback node for the recording pass only, on the premise that the interpreter would record it when the trace handed back the error (`jitcode_dispatch/mod.rs:3256`, `emit_runtime: false`).

That premise holds only for a trace **the interpreter entered**. A `CALL_ASSEMBLER` entry comes from another trace's compiled code and dispatches to `handle_fail_exit_frame_with_exception`, which republishes the exception and returns into the caller's machine code — no interpreter sees it, so the entered frame contributed no node.

Measured law: **one frame lost per `CALL_ASSEMBLER` entry**, i.e. one per `FBW_MAX_INLINE_RECURSION + 1` levels. `rec(9)` ending in `1 // 0` gave 10 `rec` frames interpreted and 9 compiled; at depth 40 it is 5; with a fatter callee body the whole 41-frame chain collapses to 1.

This reproduces with `origin/main`'s `inline_call.rs`, so it predates this branch — the foreign-callee admission here merely made it reachable from `extra_tests/parity_tests/int_div_mod_raising_specialization.py`. `fbw_publish_exit_last_instr` already publishes the coordinate the node needs, and `screen_frame_already_recorded` keeps the interpreter-entry case at one node.

### Verification

| | |
|---|---|
| `int_div_mod_raising_specialization.py` | OK |
| `rec(9)`/`1//0` cold vs hot traceback | 11 vs 11 (was 11 vs 10) |
| `r(8)` ×1M from a `while` | 1.21s → **0.10s** |
| `pyre/check.py` dynasm | **ALL PASSED 437/437** |
| `pyre/check.py` cranelift | **ALL PASSED 437/437** |

Three improved wasm jit-stats baselines re-recorded to match the native halves already in `6cafee0fe92`.

### One open item, deliberately not blessed

`wasm synth/comprehension_object_append_hot` regresses (`bridges_compiled` 14 → 17, `guard_failures` 2810 → 3410). It reproduces in isolation, so it is not noise, and it is attributable to this change adding an IR call on every top-level exception exit. It is **not** re-recorded. The narrowing that should fix it is to emit the node only for a trace that can be entered by `CALL_ASSEMBLER` rather than for every top-level exception exit.

The other wasm red (`short_circuit_value_kept_stack`) and the `dynasm generator_tree_recursion` jit-stats red are both present on `main`'s own CI at `31caaba137e`.

— *commented by Claude*

PR #1268, opened 2026-08-16
jit: correct the vstack mirror's executed-store precedence, and record the exact opcode segmentation

youknowone:
Three changes to the vstack mirror's sources of truth, found by instrumenting
the mechanism behind #1258 rather than by reading.

Background: the walk keeps an operand-stack box mirror reconstructed from
Python opcode stack effects at "boundaries". A boundary is emitted by jitcode
lowering, not by Python opcode retirement, so the walker has to infer whether
an opcode retired. #1258 fixed one consequence (a `SWAP` applied twice). These
are the rest of that family.

The virtualizable shadow is the executed-store record: every push and pop
writes it, including a NULL push (`emit_pushvalue_ref_const!`) and a pop's
clear (`emit_popvalue_ref!`).

## 1. The reorder mask covered one opcode

The mask records which slot the walk executed a store into, and the boundary
restore keeps masked slots over its pc-derived snapshot. It was set only inside
the method-form `LOAD_ATTR` branch, so a window whose only executed store came
from any other opcode reported an empty mask and the restore replaced the whole
mirror.

Measured over 444 bench programs, grading both sides of each conflict against
the shadow:

| graded vs the shadow | masked | n |
|---|---|---|
| shadow==mirror, restore overwrites | true | 184 (mask working) |
| **shadow==mirror, restore overwrites** | **false** | **28** |
| shadow==restored (restore right) | false | 9 |

The 28 span `LoadName` 9, `LoadSpecial` 6, `StoreName` 5, `BinaryOp` 4,
`LoadSuperAttr` 3, `UnaryInvert` 1.

## 2. The pcdep capture augment outranked the shadow

The per-PC color map fills operand-stack slots the mirror left as holes, but
never consulted the shadow — and omitting a slot would have left the shadow's
own value standing. 14 branch-guard captures published a box disagreeing with a
live shadow Ref; **all 14 came from this augment, none from the mirror**.
Comparing concrete values, the two boxes are different objects in all 14, and
the published `ConstPtr` is invariant across captures while the shadow box
varies — a program-point label, not a binding.

## 3. The floor tier cannot answer which opcode owns a byte

`py_floor_by_jit_pc` is derived from the first-offset-per-PC table, so a PC that
emits in two disjoint regions keeps only the earlier one. Three emission shapes
produce that: a can-raise trailing marker re-keyed to the call's fallthrough PC,
a block opened at an already-merged PC and drained after later PCs, and a
mid-opcode block split.

This records the owning PC wherever the emitted stream changes owner and ships
it as `py_exact_by_jit_pc`. **Nothing reads it outside `PYRE_VSTACK_EXACT_AUDIT`**,
which reports disagreements with the floor tier and checks the table's own
integrity at build time.

Over 442 programs / 1075 jitcodes: no crashes, and 303 disagreements at
boundaries the mirror acts on. Classified over the 13 hottest programs:

- **90 — the floor invents a boundary** (exact == current coordinate): the
  mirror replays an opcode's stack effect where the walk never left the opcode.
  Same shape as the `SWAP` defect, which `repeat_boundary` does not catch
  because the floor reports a *different* py.
- **37 — both report a boundary, different target opcode**.
- **0 — the floor misses a boundary.**

Note for reviewers: three places handle insn-index side tables and they
guarantee different things. `stream_positions_mut` covers the shifts a splice
pass applies; the assignment after the splice replaces the stream wholesale;
`remove_repeated_live` deletes ops and remaps. A table registered only with the
first stays silently empty after the second and addresses past the shortened
stream after the third. Both failure modes were hit while building this and are
now commented at each site.

## Verification

`pyre/check.py`: dynasm 436/436, cranelift 436/436. Changes 1 and 2 are
byte-identical across the 444-program corpus with a same-length-name decoy arm
as the layout-lottery control; change 3 is read only by its audit gate.

The `wasm synth/short_circuit_value_kept_stack` ratio failure is pre-existing
and unrelated — it measures 4.8-5.2x against a 3.7x gate before and after, and
the walker is backend-independent while both native backends are clean.

— *authored by Claude*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JIT stack tracking across recursive calls, exception handlers, and reordered operations.
  * Preserved valid reference values during stack recovery to prevent incorrect overwrites.
  * Improved mapping between generated JIT instructions and their corresponding Python operations.

* **Diagnostics**
  * Added optional diagnostics for auditing exact stack-segmentation behavior and comparing fallback mappings.
  * Expanded the documented diagnostic inventory to include three new default-off controls.

### DIFF
--- a/pyre/gate-triage.md
+++ b/pyre/gate-triage.md
@@ -974,7 +974,7 @@ the folds it selects, not before them.
 `PYRE_FBW_SPEC_CENSUS` in §6c is its read-only half: the per-fold
 consulted/fired tallies.
 
-### §6c — Default-OFF diagnostics, censuses and probes (64): keep, cost nothing
+### §6c — Default-OFF diagnostics, censuses and probes (65): keep, cost nothing
 
 Each is inert unless set, so none is a removal target by this file's
 already-ON criterion. They are listed so they cannot be missed again.
@@ -1000,7 +1000,7 @@ already-ON criterion. They are listed so they cannot be missed again.
 `PYRE_PORTAL_RCA`, `PYRE_PROBE_BH_STARTUP`, `PYRE_PROBE_SNAPSHOT`,
 `PYRE_PROBE_SUBSCR`, `PYRE_PROFILE_PIPELINE`, `PYRE_QMUT_MAPDICT_FORCE`,
 `PYRE_RERAISE_DIAG`, `PYRE_SIZE_SHELL_OWNERS`, `PYRE_SNAPSHOT_DIAG`,
-`PYRE_VSTACK_EXACT_AUDIT`, `PYRE_VSTACK_NO_EXACT`,
+`PYRE_VSTACK_EXACT_AUDIT`, `PYRE_VSTACK_KEEP_REORDER`, `PYRE_VSTACK_NO_EXACT`,
 `PYRE_WASM_DUMP_BAD_TRACE`, `PYRE_WASM_EXEC_TRACE`, `PYRE_WASM_FBW_CENSUS`,
 `PYRE_WASM_GUARD_CENSUS`, `PYRE_WASM_JIT_STATS`, `PYRE_WASM_CALL_HIST`,
 `PYRE_WASM_NO_CACHE`, `PYRE_WASM_STARTUP_TRACE`.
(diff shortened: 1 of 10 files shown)


youknowone:
Added the fourth commit: the mirror now resolves its coordinate from
`py_exact_by_jit_pc` rather than the floor tier, so the table added in the
third commit is no longer read only by its audit gate.

Classified over the 13 programs with the most disagreements, of the boundaries
the mirror actually acts on:

| class | n |
|---|---|
| floor reports a boundary the emission record places **inside** the current opcode | 90 |
| both report a boundary but name a **different** opcode | 37 |
| floor **misses** a boundary | 0 |

The third row being zero is what bounds the risk: the floor tier only ever
over-reports or mis-identifies, never under-reports, so switching can remove or
move a boundary but never invent one.

The 90 are the same shape as the `SWAP` defect fixed in #1258 — an opcode's
stack effect replayed at a boundary that did not retire it. `repeat_boundary`
from that fix does not catch them, because it tests `new_pypc == prev_pypc`
while here the floor tier reports a *different* py.

Verification: byte-identical output across 444 bench programs, with a
same-length-name decoy arm as the layout-lottery control and
`PYRE_VSTACK_NO_EXACT` restoring the floor tier inside the same binary (so the
two arms are not two builds). check.py dynasm 436/436, cranelift 436/436;
`cargo test -p pyre-jit-trace` green.

One thing worth stating plainly for review: **all four changes are
output-neutral on this corpus.** Each defect is measured and real, but none
produces an observable wrong answer today, so none of them can be pinned by a
regression test. The reason appears to be that the mirror's hole-fill and
shadow reseed repair a wrongly replayed effect for every class except a
non-idempotent one — `SWAP` was the single class that turned this family into a
crash, and #1258 closed that exit. These changes remove the underlying
mis-attribution rather than the symptom, which matters when the next
non-idempotent class is added or when a fold stops writing the mirror.

— *commented by Claude*

PR #1272, opened 2026-08-16
check.py: raise the wasm/dynasm ratio ceiling to 4x

youknowone:
Two commits: the wasm/dynasm ratio ceiling, and a note on a fixture's counters.

## `WASM_MAX_DYNASM_RATIO` 3x → 4x

3x was not where the backend is. Four fixtures had already been carved out of it one at a time, each for the reason `wasm_ratio_gate` documents — the wasm backend reaches CPython-level objects through interpreter round-trips and allocates each on a Rust heap the wasm path does not yet collect, so a loop dominated by that work runs several times dynasm's execution time with no regression behind it:

| fixture | allowance |
| --- | --- |
| `fannkuch` | 3.6 |
| `fib_recursive` | 3.6 |
| `raise_catch_loop` | 3.8 |
| `synth/short_circuit_value_kept_stack` | 3.7 |

And more kept arriving without one. On main's ubuntu runner, `if_else_jump_forward` fails at 3.9x and `loop_callee_shared_mutation` at 3.3x today, neither carrying an allowance.

A ceiling that the fixtures above it are exempted from individually is collecting exemptions rather than measuring the backend. This sets it to 4x and removes the four allowances now under it. `synth/global_quasiimmut_invalidation` keeps its 6, the only one still above the gate.

**What this does not fix:** `short_circuit_value_kept_stack` reads 5.2x on main's ubuntu runner and 5.7x here. It was already over its own 3.7 allowance and is over 4x too, so it stays exactly as red as it is on main. Removing its line moves its ceiling from 3.7 to 4x and changes nothing else — what that fixture should be fitted to is a separate question from where the shared gate sits, and raising a ceiling to cover it is the move that was refused before.

## A note on `const_arg_call_resume`

Comment only — no gate change, no baseline change.

Its `bridges_compiled` and `guard_failures` do not reproduce. The same binary on the same source with the same argv, the same child environment and `PYTHONHASHSEED` pinned reads six bridges / 1204 or seven / 1404, run to run — 1 of 6 runs, then 2 of 6, then 0 of 8 an hour later. CI has seen it as well: the windows runner reported `dynasm/synth/const_arg_call_resume` as `jit-stats unstable (not gated)`.

A machine settles on one value and repeats it many times over, so a local red on either counter looks like a property of the tree and is not one. The note also records the two things that do not settle it, so neither is tried again:

- the trip count reads the same counters at 220, 240, 260, 280, 300, 320, 360 and 400
- re-recording pins whichever side the recording machine was on — adding this very comment, text the parser drops, moved a wasm reading from 1404 to 1204 for eight runs in a row

`JITSTATS_STABILITY_RUNS` already excuses the fixture whenever its repeats disagree, and across six recent main runs on three hosts its counters were never gated against it, so nothing here changes the gate.

## Verification

Being re-measured on the current branch tip; the numbers will be posted here when the run finishes. (An earlier full run was discarded: the branch was rebased while it was in flight, so its source and its binaries did not come from one tree.)

Assisted-by: Claude

## Summary by CodeRabbit

- **Performance**
  - Updated WebAssembly performance validation to allow a higher execution-time variance threshold, reducing unnecessary benchmark failures.

- **Documentation**
  - Added context around benchmark measurements, including JIT behavior, test conditions, and output validation.
  - Removed outdated benchmark ratio notes and configuration comments where they no longer apply.

- **Tests**
  - Confirmed benchmark logic, outputs, and runtime behavior remain unchanged.

### DIFF
--- a/pyre/bench/fannkuch.py
+++ b/pyre/bench/fannkuch.py
@@ -1,6 +1,3 @@
-# pyre-check: max-wasm-ratio=3.6
-# Reported 2.9x here idle and 3.1x here under load, and under the gate on
-# ubuntu-24.04 twice.
 # Fannkuch-Redux benchmark (The Computer Language Benchmarks Game)
 # Ported for pyre: while-loop only, no range/list/enumerate
 
--- a/pyre/bench/fib_recursive.py
+++ b/pyre/bench/fib_recursive.py
@@ -1,6 +1,3 @@
-# pyre-check: max-wasm-ratio=3.6
-# Reported 3.1x on ubuntu-24.04, and under the gate on the run before it.
-
 def fib(n):
     if n < 2:
         return n
--- a/pyre/bench/raise_catch_loop.py
+++ b/pyre/bench/raise_catch_loop.py
@@ -1,6 +1,3 @@
-# pyre-check: max-wasm-ratio=3.8
-# Reported 3.2x and 3.3x on ubuntu-24.04.
-
 def main():
     s = 0
     i = 0
--- a/pyre/bench/synth/const_arg_call_resume.py
+++ b/pyre/bench/synth/const_arg_call_resume.py
@@ -13,6 +13,43 @@
 # and int / object element strategies each exercise the same caller-frame
 # resume; every form must reach `n` (never a dropped element or an unbound
 # local).
+#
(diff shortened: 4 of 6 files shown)


youknowone:
weird, because the test now pass on main

PR #1283, opened 2026-08-16
majit-macros: sync a flat virtualizable by resetting its token, not by copying

youknowone:
Follow-up to #1261, resolving the per-call regression flagged in https://github.com/youknowone/pyre/pull/1261#issuecomment-5308789098.

## Root cause

The macro-generated state's `virtualizable_heap_ptr` hands out the state's own address, and every array field is registered at its byte offset within it — so compiled code already reads and writes the interpreter's storage directly. Yet the default `sync_virtualizable_before_jit`/`sync_virtualizable_after_jit` hooks still copied every array element in and out around each compiled entry. Over the same memory both passes are no-ops that cost a full-array walk plus the holding vectors per entry. This is what turned #1261's armed entry contract (compiled inputs 16–42 → 2, −14 allocs/eval — both real) into a net per-call loss.

## Fix

Generate overrides in the state-field macro:

- `sync_virtualizable_before_jit`: reset the vable token through `virtualizable_heap_ptr` and return true — `warmstate.py:387-396` `execute_assembler` does only `vinfo.clear_vable_token(virt)`, one store, before the call.
- `sync_virtualizable_after_jit`: nothing — upstream does nothing after the call either.

The token store is inert while the descriptor carries no token field; it is written so a state that later declares one gets the upstream behaviour unchanged.

## Measurements (interleaved A/B, stashed binaries, min-of-4, load 5–8)

- vs the armed HEAD (#1261 as merged): faster on every percall case; `list_filter` 442→189 ns/call.
- vs the pre-arming baseline (cycle-17): 25/28 cases equal or faster (flat cases −4…−29 ns/call, `list_filter` −101 ns = −35%).
- Remaining known gap, not caused by this change (control with unarmed Vec banks matches baseline): `comprehension_scaling/100` +7.6%, `/500` +14.4% vs baseline — attributable to the armed configuration's bank-growth path, under separate investigation.

Verified: majit-metainterp + majit-macros suites green (cranelift and dynasm), allocation meters unchanged at 4.000/6.000, fmt clean.

https://claude.ai/code/session_011capRVFH2W4gwGQNWYCxEx

### DIFF
--- a/majit/majit-macros/src/jit_interp/codegen_state.rs
+++ b/majit/majit-macros/src/jit_interp/codegen_state.rs
@@ -2249,6 +2249,50 @@ fn generate_state_fields_jit_state(config: &JitInterpConfig, func: &ItemFn) -> T
                 Some(self as *const Self as *mut u8)
             }
 
+            // warmstate.py:387-396 `execute_assembler`: all a compiled entry
+            // does to its virtualizable is `vinfo.clear_vable_token(virt)` —
+            // one store, before the call, and nothing after it. There is no
+            // copy in and no copy out, because the virtualizable IS the
+            // storage the compiled code reads and writes.
+            //
+            // The default hooks copy in and copy out instead, which is what a
+            // host needs when its interpreter-side storage is a separate
+            // mirror of the virtualizable object. This state is not such a
+            // host: `virtualizable_heap_ptr` above hands out its own address,
+            // and every array field is registered at its byte offset within
+            // it, so the boxes the default reads and the boxes it writes back
+            // address the same memory. The copy in is discarded outright —
+            // nothing here overrides `import_virtualizable_boxes`, whose
+            // default drops the boxes it is handed — and the copy out reads
+            // each element and writes it back unchanged. Both are no-ops that
+            // cost a pass over every array, plus the vectors to hold it, on
+            // every entry into compiled code.
+            //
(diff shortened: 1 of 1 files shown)


youknowone:
CI triage — the ubuntu/windows `check.py` failures predate this PR. Comparing against the #1261 merge-ref run (the state that was merged, [ubuntu](https://github.com/youknowone/pyre/actions/runs/31962138614/job/95211519534) / [windows](https://github.com/youknowone/pyre/actions/runs/31962138614/job/95205149901)):

- `cranelift raise_catch`: ubuntu **3.1x on both runs, identical timings** (0.50s vs pypy 0.16s); windows 3.5x → 3.8x, same band.
- `dynasm synth/generator_tree_recursion` guard_failures 2951→2955: known main-owned drift on both x86 hosts.
- `wasm synth/short_circuit_value_kept_stack` 5.2x → 5.0x: known base-owned, under separate investigation.
- `wasm raise_catch` 4.1x vs the 4x gate is the one line not verbatim on the base run — it is the same base-owned raise_catch slowness sitting 2.5% over a gate that the cranelift leg of the same case already fails at 3.1x on the base.

macos passes on both runs. No failure is introduced by this commit; the failing set on this PR is contained in the base failing set modulo the raise_catch gate-hover.

— *commented by Claude*

PR #1293, opened 2026-08-17
majit: route both raw-hash cell readers through lookup_chain, and step the mint serial as a value

youknowone:
Two defects in the cell-unique-key celltable, found while closing out the #1261 review round.

## Two readers of one raw hash could describe different buckets

`bucket_is_chained` applied `bucket_of` to a raw green-key hash — a domain error, since `bucket_of` maps a cell key to its bucket and a raw hash is its own bucket — while `sole_cell_key` did a bare `cells.get`. When a raw hash equals a minted cell key, the two readers inspected different buckets, and `resolve_cell_key`'s greens-less fast path could answer with a cell belonging to neither the arriving key nor its sibling. Both readers now enter at `lookup_chain`, matching `counter.py:239-240` where every reader enters at `celltable[self._get_index(hash)]` with nothing in front of it.

The pin test manufactures the collision deterministically by inverting the `get_uhash` fold (the multiplier is odd, hence invertible mod 2^64), so "every u64 is a producible hash" is exercised rather than assumed.

## `mint_cell_key` hung on a bucket's second mint

The retry loop folded `(bucket ^ serial)` with the serial as the xor seed, so the serial cancelled exactly and every retry re-proposed the same candidate: the first mint in a bucket took it, the second looped forever. Three cells chained in one bucket were enough to hang the JIT. The serial is now folded as the value; distinct serials give distinct candidates and the finite live set bounds the retry. Pinned by a test that hangs before the change.

Also: the hash-only-creator comment enumerated six callers where the tree has eight; it now states the defining property (a creator is exactly a caller of `ensure_cell_by_key`) instead of a count that rots.

A raw hash that equals a live minted key while owning no cell of its own still resolves to the foreign cell via `resolve_cell_key`'s fallback; after this change that is the pre-existing 64-bit `get_uhash` collision limit already documented at `sole_cell_key`, no longer a disagreement between readers.

Verified: majit-metainterp 0 failed on cranelift and dynasm, cel consumer suite green, fmt clean.

https://claude.ai/code/session_011capRVFH2W4gwGQNWYCxEx

### DIFF
--- a/majit/majit-metainterp/src/warmstate.rs
+++ b/majit/majit-metainterp/src/warmstate.rs
@@ -1209,10 +1209,21 @@ impl WarmEnterState {
     /// one candidate, so [`Self::sole_cell_key`] is exact and a caller can skip
     /// building a `GreenKey` it would only use to confirm that — which is what
     /// keeps [`Self::resolve_cell_key`] allocation-free on the entry path.
+    ///
+    /// Resolves through [`Self::lookup_chain`] and NOT through
+    /// [`Self::bucket_of`]. The argument is a raw green-key hash, which is its
+    /// own bucket; `bucket_of` maps a CELL KEY to the bucket that cell lives
+    /// in, and applying it here sends this reader to another bucket entirely
+    /// whenever the hash coincides with a live minted key — while
+    /// [`Self::sole_cell_key`], the reader whose branch this decides, keeps
+    /// reading the bucket the hash names. Upstream cannot split the two: every
+    /// reader enters at `jitcounter.lookup_chain(hash)`
+    /// (warmstate.py:459-460, :597-598, :632-633), and `lookup_chain`
+    /// (counter.py:239-240) is a bare `celltable[self._get_index(hash)]`. See
+    /// [`a_raw_hash_equal_to_a_minted_key_resolves_through_one_bucket`].
     #[inline]
     pub fn bucket_is_chained(&self, green_key_hash: u64) -> bool {
-        self.cells
-            .get(&self.bucket_of(green_key_hash))
+        self.lookup_chain(green_key_hash)
             .is_some_and(|head| head.next.is_some())
     }
 
@@ -2416,7 +2427,7 @@ impl WarmEnterState {
     /// (see [`BaseJitCell::comparekey`]) — no accept rule fixes it here.
     #[inline]
     pub fn sole_cell_key(&self, hash: u64) -> Option<u64> {
-        let head = self.cells.get(&hash)?;
+        let head = self.lookup_chain(hash)?;
         if head.next.is_some() {
             return None;
         }
@@ -2446,8 +2457,16 @@ impl WarmEnterState {
     fn mint_cell_key(&mut self, bucket: u64) -> u64 {
         loop {
             self.mint_serial = self.mint_serial.wrapping_add(1);
+            // The serial must reach the fold as the VALUE, not pre-mixed into
+            // the accumulator: `green_uhash_step(bucket ^ serial, Int, serial)`
+            // folds `(bucket ^ serial) ^ serial`, which is `bucket` again, so
+            // every retry re-proposes one number and the second mint in a
+            // bucket cannot terminate. Folded here, distinct serials give
+            // distinct candidates — xor and the odd multiplier are both
+            // bijections — so the retry walks a fresh number each time and the
+            // finite live set bounds it.
             let candidate = majit_ir::green_uhash_step(
-                bucket ^ self.mint_serial,
+                bucket,
                 majit_ir::GreenType::Int,
                 self.mint_serial as i64,
             );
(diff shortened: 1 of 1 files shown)


youknowone:
CI triage for the three `check.py` failures.

**ubuntu / windows — inherited, not this PR.** The failing set is exactly the base-owned set documented on #1283: `cranelift raise_catch` 3.1x/4.0x vs the 2.5x gate (identical values on the #1261 merge-ref run) and `wasm synth/short_circuit_value_kept_stack` 5.3x (known, under separate investigation).

**macos — two cranelift jit-stats gates, and the counters are demonstrably unstable on this very tree.** `synth/comprehension_object_append_hot` (bridges_compiled 14→17, guard_failures 2810→3410) and `synth/const_arg_call_resume` (6→7, 1204→1404). Two full local `check.py` runs on this exact head (f85bcf59b13, macos arm64) both finish `dynasm 437/437, cranelift 437/437 ALL PASSED` — because check.py's own instability detector re-runs the same binary on a baseline mismatch, sees the counters move, and reports precisely these two cells as **"jit-stats unstable (not gated)"**. The CI leg happened to read stable within its single run, so the same mismatch gated there.

A mechanistic note on why these counters can move with this change at all: `bucket_is_chained` previously applied `bucket_of` to a raw green-key hash, i.e. inspected a rehash of the intended bucket; the committed baselines for multi-cell benches were recorded under that misread. After the fix the reader inspects the intended bucket, so on benches with chained buckets the trace/bridge counts legitimately settle differently — the old numbers pinned the bug's behavior.

Disposition is your call: re-record the two snapshots on a host where they read stable (`pyre/check.py --snapshot --backend cranelift --synthetic-pattern <name>`), or leave them to the instability detector. I have not re-recorded anything. (A filtered run cannot arbitrate here — under `--synthetic-pattern` these benches read loops=1/bridges=0, a different warm-up state than the full run that the gate is defined over.)

— *commented by Claude*

youknowone:
**Correction to my earlier triage: the two macos `jit-stats` failures are not attributable to this PR, and I was wrong to imply they were.**

In [my previous comment](https://github.com/youknowone/pyre/pull/1293#issuecomment-5314805852) I leaned on a base control — macos `check.py` is green on `main`, red here — to conclude these two gates were moved by this diff. That inference does not survive reading the green run's log.

**The green `main` run carries one of the two readings we are charged with.** Run [32006541137](https://github.com/youknowone/pyre/actions/runs/32006541137) (`db1e9bd6f17`, push to `main`, macos `check.py` = success) contains:

```
synth/const_arg_call_resume
  dynasm    UNSTABLE  jit-stats unstable — re-running the same binary moved
            bridges_compiled 7 -> 6, guard_failures 1404 -> 1204, ...
  cranelift UNSTABLE  jit-stats unstable — re-running the same binary moved
            bridges_compiled 7 -> 6, guard_failures 1404 -> 1204, ...
```

`6 -> 7` and `1204 -> 1404` are exactly the numbers this PR is failing on. They occurred on a tree that does not contain this diff, and that run is green only because its stability re-run happened to land on the other side. Whether a run passes or fails here is decided by which of two settled readings the re-run draws — that is a property of the run, not of the tree.

**The bistability is documented in-tree, ahead of time.** `pyre/bench/synth/const_arg_call_resume.py:17-40` states it and predicts these figures digit for digit: *"`bridges_compiled` and `guard_failures` here do not reproduce, so a move in either is not evidence about a diff until one configuration has been repeated"*, with the same-binary/same-argv/same-`PYTHONHASHSEED` measurement reading 6/1204 or 7/1404 at 1-of-6, then 2-of-6, then 0-of-8 an hour later. It also explains the shape: the two readings are one bridge generation apart, `402/402/400 = 1204` flooring to 2+2+2 bridges against `602/402/400 = 1404` flooring to 3+2+2. The deltas on both failing rows are exact multiples of `trace_eagerness` (200) with identical residues — `2810 = 14·200+10 -> 3410 = 17·200+10`, `1204 = 6·200+4 -> 1404 = 7·200+4` — and `loops_compiled` does not move on either.

**A mechanism check agrees with the arithmetic.** On these fixtures the diff cannot change a resolved key. `sole_cell_key`'s edit is textually the same call (`lookup_chain(h)` *is* `cells.get(&h)`). `bucket_is_chained`'s edit differs from the old code only where `bucket_of(h) != h`, i.e. on a live minted key — and there both readings hand the consumer the same number (old: chained arm, `cell_key_for(K)` = M; new: `sole_cell_key(M)` = `None`, `unwrap_or(M)` = M). What changed there is that one `GreenKey` is no longer built per warm back edge, not which key is answered. And every counter access is `bucket_of`'d first, so a minted key and its raw hash share one `JitCounter` subentry: the mint change cannot reach the timetable.

There is also a direction argument against the "the old numbers pinned the bug" reading. The only route by which a cell-key answer manufactures surplus bridges is a producer/consumer key mismatch, where `compile_bridge` fails its attach while `bridges_compiled += 1` sits outside the `if` — a fix that makes two readers agree closes that generator, so under that story the count should have **fallen**.

**Disposition — I have not re-recorded anything, and I recommend against it.** `--snapshot` here would write a reading that has been observed on both sides of this change on both trees; the house rule is that a baseline is not re-recorded to bless a cost, and "the current reading" is not "the correct reading" for a counter that flips run to run. I have re-run the macos job instead. If these two fixtures keep flapping the gate, the honest encoding of a *measured* bistability is a `# pyre-check: jitstats-band=` on these counters for these two benches (`check.py:1490`, `:1504`) — but that changes what is gated, so it is your call and a separate change, not something to fold into this PR.

`ubuntu` / `windows` remain the inherited base-owned set from my earlier comment; nothing there has changed.

— *commented by Claude*

youknowone:
**Closing the one item my first triage left open: `wasm synth/short_circuit_value_kept_stack`.** It is not merely inherited — it is **already fixed on `main`**, and this PR's CI has never been able to see the fix. A rebase clears it; nothing in this diff needs to change.

**This PR's CI merge-ref predates the fix by seven minutes.**

```
PR head       f85bcf59b13   2026-08-17 11:34:21 +0900
#1284         4383741cc42   2026-08-17 11:41:46 +0900   <- 7m25s later
#1296         0d9822644d1   2026-08-17 19:47:25 +0900
CI merge ref  a9320dbf231 = merge(f85bcf59b13 into 239070676de, i.e. #1289)
```

```console
$ git merge-base --is-ancestor 4383741cc42 refs/pull/1293/merge; echo $?
1        # the merge-ref does NOT contain #1284 (nor #1296)
```

`#1284` is the commit that expands `%` and `//` by a constant on a backend with no mul-high, and this fixture is the one it was measured on: it moved `short_circuit`'s wasm ratio **5.7x → 2.0x** (`#1296` then replaced that expansion with the parity-literal one, 1.9x). So the merge-ref runs the pre-`#1284` residual-call codegen by construction, and the **4.8x / 5.3x this PR reports is that regime**, not an effect of the diff. This PR touches one file, `majit/majit-metainterp/src/warmstate.rs`, which cannot reach wasm codegen.

**A same-commit control confirms it and rules out the runner.** I re-ran `main`'s own push run at `0d9822644d1` — [run 32021822128](https://github.com/youknowone/pyre/actions/runs/32021822128), attempt 2, [ubuntu job 95450435041](https://github.com/youknowone/pyre/actions/runs/32021822128/job/95450435041) — so the two attempts are the same run on the same commit and differ only by which runner they drew. Attempt 2 reproduces `main`'s verdict exactly: `cranelift 1 failed, 437 passed`, **wasm passes**. Both runs of this PR read `cranelift 1 + wasm 1 failed`.

Attempt 2 did draw a slower runner, but uniformly so — every column inflates together and wasm is not an outlier:

| column | attempt 1 | attempt 2 | ×    |
|--------|-----------|-----------|------|
| cpython | 1.74s | 2.54s | 1.46 |
| pypy | 0.07s | 0.10s | 1.43 |
| dynasm | 0.19s | 0.26s | 1.37 |
| cranelift | 0.31s | 0.47s | 1.52 |
| wasm | 0.32s | 0.46s | **1.44** |

`cpython` and `pypy` are free references here — no PR can move those binaries — and wasm tracks them. Both attempts pass the 4x gate on `main`.

**Disposition:** rebase onto current `main` (`81e8a2d07ba`) and re-run. Expected result is that the wasm row goes green and the only remaining red is `cranelift raise_catch` 3.0x vs the 2.5x gate, which the control run fails identically on `main` and which `#1284` does not touch (its expansion is gated on `supports_efficient_uint_mul_high=false`, so no dynasm/cranelift row moves). I have not rebased or pushed anything.

Method note for anyone re-checking a red on this repo: a PR's CI runs `refs/pull/N/merge`, so a control chosen to match the PR's own commit — or `main` at the time you look — can sit on the far side of a fix that landed minutes after the branch. `git merge-base --is-ancestor <fix> refs/pull/N/merge` settles it before any measurement, and note that `gh run rerun` re-runs the *same* stale merge-ref rather than recomputing it, so a rerun is not an independent second sample.

— *commented by Claude*

PR #1295, opened 2026-08-17
Two JIT admission gates that answered wider than their evidence

youknowone:
Two JIT admission gates that answered wider than their evidence. Each declined a whole class of work to cover a case that only arises in part of that class.

> A third commit (`88930aa7029`, the FOR_ITER per-region gate for `synth/range_ctor_in_loop`) was **dropped** from this branch by force-push: it lost a list element in `extra_tests/parity_tests/for_iter_call_bearing_comprehension.py`. See the comment below for the diagnosis.

---

# 1. Bound-method folds inside an inlined callee

Three folds in `specialize.rs` declined on `ctx.fbw_mode.inline_subwalk` alone:

- `try_walker_specialize_load_bound_method_attr`
- `try_walker_specialize_load_classmethod_attr`
- the class pin in `try_walker_fold_load_method_self`

Each cites the same reason — a guard emitted in a callee sub-walk collapses its resume to the caller's CALL, so a failure re-runs the callee from its entry and doubles whatever it sequenced before the `LOAD_ATTR`. That holds for the single-frame collapse. It does not hold for the multi-frame arm of `walker_capture_snapshot_for_last_guard_impl`, which resumes the callee at its own pc and the caller at the CALL return point, and which fires when the paused-caller chain covers the full inline depth.

That condition is factored out as `walker_inline_guard_resumes_in_callee` and is now what the three folds consult, so the emitter and its consumers cannot drift apart. The collapse case stays declined.

`synth/inlined_helper_mutation` is the case that made this visible: its `push(a, v)` binds `a.append` inside an inlined callee. Under `PYRE_FBW_DEBUG_ABORT=1` the benchmark prints no `Collapse::*` line at all — every inlined CALL in it already had a full parent chain, so the decline was pure loss.

| `synth/inlined_helper_mutation` | before | after |
|---|---|---|
| dynasm | 39.4x | **13.0x** |
| cranelift | 70.2x | **23.1x** |
| wasm | 60.8x | **18.0x** |
| `max-pypy-ratio` | 145 | **60** |

The benchmark's own acceptance check — the appended count and the attribute count must both equal the iteration count, in both helper orderings — holds at 2900000/2900000.

No jitstats baseline moves on any backend: the change turns residual calls into inline IR whose guards do not fail, so `loops_compiled` / `bridges_compiled` / `guard_failures` are unchanged. `MAJIT_LOG=1`'s per-trace `final ops: … call_may_force=N` is what tracks it.

---

# 2. `__getattr__` was the one user dunder with no inline resolver

Ten-plus call sites reach `try_walker_inline_resolved_user_call` — `__init__`, `__hash__`, `__index__`, `__getitem__`, property get/set, the user binops and compareops. `__getattr__` had none, so every hooked attribute access cost one opaque residual holding the whole `object_getattr_miss` walk plus a fresh frame for the hook. `try_walker_inline_user_call` cannot serve it: that one admits only `PyreHelperKind::{CallFn,CallKw,CallFunctionEx}`, and a `__getattr__` hook is reached from a `LoadAttr` residual.

The miss is what makes the fold possible. `getattr_hook_fast_path` in `mapdict.rs` is the miss twin of `load_attr_fast_path`: it answers "this name resolves nowhere on this type or this instance, and the type defines `__getattr__`" and hands back the two pins that make that answer constant — the type's version tag and the instance's map. `try_walker_inline_getattr_hook` guards both through `walker_guard_mapdict_instance_shape` and then inlines the hook body in place of the residual.

The hook is a special method, so it is bound through the descriptor protocol like any other (`objspace.py:710 get_and_call_function`). `HookLeading` names the three spellings: a plain function leads with the receiver, `ClassMethod.__get__` leads with the class, `StaticMethod.__get__` binds nothing and the name is the only argument. All three fold; none is admitted by an arity guess.

| `synth/getattr_hook_binding` | before | after |
|---|---|---|
| dynasm | 48.6x | **7.4x** |
| cranelift | — | **10.6x** |
| wasm | 59.5x | **9.4x** |
| `max-pypy-ratio` | 90 | **25** |

`extra_tests/parity_tests/getattr_hook_inline_deopt.py` is new and covers what the two pins owe. Each loop runs long enough to compile and then breaks exactly one pin mid-loop — a store that puts the name on the instance, a reassigned `__getattr__`, a class attribute that shadows the hook — and asserts the recorded values differ at the iteration the pin broke, which is what distinguishes a deopt from a reused compiled answer. A classmethod hook binding the receiver's own class rather than the base, an `AttributeError` raised out of the hook, and a hook that installs the attribute it was asked for are covered for the same reason: they are ordinary outcomes of an inlined body, not shapes the fold may quietly turn into a returned value.

---

## Verification

Rebuilt from clean on the current branch tip, with all four LLBC artefacts current (`extract-llbc.py --check` rc=0):

| gate | result |
| --- | --- |
| `check.py --backend dynasm` | ALL PASSED 438/438 |
| `check.py --backend cranelift` | ALL PASSED 438/438 |
| `check.py --backend wasm` | ALL PASSED 431/431 |
| `extra_tests/parity_tests/run.py --dynasm-only` | all parity tests pass |
| `cpython_tests/run.py --backend dynasm` | 205 PASS / 0 FAIL, no regressions |
| `cargo test --all --no-default-features --features dynasm` | 60 test binaries, 0 failures |

Both ceilings are twice the slowest of the three backends, rounded up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved correctness when properties, methods, functions, and `__getattr__` hooks are changed during hot loops.
  * Fixed descriptor behavior for property subclasses and dynamic class or metaclass updates.
  * Improved handling of exceptions, resumed execution, and multi-frame JIT operations.

* **Performance**
  * Expanded JIT optimization coverage for method calls, attribute access, property accessors, class instantiation, and common special methods.
  * Reduced unnecessary runtime work and improved loop compilation behavior.

* **Tests**
  * Added regression coverage for dynamic rebinding, descriptor protocols, ABC metaclasses, attribute hooks, and trace-abort scenarios.


youknowone:
Force-pushed: the FOR_ITER per-region gate commit (`88930aa7029`, `synth/range_ctor_in_loop`) is **dropped** from this branch. It shipped wrong code.

`extra_tests/parity_tests/for_iter_call_bearing_comprehension.py` lost a list element under it — `AssertionError: (46, 46, 45)`, deterministic 3/3, PASS under `PYRE_JIT=0`, PASS on CPython. `MAJIT_STATS=1` read `loops_compiled=0 bridges_compiled=0 loops_aborted=1`: nothing was compiled and the answer was still wrong, so the damage came from the *abort's* recovery path, not from a compiled loop. The trigger needed a stdlib frame (`collected.append(random.randrange(25))` failed; a local `gen(k)` in the same position passed), so what the gate newly admitted was a loop inside `random`, whose abort corrupted the caller's FOR_ITER in-flight state. Reverting only `pyre/pyre-jit/src/eval.rs` + `pyre/pyre-jit/src/jit/call.rs` made it 3/3 PASS with the other two commits still in the tree.

That regression was invisible to the three green `check.py` runs I had cited, because `check.py` runs neither `extra_tests/parity_tests/run.py` nor `cpython_tests/run.py` — both are separate CI jobs. The task#1-only CI run's CPython gate (which had reported `test.test_heapq: PASS -> FAIL`) was cancelled by the next push, so its red went unseen.

The branch is now the two remaining commits on top of `21ce53083b1`, re-gated from a clean rebuild with all four LLBC artefacts current:

| gate | result |
| --- | --- |
| `check.py --backend dynasm` | ALL PASSED 438/438 |
| `check.py --backend cranelift` | ALL PASSED 438/438 |
| `check.py --backend wasm` | ALL PASSED 431/431 |
| `extra_tests/parity_tests/run.py --dynasm-only` | all parity tests pass |
| `cpython_tests/run.py --backend dynasm` | 205 PASS / 0 FAIL, no regressions |
| `cargo test --all --no-default-features --features dynasm` | 60 test binaries, 0 failures |

Both tightened ceilings still hold on the rebuilt tree — `getattr_hook_binding` 7.4x / 10.6x / 9.4x against 25, `inlined_helper_mutation` 13.0x / 23.1x / 18.0x against 60 (dynasm/cranelift/wasm). Neither bench contains a `for` loop, so the dropped gate could not have been what produced their numbers.

— *commented by Claude*

youknowone:
Force-pushed `faa14da0f71` → `12a8ce30d75`: one new commit, *jit: build the loop region from the exception table instead of a pc interval*. All 15 previously-pushed commits are unchanged (`git cherry -v HEAD origin/single-walker` reported `+` for none of them).

### CI triage of run 32140549893

| failure | owner | disposition |
| --- | --- | --- |
| `cargo test` (ubuntu + windows): `every_live_gate_has_a_triage_entry` — `PYRE_FBW_REPLAY_DIRTY_BODY` has no `gate-triage.md` row | branch — `f4aaa0b3a28`, on this branch, not in `origin/main` | fixed; the test is 6/6 |
| `check.py` (ubuntu + windows): `cranelift synth/inline_freevar_after_mayforce` `guard_failures 1008 -> 1009` | stale baseline, not host drift — darwin reads 1009 too, so all three hosts agree and 1008 is simply out of date | re-recorded rather than banded |
| `CPython suite (gate)`: `test.test_pickle` | base — the root fix `e6e5d27c02e` lives only on branch `str` | passes here now (208 PASS / 0 FAIL) |
| cranelift `raise_catch` ratio 2.9–3.4x (previous run) | marginal headroom, not a standing red | did not recur; locally 1.06x dynasm / 1.35x cranelift against gates of 1.5 / 2.5 |

Neither automated review ran on the previous push, for the record: codex-review exited 1 on `401 token_invalidated`, and CodeRabbit hit its review limit before starting.

### The new commit

`loop_region_end` modelled the natural loop region as one `header..=end` span and grew `end` to any backward jump landing inside it. 3.14 lays an out-of-line handler *after* the code that follows its `try`, so in `synth/range_ctor_in_loop` the span grew from the while backedge (unit 253) to the handler's `JUMP_BACKWARD_NO_INTERRUPT` (unit 406) and covered the trailing comprehension's `FOR_ITER` at unit 323 — an opcode the loop never reaches. That `FOR_ITER` is a `LIST_APPEND` body holding a `len(item)` call, which `for_iter_body_is_jit_safe_at` refuses, so `main` ran interpreted end to end. The exception table names both halves exactly: `covered [117..130) -> handler 384` is the in-loop `try`, `covered [321..375) -> handler 411` is the comprehension, outside the loop. The control is static — a `try`/`except` that never raises still declines, and `try`/`finally` does not.

`loop_region_ranges` now returns the loop body plus each rejoining handler's own range, its start read from `pycode::decode_exceptiontable`.

The region fix alone made things **worse** (249x, against 158x interpreted): `maybe_compile_and_run` asked `loop_region_contains_escaping_range_append` per back edge while `eval_with_jit_inner` asked `frame_has_traceable_escaping_range_loop` for the frame, so a frame the frame gate had admitted still had every back edge but one refused. Traces closed on the inner `for` loops that could then not be entered, leaving their guards without a bridge target — `bridge_no_targets_close=89` and 17896 guard failures at N=20000, with JIT on at 9.83s against 6.08s off. Both gates now read the frame-level answer.

**This is not a re-land of `88930aa7029`.** That commit deleted `loop_region_contains_escaping_range_append` outright and admitted any frame with a safe loop region, which pulled in a loop inside `random` and dropped a list element (`AssertionError: (46, 46, 45)`). This keeps the predicate as the frame-level precondition; `for_iter_call_bearing_comprehension.py` passes 3/3, because `append(<method call>)` is not `append(range(...))`.

### Gates on the rebuilt tree

| gate | result |
| --- | --- |
| `check.py --backend dynasm` | ALL PASSED 441/441 |
| `check.py --backend cranelift` | ALL PASSED 441/441 |
| `extra_tests/parity_tests/run.py --dynasm-only` | all parity tests pass |
| `cpython_tests/run.py --backend dynasm --baseline` | 208 PASS / 0 FAIL, no regressions |
| `cargo test --all --no-default-features --features dynasm` | 140 test binaries, 0 failures |

`range_ctor_in_loop` moves 122x → 22.2x dynasm / 30.2x cranelift / 24.7x wasm on the gate's own metric; `loops_compiled` 1 → 3, `bridges_compiled` 0 → 3, and the ceiling tightens 190 → 96.

— *commented by Claude*

youknowone:
Force-pushed `12a8ce30d75` → `95be045e92e`: five new commits. `git cherry -v HEAD origin/single-walker` reported `+` for none of the previous 16, so the divergence was a peer rebase and nothing was dropped.

### The Codex review on the previous push

| § | finding | disposition |
| --- | --- | --- |
| 1 | `eval.rs` per-backedge gate widened from `cached_loop_region_contains_escaping_range_append` to `frame_has_traceable_escaping_range_loop` | **reclassified — not a parity regression** |
| 2 | `_abc_init` installs raw `set`, not `SimpleWeakSet` | fixed |
| 2 | cache entries are weakrefs without the `_remove` callback | fixed |
| 2 | `cache_attr` treats every attribute error as a cache miss | fixed |
| 2 | `function_set_func_code` suppresses invalidation for an equal-value `__code__` | fixed |
| 2 | `w_property_reinit` suppresses `w_fget`/`w_fset` invalidation likewise | fixed |
| 2 | `staticmethod`/`classmethod` `w_function` modeled as mutable + `GuardValue` | deferred |
| 3 | registry is a strong `list` | fixed |
| 3 | `_abc_register` accepts callable non-types | **won't-fix (documented)** |
| 3 | `_abc_instancecheck` collapses `instance.__class__` and `type(instance)` | fixed |
| 3 | `_reset_registry` rebinds instead of clearing | fixed |
| 3 | `_get_dump` always returns an empty tuple | fixed |

**§1 is not a regression.** `region_safe` — the header-local term — is still an unconditional `&&`, so the current loop's own region must still be all-JIT-safe. More to the point, `can_enter_jit` (`interp_jit.py:117`) carries no admission predicate at all; PyPy admits every backward jump, and *both* spellings are pyre-local. The narrower one was the defect: a frame the frame gate had admitted still had every back edge but one refused, so traces closed on inner loops that could not then be entered (`bridge_no_targets_close=89`, 17896 guard failures, JIT slower than the interpreter).

**§3 `_abc_register`** keeps admitting a callable non-type because pyre's own stdlib needs it — `contextvars` runs `Mapping.register(Context)` at import and `Context` is a builtin function here. The weak-referenceability test upstream gets for free from its `isinstance(subclass, type)` guard now sits on the Rust side so `app_abc.py` stays verbatim.

### A latent SEGV the §2 quasi-immutable fix uncovered

Removing the equal-value skip made `bench/synth/make_function_inline.py` crash. It is **not** this branch's defect: the crash also reproduces on `f.__code__ = other.__code__`, and that spelling passes the `!=` guard under both the old and the new code, so old and new take an identical path there. The bench happened to write `__code__` its own current value, which is the only reason nothing had fired.

Root cause: `Function.code_watchers` was the one pointer-shaped slot with no entry in `FUNCTION_DESCR_GROUP`, so `clear_gc_fields` emitted no NULL store behind the `NewWithVtable`. A function the inline emit built and a guard escape materialized read recycled bytes, and `fset_func_code`'s `is_installed` — a bare non-null load — took them for an installed watcher and walked it. It cannot be a `Ref` entry, since the instance behind it is an off-GC box and the same entry would have the collector follow it; the words are named as plain `Int`s and the emit writes the zero, exactly as `can_change_code` already does.

### `_abc` now holds `SimpleWeakSet`

`app_abc.py:15-44` is ported as app-level source and installed the way `_contextvars` installs its own — the `_remove` callback closes over a weakref to the set, so it has to be built where a closure can be. All three collections are reached through `add` / `in` / `clear` / iteration. Output matches CPython line for line on registration, `_get_dump`, a spoofed `__class__`, both reset paths, the spent-weakref drop, and `case [...]` / `case {...}`.

### One JIT admission widened

The inline instantiation asked whether the metatype **was** `type` while its comment asked whether the metatype **overrides** `__call__`. `ABCMeta` supplies `__instancecheck__`/`__subclasscheck__`/`register` and leaves `__call__` alone, so the identity test refused every ABC-built class. It now resolves `__call__` and pins the metaclass's own version tag, since a metaclass that gains one does not move the class's tag. `abcmeta_type_call_inline.py` covers the admission and a `__call__` installed mid-loop.

This does **not** move `synth/inline_freevar_after_mayforce` (task #4): `Fraction` defines its own `__new__`, so it declines one check later. No gated counter moves on either backend.

### Gates

| gate | result |
| --- | --- |
| `check.py --backend dynasm` | ALL PASSED 442/442 |
| `check.py --backend cranelift` | ALL PASSED 442/442 |
| `extra_tests/parity_tests/run.py --dynasm-only` | all parity tests pass |
| `cpython_tests/run.py --backend dynasm --baseline` | 207 PASS / 0 FAIL, no regressions |
| `cargo test --all --no-default-features --features dynasm` | 153 test binaries, 0 failures |

`test_abc`, `test_collections`, `test_functools`, `test_enum`, `test_dataclasses`, `test_contextlib` and `test_descr` were also run on demand for the `_abc` change and pass.

### Still open

- **task #10** cranelift `raise_catch`: no lever found. The cranelift LABEL dead-load removal in `dee43c58191` is a real removal but not this one — `raise_catch_loop` carries ints, so `demoted_ref_positions_by_label` never fires there (1.35x → 1.38x, i.e. unchanged).
- **task #4** `inline_freevar_after_mayforce` (32.0x/41.5x against 86) and **task #5** `bound_method_builtin_fold` (24.7x/32.1x against 160) both pass, and both now have a named blocker: `__new__ overridden` for the first, `no jitcode for address` on `list.append`/`dict.setdefault`/`set.add` for the second.
- **task #9** B1/B3: all four B1 candidates were adversarially refuted; B3 does not reproduce on the shipped binary.
- Codex §2 `staticmethod`/`classmethod` `w_function?`.

— *commented by Claude*

PR #1299, opened 2026-08-17
majit: Defer Dynasm bridge constant conversion to diagnostic path

notJoon:
## Summary

Move the raw constant map conversion into the `PYRE_TRACE_OPS_DIAG` branch, avoiding diagnostic-only allocation and iteration during normal bridge compilation.

`compile_loop` has similar `const_pool` collection logic, but I left it duplicated rather than extracting a helper that isn't needed yet.
https://github.com/youknowone/pyre/blob/21ce53083b17d6c941b863ae20e415e0525009b2/majit/majit-backend-dynasm/src/runner.rs#L2394-L2397

## Self-review

- [x] I did not use AI to write the code of this patch.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined internal diagnostic handling without changing the resulting output or end-user behavior.


youknowone (approved):
Thank you so much! and welcome to Pyre project!

PR #1313, opened 2026-08-17
jit(cranelift): cut the per-crossing JITFRAME marshalling on the guard->bridge path

youknowone:
Two emission-side changes to cranelift's trace-crossing path.

`pyre/bench/fannkuch.py` runs ~2.9x slower on cranelift than on dynasm with
byte-identical jitstats (`loops_compiled=6 bridges_compiled=22
guard_failures=5049`), so the whole difference is emitted code. Guard exits
execute 79,000,000 times there while `guard_failures` counts 5049 metainterp
re-entries, i.e. ~99.99% of exits dispatch straight into an attached bridge.
The steady state is `loop -> guard exit -> bridge -> closing jump -> loop`,
two crossings per iteration, each marshalling through the JITFRAME:

| site | stores | loads |
|---|---|---|
| LABEL block re-stores every ref to its home slot | 18 | |
| guard exit publishes fail args to the dense slots | 18 | 2 |
| closing jump publishes the JUMP's label args | 24 | |
| target's LABEL loader reads them back | 2 | 25 |
| bridge reads its inputargs | | 18 |
| 2x (stack check + shadowstack push + caller pop + depth check) | ~8 | ~10 |
| **total** | **~78** | **~53** |

against dynasm's one patched branch.

**`ad683eb2bbc` — skip the stack probe and shadowstack push on in-code entries.**
A patched guard jump (`x86/assembler.py:987 patch_jump_for_descr`) and a closing
jump (`:2456-2462`) transfer inside compiled code on a JITFRAME the source is
already registered with, so RPython runs neither `_call_header_with_stack_check`
nor `_call_header_shadowstack` at the target. Cranelift cannot branch into
another function's body, so both transfers reach the target through its entry
and paid the full prologue. An `IN_CODE_ENTRY_KEY_FLAG` bit in `dispatch_key`
marks those entries; both realloc paths retarget the standing root-stack entry
per `x86/assembler.py:182-184 _build_frame_realloc_slowpath`. Removes 16 of the
131 ops.

**`47b9d40ec49` — move adjacent jitframe slots as one I128 access.**
The publish loops and LABEL loaders walk a run of `Signed` slots at
`ITEM0 + i*WORD` one 8-byte access at a time. aarch64 lowers a paired I128
access to `STP`/`LDP` (`aarch64/lower.isle:2767`, `:2655`); x86 splits an I128
access back into the two halves it already emits (`x64/lower.isle:3236`), so it
is a no-op there. `fib_loop`'s dumped trace goes from 158 to 58 single-slot
stores.

## Measurements

Interleaved CPU medians, paired rounds, on base `81e8a2d07ba` (**before** #1306
landed):

| | base | +in-code entry | +paired slots |
|---|---|---|---|
| fannkuch | 2.260s | 2.156s | **2.077s** (-8.1%, 11/13 rounds ahead) |
| bridge-crossing microbench | 0.509s | 0.431s (13/13) | 0.425s (10/13) |

fannkuch's ratio to dynasm goes 2.92x -> 2.68x. #1306 has since landed and
removed the write barrier's 24.6% self-time share, which raises the JIT-code
share these changes act on from ~0.70 to ~0.92; re-measurement on the current
base is in the comments.

`check.py --backend cranelift` 438/438 and `cargo test --release -p
majit-backend-cranelift` green at each commit.

## Not done here

Batching the same I128 pairing into `spill_ref_roots` / `reload_ref_roots` and
the LABEL-header ref-root re-seed **lost 7.2% on `fib_recursive`** (8/11 paired
rounds slower) — `iconcat`/`isplit` tie two values into one live range right at
a call boundary where registers are already scarce. The committed pairing
deliberately covers only the publish/loader runs, which are not adjacent to a
call.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -3503,6 +3503,95 @@ fn emit_shadowstack_retarget(
         .store(MemFlagsData::trusted(), jf_ptr, rst, -word);
 }
 
+/// Flags for a two-slot frame access.  `trusted()` also claims 16-byte
+/// alignment, which a `Signed` slot pair does not have — a JITFRAME slot is
+/// only word-aligned — so promise no more than "cannot trap".
+fn paired_slot_flags() -> MemFlagsData {
+    MemFlagsData::new().with_notrap()
+}
+
+/// Batches consecutive JITFRAME slot stores so that two adjacent slots leave as
+/// a single `store` of an `I128`.
+///
+/// `_push_all_regs_to_frame` (x86/assembler.py:1745) writes one slot per
+/// register, and a guard exit / JUMP publish is that same shape: a run of
+/// `Signed` slots at `ITEM0 + i*WORD`.  aarch64 has `STP` for exactly this, and
+/// cranelift reaches it through an `I128` store
+/// (`aarch64/lower.isle:2767` → `aarch64_storep64`).  x86 splits an `I128` store
+/// back into the two 64-bit halves it emits today
+/// (`x64/lower.isle:3236` "store the two 64-bit halves separately"), so batching
+/// changes nothing there.
+///
+/// Only the store is deferred, never the computation of the value, so the order
+/// in which fail args are resolved is unchanged.
+#[derive(Default)]
+struct PairedSlotStores {
+    pending: Option<(i32, CValue)>,
+}
+
+impl PairedSlotStores {
+    fn store(&mut self, builder: &mut FunctionBuilder, jf_ptr: CValue, offset: i32, val: CValue) {
+        if let Some((held_ofs, held_val)) = self.pending.take() {
+            let both_word_sized = builder.func.dfg.value_type(held_val) == cl_types::I64
+                && builder.func.dfg.value_type(val) == cl_types::I64;
+            if held_ofs + 8 == offset && both_word_sized {
+                let pair = builder.ins().iconcat(held_val, val);
+                builder
+                    .ins()
+                    .store(paired_slot_flags(), pair, jf_ptr, held_ofs);
+                return;
+            }
+            builder
+                .ins()
+                .store(MemFlagsData::trusted(), held_val, jf_ptr, held_ofs);
+        }
+        self.pending = Some((offset, val));
+    }
+
+    fn flush(&mut self, builder: &mut FunctionBuilder, jf_ptr: CValue) {
+        if let Some((offset, val)) = self.pending.take() {
+            builder
+                .ins()
+                .store(MemFlagsData::trusted(), val, jf_ptr, offset);
+        }
+    }
+}
+
+/// The read direction of `PairedSlotStores`: load a run of `Signed` frame slots,
+/// taking two adjacent ones per `load` of an `I128` (`aarch64/lower.isle:2655` →
+/// `aarch64_loadp64`, i.e. `LDP`; x86 splits it into the two halves again).
+///
+/// `offsets` is in the order the caller wants the values back; only neighbours
+/// that are genuinely adjacent in the frame are paired.
+fn load_frame_slot_run(
+    builder: &mut FunctionBuilder,
+    jf_ptr: CValue,
+    offsets: &[i32],
+) -> Vec<CValue> {
+    let mut vals = Vec::with_capacity(offsets.len());
+    let mut i = 0;
+    while i < offsets.len() {
+        if i + 1 < offsets.len() && offsets[i] + 8 == offsets[i + 1] {
+            let pair = builder
+                .ins()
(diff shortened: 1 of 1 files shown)


youknowone:
Re-measured on the current base (`629e8e355a5`, i.e. with #1306 in), which is
what the PR body said was still owed. Both arms built from this tree, the
control by reverse-applying the branch diff in place; interleaved rounds, CPU
seconds, load average ~50 throughout — which is why the paired count matters
more than the medians.

| arm | median | min |
|---|---|---|
| `629e8e355a5` | 2.227s | 2.034s |
| + these 2 commits | **2.008s** | 1.878s |

**−9.8%, with the branch ahead in 13/13 paired rounds.**

Larger than the −8.1% recorded in the PR body, and in the direction the cost
model predicts: #1306 removed the write barrier's 24.6% self-time share, so the
JIT-code share these changes act on rose from ~0.70 to ~0.92, and a change that
removes a fixed number of memory operations is worth correspondingly more.

The 2.92x → 2.68x dynasm ratio in the body is still the pre-#1306 measurement —
the dynasm binary here predates the rebase, so that row is not re-measured yet.

— *commented by Claude*

youknowone:
Added `c4e42098498` — a third change on the same path, and the one that pays most.

The LABEL block's parameter loop stored each kept parameter into its ref-root
home at `ref_root_base_ofs + slot*8`. A LABEL block *is* the loop header, so
that was one store per live reference per iteration: **46 stores per iteration
on fannkuch**. `regalloc.py` publishes a reference to its frame home at
`before_call()`, not at a LABEL, and the readers here agree — a guard's own map
is `allocate_gcmap(&failarg_ref_slots)` and never names the ref-root region,
while `get_gcmap`, which does, is a call-site map that `spill_ref_roots`
refreshes on the instruction before.

**One class still needs the store, and the first version of this commit got it
wrong.** `demoted_failarg_slots` is trace-global while `loop_phi_keep` is per
LABEL, so a raw demoted at an earlier LABEL can be a kept block param at a later
one — and for that raw the home is what the readers use: `get_gcmap` marks it
with no liveness test, `spill_ref_roots` and `reload_ref_roots` both refuse to
write it, and `resolve_failarg_opref` tests the demoted offset *before* the
`use_var` fallback. On a loader re-entry nothing else writes that word, so
deleting the header store outright left the collector tracing a slot the entry
never wrote. The store now stays for exactly that class, gated on
`is_demoted_failarg`; both inserts into that map sit in the same fall-through
block as the seeding store, so the gate reads true exactly where a store to the
home already dominates.

Worth recording that **the unconditional-deletion version passed everything**:
`check.py --backend cranelift` 438/438, and seven benches byte-identical under
`PYPY_GC_NURSERY=256KB`. The defect is shape-dependent — it needs the
demote-at-one-LABEL / keep-at-a-later-LABEL trace, which occurs in 4 of the
first 120 synth fixtures, all exception-related. It was found by review, not by
the gates.

## Measurements

Interleaved CPU medians, 15 rounds, load average 15 (the earlier numbers in this
PR were taken at load ~50 and are inflated in absolute terms):

| arm | median | vs base |
|---|---|---|
| `629e8e355a5` | 1.492s | |
| + in-code entry + paired slots | 1.366s | −8.4% |
| + this commit | **1.272s** | **−14.7%** |

Ahead of base in 15/15 paired rounds; against the previous tip every round of
this arm is below every round of that one. fannkuch loses all 46 loop-header
ref-root stores and no other emitted store moves.

`check.py --backend cranelift` 438/438 and `cargo test --release -p
majit-backend-cranelift` green. One `fib_recursive` timeout on an earlier gate
run was load: re-timed directly it is 1.89s here against 1.88s on the previous
tip and 2.17s on base.

— *commented by Claude*

youknowone:
Ratio to dynasm, re-measured against a dynasm binary built from this same tree
(the figure in the PR body was against a pre-rebase binary). Three arms
interleaved in one run, so the load is common to all three:

| arm | median | vs dynasm |
|---|---|---|
| dynasm | 0.937s | 1.00x |
| `629e8e355a5` cranelift | 2.345s | 2.50x |
| this branch | 2.019s | **2.15x** |

So on the current base the branch takes fannkuch from 2.50x to **2.15x** of
dynasm — a bit over a third of the excess. The 2.92x → 2.68x pair quoted in the
PR body was measured before #1306 landed and is superseded by this row.

Absolute times here are inflated relative to the −14.7% table above (that run was
at load 15, this one at ~30); the ratio is what survives, since all three arms
take the same rounds.

What is left is the marshalling the crossing itself requires — the guard exit's
fail-arg publish, the closing jump's publish, the target loader's reads and the
bridge's inputarg reads — which no local change removes, because the two sides
are different cranelift functions.

— *commented by Claude*

PR #1325, opened 2026-08-18
majit: #1287 review follow-ups, an inline-trial geometry diagnostic, and one wasm cost removal

youknowone:
Six commits: four follow-ups from the #1287 review, one diagnostic split, and one cost removal in
the wasm backend.

## Review follow-ups from #1287

- **`majit: name has_target_tokens in the target_tokens doc`** — the doc claimed `first_target_token`
  was the list's one reader. `has_target_tokens` reads it too; it just has zero call sites. Names
  both and says so. (The reviewer's accompanying claim that this makes it a token-presence gate is
  not adopted — nothing calls `has_target_tokens`.)

- **`majit: replace the retrace seed's justification`** — the old comment asserted that upstream's
  arm without a resumekey "has no such token". It does. `compile.py:355-356` resolves
  `loop_jitcell_token = metainterp.get_procedure_token(greenkey)` before any resumekey is consulted,
  `:359` records the closing JUMP under it, and `unroll.py:321-325` walks its whole `target_tokens`.
  `unroll.py:297` then appends the retrace's own token to that same list, so the accumulation
  happens inside the optimizer rather than in either `compile_and_attach` arm. The code is
  unchanged; only the justification is, and it now states that emptying the seed on that arm is a
  deliberate deviation and why it is safe here.

- **`majit: publish the short-preamble producer wherever the builder lives`** + **`majit: skip the
  short-preamble token write after an aborted replay`** — an RAII guard re-points the walker at the
  builder's current owner and restores the previous slot on drop, and the target token's short
  preamble is no longer written when the replay left the builder partial.

## Diagnostic

- **`majit: split the four frozen-frame shortages the inline trial reports as one`** — the inline
  trial's geometry decline used to report one opaque reason. `LabelResumeData::shortage` now returns
  which of the four frozen-frame quantities fell short, with the needed/available pair, surfaced as
  a new `inline_decl_label_resume_layout` diag label and an `inline_geometry` line.

  This immediately paid for itself: **111 of 111 geometry records across the 429-fixture synthetic
  corpus are `label_resume_ref_slots=N/2`, N=3..14**, and 116 of ~119 post-build inline-trial
  failures are that one shortage — previously invisible.

## Cost removal

- **`Decouple inline bridge enablement from re-emission`** — `inline_bridge_enable()` also called
  `reemit_enable()`, which switches on the identity re-emission probe (the first bridge installed
  for a loop rebuilds it into a byte-identical module purely to prove the replacement path works).
  Inlining reaches the replacement path directly through `reemit_loop` and does not need the probe,
  so the two enables are separated and the `bridge_slots` bookkeeping the inline path does need is
  widened to cover it.

  Measured with `PYRE_WASM_INLINE_BRIDGE=1` over the whole 429-fixture synthetic corpus, both arms
  by fuel, stdout identical in 429/429:

  | | before | after |
  |---|---:|---:|
  | aggregate vs default | +1.214% | +1.178% |
  | aggregate excluding `inline_multiframe_branchy_carrier` | +0.086% | **+0.050%** |
  | fixtures losing 0.1–1% | 100 | **45** |
  | fixtures losing 1–10% | 12 | **4** |

  The four measured wins are unchanged: `global_quasiimmut_invalidation` −19.31 → −19.38,
  `kept_stack_depth_gt1_heap` −6.93 → −6.96, `kept_stack_depth_gt1` −4.67 → −4.73,
  `if_else_jump_forward` −3.19 → −3.19.

  `PYRE_WASM_INLINE_BRIDGE` remains **off by default**; this only reduces what the flag costs when
  it is set.

## Not in this PR

The characterisation behind that last commit found a single analysis defect that produces the
facility's per-iteration tax, the 116 discarded trial builds, and its one catastrophic fixture:
`LabelResumeData::collect` runs over the owner's ops concatenated with the appended bridge regions,
so a region's fresh `InputArg`s — which are `defined_before` every resumable LABEL and appear in no
LABEL arglist — become frozen-frame captures at every resumable label of the owner, even though the
entry dispatch can never land inside a region. Fixing that is a separate change with its own kill
conditions and is not attempted here.

## Summary by CodeRabbit

- **New Features**
  - Added detailed inline layout diagnostics identifying the shortage type and required versus available capacity.
  - Added a total failure count so diagnostics include failures beyond retained records.
  - Updated runtime output to display named diagnostic categories and capacity details.

- **Bug Fixes**
  - Improved handling of inline bridge and short-preamble state, helping prevent misleading or incomplete diagnostics during optimization.

- **Documentation**
  - Clarified diagnostic output and token-state behavior for easier interpretation and troubleshooting.

### DIFF
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -1035,16 +1035,38 @@ impl LabelResumeData {
         self.capture_by_id.get(r.raw() as usize).copied().flatten()
     }
 
+    fn shortage(&self, frame: FrameGeometry) -> Option<super::FrameShortage> {
+        if self.ref_slots > frame.label_ref_slots {
+            return Some(super::FrameShortage::new(
+                super::FrameShortageKind::LabelResumeRefSlots,
+                self.ref_slots,
+                frame.label_ref_slots,
+            ));
+        }
+        for storage in self.capture_by_id.iter().flatten() {
+            match storage {
+                LabelCaptureStorage::ValueSlot(slot) if *slot >= frame.value_slots => {
+                    return Some(super::FrameShortage::new(
+                        super::FrameShortageKind::LabelResumeCaptureSlots,
+                        slot + 1,
+                        frame.value_slots,
+                    ));
+                }
+                LabelCaptureStorage::RefSlot(slot) if *slot >= frame.label_ref_slots => {
+                    return Some(super::FrameShortage::new(
+                        super::FrameShortageKind::LabelResumeCaptureSlots,
+                        slot + 1,
+                        frame.label_ref_slots,
+                    ));
+                }
+                LabelCaptureStorage::ValueSlot(_) | LabelCaptureStorage::RefSlot(_) => {}
+            }
+        }
+        None
+    }
+
     fn supported_by(&self, frame: FrameGeometry) -> bool {
-        self.ref_slots <= frame.label_ref_slots
-            && self
-                .capture_by_id
-                .iter()
-                .flatten()
-                .all(|storage| match storage {
-                    LabelCaptureStorage::ValueSlot(slot) => *slot < frame.value_slots,
-                    LabelCaptureStorage::RefSlot(slot) => *slot < frame.label_ref_slots,
-                })
+        self.shortage(frame).is_none()
     }
 
     fn frame_offset(&self, storage: LabelCaptureStorage, frame: FrameGeometry) -> u64 {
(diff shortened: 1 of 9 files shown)


youknowone:
On the parity review's §1 (`pyjitpl.rs` retrace seed) — reclassifying it to §3, pre-existing, with
no code change in this PR.

The gate it names is already on `main`:

```
$ git show origin/main:majit/majit-metainterp/src/pyjitpl.rs | rg -n 'retrace_resumekey.is_some'
8335:            if retrace_resumekey.is_some() {
```

and the commit the review points at (`majit: replace the retrace seed's justification`) adds 44
lines to that file, **all of them comment** — filtering the added lines for non-comment content
returns nothing. Its message says so explicitly: "The code is unchanged."

On the substance, the review is right about upstream and that is exactly what the new comment now
says. `compile.py:355-356` resolves `get_procedure_token(greenkey)` before any resumekey is
consulted, `:359` records the closing JUMP under it, and `unroll.py:321-325` walks the whole
`target_tokens` list — plus `unroll.py:297` appends the retrace's own token to that same list during
optimization, so the accumulation happens inside the optimizer rather than in either
`compile_and_attach` arm. The previous comment claimed the arm without a resumekey "has no such
token", which was false; replacing that claim is the entire content of the commit.

Whether to also change the code was adjudicated separately and the ruling was keep-the-code,
rewrite-the-justification. The reasons are in the new comment: on this arm the close gate in
`jump_to_existing_trace_impl` admits a foreign target only when it belongs to the artifact this
compile attaches to, and on that arm there is none, so a seed's remaining consumers here are the
ownership rebind and the republication rather than candidate matching.

— *commented by Claude*

PR #1334, opened 2026-08-18
cpyext: the cyclic-collection protocol, the builtin type statics, and the entry points an extension reaches for

youknowone:
Continues the cpyext slice: the cyclic-collection protocol, the entry points an
extension actually reaches for, and one interpreter deadlock found on the way.

`markupsafe` builds and runs against these headers: its upstream suite is
39 passed / 0 failed / 1 skipped natively and 39 / 1 through the C speedups,
the one failure being `test_leak`, which asserts on `gc.get_objects()` counts.

## What is here

**The collector side.** `rawrefcount` blocks that declare `Py_TPFLAGS_HAVE_GC`
report their own references through `tp_traverse`, and the borrow this layer
owns on a container's behalf is reported alongside — a reference cycle running
through a C-defined type is collected and its `tp_clear` runs.

**The 49 builtin `PyTypeObject` statics.** `&PyList_Type` is an address fixed at
link time whose body the runtime fills in place, so `Py_IS_TYPE`,
`PyObject_TypeCheck`, the `O!` argument format and `Py_tp_base` all work. They
are bound *before* the singletons: binding a singleton resolves its own
`ob_type`, and a type reached before its static is entered would get a
synthesized block instead — which is then what `Py_TYPE(Py_True)` answers with
forever. `PyCFunction_Type` is deliberately absent, pyre having two distinct
types named `builtin_function_or_method`.

**Entry points**, each measured rather than assumed: the exception object and
error indicator, the warnings family, the codecs and `wchar_t` and filesystem
conversions, the `str` operations with a rewritten `%`-format engine, the GIL
hand-off, `bytearray` / `complex` / `weakref`, a `dict` subclass wherever
`PyDict_Check` is the gate, the errno / audit / import trio, the locale codec,
`PyUnicode_FromKindAndData`, `Py_ReprEnter` / `Leave`, `Py_HashBuffer`,
`PyDict_SetDefaultRef`, `_PyType_Name`, `Py_GenericAlias`, `PyMutex`, the
`PyThread_type_lock` family, `_PyBytes_Resize`, `_PyErr_ChainExceptions1` and
`Py_FatalError`.

Of the 746 public `PyAPI_FUNC` entry points CPython 3.14.7 declares in its
top-level `Include/*.h`, 455 are now present in some form `Python.h` offers.

## Two defects fixed along the way

- **`O!` admitted a `__class__` liar.** It asked `PyObject_IsInstance` where
  `convertsimple` tests the layout with `PyType_IsSubtype(Py_TYPE(arg), type)`.
  Measured: CPython raises `TypeError: argument 2 must be list, not Liar`, pyre
  admitted the object and the caller went on to read it through `list`'s
  fields.
- **`register_exc_class` deadlocked the interpreter.** It held
  `EXC_CLASS_REGISTRY`, a plain `std::sync::Mutex`, across the method stamping,
  which takes the striped dict locks — two unrelated dictionaries can share a
  stripe, so a thread holding one and waiting for the registry closed a cycle
  with the registry's holder waiting for a stripe. Seen live: 16 threads
  blocked, no progress for 90 minutes. `builtins::tests::
  concurrent_builtin_installs_all_finish` waits with a deadline rather than
  joining, a deadlock having no failing assertion of its own.

## The ABI gate

`pyre/scripts/cpyext-abi.py` compares every export and header inline against a
snapshot of CPython 3.14.7's own declarations. It gained `PyAPI_DATA` in this
branch — it had never looked at exported data at all — and, in the last commit,
pointer typedefs (`typedef void *PyThread_type_lock;`) and object-like type
macros (`#define PY_TIMEOUT_T long long`), which the word-only alias pattern
could not see. 448 exports, 22 header inlines and 121 data objects are checked;
all match.

## Testing

Every cpyext test file states its expectations as values taken from CPython
3.14.6 running the same script against the same fixture.

- CI cpyext leg: 91 `test result: ok`, 0 failed
- `cpyext-abi.py check` and `generate --check`: green
- `pyre/extra_tests/run.py --gated-only`: 62/62 on cpython, dynasm and cranelift
- `cargo fmt --all --check`: clean

## Known divergences

Documented at their definitions and listed in `pyre/docs/cpyext.md`. The ones
worth naming here: a capsule destructor is recorded and never called; the module
state block is never released; `PyList_New(n)` fills with `None` rather than
NULL; a type is built on a single base; the argument parser words a bad argument
its own way; `PyThread_acquire_lock_timed` never answers `PY_LOCK_INTR`, its
wait not being interruptible; and the locale codec converts as UTF-8 rather than
following `LC_CTYPE`, pyre reading no locale.

https://claude.ai/code/session_01VKvxTiG1M3K7KuKxVVezaX

## Summary by CodeRabbit

* **New Features**
  * Expanded CPython extension compatibility with broad support for Unicode, bytes, bytearrays, complex numbers, dictionaries, exceptions, warnings, weak references, locks, thread state, GIL handling, imports, mappings, and generic aliases.
  * Added cyclic garbage-collection support for C-extension objects and improved raw-reference-counting integration.
  * Added ABI validation for exported functions, data objects, constants, and type information.
* **Bug Fixes**
  * Improved dictionary subclass behavior, Unicode error reporting, warning handling, buffer resizing, and reference management.
* **Tests**
  * Added extensive cross-platform integration coverage for the expanded extension APIs and garbage collection behavior.


youknowone:
Rebased onto `main` (`8c872c1d9ce`), no conflicts, plus three new commits.

### The ubuntu `cargo test` failure on the previous head — root cause and fix

`pyrex --test cpyext_runtime` aborted in `cpyext::pyobject::incref` on
`0x694f006f7774004f`. Those eight bytes are the fixture's own rodata read as a
pointer: `4f 00 74 77 6f 00 4f 69` is `"O" NUL "two" NUL "Oi"`, the three
literals `audit()` passes to `PySys_Audit`.

`Py_VaBuildValue` took the address of its own `va_list` **parameter**. Where
`va_list` is an array type the parameter decays to a pointer, so `&va` is a
pointer to that pointer, not a `va_list *`, and the walker read the format
string itself as the first argument. `-w` on the fixture compile hid the type
mismatch.

`__builtin_types_compatible_p` under `-fsyntax-only`, per target:

| target | `&va` (parameter) | `&copy` (`va_copy`) |
| --- | --- | --- |
| `arm64-apple-darwin` | `va_list *` | `va_list *` |
| `aarch64-unknown-linux-gnu` | `va_list *` | `va_list *` |
| `x86_64-unknown-linux-gnu` | `struct __va_list_tag **` | `va_list *` |

Which is why macOS was green and only ubuntu-x86_64 aborted. `1a569869907`
walks a `va_copy` instead, the spelling `PyUnicode_FromFormatV` in the same
header set already uses.

### `dict.pop` with a key it cannot rule out

`{1: 2}.pop(unhashable)` answered a miss where `in`, `[]`, `get`, `setdefault`
and `del` all raise `TypeError`: `w_dict_pop_checked` dispatched to the typed
strategy's `pop`, which reaches the object strategy through the *infallible*
lookup and so swallows the hash error. `a36fd985b2b` moves the dictionary to the
object strategy before the dispatch when the key is neither the strategy's own
type nor one it can never compare equal to — the third arm of
`AbstractTypedStrategy.pop` (`dictmultiobject.py:1136-1138`), and the arm that
hashes. int-, str- and bytes-keyed dictionaries now answer the `TypeError`
`test_dict.test_unhashable_key` asks for; an empty dictionary still answers the
miss without hashing. `test_dict` stays green, and the dictionaries survive the
switch intact.

### The string writer and ten container entry points

`8421558319d` adds `cpyext/unicodewriter.rs` — `PyUnicodeWriter_Create` /
`Discard` / `Finish` and `WriteChar` / `WriteUTF8` / `WriteASCII` /
`WriteWideChar` / `WriteUCS4` / `WriteStr` / `WriteRepr` / `WriteSubstring`,
with the variadic `PyUnicodeWriter_Format` written on the header over the
format engine already there — and `PyDict_Pop` / `PyDict_PopString`,
`PyDictProxy_New`, `PyMapping_GetOptionalItem` / `GetOptionalItemString`,
`PyList_Clear` / `PyList_Extend`, `PyBytes_Join` / `PyBytes_Concat` /
`PyBytes_ConcatAndDel`.

`PyBytes_Concat` reads both sides as buffers rather than adding them, so
`__radd__` is never consulted and a right side exporting no buffer is the
`can't concat str to bytes` refusal CPython gives.

`pyrex/tests/cpyext_writer.rs` takes its expectations from CPython 3.14.6
running the same script against the same fixture. Of the 58 rows measured, the
three that differ are the `PyErr_BadInternalCall` reports already recorded as a
divergence — the runtime's own call has no caller file and line to name.

The public census is 460 of the 746 public `PyAPI_FUNC` declarations, up from
455.

— *commented by Claude*

PR #1336, opened 2026-08-18
jit: port Function quasi-immutability, and carry const heap short boxes on their own export channel

youknowone:
## What this is

`pypy/interpreter/function.py:34-42` declares nine `Function` fields quasi-immutable:

```python
_immutable_fields_ = ['code?', 'w_func_globals?', 'closure?[*]', 'defs_w?[*]',
                      'name?', 'qualname?', 'w_objclass?', 'w_text_signature?',
                      'w_kw_defs?']
```

pyre had none of it. `descr.rs` documented the gap and stood a `GuardValue` in for the
invalidation. This branch ports the machinery: a `mutate_<name>` slot per field, the
notify before every write, the descriptor declaration, and both watcher halves.

`(is_immutable, is_quasi_immutable) = (true, true)`, because `descr.py:229 is_pure =
STRUCT._immutable_field(fieldname) != False` is true for a quasi field, and
`is_always_pure` is `is_immutable && !is_quasi_immutable` — so the fold still stays
behind `record_quasiimmut_field`, per `jtransform.py:895-903`.

## The uninitialised-slot defect, and the fix

`emit_make_function_inline` allocates a `Function` with `NewWithVtable`, and
`rewrite.py:498-504 clear_gc_fields` emits delayed NULL stores for the census's
`gc_fielddescrs` only. incminimark does not zero-fill (`gc.rs malloc_zero_filled =
false`), so every non-GC byte of a JIT-allocated `Function` keeps whatever the address
last held. Upstream has no such hole: its `mutate_<field>` **is** a GC pointer, so
`clear_gc_fields` covers it.

The first cut put nine inline `QuasiImmutField`s on the struct — 144 bytes,
`AtomicPtr` + `parking_lot::Mutex` each, no descriptor. Nothing initialised them.
`QuasiImmutField::new()` is a `const fn` whose value is all zero bytes, so they are
*supposed* to start zeroed; in practice word1 read `0x442000000` object after object,
and eventually word0 landed non-null and `invalidate` chased it. `function_destructor`
also ran `drop_in_place` over that garbage on **every** sweep of a JIT-allocated
function, which is why the failure was not confined to any probe: `pyre/check.py` was
red on nine synth fixtures in the **default** configuration on both backends
(`synth/closure_per_call` at `exit -11`), and the CPython gate showed
`test.test_itertools PASS -> CRASH` and `test.test_strtod PASS -> TIMEOUT`.

`aa1b2f5dda8` collapses the nine slots to one lazily allocated block pointer,
`Function.mutate_slots: AtomicPtr<FunctionQuasiImmutSlots>`, and `1c06b1e598f` puts
that word on upstream's own footing. `gc_fielddescrs` is derived from
`is_pointer_field()`, and the collector's census is the separate hand-maintained
`FUNCTION_GC_PTR_OFFSETS`, which the word is deliberately absent from. Declaring it
`Type::Ref` therefore lets `clear_gc_fields` emit the NULL — exactly as it does for
upstream's `mutate_<field>`, that one being a real GC pointer — while the collector
still never follows the `Box`. Null is not a sentinel invented here: it is
`quasiimmut.py:130`'s null `mutate_<name>`, meaning nothing has folded the field, so a
write revokes nothing. The block is created on first install (CAS, loser frees its own
allocation) and freed by `function_destructor` via `swap`, so a function reached twice
frees once. `size_of::<Function>()` moves 176 → 184, and the
`fold_size_const_real_function_object_size` anchor is re-taken from the artefact.

The first cut of that word was declared `Type::Int` with a hand-written NULL store,
which put it in resume data int-tagged and made
`synth/call_loop_local_function` — the fixture whose own header names the
`_immutable_fields_` set this port implements — panic in `resume.rs` with `decode_ref:
unexpected tag 1`, alongside a timeout on `synth/range_ctor_in_loop`. Both pass now.

Nursery size was never the variable — `PYPY_GC_NURSERY` at 256KB, 512MB and default all
crashed identically.

Section 1 was empty. Section 2 carried four findings:

- **`jit_force_quasi_immutable` missing before the setters** — *won't-fix, documented*.
  pyre's `SET_FUNCTION_ATTRIBUTE` compiles to a `dont_look_inside` residual
  (`runtime_ops.rs jit_set_function_attribute`), so the tracer never sees the store and
  has nowhere to hang a pre-store force. The invalidating half **is** present: the
  residual really runs and really calls `invalidate()`, and `heap.rs:3265` emits
  `GUARD_NOT_INVALIDATED` which the backends patch. The correctness backstop is
  `quasiimmut_field_still_valid` (`heap.py:798-804`), which re-reads the field and
  discards any loop whose recorded `?` value moved during tracing — so the missing abort
  costs a wasted trace attempt, never a stale fold. This is also the position pyre
  already takes for the analogous mapdict force, which ships behind a default-OFF
  `PYRE_QMUT_MAPDICT_FORCE`.
- **The other three** are the temporary probe gates, addressed below.

## The five investigation gates are gone

`a4ac3bf3f65` removes the four inert ones with their ON paths.
`virtualstate.rs` returns to its base content — the branch had only wrapped the five
`_jump_to_existing_trace` runtime fallbacks in `fallback_enabled(...)` and added a
per-entry verdict print, and `PYRE_VSTATE_FALLBACK` already defaulted to "all fallbacks
on", so the removal is not a behaviour change. `trace_inputarg_values` goes with
`PYRE_INPUTARG_VALUE_MODE`: the channel is inert in the default mode, `canonical`
delivers nothing because the bridge's runtime boxes name shifted positions, and
`shifted` is measured wrong. `bind_canonical_inputarg` returns to its base shape, and
the pre-existing `ia.set_value` on the bridge frontend inputargs (`history.py:803`) is
untouched.

The fifth, `PYRE_CONST_SHORT_BOXES`, stays — and now says why. `967245757e5` made the
const-result append unconditional, as `shortpreamble.py:272-281 ← :369-373` does, on the
strength of three fixtures passing with the gate on. The 431-fixture run overruled it:
8 failed, 423 passed, seven of them JIT panics that append introduces —
`make_equal_to: cross-type forward (Box.type invariant)` on
`check_exc_match_invalid_class`, `exc_mixed_classes_bridge_flavor`,
`except_tuple_clause_hot`, `exception_traceback_frame_lineno`,
`frame_lineno_mid_replay_regression`, `jit_callee_raised_exc_value`, and `resume.rs
decode_ref: unexpected tag 1` on `call_loop_local_function`. `9276b765528` reverts it
and the §5b row now names both panics as what retires the gate.

## Verification

`pyre/check.py --backend dynasm --synthetic-only`: **431/431 passed**.
`cargo test -p pyrex --test gate_triage_complete`: 6/6.

## Deliberately not done

`#[majit_macros::jit_immutable_fields("code?", ...)]` — the declarative spelling — is
not applied. It would make the codewriter emit `record_quasiimmut_field` into the
analyzer's jitcodes, and that bytecode has no arm in `pyjitpl/dispatch.rs` (unknown
bytecodes panic) and no registration in the production blackhole builder, which
`production_bh_builder_overlay_only_gap_snapshot` pins. The declaration lives on
`FUNCTION_DESCR_GROUP` instead, which is what every hand-written trace path resolves
through.

— *authored by Claude*

## Summary by CodeRabbit

* **New Features**
  * Improved JIT responsiveness when function metadata, code, defaults, closures, or related attributes change.
  * Added experimental support for preserving constant short-preamble results, controlled by `PYRE_CONST_SHORT_BOXES` (disabled by default).

* **Bug Fixes**
  * Improved handling and replay of constant heap values during optimized execution.

* **Documentation**
  * Documented the experimental configuration, current limitations, and known failure conditions.

### DIFF
--- a/majit/majit-metainterp/src/optimizeopt/mod.rs
+++ b/majit/majit-metainterp/src/optimizeopt/mod.rs
@@ -80,6 +80,14 @@ pub(crate) fn next_snapshot_pos<T>(store: &[Option<T>]) -> i32 {
 
 pub(crate) use crate::majit_log_enabled;
 
+// Temporary probe: the const heap short-box path has never executed, so
+// enabling it moves the compiled short preamble on every fixture.
+pub(crate) fn const_short_boxes_enabled() -> bool {
+    static ON: std::sync::LazyLock<bool> =
+        std::sync::LazyLock::new(|| std::env::var_os("PYRE_CONST_SHORT_BOXES").is_some());
+    *ON
+}
+
 /// info.py:865-894 `getrawptrinfo` / `getptrinfo` return shape, with
 /// RPython `_forwarded` object identity preserved.
 ///
--- a/majit/majit-metainterp/src/optimizeopt/mod.rs
+++ b/majit/majit-metainterp/src/optimizeopt/mod.rs
@@ -792,6 +792,15 @@ pub struct OptContext {
     /// optimizer.py:34 `self.inputargs = inputargs` parity.
     /// Typed InputArg OpRefs; slot `i` is `OpRef::input_arg_typed(i, tp)`.
     pub inputargs: Vec<majit_ir::OpRef>,
+    /// Bridge frontend values by logical inputarg index. Empty means the
+    /// current optimization run has no runtime-value channel. The values are
+    /// copied onto fresh `InputArgRc` hosts without sharing their forwarding
+    /// identity (`unroll.py:183-193`).
+    pub(crate) trace_inputarg_values: Vec<Option<majit_ir::Value>>,
+    /// Number of distinct inputarg positions that received a concrete value
+    /// while canonical bindings were installed. Used only by the temporary
+    /// bridge diagnostic in `optimizer.rs`.
+    pub(crate) inputarg_value_stamped_positions: usize,
     /// Strong `InputArgRc` ownership for the inputargs seeded by
     /// `with_inputarg_types`. Production traces own their `InputArgRc`s
     /// via `TreeLoop.inputargs`; the test-and-fallback helper
@@ -1744,6 +1753,8 @@ impl OptContext {
             snapshot_frame_pcs: Vec::new(),
 
             inputargs: Vec::new(),
+            trace_inputarg_values: Vec::new(),
+            inputarg_value_stamped_positions: 0,
             inputarg_refs: Vec::new(),
             resop_refs: indexmap::IndexMap::default(),
             live_synthetics: Vec::new(),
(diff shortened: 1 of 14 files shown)


youknowone:
## Correction: the crash is NOT probe-only

The PR body says the uninitialised-slot crash is "reachable today only with
`PYRE_INPUTARG_VALUE_MODE=shifted` **and** `PYRE_CONST_SHORT_BOXES=1` together".
That is wrong, and CI shows it. `pyre/check.py` fails on both backends in the
**default** configuration:

```
FAIL dynasm    synth/closure_per_call  crash (exit -11)
FAIL cranelift synth/closure_per_call  timeout (>20s)
```

plus `synth/make_function_inline`, `synth/foriter_make_function_body`,
`synth/del_cellvar_walk_commit`, `synth/exception_as_cell_cleanup`,
`synth/gc_iterator_source_drop`, `synth/hash_subclass_disabled`,
`synth/list_length_hint_validate`, `synth/surrogate_class_kwargs` — nine
fixtures, dynasm and cranelift alike. The MAKE_FUNCTION / closure fixtures are
exactly the ones that drive `emit_make_function_inline`.

`cargo test` is red for a separate, unrelated reason: the five `PYRE_*` probe
gates this branch adds have no row in `pyre/gate-triage.md`
(`every_live_gate_has_a_triage_entry`).

The mechanism is the one already described in the PR body — the JIT-allocated
`Function`'s `mutate_<name>` slots are never initialised — but the trigger is
broader than stated: `function_destructor` runs `drop_in_place` over those slots
on **every** sweep of a JIT-allocated `Function`, so a garbage word is freed as
a `QuasiImmut` pointer regardless of any gate.

Fixing this branch, not merging it as-is.

— *commented by Claude*

PR #1341, opened 2026-08-18
majit: remove transient allcations from regalloc walk

notJoon:
## Summary

Remove avoidable allocations and clones from the register-allocation walk

- Take ownership of `j2_ops` once at the start of `walk_operations`.
- Borrow each `LirOp` during dispatch instead of cloning it.
- Iterate over uses and fail arguments directly in `_free_j2_op_vars`, dropping the temporary `uses` and `fail_uses` vectors.
- Extract the shared operand-freeing logic into `_possibly_free_j2_vars`.
- Preserve the existing use → fail arguments → result freeing order and the register-allocation policy.

## Self-review

  - [x] Auto-review section 1 is clear. This check is mandatory.
  - [x] Auto-review section 2 is clear. If this is not checked, please add a comment explaining why.
  - [ ] I added `Assisted-by` to commit messages to the commits AI wrote.

## Summary by CodeRabbit

* **Performance**
  * Improved internal operation processing efficiency by reducing unnecessary data copying.
  * Streamlined variable cleanup while preserving existing behavior across supported operations.

### DIFF
--- a/majit/majit-backend-dynasm/src/regalloc.rs
+++ b/majit/majit-backend-dynasm/src/regalloc.rs
@@ -2396,6 +2396,8 @@ impl<'a> RegAlloc<'a> {
     pub fn walk_operations(&mut self) -> Vec<RegAllocOp> {
         let operations: &'a [Op] = self.operations;
         let inputargs: &'a [InputArg] = self.inputargs;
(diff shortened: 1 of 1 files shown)


notJoon:
planning to check CI failures in this evening

youknowone (approved):
great, thanks!

PR #1363, opened 2026-08-19
cpyext: what a Cython module reaches for, and the descriptor-cache order it made visible

youknowone:
Two commits. The first is the cpyext slice a Cython-generated module needs; the
second is a JIT descriptor-cache fix that slice made visible.

# 1. `cpyext`: the callables, thread state and headers a Cython module reaches for

Cython's generated C takes its `CYTHON_COMPILING_IN_PYPY` branch when
`PYPY_VERSION` is defined, and that branch names entry points and header
spellings this layer did not carry. Against a `cythonize`d module's C, the
compile went from 133 errors to 8, measured after each tier rather than
asserted.

## What an extension can now reach

Bound methods (`PyMethod_New` / `PyMethod_Function` / `PyMethod_Self`, a new
`funcobject.rs`), the `PyMethodDef`-backed callables (`PyCFunction_New` /
`_NewEx` / `_GetFunction` / `_GetFlags` / `_GetSelf`), unraisable-exception
reporting (`PyErr_WriteUnraisable` and the `_PyPyre_WriteUnraisable` core
under `PyErr_FormatUnraisable`), imports (`PyImport_GetModuleDict`,
`PyImport_ImportModuleLevel`, `PyImport_ImportModuleLevelObject`), the object
protocol pieces a `tp_dealloc` and the managed-dict paths call
(`PyObject_CallFinalizerFromDealloc`, `PyObject_VisitManagedDict`,
`PyObject_ClearManagedDict`, `PyObject_VectorcallDict`),
`PyObject_ClearWeakRefs`, `PyDict_SetDefault`, `PyInterpreterState_Get` /
`_GetID`, and the `Py_Version` data symbol.

## `PyCFunction_Type`, and why the check stays narrow

`typeobject.rs` carried a note that `PyCFunction_Type` could not be bound
because pyre has two `builtin_function_or_method` types. It now names
`methodobject`'s own one — the type a method an extension defines carries, and
so the base a type derived from it is derived from.

`PyCFunction_Check` is a header macro over `PyObject_TypeCheck` rather than an
export. The broad PyPy-style check answers yes for the interpreter's `len`,
which carries no `PyMethodDef`, so `PyCFunction_GetFunction` can only set
`SystemError` for it — and Cython's `__Pyx_IsSameCFunction` spells
`PyCFunction_Check(f) && PyCFunction_GET_FUNCTION(f) == c` as one expression
and never looks for an error. Measured: the broad form leaves `SystemError:
cpyext function returned a result with an exception set` on the first call.

## Declarations an extension reads through rather than calls

`PyThreadState` gains its `interp` member — `PyThreadState_Get()->interp` is a
field read — and `PyCFunctionObject` / `PyCMethodObject` are declared so a type
derived from `PyCFunction_Type` can embed one. Nothing pyre makes is read
through those fields; they are declared because a C-defined type's instance
block is `tp_basicsize` bytes, so an extension that embeds one writes its own
storage.

## Headers

New: `code.h` (the eleven `CO_` flags, each checked against the value pyre's
own `inspect` reports), `funcobject.h`, `traceback.h`, `compile.h`,
`frameobject.h`. `pyport.h` gains the `PY_LONG_LONG` / `PY_*INT*_T` spellings,
`lock.h` the critical-section macros (no-ops behind pyre's GIL), `objimpl.h`
`_PyGC_FINALIZED`. `patchlevel.h` moves after `pyport.h` in `Python.h` because
`Py_Version` is declared with `PyAPI_DATA`.

# 2. `jit`: a declared descriptor group takes its fields' cache slots before a serialized `BhDescr` rebuilds them

`GcCache` is keyed by `(STRUCT, fieldname)` and returns the first descriptor
minted for a pair unchanged (`descr.py:220-221`). Two producers reach that
namespace with different information:

- a `*_DESCR_GROUP` in `pyre-jit-trace/src/descr.rs` spells out `descr.py:229
  STRUCT._immutable_field(fieldname)` per field;
- a serialized `BhDescr` does not — the codewriter's
  `immutable_fields_by_struct` is empty for the whole LLBC pipeline, since
  Charon does not carry the declaration, so every field it describes reads as
  plain mutable.

`make_descr_from_bh` rebuilds a `BhDescr`'s whole parent group through
`simple_descr_group_from_bh_size` **before** looking anything up, so one
rehydrated field takes the declaration off every field of that STRUCT.
`get_field_descr` keeps the cached flags when a caller declares nothing, so the
group forced afterwards disagrees with its own slots and trips the identity
assertion:

```
get_field_descr cache hit for code disagrees with the caller:
cached "Function.code" (offset 16, size 8, type Ref, immutable false, quasi false, ...)
vs requested Some("Function.code") (... immutable true, quasi true ...)
```

Reached by importing a C extension and then running a Python `escape` loop hot
(markupsafe 3.0.3 with `_escape_inner` swapped back to `_native`): the
extension changes which jitcode is resolved first, so a body reading a
`Function` field arrives before `FUNCTION_DESCR_GROUP` is forced. Without an
extension loaded the group wins and nothing is visible, which is why this
surfaces only now. In a build without debug assertions the group silently keeps
the stripped flags instead — a lost fold, not a miscompile, since
`is_always_pure` gates on both flags.

`force_declared_group` forces this module's group for the STRUCT about to be
rebuilt, so the declaration takes the slots and both orders agree. **Nothing
already minted is mutated**, so a descriptor a compiled trace holds cannot
change under it; that was chosen over upgrading the cached flags in place,
which would change folding for already-compiled traces. `DECLARED_GROUPS`
lists the twenty-eight groups registered under a def path; the three that are
not (`PYCODE`, `EC`, `PYTRACEBACK`) mint through the unkeyed factory or under
`path_hash("")`, so they name no STRUCT a `BhDescr` resolves into.

This is main-owned: it reproduces on the base without any of the cpyext
changes, which mint no field descriptors.

# Gates

- `cpyext-abi.py check`: 489 exports / 24 header inlines / 123 data objects,
  every entry point matches.
- `cargo test -p pyrex --features cpyext,dynasm --test 'cpyext_*'` — 25 tests
  across 18 files.
- `cpyext_callables` is a new fixture; the same C source and the same script
  were built against CPython 3.14's headers and run on CPython, and give the
  same output there.
- markupsafe 3.0.3's own test modules: panic → 39 passed / 1 failed. The
  remaining failure is `test_leak.test_markup_leaks`, which wants fewer than
  three distinct `gc.get_objects()` counts and gets three — four before this.
- `cargo fmt --all -- --check`.
- `cargo test --all --features dynasm` is re-running locally after the last
  rebuild; every result line it has printed so far is `ok`, and it had not
  reached a failure when it was last interrupted. CI is the verdict here.

## Summary by CodeRabbit

* **New Features**
  * Expanded CPython C API compatibility for code objects, compilation modes, frames, methods, functions, imports, interpreter state, weak references, and object finalization.
  * Added support for vector calls with keyword mappings, unraisable-exception reporting, dictionary defaults, and module dictionary access.
  * Added runtime version and PyPy compatibility metadata.
  * Improved descriptor handling and added comprehensive C-extension compatibility coverage.

### DIFF
--- a/include/pyre3.14t/Python.h
+++ b/include/pyre3.14t/Python.h
@@ -5,8 +5,11 @@
  * from the exports themselves; the rest is the hand-written half --
  * the structs an extension lays out, and the macros it expands.
  */
-#ifndef PYRE_PYTHON_H
-#define PYRE_PYTHON_H
+/* The guard is the name CPython and PyPy both give it: an extension that
+ * wants to know whether it has Python's headers at all tests `Py_PYTHON_H`,
+ * and Cython's generated C refuses to compile without it. */
+#ifndef Py_PYTHON_H
+#define Py_PYTHON_H
 
 #include <assert.h>
 #include <inttypes.h>
@@ -19,8 +22,10 @@
 #include <string.h>
 #include <wchar.h>
 
-#include "patchlevel.h"
+/* `pyport.h` first: `patchlevel.h` declares one object, and `PyAPI_DATA` is
+ * what declares it. */
 #include "pyport.h"
+#include "patchlevel.h"
 #include "pymacro.h"
 #include "pytypedefs.h"
 #include "object.h"
@@ -59,9 +64,12 @@
 #include "sliceobject.h"
 #include "memoryobject.h"
 #include "pycapsule.h"
+#include "code.h"
+#include "funcobject.h"
+#include "traceback.h"
 #include "import.h"
 #include "modsupport.h"
 #include "audit.h"
 #include "abstract.h"
 
-#endif /* !PYRE_PYTHON_H */
+#endif /* !Py_PYTHON_H */
--- a/include/pyre3.14t/code.h
+++ b/include/pyre3.14t/code.h
@@ -1,0 +1,33 @@
+/* The flag bits a code object's `co_flags` is made of.
+ *
+ * One of the headers `Python.h` includes; an extension includes only
+ * `Python.h`. The exported entry points are declared together in
+ * `pyre_decl.h`, which is generated.
+ */
+#ifndef PYRE_CODE_H
+#define PYRE_CODE_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/* Each bit is the one `inspect` publishes under the same name, so a value an
+   extension builds here and a value it reads from Python mean the same thing.
+   A code object itself stays opaque -- `PyCodeObject` is named in
+   `pytypedefs.h` and has no fields an extension may read. */
+#define CO_OPTIMIZED            0x0001
+#define CO_NEWLOCALS            0x0002
+#define CO_VARARGS              0x0004
+#define CO_VARKEYWORDS          0x0008
+#define CO_NESTED               0x0010
+#define CO_GENERATOR            0x0020
+#define CO_COROUTINE            0x0080
+#define CO_ITERABLE_COROUTINE   0x0100
+#define CO_ASYNC_GENERATOR      0x0200
+#define CO_HAS_DOCSTRING        0x4000000
+#define CO_METHOD               0x8000000
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* !PYRE_CODE_H */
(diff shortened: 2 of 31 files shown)


youknowone:
@coderabbitai review

The earlier run stopped at the review limit, so this PR has not been reviewed yet.

— *commented by Claude*

PR #1370, opened 2026-08-20
jit, front: a retrace's closing JUMP keeps its own args; the swap element proves it is an object pointer

youknowone:
Follows up the P1 Codex raised on #1365 after it merged (`dc8ee02f15a`), so the defect is currently in `main`.

### The defect

#1365 gated `PYOBJECT_GCARRAY_TYPE_ID` on the element's `ValueType` coming out `Ref`. `Ref(None)` is also `tyref_to_value_type`'s tail — the fallback for every shape it cannot resolve — so on the `slice::swap` arm that condition read "I could not tell" as "this is the object block".

Charon runs `monomorphize:false` (stated at `mir.rs`'s `slice_index_call` note and `front::slice_index`). A `core::slice::<Impl>::swap` call inside a generic body therefore carries a `TypeVar` in `generics.types[0]`, and `listsort::sort_with<T: Copy, L: SortLt<T>>` reaches `swap` through `reverse_slice` — so the `i64`, `f64` and `usize` sort paths all arrive unresolved and got the object-array identity. That aliases them onto the real object array and gives them a pointer itemsize: 4 bytes on wasm32, where those banks stride 8.

This is strictly worse than the pre-#1365 state, where those sites carried `array_type_id: None` and so could not alias at all.

### The fix

The swap arm gates on `output_type_is_objectptr` — a positive proof that the type is a raw pointer onto `PyObject`, the same probe that recognises `w_list_new`'s `PyObjectRef` return. `FixedObjectArray::swap` (`&mut [PyObjectRef]`) resolves and keeps the identity; the `array` module's `Vec<u8>` buffer, the scalar banks, and every unresolved element carry `None`, whose `arraydescrof_concrete` arm mints locally without a cache publish.

`slice_swap_elem_value_type` becomes `slice_swap_elem_tyref` so the item kind can keep using the fallback — what the decomposition assumed unconditionally before #1365 — while the identity requires the proof.

### What this deliberately does not change, and why

The workspace `Index` arm keeps its item-kind condition. I tried extending the same proof to it and measured the cost:

| fixture | counter | backends |
| --- | --- | --- |
| `pickle_terminal_raise_resume` | `guard_failures` 298 → 338 | dynasm, cranelift, wasm |
| `list_append_write_barrier_gc` | `loops_compiled` 13 → 12 | cranelift, wasm |
| `global_quasiimmut_invalidation` | — | all three |

So some site behind that arm resolves under `tyref_deref_value_type` but not under `output_type_is_objectptr`, and loses an identity it was using. `strip_ty_wrappers` peels `Ref` but not `RawPtr`, so a plain `&PyObjectRef` `Output` satisfies both probes — the divergent site is something else, most likely a `pyre_`-pathed `index`/`index_mut` whose element does not resolve. I have not identified it, so I am not touching that arm on a guess. Named as follow-up work rather than folded into this fix.

### Gates

`pyre/check.py` — dynasm 442/442, cranelift 442/442, wasm 435/435, 3/3 backend runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of array and slice swaps to preserve object-array behavior only when the element type is confirmed.
  * Prevented unresolved types from being incorrectly treated as object arrays.
  * Improved scalar sorting behavior for arrays with ambiguous element types.

---

## Second commit: `jit: leave a closing JUMP alone when it targets another TargetToken`

Unrelated to the swap fix above; it lands here because it was found while
suppressing the hand-written trace-time specialization folds and it is the
last thing standing between that corpus and a clean all-folds-off run.

### The defect

`normalize_closing_jump_args` (`majit-metainterp/src/compile.rs`) repairs a
JUMP argument that names no definition in the trace by taking the trailing
LABEL's argument at the **same index**. That correspondence exists only while
the JUMP closes onto that LABEL.

`compile_retrace` also reaches the pass with a JUMP retargeted at another
TargetToken (`unroll.py:156/171 jump_to_preamble`). Its argument order comes
from **that** token's virtual state, so slot `idx` names a different live
value in each list. `MAJIT_LOG=1` prints the same trace on both sides of the
pass:

```
--- peeled trace (assembled) ---
  Label(v174, v183, v175, v176, v177, v178, v179, v192, v200, v187) descr=<LoopTargetDescr(2)>
  Jump (v174, v183, v176, v178, v179, v294, v297, v290, v187)       descr=<LoopTargetDescr(1)>
--- retrace combined (after opt) ---
  Jump (v174, v183, v175, v176, v177, v294, v297, v290, v187)       descr=<LoopTargetDescr(1)>
                       ^^^^^^^^^^^^^^^^^^  = label_args[2..5]
```

LABEL descr 2 carries state order `[0,1,3,5,6,7,8,9,10]`; the target, descr 1,
carries `[0,1,5,7,8,...]`. Slots 2, 3, 4 were rewritten from positions 5, 7, 8
to positions 3, 5, 6 — and positions 3 and 6 are the frame's globals and
builtins dicts, so a Python local came back holding a `dict`:

```
TypeError: '>' not supported between instances of 'int' and 'dict'
UnboundLocalError: cannot access local variable 'n' where it is not associated with a value
```

The pass cites `optimizer.py:651-652` for this, but that upstream loop is
`op.setarg(i, self.force_box(op.getarg(i)))` — it resolves an argument through
its forwarding chain and never substitutes a LABEL argument. The producing
site, `assemble_peeled_trace_with_jump_args`, already documents the rule this
pass was breaking: *"A JUMP redirected to another token already carries the
arity produced by that token's virtual state and must pass through
unchanged."*

### The fix

Compare `jump.getdescr().index()` with the trailing LABEL's and return the
trace unchanged when they differ. Same-token repair is untouched — the added
test pins both halves.

### Why it only shows with two folds off

ddmin over the 56-label suppression mask gives a minimal set of exactly
`{binary_op_int, binary_op_float}`; neither alone reproduces. With the folds
on, the accumulator and induction variable are unboxed `Int` LABEL arguments,
where a mismap is caught by the type check. With both off they are boxed
`Ref`s, and a `Ref`↔`Ref` shuffle is invisible to every type check in the
pipeline. `mc_diag` is byte-identical between the failing and passing runs
(`retrace_entered=1`, `close_hdr_fallback=1`, `all_descrs=9668`), so the JIT's
decisions matched and only the trace content differed.

The bug is not fold-specific — any retrace whose closing JUMP is retargeted
while its live values are Refs can hit it.

### Localisation, before any rebuild

| lever | result |
| --- | --- |
| `PYRE_JIT=0` | passes ⇒ not the interpreter |
| dynasm **and** cranelift | both fail ⇒ not a backend |
| `PYRE_NO_UNROLL=1` | passes ⇒ the unroll pass |
| `retrace_limit=0` / `=1` | passes / already fails ⇒ one retrace is enough |

### Gates

- `pyre/check.py`: `dynasm 442/442`, `cranelift 442/442`, `wasm 435/435` — no
  counter moved, so no baseline re-record.
- `cargo test -p majit-metainterp --lib --features dynasm`: new test passes.
- `cargo test --all --features dynasm`: one failure,
  `gc_stress::class_body_and_exec_loops_keep_the_vable_shadow_synchronized`,
  panicking at `compiler.rs:1701` *"no JITFRAME type id published"*. Not from
  this change — 3/3 in isolation and 2/2 (34/34) as its own suite; it is the
  cross-suite race under load already seen on this branch.
- All-folds-off corpus sweep over `pyre/bench/synth/*.py`:
  `TOTAL=432 CONTROL_NONZERO=0 ALLFOLDSOFF_NONZERO=0`. Before this commit the
  same sweep read `431/432` with `retrace_accumulator_type_flip` as the single
  holdout.


youknowone:
Added a second commit, `50c80ae9e30`, and rewrote the PR title/body to cover both. It is unrelated to the swap fix — it is the last blocker to a clean all-folds-off run of the synthetic corpus, which is what I was measuring when it surfaced.

Short version: `normalize_closing_jump_args` repairs a JUMP argument from the trailing LABEL's argument **at the same index**, which is only meaningful while the JUMP closes onto that LABEL. `compile_retrace` also feeds it a JUMP retargeted at another TargetToken, whose argument order comes from that token's virtual state — so the repair moved the frame's builtins dict into a Python local (`'>' not supported between instances of 'int' and 'dict'`). Fixed by comparing the two descrs; the producing site `assemble_peeled_trace_with_jump_args` already documented the invariant this pass was undoing.

`check.py` is `442/442 / 442/442 / 435/435` with no counter moved, and the corpus sweep goes from `431/432` to `432/432`. Details, including the before/after `MAJIT_LOG` dump that pins the substitution, are in the PR body.

— *commented by Claude*

youknowone:
Two more commits, `bb04994948f` and `efef29551aa`, carrying the hand-written fold retirement to the point the gate can defend.

**`jit-trace: retire the eight specialization folds that never fire`** (−1589 lines)

A `PYRE_FBW_SPEC_CENSUS=1` census over all 432 `pyre/bench/synth/*.py` (24192 rows = 432 × 56) records **zero fires** for `zip_two_tuple_iters`, `builtin_zip`, `compare_op_long_int`, `truediv_op_long`, `binary_op_long_int_pow`, `binary_op_long_int_shift`, `math_frexp` and `math_ldexp`. Suppressing exactly those eight leaves `check.py` unmoved on all three backends.

`fired=0` is not "dead code" here and I did not treat it as such — none of the eight is statically dead (`math_frexp`/`math_ldexp` are consulted 2272 times each, `builtin_zip` 17). It is **gate coverage**, so the retirement rests on a purpose-built covering fixture instead: it makes **seven of the eight actually fire**, and on it the retired build prints exactly what CPython prints. `builtin_zip` declines even on the exact `zip(tuple0, tuple1, strict=True)` shape its own comment names, which is consistent with 17 consults and 0 fires across the corpus.

Removes the eight `spec_gate` sites, the eight `try_walker_specialize_*` bodies, and their `SPEC_FOLD_ROWS` rows (56 → 48), and rewords the four doc comments that named a removed function so no citation dangles.

**`bench: cover the paths whose specializations were retired`**

That covering fixture, added to the corpus. Without it the now-generic paths would carry no gate at all — which is precisely the hole that let these eight sit unexercised in the first place. Float legs are folded to `int(round(x * 1e6))` so backend float formatting cannot leak into the comparison.

**Gates**

- `check.py`: `dynasm 444/444`, `cranelift 444/444`, `wasm 437/437`, no counter moved. Re-run a second time after reverting an unintended re-record (see below), so the macro baselines in the tree are the original ones.
- All-folds-off corpus sweep: `TOTAL=432 CONTROL_NONZERO=0 ALLFOLDSOFF_NONZERO=0`.
- `cargo test --all --features dynasm`: the known `gc_stress::class_body_and_exec_loops_keep_the_vable_shadow_synchronized` cross-suite race (33 passed; 34/34 when the suite runs alone).

**One thing worth flagging for anyone recording a baseline:** `check.py --snapshot --backend B --synthetic-pattern <one-name>` reported `11/11` and re-recorded all ten macro benchmarks under `pyre/bench/` as well — 29 `.jitstats` files across three backends. The diffs were purely additive (a new `retraces_compiled=0` line the older baselines predate), but I reverted them all: that would start gating a counter this PR did not introduce, on benchmarks it does not touch. Only the new fixture's own three baselines are committed.

**Not done, deliberately.** Retiring the remaining 48 is not free and I did not attempt it. Measured: suppressing all 56 moves jitstats on **198 benches (~99 fixtures on both native backends), including `nbody` and `fannkuch`** — no wrong answers anywhere, but it would require re-recording those baselines, which discards regression detection on all of them. That is a policy call about the corpus, not something to slip into a retirement PR.

— *commented by Claude*

PR #1374, opened 2026-08-20
jit: port arraylen_vable, drop the vable array getfield, and close the three #1358 follow-ups

youknowone:
`rewrite_op_getfield`'s `except VirtualizableArrayField:` handler ends in
`return []` (`jtransform.py:848-857`): registering the base in
`vable_array_vars` **is** the whole rewrite. Porting that faithfully turned
out to need an upstream instruction the codewriter never emitted.

The first three commits are the follow-ups #1358's body listed.

## The three #1358 follow-ups

**A — `to_optimizer_config`'s `array_lengths: vec![]` is not a missing seed.**
Refuted. It is a placeholder its sole caller patches on the next line:
`MetaInterp::current_virtualizable_optimizer_config` fills it from
`TraceCtx::virtualizable_array_lengths`, and the sibling field
`vable_input_offset` right below it is the same shape with the same
convention already documented. A length is not a property of the shape —
upstream reads `len(lst)` off the live object every time it needs one
(`virtualizable.py` `read_boxes`, `get_array_length`) and stores it nowhere.

What was actually missing was the diagnosis if the patch is ever skipped:
`VirtualizableTracker::init`'s zip runs zero times, no element state is
seeded, and `tracked_array_element` silently cannot hit. A `debug_assert!`
now names that state. Its two escapes are load-bearing (the state-field macro
JIT sets `track_array_elements = false`; an arrayless config has nothing to
seed) and a second test pins them.

**B — `_unpackiterable_known_length_jitlook` must NOT carry `unroll_safe`.**
Settled from upstream source, not inference. Upstream fences that body behind
two functions: `unpackiterable` → `_unpackiterable_known_length`
(`@jit.dont_look_inside`) → `_unpackiterable_known_length_jitlook`
(`@jit.unroll_safe`). Only `unpackiterable_unroll`, whose `expected_length` is
an UNPACK_SEQUENCE oparg, calls the hinted body directly. pyre has neither the
shim nor `unpackiterable_unroll`, so `unpackiterable` is the body's only
caller — the path upstream keeps closed. Carrying the attribute alone inverts
that decision instead of matching it.

The doc comment quoting `@jit.unroll_safe` reads as an unfinished port and has
been picked up as one twice. `tests/test_unroll_safe_inventory.rs` makes
adding it fail with the reason. It is a *subset* check with a positive
control, not an equality check: a developer's `build/llbc` is routinely older
than the source, and a stale artefact must only be able to under-report.

**C — `is_iter_op_segments` admits non-GC slices.** A live defect, not a
latent one. `iter_next_item_type` answered `Ref(None)` for every non-range
iterator, so a `for &v in slice` over `&[i64]` typed its element as a GC
reference. `charon-corpus`'s `branch_loop_sum` is exactly that shape and folds
today. The element type now rides from the `Option<T>` payload at the `next`
call site through `next_call_results` to the fold. Declining the fold for
non-GC slices — my first design — would have regressed a shape a test already
required.

## `arraylen_vable`, and why the getfield could not be dropped alone

Dropping the getfield made `fast2locals_assembles` die in liveness:

```
variable_to_register: graph declared kind Ref for Variable(... id: 3797 ...)
but regallocs[Ref] has no coloring (other classes with a coloring: [])
```

That message means a variable is **used but never defined**. The user was
`ArrayLen`.

`vable_array_vars` has three consumers upstream, not two:

| upstream | pyre codewriter before this PR |
| --- | --- |
| `rewrite_op_getarrayitem` (`jtransform.py:760`) | `VableArrayRead` |
| `rewrite_op_setarrayitem` (`:794`) | `VableArrayWrite` |
| `rewrite_op_getarraysize` (`:811`) → `arraylen_vable` | **absent** |

So `locals_w!(frame).len()` stayed an `arraylen_gc` **on the raw array
pointer** — the one route still needing the op the drop removes.
`check_no_vable_array` does not catch it: its four routes are link arg, fused
exitswitch, call arg and setfield, and an `arraylen` in the same block is none
of them.

`arraylen_vable` was already ported eight times over: the
`arraylen_vable/rdd>i` key in `insns.rs` (byte 74 assigned),
`blackhole.rs`'s `wire_handler`, `jitcode/mod.rs`,
`jitcode_dispatch/mod.rs`, `opimpl_arraylen_vable` in `pyjitpl.rs` /
`trace_ctx.rs` / `jitdriver.rs`, `bhimpl_arraylen_vable` in
`virtualizable.rs`, `MIFrame::read_vable_arraylen`, and the macro lowering's
`lower_vable_array_len`. Only the codewriter path never reached it.

The port cost nine exhaustive matches the compiler lists, plus four tables
that fall through silently and had to be found by hand — `type_state.rs`,
`format.rs`, `flowspace_adapter.rs` (both name tables) and
`legacy_resolve.rs`. All four answer **Int**: a length is a `Signed` whatever
the element kind is.

`item_ty` on the len op has no honest per-access source, because `arraylen_gc`
carries no element type. Upstream passes `vinfo`'s own `arraydescr`
(`:816`). `vable_arraydescrof` already asserts the block behind a
virtualizable array is a `FixedObjectArray` of word-wide `PyObjectRef`s, so
that is the element type, and `vable_array_index_pair_at` — which only
requires the pair be `(VableArray, Array)` and takes the length off the first
— accepts it.

**Ordering is load-bearing.** The drop is unsound without the port, so the
port is the earlier commit.

## Gates

Run on base `712fc16c29b` with freshly extracted LLBC, after a concurrent
rebase invalidated an earlier round taken against `1f7229aed9b`:

- `cargo check -p pyre-jit-trace` — green. `build.rs` calls `generate_into`
  with no `catch_unwind`, so a `_check_no_vable_array` escape or a missing
  assembler key is a hard build failure over the real interpreter.
- `cargo test --all --no-default-features --features dynasm` — green.
- `cargo test --no-fail-fast --no-default-features --features cranelift,cpyext`
  over the workflow's package list verbatim — green.
- `cargo fmt --all -- --check` — clean.

## Note on main's macOS red

`cargo test (macos-latest)` is red on main at `1f7229aed9b`
(`test_deadframe_exception_ref_survives_collection_after_execute_token`,
`INVALID encountered` from cranelift). It passes here, in the same feature set
and package list, on darwin-arm64. That is not an acquittal of that commit —
this branch's base is three commits later. Worth recording: since that red,
seven main runs were cancelled and `712fc16c29b` has been queued without a
verdict, so nothing at or after the red commit has a completed CI answer.

https://claude.ai/code/session_017wapwfqfqNe7kFcxQRx85P

## Summary by CodeRabbit

- **New Features**
  - Added optimized handling for virtualizable-array length operations.
  - Improved iterator result typing for ranges, references, and collection elements.

- **Bug Fixes**
  - Fixed iterator values being assigned incorrect reference or integer types.
  - Improved validation and safety for virtualizable-array accesses and snapshots.
  - Preserved correct behavior for raw array access and variable-index writes.

- **Documentation**
  - Clarified virtualizable-array tracking and safe unpacking behavior.

- **Tests**
  - Added coverage for array lengths, iterator typing, identity tracking, invalid accesses, and configuration validation.


youknowone:
All five findings were valid and are fixed in `9dfadd2e3f2`. The two P2s were
regressions this PR introduced, and both are now pinned by tests.

**Codex — honor `lower_virtualizable` before dropping the array field read.**
Correct, and it is the same failure this PR's own `arraylen_vable` commit
exists to prevent — I hit it as an uncolored-Ref panic in `fast2locals` and
then reintroduced it through a different door. `rewrite_op_getfield` is the one
member of the family that runs ungated, and the dispatch comment says why: the
quasi-immutable `-live-` + `record_quasiimmut_field` tail does not depend on
virtualizable lowering. Its two virtualizable arms do, and the field's own doc
covers "field/array accesses" — both. Gated both on
`self.config.lower_virtualizable`, not just the array arm, so the scalar arm
stops disagreeing with `rewrite_op_setfield` at the same time.
`a_vable_array_read_is_kept_when_virtualizable_lowering_is_off` pins it;
reverting the gate makes it fail.

**Codex — preserve nested references when deriving iterator item types.**
Correct. `strip_ty_wrappers` peels `Ref` in a loop, so `Option<&&i64>` came
back as `i64`. Worse, it is a regression: before this PR the non-range arm
answered `Ref(None)` unconditionally, so `&[&i64]` happened to be right, and my
change put that pointer in the integer register bank — the exact inversion of
the defect the commit was fixing. New `iterator_payload_element` strips the
pure indirection wrappers (`Deduplicated`, `HashConsedValue`), peels exactly
the one `Ref` the slice iterator adds, and stops. A payload that is not a `Ref`
is already the element, which keeps by-value iterators (`[i64; N]`, `Vec<i64>`)
answering `Int`.

**CodeRabbit — validate the complete parallel-array invariant.** Taken. The
assertion checked non-emptiness, which only catches the loudest case; `zip`
pairs what it can and drops the rest, so a short vector seeds some arrays and
silently leaves the others untracked. It now requires
`array_lengths.len() == array_field_offsets.len()` and reports both counts plus
how many the zip would actually pair. The `array_field_offsets.is_empty()`
escape is subsumed by the equality. Added
`a_partial_array_lengths_vector_is_named_not_truncated` (two fields, one
length). I did not add the same check on the `virtualizable.rs` side: that
site is `to_optimizer_config`'s documented placeholder, which is *supposed* to
be short until its sole caller patches it, so asserting there would fire on the
intended state.

**CodeRabbit — regression test for the `VableArrayLen` wire shape.** Added
`assemble_vable_arraylen_emits_the_rdd_to_i_wire_shape`. It drives the whole
pipeline and asserts `arraylen_vable/rdd>i`, that no `arraylen_gc` survives,
that the `VableArray` descr precedes the `Array` descr, and that the array
descr carries the flat-word base / word itemsize / `len_offset = Some(0)` /
pointer-items shape `MIFrame::read_vable_arraylen` decodes against.

**CodeRabbit — detect leaf-name collisions.** Added, in the shape suggested.
The file's whole purpose is failing loudly on an unreviewed hint, and matching
by leaf on an unverified uniqueness assumption is the one way it could pass on
another function's review.

Gates after the fixes: `cargo test -p majit-translate` green (lib 3241 → 3243),
`cargo test -p majit-metainterp --features dynasm --lib optimizeopt::virtualize`
65 passed, `cargo fmt --all -- --check` clean.

— *commented by Claude*

youknowone:
Rebased onto `39d57244d31` and force-pushed. All four findings from the second
review round are fixed in `14dc6371760`.

### The Codex P2 was the mirror image of the previous round's fix

The last round's peel stopped `&[&i64]` from putting a pointer in the integer
bank. It did so unconditionally, and the by-value iterators
`is_concrete_iter_constructor` admits have no iterator-added reference to peel:
`alloc::vec::into_iter::IntoIter` and `core::array::iter::IntoIter` yield
`Option<T>`, so that one reference was the element's own — the same defect,
pointing the other way.

The `next()` receiver names the iterator ADT, so the decision reads that
instead of guessing from the payload's shape. Only `core::slice::iter::Iter`
and `IterMut` peel.

Two corpus fixtures carry both shapes, and each direction is red-checked:

| arm | `slice_of_refs_sum` (`&[&i64]`) | `array_of_refs_sum` (`[&i64; 3]`) | `branch_loop_sum` (`&[i64]`) |
|---|---|---|---|
| peel always | ok | **`Int` — pointer in int bank** | ok |
| never peel | ok | ok | **`Ref` — int in ref bank** |
| receiver-keyed | ok | ok | ok |

No fixed answer passes all three, which is why the receiver has to be read.
Regenerating `corpus.ullbc` showed no format drift; only the local-fn count
assertion moved (14 → 16).

### The other three

- **`body.code` decode** — valid; the pool order only said which descr was
  minted first. Red-checked by swapping the expected indexes (`left: 0,
  right: 1`).
- **Scalar-field gate test** — valid, added.
- **CONTROL guard ordering** — valid, the collision check now runs after it.

### One addition beyond the review

`check_no_vable_array` enumerated four operand positions. Since this PR drops
the `getfield` behind a virtualizable array, "every consumer lowers to the
vable form" became load-bearing with nothing checking it — and this PR has
already been bitten by that invariant twice, each time surfacing as an
uncoloured register in regalloc rather than as the missing read.

A fifth route, `"surviving operation operand"`, scans every operand of every
operation the block kept. This is safe to assert rather than warn because
nothing between `transform` and regalloc prunes dead operations, so a
surviving operand always reaches regalloc anyway — the catch-all cannot fire
on a graph that would otherwise have compiled. The function's own doc already
recorded why enumeration is the wrong shape here (`_check_stack_index` escaped
by the one route its test did not look at).

While investigating I also settled a gap I had previously reported as open:
`_handle_list_call` carries no `vable_array_vars` check and is owed none.
Upstream splits on `resizable` and puts the check only on the
`do_fixed_list_*` arms, whose receiver is a `GcArray`; every spelling pyre
ports is of the resizable family with a `W_ListObject` receiver. That
reasoning is now recorded on the function.

### Verification

- `cargo test --all --no-default-features --features dynasm` — **8027 passed,
  159 suites, 0 failed**, checked against the full log rather than a tail.
- `cargo test -p majit-translate` lib 3243 → **3245**; `optimizeopt::virtualize`
  **65**; `majit-charon-reader` green.
- LLBC re-extracted at the new base and reported current.
- `cargo fmt --all --check` clean.

— *commented by Claude*

PR #1385, opened 2026-08-20
gc: close the rooting holes a joined-artefact collection census finds, and settle gc.collect()'s answer

youknowone:
Sixteen commits: fifteen close the GC rooting holes a whole-program collection census reports, and one settles what `gc.collect()` is supposed to answer.

## The census, and what it found

`gc-root-reachability` answered per charon artefact, and one artefact is not one program. `pyre-jit.ullbc` carries the portal and the blackhole but only 575 of `pyre-interpreter`'s bodies, so its closure was **77 / 7615** and the liveness scan over it reported nothing — because nothing in it reached a seed, not because the crate was clean.

`framework::Joined` merges `CallGraph`s across artefacts and `project()` maps a joined node set back onto one artefact's def ids, so `liveness::scan` gets the whole-program reachability. `GC_JOIN_WITH=build/llbc/pyre-interpreter.ullbc` lifts pyre-jit to **240 / 7615**.

| | alone | joined |
|---|---|---|
| pyre-jit closure | 77 / 7615 | **240 / 7615** |
| pyre-jit **frame** findings | 0 (vacuous) | **3 real → fixed → 0** |
| pyre-interpreter closure | 5316 / 41565 | 5316 — **the join admits 0** |

So the reverse direction is closed too: pyre-interpreter's census was never vacuous, and only pyre-jit needed the join.

### The trap in the join

`ItemMeta::name_path` renders an inherent `impl` block as the opaque segment `<Impl>`, so **`PyFrame::new` and `FrameDebugData::new` are both `pyframe::<Impl>::new`**. Joining on that spelling merged them and invented an edge from every caller of one to every callee of the other — `getorcreate_debug_data`, which only allocates, came out reaching `call_user_function_with_args`, and pyre-interpreter grew 7 frame findings that do not exist.

"Widening is safe for a may-collect analysis" is false when the widening comes from a fabricated identity. Over-bracketing is conservatism; a fabricated path is a wrong answer with a citation. The rule the last commit implements: **a name one artefact carries more than once is not a join key** — each occurrence keeps its own node. The join line now reports both counts (1341 spellings merged across artefacts, 2430 held apart).

### The repairs

Each one re-reads its operand after the call that can collect, per `rpython/memory/gctransform/framework.py:1501-1508 get_livevars_for_roots`, and each bracket is narrow — per collecting operation, not per function — so the tail of the body stays scanned.

- `bh_load_global_fn` / `bh_load_from_dict_or_globals_fn` read the frame **and** the globals again in the builtins leg after `finditem_str`
- `bh_call_fn_impl` re-derefs the frame each iteration across `descr_repr`, which fsdecodes the filename into `call_registered_decode_error_handler`
- `bh_normalize_raise_varargs_with_frame` carries `exc` across both class calls and the normalized cause across the second
- `frontend_global_object` re-reads the globals for `__builtins__`
- the interpreter side anchors the running frame across the opcode steps that run Python, hands the frame back from `bytecode_only_trace`, and roots the descriptor operands `get_and_call_function` re-reads after `__get__`

Two sites were adjudicated rather than fixed: `dict_eq_hook::signal_eq_error` and `signal_hash_error` store the pointer as a presence token — `take_*_error` only tests it for null — so a stale one is never followed. That is recorded at the call.

## `gc.collect()` — measured, not inferred

A forwarded report said `gc.collect()` returning None violates the 3.14 contract and that `extra_tests/snippets/stdlib_gc.py` proves it. Both halves needed checking.

**What pypy actually answers.** `interp_gc.py` reads as though `collect` has no return value: the `return space.newint(0)` that closed it from `01b045acd645` (2016) is still in the file, but `b5632df5b6e0` (2018) split `_run_finalizers` out of `collect` and the statement went with the extracted half, where neither caller reads it. Rather than infer intent from that, I ran the interpreter:

```
$ pypy3 -c 'import gc; print(repr(gc.collect()), repr(gc.collect(-1)))'
None None                       # PyPy 7.3.22 / 3.11.15
```

A full differential probe of the module against pyre agrees on every other `collect` and `get_objects` case, including `__index__` conversion and every TypeError. The two remaining differences are error-message wording, which is reference-styled across the whole interpreter by design.

**Why the snippet is not evidence.** `.github/workflows/pyre-ci.yml` runs `extra_tests/run.py --gated-only`; `stdlib_gc.py` carries no `# pyre-check: gate=1`, so **CI has never run it**, and `git log` shows it landed in #822 — the same commit as the implementation it contradicts. It has failed at line 6 since the day it was added.

**What this PR changes.** Nothing in `module/gc/mod.rs`: pyre already answers exactly what pypy answers. The snippet asserted a contract the tree deliberately does not implement, in a file CI cannot enforce, while the pypy behaviour was already pinned — correctly, and graded against the pypy oracle — in `bench/synth/gc_pypy_frontend.py`. It now asserts the argument binding, the `__index__` conversion, the TypeError cases and `get_objects()`, all of which cpython 3.14.6, pypy 7.3.22 and both backends answer alike, and it carries `gate=1` so CI runs it.

> **One line is still open for the reviewer.** If the intent is that `gc.collect()` should answer an int, the change is `Ok(w_none())` → `Ok(w_int_new(0))` in `module/gc/mod.rs`. It diverges from the running pypy, and it requires deleting `assert gc.collect(generation) is None` from `gc_pypy_frontend.py`, whose only oracle is pypy. I did not take it, because every authority I could check — the vendored source, the shipped interpreter, and the gated fixture — says None.

## Gate

Rebased onto `origin/main` (`01b740aedaf`, #1370); one conflict, in `pytraceback.rs:159`, resolved to origin/main, which applies the read-back rule our side only documented.

| | |
|---|---|
| llbc, 4 crates | `--check` RC=0, fingerprints match the tree |
| `cargo test --all --no-default-features --features dynasm` | RC=0 |
| `check.py --backend dynasm,cranelift` | **446/446 · 446/446 ALL PASSED** |
| `extra_tests/run.py --gated-only` | **68/68** on cpython, dynasm and cranelift |
| jitstats drift | none |
| raise-from + frame stress, 2 backends × 2 nursery sizes | 40/40 |
| census, re-run after the repairs | 240/7615, frame findings **0** |
| `get_objects()` marker assertion, 2 backends × 3 nursery sizes × 10 | 60/60 |

— *authored by Claude*

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -3689,15 +3689,51 @@ impl MiniMarkGC {
     #[inline]
     fn validate_type_id(&self, type_id: u32, obj_addr: usize, site: &str) {
         if type_id as usize >= self.types.len() {
+            self.report_invalid_type_id(type_id, obj_addr, site);
+        }
+    }
+
+    /// Out-of-line half of [`validate_type_id`](Self::validate_type_id).
+    ///
+    /// A type id past the end of the type table has two very different
+    /// causes, and the bare number does not separate them.  When the header
+    /// word is `FORWARDED_MARKER` the object is not mistyped at all: it is a
+    /// nursery corpse whose live copy sits at the forwarding address, and the
+    /// caller reached it through a field nothing updated.  Say so, and give
+    /// the copy plus the nursery geometry and collection cadence, so the
+    /// surviving question is only which holder kept the stale pointer.
+    #[cold]
+    #[inline(never)]
+    fn report_invalid_type_id(&self, type_id: u32, obj_addr: usize, site: &str) -> ! {
+        let header_addr = obj_addr - GcHeader::SIZE;
+        let nursery_start = self.nursery.start_ptr() as usize;
+        let nursery_end = nursery_start + self.nursery.size();
+        let in_nursery = (nursery_start..nursery_end).contains(&obj_addr);
+        let header = unsafe { *(header_addr as *const GcHeader) };
+        if header.is_forwarded() {
+            let moved_to = unsafe { GcHeader::forwarding_address(header_addr as *const GcHeader) };
+            let moved_to_type_id = self
+                .is_managed_heap_object(moved_to)
+                .then(|| unsafe { (*header_of(moved_to)).type_id() });
             panic!(
-                "GC BUG: invalid type_id={} at obj_addr={:#x} (header_addr={:#x}, nursery_start={:#x}, site={})",
-                type_id,
-                obj_addr,
-                obj_addr - GcHeader::SIZE,
-                self.nursery.start_ptr() as usize,
-                site,
+                "GC BUG: {site} was handed {obj_addr:#x}, a STALE address whose object a                  minor collection already moved to {moved_to:#x}                  (live copy type_id={moved_to_type_id:?}).                  nursery=[{nursery_start:#x}, {nursery_end:#x}) obj_in_nursery={in_nursery}                  offset_from_nursery_start={:#x} minor_collections={} major_collections={}.                  FORWARDED_MARKER sets every flag bit, so an inline TRACK_YOUNG_PTRS test                  cannot tell this corpse from a live old object — the defect is the holder                  that kept the pre-collection pointer, not this barrier.",
+                obj_addr.wrapping_sub(nursery_start),
+                self.minor_collections,
+                self.major_collections,
             );
         }
+        panic!(
+            "GC BUG: invalid type_id={} at obj_addr={:#x} (header_addr={:#x}, nursery_start={:#x}, site={})              header_word={:#x} obj_in_nursery={} minor_collections={} major_collections={}",
+            type_id,
+            obj_addr,
+            header_addr,
+            nursery_start,
+            site,
+            header.tid_and_flags,
+            in_nursery,
+            self.minor_collections,
+            self.major_collections,
+        );
     }
 
     /// `base.py:134-144 _get_size_for_typeid` — the payload size of `obj_addr`,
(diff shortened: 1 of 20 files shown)


youknowone:
Four P1s on this PR. Two were right and are fixed in **#1389**; two do not reproduce. Evidence for each below.

---

### ✅ `framework.rs:386` — "Reject cross-artifact collisions in opaque impl names"

Correct, and it is the same defect this join was written to remove, arriving by a route the rule could not see. `Joined::build` marked a name ambiguous when **one artefact** carried it more than once; two artefacts each carrying `pyframe::<Impl>::new` exactly once, for two different functions, passed that test.

Fixed: a name carrying an opaque `<…>` segment is held apart as well — the rendering has already discarded what distinguished it, so it is not an identity in either direction.

| pyre-jit ⋈ pyre-interpreter | before | after |
|---|---|---|
| spellings merged across artefacts | 1341 | 551 |
| held apart | 2430 | 20875 |
| closure | 240 / 7615 | 192 / 7615 (77 unjoined) |
| findings | 3 PyObjectRef, 0 frame | **the same 3 rows, still 0** |

48 nodes of closure for soundness, no finding lost.

---

### ✅ `stdlib_gc.py:17` — "Restore the CPython gc.collect return contract"

Correct, and the `AGENTS.md` citation is what settles it. I had taken the running interpreter as the authority (`pypy3 -c 'import gc; print(repr(gc.collect()))'` → `None`), which is the wrong axis: the return value is in the caller-observable set the 3.14t spec governs, and only the value is engineered the way upstream would.

`collect` now answers `w_int_new(0)`, the snippet asserts `isinstance(..., int)` again, and `gc_pypy_frontend.py` stops asserting the return value since its only oracle is pypy.

On the constant: `interp_gc.py:48` still carries the `return space.newint(0)` that closed `collect` from `01b045acd645` until `b5632df5b6e0` split `_run_finalizers` out of it and the statement went with the extracted half, where neither caller reads it. That is why the shipped interpreter answers `None`, and why `0` is the honest value — incminimark never computes a count of unreachable objects. So `test_gc.py:99` stays out of reach: the type is on-spec, the count is not, and closing that would mean making the collector count.

---

### ❌ `thread/mod.rs:493` — "Keep the pinned frames outside the temporary anchor scope"
### ❌ `pyframe.rs:1178` — "Preserve coroutine summaries beyond each frame anchor"

Both rest on the anchor and the pinned roots sharing one stack. They do not.

- `FrameAnchor::from_raw` → `majit_gc::shadow_stack::push`, and `Drop` → `majit_gc::shadow_stack::try_pop_to`, which truncates `SHADOW_STACK` (`majit-gc/src/shadow_stack.rs:215`, a `RefCell<ShadowStack>` over `entries: Vec<GcRef>`).
- `pin_root` / `shadow_stack_len` / `shadow_stack_get` address `ROOT_STACK` (`pyre-object/src/gc_roots.rs:218`, a `RootStack` with its own base/top cells); `shadow_stack_get` is `with_shadow_stack(|stack| *stack.slot(index))`.

Two thread-locals. Dropping the anchor cannot truncate a pinned root, and `base + index` indexes a stack the anchor never touched. The co-naming is genuinely misleading and is a fair thing to have been caught by.

Stressed rather than argued: 200 rounds of `sys._current_frames()` with five live threads and a full collection between rounds, and 150 rounds of `set_coroutine_origin_tracking_depth(5)` + coroutine creation with a collection between the frames — on dynasm and cranelift, at the default nursery and `PYPY_GC_NURSERY=65536`. All pass, and both outputs match cpython 3.14.6 and pypy 7.3.22 exactly.

— *commented by Claude*

PR #1389, opened 2026-08-20
gc: answer collect() with an int, and stop the artefact join merging two functions

youknowone:
Follow-up to #1385: the two review findings that were real, plus the fixture the merged raise-from fix landed without.

## `gc.collect()` answers an int

I had left this open in #1385, having taken the running interpreter as the authority:

```
$ pypy3 -c 'import gc; print(repr(gc.collect()))'
None            # PyPy 7.3.22 / 3.11.15
```

That was the wrong axis. `AGENTS.md` — *"Pursue PyPy parity to the extreme as engineering, and CPython 3.14t as spec; where they collide, take the 3.14t behaviour and engineer it the way PyPy would"* — puts **the return value** in the caller-observable set the spec governs. So the type is CPython's and only the value is upstream's.

`collect` now answers `w_int_new(0)`. The constant is not invented: `interp_gc.py:48` still carries the `return space.newint(0)` that closed `collect` from `01b045acd645` (2016) until `b5632df5b6e0` (2018) split `_run_finalizers` out of it and the statement went with the extracted half, where neither caller reads it. That is why the shipped interpreter answers `None`, and it is also why `0` is the right value here: incminimark never computes a count of unreachable objects.

So the type is on-spec and the count is not — `test_gc.py:99`'s `assertEqual(gc.collect(), 1)` stays out of reach, and closing it would mean making the collector count.

- `extra_tests/snippets/stdlib_gc.py` asserts `isinstance(gc.collect(...), int)` again, and stays gated.
- `bench/synth/gc_pypy_frontend.py` no longer asserts the return value: its only oracle is pypy.

## The join could still merge two functions

`Joined::build` marked a name ambiguous when **one artefact** carried it more than once. That test can only see a collision an artefact already contains — two artefacts may each carry `pyframe::<Impl>::new` exactly once, for two different functions, and the join merged them with nothing to warn on. Same fabricated edges, different route in.

A name carrying an opaque `<…>` segment is now held apart as well: the rendering has already discarded what distinguished it, so it cannot be an identity in either direction.

| pyre-jit ⋈ pyre-interpreter | before | after |
|---|---|---|
| spellings merged across artefacts | 1341 | 551 |
| held apart | 2430 | 20875 |
| closure | 240 / 7615 | **192 / 7615** (77 unjoined) |
| PyObjectRef findings | 3 | **3 — the same rows** |
| frame findings | 0 | **0** |

48 nodes of closure for soundness, and no finding lost.

## The raise-from fixture

`c6438765d7d` moved the cause check from `normalize_raise_cause` to `attach_raise_cause` — `error.py:376-385 set_cause`'s position — but landed without a test. `extra_tests/parity_tests/raise_from_invalid_cause.py` covers the three shapes that differed from **both** references:

| | cpython 3.14.6 | pypy 7.3.22 | pyre, before |
|---|---|---|---|
| `raise Cls from 42` | `Cls()`, then TypeError | `Cls()`, then TypeError | TypeError, **`Cls()` skipped** |
| `raise 5 from 42` | the value's TypeError | the value's TypeError | **the cause's** TypeError |
| `raise Cls from Ctor` (Ctor raises) | that exception | that exception | **swallowed**, `Cls` raised |

The raise runs hot so the JIT residual and the tracer answer alongside the interpreter. Whether `Cls()` has run when a cause constructor raises is deliberately not asserted — `pyopcode.py:757-760` instantiates the cause before popping the value and the reference does it the other way round, so the two disagree on that and only that.

Side effect: CPython's own `test_raise` goes from `failures=2, errors=2` to `failures=2, errors=1` (`TestCause.test_erroneous_cause` now passes). `test.test_raise` is baselined `FAIL`, so it never gated; the two remaining failures are `tb_lasti` and the `__new__`-returns-a-non-exception message, neither related.

## Not reproduced

Two P1s on #1385 said a per-iteration `FrameAnchor`'s `Drop` truncates roots pinned after it, in `thread/mod.rs:493` and `pyframe.rs:1178`. They are two different thread-locals: `FrameAnchor` pushes and `try_pop_to`s `majit_gc::shadow_stack::SHADOW_STACK`, while `pin_root` / `shadow_stack_len` / `shadow_stack_get` address `pyre_object::gc_roots::ROOT_STACK`. Dropping the anchor cannot reach a pinned root. Stressed anyway — 200 rounds of `sys._current_frames()` with five live threads, 150 rounds of coroutine-origin depth 5, a full collection between rounds, two backends × two nursery sizes — all matching cpython and pypy exactly.

## Gate

| | |
|---|---|
| llbc, 4 crates | `--check` RC=0 |
| `cargo test --all --no-default-features --features dynasm` | RC=0 |
| `check.py --backend dynasm,cranelift` | **447/447 · 447/447 ALL PASSED** |
| `extra_tests/run.py --gated-only` | **69/69** on cpython, dynasm, cranelift |
| `extra_tests/parity_tests/run.py` | all pass, 3 runners |
| sandbox clippy fence | `pyframe.rs` no longer trips `std::io::_eprint` |

— *authored by Claude*

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed name handling so distinct rendered names with opaque path segments are no longer incorrectly merged.
  - Updated `gc.collect()` to return an integer result consistently across supported invocation forms.
  - Corrected exception chaining behavior for invalid causes, including evaluation order and error reporting.
  - Improved traceback source location and instruction-offset reporting.
  - Prevented errors when deleting traceback links.

- **Tests**
  - Expanded coverage for garbage collection, exception chaining, and traceback behavior across execution modes.

### DIFF
--- a/majit/majit-translate/src/memory/gctransform/framework.rs
+++ b/majit/majit-translate/src/memory/gctransform/framework.rs
@@ -333,11 +333,22 @@ impl CallGraph {
 /// `framework.py` over-brackets rather than under-brackets, but a fabricated
 /// path is not conservatism, it is a wrong answer with a citation.
 ///
-/// So a name **one artefact carries more than once** is not a join key at all:
-/// each occurrence keeps its own node, exactly as it had before the join.  What
-/// is joined is the unambiguous majority — the free functions the JIT reaches
-/// the interpreter through (`finditem_str`, `call_function_impl_result`) — and
-/// [`Self::ambiguous_names`] reports what that rule held back.
+/// A name **one artefact carries more than once** is therefore not a join key:
+/// each occurrence keeps its own node, exactly as it had before the join.  That
+/// test alone is not enough, because it can only see a collision an artefact
+/// already contains: two artefacts may each carry `pyframe::<Impl>::new`
+/// exactly once, for two different functions, and the join would merge them
+/// with nothing to warn on.  So **any name carrying an opaque `<…>` segment**
+/// is held apart as well — the rendering has already discarded what
+/// distinguished it, and a spelling that cannot fail to collide is not an
+/// identity.
+///
+/// What is joined is what stays fully spelled — the free functions the JIT
+/// reaches the interpreter through (`finditem_str`,
+/// `call_function_impl_result`) — and [`Self::ambiguous_names`] reports what
+/// the two rules held back.  Measured on `pyre-jit` joined with
+/// `pyre-interpreter`, holding the opaque names apart costs 48 nodes of closure
+/// (240 → 192, against 77 unjoined) and no finding at all.
 pub struct Joined {
     /// The joined graph.  Its `opaque` census is left empty: an opaque *count*
     /// is per-artefact accounting, and summing two of them double-counts every
@@ -349,7 +360,8 @@ pub struct Joined {
     /// Unambiguous names carried by more than one artefact — what the join
     /// actually merged.
     pub joined_names: usize,
-    /// Names some artefact carries more than once, kept apart.
+    /// Names held apart: carried more than once by some artefact, or spelled
+    /// with an opaque `<…>` segment that cannot identify a function.
     pub ambiguous_names: usize,
 }
 
@@ -359,7 +371,15 @@ impl Joined {
         for part in parts {
             let mut seen: HashSet<&str> = HashSet::new();
             for name in part.names.values() {
-                if !seen.insert(name.as_str()) {
+                // A repeat inside one artefact is the visible half of the
+                // problem.  The other half is invisible to it: `name_path`
+                // renders a non-ident segment as the opaque `<Variant>`
+                // (`majit-charon-reader/src/ullbc.rs`), so two artefacts can
+                // each carry `pyframe::<Impl>::new` exactly once for two
+                // different functions, and the repeat test would merge them.
+                // A spelling that has already lost the part that distinguished
+                // it is not an identity in either direction.
+                if name.contains('<') || !seen.insert(name.as_str()) {
                     ambiguous.insert(name.as_str());
                 }
             }
--- a/majit/majit-translate/src/memory/gctransform/framework.rs
+++ b/majit/majit-translate/src/memory/gctransform/framework.rs
@@ -50,6 +50,10 @@ use std::collections::{HashMap, HashSet};
 pub struct CallGraph {
     /// `def_id -> fully qualified name`
     pub names: HashMap<u64, String>,
+    /// `def_id -> the spelling [`Joined`] identifies a function by`.  The name
+    /// for everything charon spelled out, the name plus the item's source line
+    /// where it did not.
+    pub keys: HashMap<u64, String>,
     /// `def_id -> callees`, from `Fun::Regular` Call terminators only.
     pub callees: HashMap<u64, HashSet<u64>>,
     /// `def_id -> callers`, the reverse index.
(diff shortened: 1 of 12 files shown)


youknowone:
All five findings adjudicated. Four fixed, one refuted with measurements. The branch is rebased onto `6cc6de4a536` and force-pushed.

---

### ✅ Codex P2 · `framework.rs:383` — "Preserve joins for corresponding impl methods"

Right, and the important half of it: an unjoined declaration is a childless leaf, so a caller reaching a collection only through it comes back "cannot reach" — under-approximation, which is how a required root bracket gets reported unjustified. My "findings identical" measurement could not have caught that; findings are the positives.

Blacklisting the spelling was the wrong instrument. The key is now the name plus **the item's own source line** where the name went opaque (`CallGraph::keys`) — charon stamps that line from the definition, so it survives into every artefact carrying the item, while two items collapsing onto one spelling are written at two lines. That is the "provably corresponding declarations" this asked for, and it needs no impl-owner id (those are artefact-local and would break every join).

Measured on `pyre-interpreter` ⋈ `pyre-jit`, over the 1120 opaque names both carry:

| | agree on line | disagree | already repeat in an artefact |
|---|---|---|---|
| opaque names in both artefacts | **779** | **7** | 334 |

779 real joins the blacklist was discarding; 7 genuine collisions, which is the defect the key exists to catch.

| join key | merged | held apart | reachable | justified | cannot-reach |
|---|---|---|---|---|---|
| bare name | 1337 | 2417 | 5317/41572 | 583 | 256 |
| name, `<…>` blacklisted | 551 | 20859 | 5317/41572 | 583 | 256 |
| **name + line** | **1913** | **215** | 5317/41572 | 583 | 256 |

Strictly more precise than either, and every verdict is unchanged.

---

### ✅ CodeRabbit · `framework.rs:363` — tests, and don't lean on `name.contains('<')`

Both taken. `contains('<')` is gone from the join rule; it now only selects which key gets a line appended, and the ambiguity test is the plain repeat test over keys. Five tests added covering the cases named: a declaration and a definition of `pyframe::<Impl>::new` at one line join and reachability flows through; the same spelling at two lines stays two nodes and no path is invented; a key repeated inside one artefact joins nothing; a fully spelled name joins on the name; `project` round-trips.

---

### ✅ CodeRabbit · `raise_from_invalid_cause.py:47` — exact messages
### ✅ CodeRabbit · `raise_from_invalid_cause.py:79` — don't swallow BaseException

Both taken. Five assertions are now `==` against the full message, and the cause-constructor handler catches `Exception`, so a KeyboardInterrupt or SystemExit keeps travelling.

---

### ❌ Codex P2 · `gc/mod.rs:1224` — "Return the actual collected-object count"

Three things, any one of which is enough:

1. **`test_gc.py:99` is not a gate.** `baseline.json` records `test.test_gc` as `SKIP`, reason `"implementation detail"`. A SKIP entry is not run in any lane, so nothing is being knowingly preserved.
2. **There is no count to thread.** `gc_hook::try_gc_collect`'s hook type is `fn()`. No stage of the collection pipeline computes a number of reclaimed objects.
3. **The number would not be that number.** incminimark reclaims everything unreachable, not the cycle members a refcounted collector is left with, and it is incremental — one `collect()` need not finish a major cycle. `assertEqual(gc.collect(), 1)` would still fail, and the answer would now be a nondeterministic count with different semantics rather than an honest constant. Upstream, with the same collector, answers a constant for exactly this reason.

The return **type** is the axis `AGENTS.md:146-150` puts under the spec, and that is what the earlier finding correctly caught and this PR fixed.

---

### Also in this push

`test.test_raise` goes **FAIL → PASS** (37 tests, both backends). Two more of its failures were the same area:

- `PyTraceback.lasti` held an instruction index while `tb_lasti` is defined in bytes, so the getter doubled and the setter and constructor halved — `TracebackType(tb, f, 1, 2).tb_lasti` read back `0`. The slot now holds the byte form, as `descr_get_tb_lasti` hands its slot straight out upstream; the two producers that count instructions double, the two readers that index by instruction halve. Verified that traceback rendering is byte-identical to CPython 3.14.6 after JIT warmup on both backends.
- `del tb.tb_next` had no deleter and fell through to the immutable-type path, reporting AttributeError about the type. It now answers `TypeError: can't delete tb_next attribute`.

The baseline is promoted in the shared file rather than a darwin overlay, since the fix is not host-specific — if ubuntu or windows disagree, this PR's CI is where that shows.

Gates: `check.py` 447/447 · 447/447, snippets 69/69 ×3, parity all pass, `cargo test --all --no-default-features --features dynasm` clean, `cargo fmt --check` clean, LLBC 4 crates current.

— *commented by Claude*

PR #1401, opened 2026-08-21
jit: trace a nested _getframe chain through its virtualref pairs

youknowone:
`sys._getframe(n)` with `n > 0` from an inlined MIFrame was residual: the walker
declined as soon as the recording-time `f_backref` chain crossed a
`JitVirtualRef`, so the specialized arm covered only depth 0. The generic
residual then forced the published callee frame and the loop aborted.

## What changed

**`try_walker_specialize_sys_getframe`** now walks the whole concrete chain
before emitting anything, and admits a vref hop when the vref is still one of
`MetaInterp.virtualref_boxes`. For each such hop it runs the orthodox bracket
around the concrete force — `vrefs_before_residual_call`, the
`CALL_MAY_FORCE` + `GUARD_NOT_FORCED`, then `vrefs_after_residual_call`, which
publishes `VIRTUAL_REF_FINISH(vref, virtual)` and replaces the tracked vref
with `CONST_NULL` (`pyjitpl.py handle_possible_exception` neighbourhood:
`vable_and_vrefs_before_residual_call` / `vrefs_after_residual_call`). With that
proof in the trace, `optimize_jit_force_virtual` forwards the force to the
paired virtual frame instead of materialising a vref with a null `forced`
field, and the `JIT_FORCE_VIRTUAL`/`GUARD_NOT_FORCED` pair leaves the optimized
loop.

Two supporting resolvers on `TraceCtx`: `live_virtualref_pair_for_ptr` and
`virtualref_virtual_for_object_ptr`, which still finds the virtual box of a
pair whose vref half `stop_tracking_virtualref` has already replaced with
`CONST_NULL`. Both read a pair's address back through `concrete_of_opref`
rather than through the `usize` pushed beside the box: that copy is the address
the object had at push time, and a minor collection relocates the object and
forwards the box's stamp, which is the hazard `opimpl_virtual_ref_finish`
documents on the same list.

The chain census is all-or-nothing and runs **before** the first emission, so a
decline never leaves the residual `getframe` a shorter chain than the
interpreter's. It also declines on a `hidden_applevel` frame, because
`ExecutionContext.getnextframe_nohidden` skips those without consuming a depth
level and the emitted per-hop shape consumes one per raw `f_backref`.

**Scope of the positive-depth admission.** It is gated on the walk landing on
the standard portal frame whose result is immediately consumed by `f_locals` —
statically preflighted by `next_op_is_f_locals_for_getframe_result`. That getter
only constructs the write-through proxy and never reads fast locals, so it
crosses no residual boundary. Generic positive-depth consumers, and inline
`f_lineno` / `f_lasti`, stay residual: their single live-coordinate slot cannot
describe a nested caller chain. `try_walker_specialize_load_attr`'s `f_locals`
arm accordingly accepts a second provable receiver — the standard virtualizable,
gated on **both** its red box and its concrete pointer, so no arbitrary inline
callee collapses onto the portal anchor.

**The locals write-back.** `pyframe.py fast2locals` is `@jit.unroll_safe`, so
upstream reads the virtualizable BOXES and never forces; pyre answers `f_locals`
with the 3.14 `FrameLocalsProxy`, which reads the frame's array lazily. Folding
the getter therefore has to emit `virtualizable.py write_boxes` for the
locals/cells region itself (`vable_array_region_write_back`), mirroring
`gen_store_back_in_vable`'s array loop. Without it every local the traced body
assigned read back unbound.

**`optimizeopt/heap.rs field_slot_index`** — a field descriptor the parent's
slot list does not place fell back to `Descr::index()`. Every
`VirtualizableInfo` field descriptor is minted with index `0`, so the
array-pointer read the write-back emits shared a `PtrInfo` slot with the vable
token store and was answered with the force token; the write-back then stored
through it. dynasm tolerated the bad base, cranelift deopted every iteration,
and wasm — which bounds-checks linear memory — crashed. Unplaced descriptors now
key by offset in a band of their own.

**`optimizer.rs`** — `drain_extra_operations_from` discarded the level's pending
queue when the drain returned `InvalidLoop`. `InvalidLoop` unwinds the recursive
Rust drain in place of RPython's exception unwinding, so the operations the
failing propagation emitted, plus this level's untouched tail, must stay visible
to the caller's unwind. They now move to `extra_operations_after`.

## Measured

Uniform across dynasm, cranelift and wasm:

| fixture | loops_compiled | loops_aborted | guard_failures | bh adopted |
|---|---|---|---|---|
| `getframe_bridge_force_after_store_declined` | — (+1 bridge) | 20 → 0 | 4114 → 201 | 20 → 0 |
| `getframe_bridge_force_plain_declined` | — (+1 bridge) | 20 → 0 | 4114 → 201 | 20 → 0 |
| `getframe_inline_subwalk_multiframe` | — | 5 → 0 | 0 → 1 | 5 → 0 |
| `getframe_residual_callee_own_frame_declined` | 1 → 2 | 5 → 0 | — | 5 → 0 |
| `getframe_root_loop_force_blackhole_crn_declined` | 0 → 1 | 5 → 0 | 0 → 1 | 5 → 0 |
| `..._crn_nonidempotent_declined` | 0 → 1 | 5 → 0 | 0 → 5 | 5 → 0 |

`getframe_inline_subwalk_multiframe` is the regression fixture for the new path:
`leaf` reads `_getframe(0).f_locals["x"]` and `_getframe(2).f_locals["base"]`,
so collapsing either lookup onto the portal loses a distinct name.

The `field_slot_index` change is in the shared heap optimizer, so its blast
radius was measured rather than assumed: one binary carrying both fallbacks
behind a switch, run over every synthetic fixture whose jit-stats differ from
its baseline on the development host (18 on dynasm). **All 18 read byte-identical
under both fallbacks** — the only fixture the change moves is
`getframe_inline_subwalk_multiframe` (cranelift `guard_failures` 9480 → 1,
`loops_compiled` 2 → 1; wasm from a crash to a clean run).

The five `*_declined` fixtures keep their historical names; each header records
what its shape now does and why the `_declined` half of the name is history.

## Reviewer findings

**P2, hidden frames in the census** — fixed; see the census paragraph above.

**P2, virtualref pairs resolved through a pushed address** — fixed; see the
resolvers paragraph above.

**P1, proxy/shadow coherence** — real, and **pre-existing on `main`**, not
introduced here. `p["x"] = 999` through a retained `FrameLocalsProxy` lands in
the frame array while a later compiled `LOAD_FAST` still reads the vable shadow.
Measured: gating the `f_locals` arm to the **inline** receiver only — i.e.
removing this PR's standard-virtualizable admission entirely — still reproduces
it, so the defect belongs to the inline-callee arm already on `main`. Scoped
measurement of the divergence: only `proxy.__setitem__` diverges;
read-after-`STORE_FAST`, `len(p)`, `del p[...]` and the `locals()` snapshot all
agree with `PYRE_JIT=off` and CPython. Closing it needs either a de-virtualized
frame for the proxy's lifetime — which is the residual force this fold exists to
avoid — or a proof that the proxy cannot outlive the fold point; both are larger
than this PR and neither is a regression from it.

## CI

Of the five gate failures on the previous head, two were this branch's and are
fixed by the `field_slot_index` change: `cranelift
getframe_inline_subwalk_multiframe` (`guard_failures 1 → 9480`, red on all three
runners) and `wasm getframe_inline_subwalk_multiframe` (crash, exit 1).

The other three are ubuntu-only ratio gates. A sibling PR (#1404, whose base is
an ancestor of this branch's) is the control:

| fixture | here | sibling #1404 | note |
|---|---|---|---|
| `wasm str_getitem_len_hot` | 3.7x > 3.5x FAIL | 3.6x > 3.5x **FAIL** | red on both |
| `cranelift mapdict_frozen_unboxing_fold` | `?91.3x`, FAIL at 100.9x | `?35.2x`, no FAIL | `?` = pypy exec under the floor-gate baseline; pypy 0.01s here vs 0.05s there, our own exec 1.16s vs 1.61s |
| `dynasm dict_update_hot` | 17.2x, FAIL at 16.0x | 14.2x | pypy 0.12s vs 0.20s; the fixture sits on its 15x gate and has flipped on other branches |

In all three the pyre-side absolute time is **lower** than the control's; what
moved is the denominator.

## Notes

Rebased onto `origin/main` after #1399. Every upstream citation this branch adds
names a symbol; `scripts/check-new-line-citations.py --base origin/main` is
clean.

A third commit on this branch fixed the two dead-token `warmstate` fixtures; it
was dropped during the rebase as a duplicate of #1398, which had landed the same
fix. The resolution was verified byte-identical to `origin/main`.

— *authored by Claude*

## Summary by CodeRabbit

- **New Features**
  - Improved JIT handling for `sys._getframe()` and `f_locals`, including inline and virtualizable frames.
  - Preserved frame-local state, virtual references, and array updates during optimization.

- **Bug Fixes**
  - Prevented field-slot collisions in optimized object layouts.
  - Preserved pending operations when optimization exits early.
  - Improved frame instruction tracking before residual calls.

- **Documentation**
  - Updated frame-tracing guidance and regression benchmark expectations to reflect current behavior.


youknowone:
Pushed `dd0d15b1e86`. Both reviewer findings and the two branch-attributable CI failures are addressed.

**P2 — hidden frames in the census.** Fixed in `42810821973`: the preflight walk now declines outright when a hop lands on a `hidden_applevel` frame, because `ExecutionContext.getnextframe_nohidden` skips those without consuming a depth level while the emitted per-hop shape consumes one per raw `f_backref`. Declining is the conservative half of the two the finding names; implementing the skip loop in the emitted traversal is the other, and is not needed to make the gate sound.

**P1 — proxy/shadow coherence.** Real, and reproduces on `main` without this PR. Repro:

```python
def g(n):
    x = n
    p = sys._getframe(0).f_locals
    p["x"] = 999
    return x
```

driven 4000 times yields `{2, 999}` under the JIT and `{999}` under `PYRE_JIT=off` and CPython. Gating `try_walker_specialize_load_attr`'s `f_locals` arm to the **inline** receiver only — i.e. removing this PR's standard-virtualizable admission entirely — still reproduces it, so the defect belongs to the inline-callee arm already on `main`. Scoped: only `proxy.__setitem__` diverges; read-after-`STORE_FAST`, `len(p)`, `del p[...]` and the `locals()` snapshot all agree with the interpreter. Closing it needs either a de-virtualized frame for the proxy's lifetime — the residual force this fold exists to avoid — or a proof that the proxy cannot outlive the fold point.

**CI.** Of the five gate failures on the previous head, two were this branch's, and they had one root cause. `OptHeap::field_slot_index` answered a field descriptor the parent's slot list does not place with `Descr::index()`; every `VirtualizableInfo` field descriptor is minted with index `0`, so the vable token store and the array-pointer read the new write-back emits shared a `PtrInfo` slot, and the read came back with the force token. dynasm tolerated the bad base, cranelift deopted every iteration (`guard_failures 1 -> 9480`, red on all three runners), and wasm — which bounds-checks linear memory — exited 1. `getframe_inline_subwalk_multiframe` now reads `loops_compiled=1 guard_failures=1` on dynasm, cranelift and wasm alike, matching its recorded baselines; `check.py --backend wasm --synthetic-pattern 'getframe_*'` is 27/27.

That change is in the shared heap optimizer, so its blast radius was measured rather than argued: one binary carrying both fallbacks behind a switch, run over every synthetic fixture whose jit-stats differ from its baseline on the development host (18 on dynasm). All 18 read byte-identical under both fallbacks, so the change moves exactly one fixture — the one it fixes.

The remaining three are ubuntu-only ratio gates, and #1404 (base an ancestor of this branch's) is the control: `wasm str_getitem_len_hot` is red there too (3.6x vs our 3.7x, gate 3.5x); `cranelift mapdict_frozen_unboxing_fold` carries the `?` marker on both runs — pypy's own exec was 0.01s here against 0.05s there — and `dynasm dict_update_hot` sits on its 15x gate with pypy at 0.12s here against 0.20s there. In all three the pyre-side absolute time is lower than the control's; what moved is the denominator.

— *commented by Claude*

youknowone:
Pushed `ef91ebe97ea`, rebased onto `41d3ff671aa`. Both new findings addressed; one skipped with a reason.

**Codex P2 — moved virtual frames resolved through the pushed address.** Fixed. `live_virtualref_pair_for_ptr` and `virtualref_virtual_for_object_ptr` compared against the `usize` pushed beside each `virtualref_boxes` box, which is the address the object had at push time; a minor collection relocates the object and forwards the box's stamp, leaving that copy naming the old address. This is the same hazard `opimpl_virtual_ref_finish` documents a few lines above, and it is now resolved the same way — through `concrete_of_opref`, pyre's `getref_base()` — with the pushed copy kept only for an entry that carries no stamp. All six `getframe_*` fixtures read identically before and after on dynasm and cranelift, so no recorded pair in the suite was stale; the fix is for the window, not for a fixture.

**CodeRabbit — `getframe_residual_callee_own_frame_declined` header says one loop.** Fixed. Its three baselines say `loops_compiled=2`. The sentence also asserted a PyPy loop count that nothing in the fixture measures, so it now states the recorded shape instead of attributing it to PyPy.

**CodeRabbit — preflight accepts hops the emission loop rejects.** Skipped, deliberately. The census accepts any non-null, non-hidden, non-vref hop while the emission loop accepts a non-vref hop only when it equals `standard_vable_ptr`; the finding notes this is safe today because the emission path rolls back through `cut_trace_with_snapshots` and declines. Tightening the census would be a strict subset — every chain it would newly reject is one emission already rejects — so it changes no outcome, only the point of decline. I would rather land that as its own change with its own fixture run than fold an untested gate narrowing into a PR whose gates are already green; a narrowed census that is wrong declines shapes that currently specialize, and nothing in this suite would show it.

**Verification on this head**, all three backends built from these sources: the six `getframe_*` fixtures match their recorded baselines on dynasm and cranelift, and `check.py --backend wasm --synthetic-pattern 'getframe_*'` is 27/27.

— *commented by Claude*

PR #1405, opened 2026-08-21
_abc: weak-set membership without the app-level frame

youknowone:
Twelve commits. One lands something; the rest are review response and support.

**`_abc`** — `weak_cache_contains` bounced out to app-level
`SimpleWeakSet.__contains__` for every membership test, and that one frame was
1.11 µs of `_abc_instancecheck`'s 1.67 µs. `simple_weak_set_contains` answers
natively and declines to the membership protocol otherwise.

Measured warmed, 3 interleaved rounds, 3/3 consistent:

| | before | after |
|---|---|---|
| `_abc_instancecheck` hit | 1.261 µs | 0.627 µs |
| `_abc_instancecheck` miss | 2.319 µs | 1.039 µs |
| `isinstance(obj, ABC)` | 1.636 µs | 1.034 µs |

Answering natively is sound only while every part of the collection is still
the one `_abc_init` installed, and review found four ways it was not. It now
declines unless the receiver's class is exactly `SimpleWeakSet`, that class's
`__contains__` is still the method `app_abc.py` installed *and still holds the
code object it was defined with*, and `data` is exactly a `set` — not merely one
by layout, which a subclass overriding `__contains__` also satisfies. A failed
probe hash recovers the pending exception instead of re-asking through the
protocol, which would have run an observable metaclass `__hash__` a second time,
and it recovers it through the set spelling, so the message names a set element
rather than a dict key.

The guard is the triple `(method, code, ref_global)`. The app-level body is
`try: wr = ref(item) / except TypeError: return False / return wr in self.data`,
so `ref` is its whole free-variable surface and it takes no defaults — pinning
those three is exhaustive for this body rather than a narrowing. Reading the code
object off the installed method is defined only for a function, so a
`__contains__` rebound to a non-function is rejected before that read, not by
comparing what the read returned.

**Withdrawn: both walker inline routes.** The branch added two and takes both
back out, so the net walker diff is empty and each case declines as `main` does.

- *`__contains__` from the `CONTAINS_OP` residual* (`723fe5edaed`, removed in
  `1312fe209b5`). Its `CompareOp` descriptor cannot be latched by
  `latch_abort_call_resume`, which reconstructs only `CallFn`/`CallFunctionEx`,
  so a non-bool result aborted the walk after the body's effects had run and the
  loop replayed from entry: 3002 body runs for a 3000-turn `while` loop, on both
  backends. Gating on `exc_override_sample_safe` as review proposed was measured
  to admit constant-returning bodies and nothing else — `self.a`, `x + 1` and
  `len(self.items)` all decline — which retires the route's whole win. A sound
  fix needs the object-level truth conversion (`PyreHelperKind::Truth`) minted
  from a walker route, which no helper does today. Removing the route also lets
  that loop compile: `loops_compiled=1 loops_aborted=0`, previously `0` and `2`.
- *An overridden `__new__` in the walker's type call* (`5af05b0b78a`, removed in
  `06ee51ff535`). It decided `__init__` from the requested class, while
  `descr_call` calls it on the returned object's class: a `__new__` returning an
  instance of a subclass with its own `__init__` ran that initializer 1041 of
  40000 times, the interpreted iterations only. Guarding the result needs the
  `w_class` word rather than the layout `ob_type` that a `GuardClass` reads,
  which is more than the route is worth while its original target fixture is
  blocked upstream of it for an unrelated reason.

The sibling rich-compare route `try_walker_inline_user_compareop` on `main` has
the same unlatched-abort shape — an effectful `__lt__` returning `NotImplemented`
also runs 3002 times against CPython's 3000. That is pre-existing and belongs in
its own issue.

**Fixture** — no bench in the corpus reached `_abc` at all: counting
`ABCMeta.__instancecheck__` calls across `pyre/bench/synth` returned zero for
every script. `abc_instancecheck_weak_cache` drives both caches plus a mid-loop
`register`, with baselines for all three backends. Note its ratio prints with a
`?` when pypy's exec time lands under `FLOOR_GATE_MIN_BASELINE_S`, which happens
on a quiet machine; the ceiling applies either way.

**Tests** — `pyre/extra_tests/parity_tests/abc_weak_cache_membership.py` covers
the empty-collection answer, an item that cannot carry a weakref, a metaclass
`__hash__` that raises (once, not twice) and is reported as a set element, a
collected entry, and the arms that must decline: a rebound collection, a
replaced `data`, a `data` that is a `set` subclass whose `__contains__` lies,
and a `__contains__` whose `__code__` was reassigned in place. The arms that
name the collection are guarded, since parity tests run under CPython too and
CPython does not expose `_abc_cache`.

**Known, not fixed here.** Both predate this branch and both are confirmed:

- A `staticmethod` subclass overriding `__get__`, installed as a class's
  `__new__`, is bypassed once the loop compiles — `{'bound': 2104, 'real':
  37896}` where CPython gives `{'bound': 40000}`. It reproduces with the
  withdrawn route entirely removed, and `pypyjit.set_param("off")` makes it
  correct, so it is a trace-time fold elsewhere in the type-call path. Repro in
  the review thread.
- `_abc._reset_registry(cls)` leaves a registered virtual subclass answering
  `True` where CPython answers `False`.

https://claude.ai/code/session_01QCct5Z2Yn3Be6Pz22GgNZw

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ABC cache membership handling, including cache updates, invalidation, and unusual class behaviors.
  - Preserved accurate error reporting when hashing fails for dictionary keys or set elements.
  - Improved thread safety for cached string hash values.
  - Strengthened cleanup behavior during concurrent code execution.
  - Prevented unsafe replay behavior during certain optimized operations.

- **Tests**
  - Added comprehensive compatibility tests for ABC weak-cache behavior.
  - Added a benchmark covering positive and negative instance checks, cache invalidation, and runtime performance statistics.
  - Updated performance statistics for loop compilation benchmarks.

### DIFF
--- a/pyre/bench/synth/abc_instancecheck_weak_cache.cranelift.jitstats
+++ b/pyre/bench/synth/abc_instancecheck_weak_cache.cranelift.jitstats
@@ -1,0 +1,15 @@
+bridges_compiled=1
+descr_set_absent=0
+descr_set_ambiguous=0
+descr_set_stale_absent=0
+fbw_blackhole_adopted_multi_frame=0
+fbw_blackhole_adopted_single_frame=0
+fbw_rolled_back_with_effects=0
+fbw_store_journal_rollback_failed=0
+field_pos_attached_misplaced=0
+field_pos_spec_misplaced=0
+guard_failures=202
+internal_compile_panics=0
+loops_aborted=0
+loops_compiled=1
+retraces_compiled=0
--- a/pyre/bench/synth/abc_instancecheck_weak_cache.dynasm.jitstats
+++ b/pyre/bench/synth/abc_instancecheck_weak_cache.dynasm.jitstats
@@ -1,0 +1,15 @@
+bridges_compiled=1
+descr_set_absent=0
+descr_set_ambiguous=0
+descr_set_stale_absent=0
+fbw_blackhole_adopted_multi_frame=0
+fbw_blackhole_adopted_single_frame=0
+fbw_rolled_back_with_effects=0
+fbw_store_journal_rollback_failed=0
+field_pos_attached_misplaced=0
+field_pos_spec_misplaced=0
+guard_failures=202
+internal_compile_panics=0
+loops_aborted=0
+loops_compiled=1
+retraces_compiled=0
--- a/pyre/bench/synth/abc_instancecheck_weak_cache.py
+++ b/pyre/bench/synth/abc_instancecheck_weak_cache.py
@@ -1,0 +1,84 @@
+# pyre-check: max-pypy-ratio=40
+# `_abc_instancecheck` answers from two per-class weak-set caches, and a
+# question that hits one is meant to cost the probe and nothing else.  This
+# drives both answers in a hot loop: a class that matches (positive cache) and
+# one that never will (negative cache), plus the stdlib `collections.abc`
+# checks, whose caches are shared with everything else that asks them.
+#
+# A `register` partway through discards every negative cache by bumping the
+# invalidation counter, so the second half re-walks and refills rather than
+# reading the entries the first half recorded.  Deterministic.
+import collections.abc as cabc
+from abc import ABCMeta
+
+
+class Shape(metaclass=ABCMeta):
+    pass
+
+
+class Circle(Shape):
+    pass
+
+
+class Square(Shape):
+    pass
+
+
+class Blob:
+    pass
+
+
+class LateBlob:
+    pass
+
+
+N = 72000
+SWITCH = N // 2
+
+
+def main():
+    circle = Circle()
+    square = Square()
+    blob = Blob()
+    late = LateBlob()
+    seq = [1, 2, 3]
+    mapping = {"k": 1}
+    hits = 0
+    misses = 0
+    for i in range(N):
+        # Positive cache: two classes so the probe is not answered by a single
+        # resident entry.
+        if isinstance(circle, Shape):
+            hits += 1
+        if isinstance(square, Shape):
+            hits += 1
+        # Negative cache, and the class the mid-loop `register` moves across.
+        if isinstance(blob, Shape):
+            hits += 1
+        else:
+            misses += 1
+        if isinstance(late, Shape):
+            hits += 1
+        else:
+            misses += 1
+        # The stdlib ABCs, asked against a matching and a non-matching type.
+        if isinstance(seq, cabc.Sequence):
+            hits += 1
+        if isinstance(mapping, cabc.Mapping):
+            hits += 1
+        if isinstance(seq, cabc.Mapping):
+            hits += 1
+        else:
+            misses += 1
+        if i == SWITCH:
+            Shape.register(LateBlob)
+    return hits, misses
+
+
+hits, misses = main()
+# `late` misses until the registration and hits after it; `blob` never hits.
+expected_hits = N * 4 + (N - SWITCH - 1)
+expected_misses = N * 2 + (SWITCH + 1)
+assert hits == expected_hits, "hits %r != %r" % (hits, expected_hits)
+assert misses == expected_misses, "misses %r != %r" % (misses, expected_misses)
+print(hits, misses)
--- a/pyre/bench/synth/abc_instancecheck_weak_cache.wasm.jitstats
+++ b/pyre/bench/synth/abc_instancecheck_weak_cache.wasm.jitstats
@@ -1,0 +1,15 @@
+bridges_compiled=1
+descr_set_absent=0
+descr_set_ambiguous=0
+descr_set_stale_absent=0
+fbw_blackhole_adopted_multi_frame=0
+fbw_blackhole_adopted_single_frame=0
+fbw_rolled_back_with_effects=0
+fbw_store_journal_rollback_failed=0
+field_pos_attached_misplaced=0
+field_pos_spec_misplaced=0
+guard_failures=202
+internal_compile_panics=0
+loops_aborted=0
+loops_compiled=1
+retraces_compiled=0
--- a/pyre/bench/synth/inline_freevar_after_mayforce.cranelift.jitstats
+++ b/pyre/bench/synth/inline_freevar_after_mayforce.cranelift.jitstats
@@ -11,5 +11,5 @@ field_pos_spec_misplaced=0
 guard_failures=1008
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=6
+loops_compiled=7
 retraces_compiled=0
--- a/pyre/bench/synth/inline_freevar_after_mayforce.dynasm.jitstats
+++ b/pyre/bench/synth/inline_freevar_after_mayforce.dynasm.jitstats
@@ -11,5 +11,5 @@ field_pos_spec_misplaced=0
 guard_failures=1004
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=6
+loops_compiled=7
 retraces_compiled=0
--- a/pyre/bench/synth/inline_freevar_after_mayforce.wasm.jitstats
+++ b/pyre/bench/synth/inline_freevar_after_mayforce.wasm.jitstats
@@ -11,5 +11,5 @@ field_pos_spec_misplaced=0
 guard_failures=1004
 internal_compile_panics=0
 loops_aborted=0
-loops_compiled=6
+loops_compiled=7
 retraces_compiled=0
--- a/pyre/extra_tests/parity_tests/abc_weak_cache_membership.py
+++ b/pyre/extra_tests/parity_tests/abc_weak_cache_membership.py
@@ -231,6 +231,60 @@ def a_mutated_contains_body_is_consulted():
         collection.__contains__.__code__ = original
 
 
+def a_non_function_contains_declines():
+    class Guarded(metaclass=ABCMeta):
+        pass
+
+    class Impl(Guarded):
+        pass
+
+    assert isinstance(Impl(), Guarded) is True
+
+    collection = type(Guarded._abc_cache)
+    original = collection.__contains__
+
+    class Descriptor:
+        def __get__(self, obj, objtype=None):
+            return lambda item: True
+
+    # Not a function at all.  Reading the code object off the installed method
+    # is only defined for a function, so the shortcut has to reject this before
+    # the read rather than by comparing whatever the read returned.
+    collection.__contains__ = Descriptor()
+    try:
+        assert isinstance(Plain(), Guarded) is True, 'a non-function __contains__ was read past'
+    finally:
+        collection.__contains__ = original
+
+
+def a_mutated_ref_global_is_consulted():
+    class Watched(metaclass=ABCMeta):
+        pass
+
+    class Impl(Watched):
+        pass
+
+    assert isinstance(Impl(), Watched) is True
+
+    collection = type(Watched._abc_cache)
+    namespace = collection.__contains__.__globals__
+    original = namespace['ref']
+    calls = []
+
+    def watching_ref(item, *rest):
+        calls.append(item)
+        return original(item, *rest)
+
+    # `ref` is the whole free-variable surface of the app-level body, and
+    # rebinding it moves neither the method nor its code object.
+    namespace['ref'] = watching_ref
+    try:
+        isinstance(Impl(), Watched)
+        assert calls, 'the rebound `ref` global was not consulted'
+    finally:
+        namespace['ref'] = original
+
+
 empty_collection_answers_then_fills()
 unweakreferenceable_class_reaches_the_subclass_check()
 a_raising_hash_runs_once()
@@ -241,4 +295,6 @@ if hasattr(ABCMeta('_Probe', (), {}), '_abc_cache'):
     replaced_data_is_consulted()
     a_data_subclass_keeps_its_override()
     a_mutated_contains_body_is_consulted()
+    a_non_function_contains_declines()
+    a_mutated_ref_global_is_consulted()
 print('OK')
--- a/pyre/extra_tests/parity_tests/abc_weak_cache_membership.py
+++ b/pyre/extra_tests/parity_tests/abc_weak_cache_membership.py
@@ -1,21 +1,22 @@
 # CPython-suite gap: `test_abc` asks membership only through classes whose
-# caches it never inspects, so it covers neither the empty-collection answer
-# nor an item that cannot carry a weakref, and it cannot cover a replaced
-# collection at all -- the reference build keeps the three collections in the
-# type's own struct, so there is no attribute there to rebind.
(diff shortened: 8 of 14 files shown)


youknowone:
Worked all three findings. Two were real and are fixed; the third names a real wrong answer whose stated mechanism does not survive a diagnostic.

**1. Decline set subclasses with custom membership — confirmed, fixed.**

Reproduced against the pre-change binary, which always takes the membership protocol:

```
ctl (protocol):  answer: True
fast path:       answer: False     # a set subclass whose __contains__ returns True, empty table
```

`is_set` is `py_type_check`, an `ob_type` test, and a subclass keeps the base layout while retagging `w_class` — so it passed. `contains` dispatches such an override through `subclass_special_override`, and reading the table answered past it. Now gated on `is_exact_type(data, &SET_TYPE)`.

**2. Propagate the first weakref hash failure — confirmed, fixed.**

Metaclass `__hash__` that raises, probed against a populated collection:

```
cpython:     hash calls: 1
ctl:         hash calls: 1
before fix:  hash calls: 2
after fix:   hash calls: 1
```

`Ok(None)` sent the caller back through the protocol, which built a second probe. Now recovers the pending exception via `take_pending_dict_key_error`.

**3. Invoke overridden staticmethod descriptors — the hardening is right, the attribution is not.**

The unwrap is now gated on `is_exact_staticmethod` and a subclass declines. But the wrong answer *survives that decline*:

```
$ PYRE_FBW_INLINE_DIAG=1 ./pyre-dynasm repro.py
[type-call-decline] __new__ wrapped in a staticmethod subclass
[type-call-decline] __new__ wrapped in a staticmethod subclass
hot: [('fake', 2104), ('real', 37896)]     # cpython: [('fake', 40000)]
```

So this route is not what produces it. `pypyjit.set_param("off")` gives `[('fake', 40000)]`, so it is JIT-side; the optimized trace lowers the type call to two residuals carrying baked constant pointers, i.e. the `__new__` resolution is folded at trace time somewhere other than this walker route. A plain `staticmethod`-subclass *method* attribute is answered correctly, so it is specific to the type-call path.

Repro, for whoever picks it up:

```python
def real(cls):
    o = object.__new__(cls); o.tag = 'real'; return o
def bound_instead(cls):
    o = object.__new__(cls); o.tag = 'bound'; return o

class Rebinding(staticmethod):
    def __get__(self, obj, objtype=None):
        return bound_instead

class Point:
    __new__ = Rebinding(real)

tags = {}
for _ in range(40000):
    t = Point().tag
    tags[t] = tags.get(t, 0) + 1
print(tags)   # cpython {'bound': 40000}; pyre {'bound': 2104, 'real': 37896}
```

I wrote a parity test for it and then withdrew it from this branch rather than land a red gate for a defect this branch does not introduce and I have not located. Flagging it here so it is not lost.

— *commented by Claude*

youknowone:
Second round: three reviewers, seven findings. Two confirmed and fixed, one refuted, one route withdrawn, three left with reasons.

### Confirmed and fixed

**Rebound `SimpleWeakSet.__contains__`** (Codex, and the parity review independently). The receiver stays an exact instance across the rebind, so the type gate passed and the native path answered anyway:

```
ctl (protocol):  True
before:          False
after:           True
```

The method `app_abc.py` installs is now stashed beside the type and compared on every call.

**`__new__` route skipped `descr_call`'s `call_init`** (parity review). `__new__` returning an instance of a subclass with its own `__init__`:

```
cpython:  Sub.__init__ ran 40000 / 40000
before:   Sub.__init__ ran  1041 / 40000     # interpreted iterations only
```

The route tested `__init__` on the *requested* class; `descr_call` calls it on the returned object's class. Fixing it properly needs a guard on the result's `w_class` — a `GuardClass` reads the layout `ob_type`, which every user instance shares, so it cannot tell `Base` from `Sub`. That is a bigger change than the route is worth right now: this is the second soundness hole found in it in one review pass, and its original target fixture is blocked upstream of it for an unrelated reason. **Route withdrawn**; `__new__ overridden` declines again, as on `main`.

### Refuted

**"Hash the weakref before answering an empty cache."** Not reproducible. Empty cache, metaclass `__hash__` that raises:

```
cpython: raised ValueError, hash calls: 1
ctl:     raised ValueError, hash calls: 1
new:     raised ValueError, hash calls: 1
```

The shortcut is not observable on any arm I could construct; the pre-change binary, which never takes it, behaves identically.

**Staticmethod subclass, closing the loop from last round.** The route is now gone entirely and the wrong answer is unchanged (`{'bound': 2104, 'real': 37896}`). It is a pre-existing trace-time fold in the type-call path, not this branch's. Repro is in my previous comment.

### Left, with reasons

**`cut_trace` → `cut_trace_with_snapshots`** (CodeRabbit). Not changed. There are six `cut_trace(pre_fold_pos)` sites in this file and five predate this branch (`#878`, `#918`, `#1295`×3). The two helpers' own docs contradict each other — `cut_trace` says stale snapshots are harmless and that truncating regresses bench; `cut_trace_with_snapshots` says they expose stale boxes. Changing one site without a reproduction, against a documented bench cost, would be a guess. Worth settling repo-wide by whoever owns the optimizer's snapshot export.

**Atomic JIT read of the Unicode hash memo** (CodeRabbit, Major). Not changed here. The parity review looked at the same lines and classified them as a sound free-threading adaptation rather than a mismatch. The narrower point — that the JIT lowers `Unicodehash` to a plain load while `hash_slot` stores through `AtomicI64::from_ptr` — is a real formal-UB mismatch, but the fix is in the majit lowering, not this file, and the field is deliberately a plain `i64` because an `AtomicI64` would renumber the field descrs the JIT addresses it by.

**Three `require_*` bools → one enum** (CodeRabbit). Agreed in principle, not done here: it retypes a parameter list threaded through four call sites in the hot dispatch path, which is a refactor orthogonal to what this PR is about. The immediate hazard it names — an arm reachable without latching — is closed by the assertion described below.

**Dead result arms** (CodeRabbit). Correct: `require_bool_result` validates both properties before `Continue`, where the CALL boundary can still be latched, so both arms are unreachable and their unlatched abort is the weaker disposition. Kept as the shape's classification with the invariant asserted, and the stale comment claiming they were reachable is corrected.

— *commented by Claude*

youknowone:
Third round: three findings — one Major from CodeRabbit, two P2s from Codex. (The parity review reported no regressions on this diff.) All three reproduce against a control binary. Two are fixed; the route the third names is withdrawn.

### 1. `CONTAINS_OP` can run an effectful `__contains__` twice — confirmed (CodeRabbit, Major)

The stated mechanism is exactly right. `reconstructed_all_ref_call_stack` hard-returns `None` unless the helper is `CallFn | CallFunctionEx`, so for a `CompareOp` descriptor `latch_abort_call_resume` is structurally a no-op, and the `Err(callee_inline_unsupported)` that follows a non-bool result rolls the walk back to the loop entry with the body's effects already applied.

```python
class Holder:
    def __init__(self):
        self.n = 0
        self.answer = 3           # non-bool
    def __contains__(self, x):
        self.n = self.n + 1       # observable, not sample-safe
        return self.answer
```

driven by `while i < N: 1 in h` for `N = 3000`:

| | body_runs |
|---|---|
| CPython | 3000 |
| pyre, JIT off | 3000 |
| pyre dynasm | **3002** |
| pyre cranelift | **3002** |

`MAJIT_STATS=1` reads `loops_compiled=0, loops_aborted=2` — the two extra body runs are the two rolled-back
attempts. With the route removed the same script reads `loops_compiled=1, loops_aborted=0`, so on this shape
the route was not merely unsound, it was also stopping the loop from compiling at all.

One correction to the report's framing: the caller's loop shape gates this independently of the route. Under a `for` loop the FOR_ITER admission gate (`inline_call.rs:3693`) already declines `CalleeReplaySafety::Dirty` bodies unless a bound method is present, so my first three repro attempts read 3000 and looked like a refutation. It only reproduces under `while`.

**The prescribed fix was measured and is inert.** Gating on `!body_facts.exc_override_sample_safe` leaves only bodies whose residuals are `LoadConst`/`BoxInt`:

| body | `exc_override_sample_safe` |
|---|---|
| `return 3` | admitted |
| `return self.a` | declined |
| `return x + 1` | declined |
| `return len(self.items)` | declined |

That is, the gate admits constant-returning `__contains__` and nothing else, which retires the route's whole measured win (`(i % 7) in c`, 0.635 → 0.004 us/iter). The only sound handling of a non-bool result after an effectful body has run is to emit the object-level truth conversion — `PyreHelperKind::Truth` — and no walker-side helper mints that descr today; adding one is a new mechanism, not a review fix.

So **the `__contains__` inline route is withdrawn from this PR** (`inline_call.rs` −280, `residual_call.rs` −7), on the same reasoning as the `__new__` route last round. The `_abc` native membership path, which is where the fixture's win actually comes from, is unaffected.

**This is a pre-existing family, not something the branch introduced.** The sibling rich-compare route `try_walker_inline_user_compareop` is on `origin/main` and double-executes identically when `__lt__` has an effect and returns `NotImplemented` — 3002 against CPython's 3000, and its `Err` is not latched either. That belongs in its own issue rather than this PR.

### 2. Preserve set-element context for hash failures — confirmed, fixed (Codex, P2)

`wr in self.data` is a set membership test, but the low-level probe recorded the failure through the dict hook and recovered it with `take_pending_dict_key_error`, so the message renamed the operand:

```
control / CPython:  cannot use 'weakref.ReferenceType' as a set element (unhashable type: 'Meta')
before:             cannot use 'weakref.ReferenceType' as a dict key    (unhashable type: 'Meta')
```

Added `take_pending_set_element_error` as the set spelling of the same recovery; both containers now share one body (`take_pending_hash_error_for`) and differ only in the wrapper they pass. Covered by a new parity arm asserting `'as a set element' in message`.

Note the repro needs a **populated** cache — an empty one takes the `w_set_len == 0` early-out and never reaches the hash, which is why a first attempt at this repro showed nothing.

### 3. Guard mutations inside the installed contains function — confirmed, fixed (Codex, P2)

Assigning `__contains__.__code__` leaves the same function object at the same address, so an identity test over the method alone still matched while the body being run no longer did.

CPython cannot serve as the control here — it exposes no `_abc_cache` attribute, so the script only runs on pyre. The control is instead the *rebind* spelling of the same evil body, which the previous round already made decline, measured on the same pre-fix binary:

| spelling, same evil body | pre-fix binary |
|---|---|
| `SWS.__contains__ = evil` (rebind, already declined) | `unregistered -> True` ← app-level answer |
| `SWS.__contains__.__code__ = evil.__code__` | `unregistered -> False` |
| `__code__`, after this fix | `unregistered -> True` |

The stash is now the pair `(method, code)` and is compared as a pair on every call.

**The residual is real and I have not closed it.** A function's globals and defaults are also mutable in place, and the pair does not cover them — `__globals__['ref']` is exactly the case in your report that this does not catch. What the fix buys is narrowing the native answer to a collection whose method *and* body are both untouched; it is not a proof that the app-level body could not have been redirected. That limitation is stated in the doc comment rather than left implicit, and a `__code__` parity arm is added.

### On the red `pyre/check.py (macos-latest)`

It is `re_jit_call_resume.py [dynasm]: timed out after 30s`, which is not this diff. That fixture runs **3.3s and 4.0s** on dynasm locally against a 30s watchdog — the macOS runner was roughly 8x slower than local on that job (its `cargo test` took 27m). The same fixture has previously failed this watchdog on one run and passed on a re-run of the identical commit. The one commit here with a path to string hashing (`object: read and write the str hash memo atomically`) uses `Relaxed` ordering, which on arm64 lowers to the same `ldr`/`str` as the plain field access it replaced.

### Verification

- parity test `abc_weak_cache_membership.py`: `cpython=OK dynasm=OK cranelift=OK`
- `check.py --synthetic-pattern abc_instancecheck_weak_cache`: `ALL PASSED 2/2`, dynasm 14.6x / cranelift 20.5x against the fixture's `max-pypy-ratio=40`; the dynasm and cranelift `.jitstats` baselines unchanged (the wasm baseline was not exercised locally — CI covers that lane)
- the `__code__` parity arm sits behind the existing `hasattr(..., '_abc_cache')` guard, so like its sibling rebind arms it exercises pyre only — CPython does not expose that attribute
- every defect above was measured against a control binary and re-measured after the fix, on a binary whose mtime was checked against the source edits

— *commented by Claude*

youknowone:
The parity review found two regressions on `1312fe209b5` that the other two reviewers did not, and both were mine from this round. Fixed in `abfcdb296ad`.

Worth noting for anyone reading the check list: **the parity review reported `pass` as a CI check while its comment body carried two regressions.** It also updates its existing comment in place rather than posting a new one, so neither the check status nor a "new comments since" query surfaces it. It has to be read.

### 1. A non-function `__contains__` reached an unchecked field read — confirmed, fixed

`simple_weak_set_contains_identity` handed whatever `lookup_in_type` returned to `function_get_code`:

```rust
pub unsafe fn function_get_code(obj: PyObjectRef) -> *const () {
    unsafe { (*(obj as *const Function)).code }
}
// # Safety: `obj` must point to a valid `Function` ... Only call on user-defined functions.
```

No type check, and the contract says function-only. Rebinding `SimpleWeakSet.__contains__` to an `int`, a `str`, `len`, a `classmethod` or a plain descriptor instance all reach it.

Measured before the fix: **no crash, and every answer correct** — `int`/`str`/`builtin`/`None` gave 0 hits, `classmethod` gave 2000, matching the app-level protocol. That is the read landing inside some allocation and returning a value that happened not to match the stash, so it declined for the wrong reason. The defect is the undefined read, not a wrong answer today; had the garbage matched, it would have taken the native path with a rebound method installed.

`is_function` now gates the read and a non-function reads as the `(0, 0, 0)` that no stash can equal — rejected *before* the read rather than by comparing what the read returned.

### 2. `__globals__["ref"]` — confirmed, and this one closes the residual I said I was leaving open

Last round I said the `(method, code)` pair could not cover globals and left it as a stated limitation. That was the wrong call: I had not checked how large the surface actually is. The app-level body is

```python
def __contains__(self, item):
    try:
        wr = ref(item)
    except TypeError:
        return False
    return wr in self.data
```

`ref` is its entire free-variable surface, and the body takes no defaults. So the stash is now the triple `(method, code, ref_global)` and rebinding `__globals__["ref"]` declines. There is no residual left on this path — the doc comment says what is pinned and why that is exhaustive, instead of hedging.

### Verification

- parity test `abc_weak_cache_membership.py`: `cpython=OK dynasm=OK cranelift=OK`, with two arms added — a non-function `__contains__` whose lying body must be reached, and a rebound `ref` that must be consulted
- the three findings from the previous round re-measured on the new binaries and still fixed: `body_runs 3000`, `as a set element`, `unregistered -> True`
- `check.py --synthetic-pattern abc_instancecheck_weak_cache`: `ALL PASSED 2/2`, dynasm 16.6x / cranelift 22.5x against `max-pypy-ratio=40`, `.jitstats` baselines unchanged
- binary mtimes checked against the source edits before every measurement

— *commented by Claude*

PR #1420, opened 2026-08-22
shutdown: skip the sweep after the `__main__` release loop when the one before it found nothing

youknowone:
`finalize_runtime`'s teardown runs a **whole-heap** sweep plus the finalizer queue (`collect_and_run_finalizers`) from five sites, and a run whose `__main__` globals are all scalars or live `sys.modules` entries still pays four of them. That teardown is about **13% of a short process**, and the cost is the sweeps, not the per-name dict clearing. #1393 removed one sweep by proving it could reach nothing; this removes a second one under a narrower condition.

### The sweep that closes the `__main__` release loop

The loop releases each `__main__` global newest-first, and sweeps after each release that `release_frees_nothing` did not rule out. It then sweeps once more unconditionally.

That closing sweep can only find something the previous one missed if a `__del__` the previous sweep *ran* dropped the last reference to something else. A release the loop skipped removes nothing from the reachable set — that is exactly what `release_frees_nothing` decides — and a release it did not skip is followed immediately by its own sweep.

So `_run_finalizers` now returns whether it took anything off the queue, `collect_and_run_finalizers` forwards that, and the closing sweep runs only when the last one did in fact run a finalizer.

### Measurement — and the honest result

Child CPU time (`os.wait4` rusage), min of 21 interleaved runs against a build of the parent commit. Wall clock is unusable on this box; see the note below.

| workload | min | median |
|---|---|---|
| exits with the queue already drained (`import gc; gc.collect(); gc.collect()`) | **−6.61%** | **−5.02%** |
| `import unittest, json, argparse, email, logging, http.client` | +0.02% | −1.51% |
| `python -c "pass"` | +0.65% | +1.16% |

**On ordinary runs this is worth nothing.** The gate is correct and the sweep is genuinely removable, but the *first* shutdown sweep finds something finalizable on every ordinary run, so the gate never opens. What blocks the wider win is whatever is always pending in that queue at exit — identifying it is the follow-up, not another gating argument.

I am proposing it anyway because it is a small, self-contained ordering argument that stands on its own, and because it makes the remaining cost attributable: after this, every sweep that still runs at shutdown runs because something was actually finalizable.

### Verification

`cargo fmt --all --check` and `cargo check --all --all-targets --no-default-features --features dynasm,cpyext` clean. Three finalizer repros — a module dict holding a `__del__`, a non-module value under a `sys.modules` key, and a `__main__` global with a `__del__` — behave identically to the parent build. `test_struct` and `test_threading` pass with no regressions.

Measurements were taken with this box at load 25–63 from a sibling session's three-backend `check.py`, which is why they are child CPU time rather than wall clock; the wall-clock A/B on the same builds read pure noise, with absolute times doubling between runs.

Assisted-by: Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime shutdown handling by tracking whether cleanup actions actually ran.
  * Finalizer processing now repeats collection only when necessary, avoiding redundant cleanup work.
  * Ensured cleanup chains close correctly during application teardown.

### DIFF
--- a/pyre/pyre-interpreter/src/executioncontext.rs
+++ b/pyre/pyre-interpreter/src/executioncontext.rs
@@ -1,5 +1,6 @@
 use pyre_object::PyObjectRef;
 use std::sync::OnceLock;
+use std::sync::atomic::{AtomicUsize, Ordering};
 
 use crate::PyFrame;
 
@@ -213,7 +214,23 @@ impl WRootFinalizerQueue {
     }
 }
 
+/// How many times the collector has run the death-queue trigger below.
+///
+/// The collector skips a queue whose deque is empty
+/// (`majit-gc` `Collector::execute_finalizer_triggers`), so a collection bumps
+/// this exactly when that collection put something finalizable on the queue.
+/// Reading it either side of a collection is therefore the collector's own
+/// answer to "did that sweep find anything to finalize", which is why the
+/// drain it feeds keeps its upstream shape and reports nothing.
+static FINALIZER_TRIGGER_COUNT: AtomicUsize = AtomicUsize::new(0);
+
+/// Read [`FINALIZER_TRIGGER_COUNT`].
+pub fn finalizer_trigger_count() -> usize {
+    FINALIZER_TRIGGER_COUNT.load(Ordering::Relaxed)
+}
+
 fn finalizer_queue_trigger() {
+    FINALIZER_TRIGGER_COUNT.fetch_add(1, Ordering::Relaxed);
     let action = space_user_del_action();
     if !action.is_null() {
         unsafe { (*action).fire() };
@@ -907,10 +924,11 @@ impl ExecutionContext {
         Ok(anchor.live())
     }
 
-    /// Returns whether any queued finalizer was taken; see [`UserDelAction::_run_finalizers`].
-    pub fn _run_finalizers_now(&mut self) -> bool {
+    pub fn _run_finalizers_now(&mut self) {
         let action = space_user_del_action();
-        !action.is_null() && unsafe { (*action)._run_finalizers() }
+        if !action.is_null() {
+            unsafe { (*action)._run_finalizers() };
+        }
     }
 
     /// gh-142766 compatibility for `generator.close()`: once the close path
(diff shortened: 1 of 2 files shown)


youknowone:
**On the P1 — accepted, and the bookkeeping now comes from the collector.**

The objection is right: `UserDelAction::_run_finalizers` is a port of `executioncontext.py`'s method, which returns nothing, and giving it a `bool` invented a contract upstream does not have purely to serve the `pyrex` teardown loop.

`d32dcf6b0ee` puts both `_run_finalizers` and `_run_finalizers_now` back to their previous bodies — each is now **byte-identical to `origin/main`** again.

The suggested alternative turned out to be available almost for free, because the collector already produces exactly this signal. `majit-gc`'s `Collector::execute_finalizer_triggers` runs a death queue's trigger only for a queue whose deque is non-empty:

```rust
for handler in &self.finalizer_handlers {
    if !handler.deque.is_empty() {
        (handler.trigger)();
    }
}
```

So a collection runs the trigger exactly when that collection put something finalizable on the queue. `finalizer_queue_trigger` now bumps a counter, and `collect_and_run_finalizers` reads it either side of its own `try_gc_collect`:

```rust
let triggers_before = pyre_interpreter::executioncontext::finalizer_trigger_count();
pyre_object::gc_hook::try_gc_collect();
let made_finalizable =
    pyre_interpreter::executioncontext::finalizer_trigger_count() != triggers_before;
```

That reads "did *this* sweep find anything", which is a slightly different question from the one the `bool` answered ("did a finalizer actually run") — and the difference only ever errs toward taking the sweep: a queued object whose finalizer is deferred by `gc_disabled` now counts, where before it did not.

**Verified unchanged**, with a temporary tag at each of the five sweep sites:

| run | sweeps taken |
|---|---|
| `python -c "pass"` | A, B, **C**, D |
| `import unittest, json, argparse, email, logging, http.client` | A, B, **C**, D |
| `import gc; gc.collect(); gc.collect()` | A, B, D — **C skipped** |

and the three finalizer repros (a module dict holding a `__del__`, a non-module value under a `sys.modules` key, a `__main__` global with a `__del__`) all still run.

---

Separately, while measuring this I ran the census the commit message asks for — *what is always pending in that queue at exit*. The answer is **one `_io.BufferedReader` plus one `_io.FileIO` per imported module**, held from the moment the importer drops the handle until the process exits, alongside 9 `_weakref.WeakrefLifeline`s and one coroutine. Adding 89 modules adds exactly 89 of each.

They are *closed* files, and upstream withdraws the registration for precisely this case — `W_IOBase.maybe_unregister_rpython_finalizer_io` calls `rgc.may_ignore_finalizer(self)` — while `rg may_ignore_finalizer` over pyre's Rust returns nothing. That is the follow-up, and it is what would let this gate fire on ordinary runs rather than only on a manually-collected one. Not in this PR.

— *commented by Claude*

PR #1429, opened 2026-08-22
cpyext: type mirrors, types built from C, and slots bound to the type that owns the method

youknowone:
Forty-five commits on the cpyext slice, rebased onto `origin/main`.

## What is here

**Type mirrors.** A type this runtime defines now crosses into C with the
blocks and scalars its header declares: the unary, length, operand-taking,
iteration and buffer slots; `tp_itemsize` and the mirror's `ob_size`; the
header size in `tp_basicsize`; the complex and `type` blocks; the namespace
`tp_dict` names; the ancestry tuples `tp_bases` and `tp_mro`; and `tp_dealloc`.
An inherited slot is taken from the base rather than filled, which is
`inherit_slots`.

**Types built from C.** A readied type takes its base's layout and
constructor, is built through the metaclass a spec names, is marked
weakrefable, and reports the four layout fields it declared —
`tp_basicsize`, `tp_itemsize`, `tp_dictoffset`, `tp_weaklistoffset` — as the
values C reads back, on all four construction routes. A spec has no slot id
for `__vectorcalloffset__`, `__dictoffset__` or `__weaklistoffset__`, so
those are read off `tp_members` as `_PyType_FromMetaclass_impl` reads them.

**Slot binding.** `PyType_GetSlot` on a builtin used to hand back a shared
trampoline that re-resolves on the *receiver*. A C subclass calling its
base's slot therefore re-entered its own method and overflowed the stack —
the commonest thing a C extension does. Slots of a runtime-defined type are
now bound to the type that owns the method, through a pool of monomorphised
thunks; a class written in Python keeps the shared trampoline, which is what
the language asks for there. Pool exhaustion falls back to the old
behaviour.

**A `dict` subclass defined in C.** `dict` is the one receiver predicate that
tests for a *backing* rather than a layout: instances carry the mapping
payload in a reserved layout slot. `create_all_slots` reserved it and the
readied-from-C route did not, so every `dict` method rejected such an
instance as a foreign receiver. Both routes now ask one predicate.

**Concrete storage, not the method of the same name.** `PySet_Add`,
`Discard`, `Pop`, `Clear` and `Contains` dispatched on the receiver, so a
subclass whose `add` calls `PySet_Add` recursed. They reach the storage now,
as `set_add_key` and friends do, and as the `dict` entry points already did.

**Vectorcall.** `PyVectorcall_Call` delegated to `PyObject_Call`, which is a
self-loop by construction: a type declaring a `tp_vectorcall_offset` has
`PyVectorcall_Call` *as* its `tp_call`. It now reads the `vectorcallfunc` at
the offset and unpacks tuple+dict into one vector, as `_PyStack_UnpackDict`
does.

**Deallocation.** Every mirror left `tp_dealloc` null while filling `tp_alloc`
and `tp_free`. A deallocator written for a type derived from a builtin ends in
the base's, read off the base's block, so the null was a call through address
zero. `type_mirror_edges` reports `tp_dict`, `tp_bases`, `tp_mro` and a heap
type's `tp_base` to a collection, as `type_traverse` visits them — a type's
MRO names the type, so the tuple minted for `tp_mro` refers back through the
mirror holding it.

**Entry points added.** The descriptor block and its entry points, the
datetime C API, the four `PyContextVar` calls, `PyFloat_FromString`,
`PyFloat_Pack`/`Unpack`, `PyUnicode_Format`, `Py_SET_REFCNT`,
`PyEval_GetBuiltins`, `PyMemoryView_GetContiguous`, `PyErr_Print`/`PrintEx`,
frame and code-object mirrors, `Py_intptr_t`/`Py_uintptr_t`,
`Py_EnterRecursiveCall`/`Py_LeaveRecursiveCall`, `PyDict_GET_SIZE`,
`PySet_CheckExact`, `PyFrozenSet_CheckExact`, `PyAnySet_CheckExact`.

**Fixes outside cpyext, found by running C extensions.**

* `pyframe` no longer merges cells into a frame that is not optimized — a
  class body carries `__classdict__` as a cell whose value is the namespace
  itself, and writing it back left the namespace holding itself.
* `memoryview` refuses a writable request over read-only storage.
* A thread entering from foreign code runs the thread-entry hook.
* `typedef` keeps a namespace's own `__module__` when building a type.
* `bytearray.__repr__` names the receiver type's own name component, which is
  what `_PyType_Name` reports.

## What it runs

SQLAlchemy 2.0.52's `cyextension` package — five Cython 3.3 modules, ~2.5MB of
generated C — compiles against these headers, imports with
`HAS_CYEXTENSION: True`, and the process exits 0. Exercised: `immutabledict`
(a `cdef class` derived from `dict`) construct / read / `union` /
`merge_with` / write-refusal, `OrderedSet` (derived from `set`) add / union /
intersection / discard / ordering, `IdentitySet`, `unique_list`, the
`processors` converters, `prefix_anon_map` and `cache_anon_map`.

## Gate

`cargo test --all --no-default-features --features dynasm,cpyext` —
**8202 passed, 0 failed**.
`cargo fmt --all -- --check` clean.

— *authored by Claude*

## Summary by CodeRabbit

- **New Features**
  - Expanded CPython C API compatibility for datetime, context variables, frames, descriptors, complex numbers, sets, Unicode, floating-point conversion, slices, thread state, tuples, and error handling.
  - Added support for contiguous memoryview access, vectorized calls, type dictionaries, metaclasses, heap types, code-object creation, and enhanced Cython interoperability.
  - Added string parsing, Unicode formatting/splitting, tuple item access, and `__pypy__.newlist_hint`.

- **Bug Fixes**
  - Correctly reject writes to read-only buffers.
  - Improved `StopIteration.value`, frame line reporting, set operations, reference counting, and garbage collection.


youknowone:
Pushed one more commit past the cpyext slice, found by running SQLAlchemy.

**`baseobjspace: run __getattr__ for a dict-less receiver's descriptor miss`**

`getattr`'s `hasdict` arm routed an `AttributeError` from `get(descr, ...)` to
`instance_getattr_hook_or_err`; its twin for a receiver carrying no instance
`__dict__` propagated it instead. An unset `__slots__` member raises exactly
there, so a class that shadows an inherited method with a slot of the same
name — so that `__getattr__` can install one on first read — reported the miss
rather than reaching the hook. `_handle_getattribute` wraps the whole of
`__getattribute__` and does not consult the receiver's storage.

Reproduces with no C extension at all:

```python
class Hook:
    __slots__ = ()
    def __getattr__(self, key): return "hook(%s)" % key

class B(Hook, str):
    __slots__ = ("lower",)          # shadows str.lower with an unset member

B("AbC").lower     # CPython 3.14: 'hook(lower)'   pyre before: AttributeError
```

`sqlalchemy/sql/elements.py quoted_name` is exactly that shape, and every
compiled statement goes through `IdentifierPreparer.quote`, which calls
`.lower()`. Snippet: `pyre/extra_tests/snippets/class_slots_shadowing_getattr.py`
(`gate=1`, green under CPython 3.14 and pyre).

## What runs now

SQLAlchemy 2.0.52 end to end on this branch, ORM included — declarative
mapping, `create_all`, `add_all` / `commit`, `select` with `order_by`,
`group_by` + `func.count`, `scalars().one()`, UPDATE by attribute assignment,
DELETE, and `Row._mapping` / index / attribute access on the `cyextension`
`BaseRow`, all with `HAS_CYEXTENSION: True`.

⚠️ One caveat on that run: pyre has **no `_sqlite3`**, so `create_engine`
cannot open a DBAPI. The run above used a throwaway ctypes-over-`libsqlite3`
`_sqlite3` kept outside the repo, purely to see whether anything else was
still broken behind it. Nothing was. `lib_pypy/_sqlite3.py` is vendored but
needs `_sqlite3_cffi`, and pyre has `_ctypes` but no `_cffi_backend` — that
gap is its own piece of work, not part of this PR.

## Gate

`cargo test --all --no-default-features --features dynasm,cpyext` —
**8202 passed, 0 failed**, exit 0.

— *commented by Claude*

youknowone:
Four more commits, all found by running cffi 2.1.1's own test suite on pyre.

**`cpyext: leave tp_new empty where a static type declares none`**

`type_ready_set_new`: a static type that derives straight from `object` and
declares no constructor takes `Py_TPFLAGS_DISALLOW_INSTANTIATION` and keeps the
slot empty, and a base that refuses instantiation hands the empty slot on rather
than the `PyType_GenericNew` stand-in pyre used to fill it with. The
interpreter-side refusal keys on the empty slot rather than on the flag, which
is `type_call`'s own test — a subtype inherits the empty slot *without* the
flag, and is just as unable to build an instance.

`_cffi_backend.CType()` used to hand back an instance made of zeroed storage,
and `_CDataBase()` / `__CDataOwn()` **segfaulted** from pure Python.

**`gc: run a dying mirror's finalizer before the collection reclaims it`**

A collection that found a mirror's linked object dead cleared the link and
queued the deallocator in the same pass, and the references the block's C fields
hold stopped rooting the moment its own object did — `mark_c_edges` re-grants
only from a block whose object survived. A finalizer reading those fields
therefore ran with them already reclaimed: `ffi.gc`'s destructor was handed a
null object and answered `SystemError: bad argument to internal function`.

`gcmodule.c finalize_garbage` runs before `delete_garbage`, and this is that
order. `rrc_claim_finalizers` (and its minor half) asks the embedder through a
new `rawrefcount::FinalizerClaimFn` for every mirror whose link is dead, marks
or drags out **the block's own object** so the free pass reads the mirror as
surviving, and queues the block; `mark_c_edges` then re-grants what the block
references for free. The drain runs the queued finalizers before `tp_clear` and
the deallocators, and the next collection reclaims what they released. The claim
answers once per block, so the retention lasts exactly one collection. A type
declaring no `tp_finalize` is never asked, and `_rrc_free` is unchanged.

Two things this deliberately is *not*, both refuted by measurement:

* Retaining the **referents** of a dying block instead leaks
  `cpyext_cycles::a_cycle_through_a_c_held_reference_is_collected` forever — the
  referent's attribute points back at the block's own object, so every round
  re-marks it.
* Gating the claim on `ob_refcnt == REFCNT_FROM_PYRE` ("the free pass will queue
  the deallocator") skips any block a borrow still holds, and
  `PyArg_ParseTuple(args, "O")` leaves one: `ffi.cast("char *", p)` alone turned
  `ffi.new_allocator`'s free callback into `myfree(None)`.

**`cpyext: apply rustfmt to the files this branch touches`**, and the
`PyNumber_Long` / writable-buffer / `PYPY_VERSION` commits described earlier.

## Measured

`PYRE_LLBC_SKIP_FINGERPRINT_CHECK=1 cargo test --all --no-default-features
--features dynasm,cpyext` — **8255 passed, 0 failed**, the same as before the
change.

cffi 2.1.1's own suite, run on this build:

| file | before | now |
| --- | --- | --- |
| `cffi0/test_ffi_backend.py` | 10 failed, 204 passed | **1 failed, 213 passed** |
| `cffi0/test_verify.py` | 3 failed, 150 passed | **0 failed, 153 passed** |
| `cffi1/test_new_ffi_1.py` | 4 failed, 118 passed | **1 failed, 121 passed** |
| `cffi0/test_vgen.py`, `test_zdistutils.py`, `cffi1/test_dlopen.py`, `test_re_python.py`, `test_parsing.py`, `test_function.py`, `test_model.py`, `test_commontypes.py`, `test_realize_c_type.py`, `test_cdata.py`, `test_version.py`, `test_unicode_literals.py` | | **0 failed** |

Both remaining failures are the same `test_char16_t`, and neither is a cffi
defect: the test source spells lone surrogates (`'\ud808'`), pytest rewrites
assertions by `ast.parse` → `compile(tree)`, and `ruff_python_ast`'s string
constant is a Rust `str`, which cannot hold one. Run outside pytest, cffi's
`char16_t` conversion is correct on pyre in every direction.

New regression coverage: `cpyext_types.Sealed` (a type only a module-level
factory builds) and `cpyext_cycles.Doomed` (a block whose finalizer hands its
own two fields back to the interpreter).

— *commented by Claude*

youknowone:
### cffi: the extern "Python" abort and the `lib.<func>` introspection wall are closed

Six commits since the last update, all cpyext parity fixes found by running cffi 2.1.1's own test suite on `pyre-dynasm`.

| | before | after |
|---|---|---|
| `test_recompiler.py` | **aborted** (SIGABRT) at `test_extern_python_1`, 14 failed before it | **165 passed, 3 skipped, 0 failed** |
| `test_verify1.py` | 2 failed, 136 passed | **138 passed, 0 failed** |
| whole suite (21 files) | — | **2 failed**, both `test_char16_t` |

The two remaining failures are not cffi defects: the test body gives all-correct results outside pytest. pytest's assertion rewriting round-trips the source through `ast.parse` → `compile(tree)`, and `ruff_python_ast`'s string constant is a Rust `str`, so a lone surrogate does not survive. That needs an upstream RustPython change.

#### What was wrong

1. **`PyGILState_Ensure` did not make the thread state current** (`be2be82cc90`). It took the GIL but left `cpyext_threadstate_is_current` as `PyEval_SaveThread` had cleared it, so `PyInterpreterState_Get` / `PyThreadState_Get` called `Py_FatalError`. Every `extern "Python"` callback hit it: `_cffi_backend` releases the GIL around the C call and the C stub calls back in. Upstream's `pystate.py PyGILState_Ensure` ends with the same assignment.

2. **A C function carrier's mirror fields were never filled** (`a48f8a12e91`). `PyCFunctionObject`'s `m_ml`, `m_self` and `m_module` were declared in the header but written nowhere. cffi's `lib_obj.c _cpyextfunc_get` reads `m_ml` off the block and tests `m_module` for pointer identity against the reference it handed `PyCFunction_NewEx` — so `ffi.typeof(lib.f)`, `ffi.addressof(lib, "f")` and `ffi.cast` of a lib function all failed. 13 tests.

3. **`PyArg_UnpackTuple` reported its own message** (`8428b9478a6`) — `foo2() expected 2-2 arguments (0 given)` where `getargs.c _PyArg_CheckPositional` reports `foo2 expected 2 arguments, got 0`. `_PyArg_CheckPositional` is now split out under that name, which Argument Clinic output also calls.

4. **The unraisable report's first line was missing its colon** (`8ec6a1603d3`). `errors.c write_unraisable_exc_file` ends the stated message with `":"` whether an object repr follows it or not; the printer here ended it with nothing, so *every* report — deallocator, generator close, thread, ctypes callback — was one character short of the reference text. Three call sites had baked that colon into their own literal and no longer do; the `_lsprof` timer, the `at_fork` callback and the async generator finalizer stated their object twice and now name it once, in the message.

5. **A frame mirror's `f_lineno` held an object address** (`e6f8598e67b`). `fget_f_lineno` has answered a `PyObjectRef` since 3e3b0a50a87 (#1437); the mirror fill still cast it to `c_int`, which still compiles. `get_last_lineno` is the number, and `frame_attach`'s own upstream comment names it as what the field is owed. This one is not cffi's — any extension reading a running frame's block read the low half of a pointer.

6. **The carrier mirror fill minted its own type** (`56663aa2502`) — minting allocates, and the fill runs inside another mirror's, whose caller holds raw reads.

Every expectation added to the pyrex fixtures was taken from CPython 3.14.6 running the same C against the same script.

`cargo test --all --no-default-features --features dynasm,cpyext`: **8265 passed, 0 failed**.

— *commented by Claude*

youknowone:
Second review pass. Spec is CPython 3.14, implementation follows PyPy; each decline names the upstream that decides it.

### Fixed

- **`object.h:246` — `Py_RELATIVE_OFFSET` / `Py_AUDIT_READ`.** `from_spec` now copies the spec's member table, adds the type-data offset to every relative member, clears the flag, and rejects the two spellings `_PyType_FromMetaclass_impl` rejects ("With Py_RELATIVE_OFFSET, basicsize must be negative.", "Member offset out of range (0..-basicsize)"). `adopt_offset_members` validates the three special offset members the way `special_offset_from_member` does. `read_member`/`write_member` refuse a member still carrying the flag, as `PyMember_GetOne`/`SetOne` do. `member_descr_get` emits `object.__getattr__` for a `Py_AUDIT_READ` member, as `member_get` does. `Extra` in `cpyext_types.c` declares two relative members, one audited, and the script reads them back.
- **`pyerrors.rs:1040` — `PyErr_PrintEx`.** The `_PyErr_PrintEx` else-arm: `"sys.excepthook is missing\n"` on stderr and then the exception is printed anyway. A `None` hook is now called rather than skipped; a lookup failure that is not `AttributeError` is reported unraisable, as `_PySys_GetOptionalAttr` returning -1 is.
- **`object.rs:1639` — `PyVectorcall_Call`.** A non-tuple `tuple` or non-dict `dict` is refused with `PyErr_BadInternalCall` instead of read through. (`_PyVectorcall_Call` asserts; an assertion is a crash in one build and a wild read in every other.)
- **`object.rs:32`** — the three copies of the current-frame lookup are one `call_named`.
- **`complexobject.rs:25`** — `CPyComplexObject`'s offsets pinned with a compile-time assertion, like its sibling mirrors.
- **`dictobject.h:20`** — `PyDict_GET_SIZE` is documented as a call that can fail.
- **Fixtures** — `point_declared_in` reads its `long` with `l`; `caller_spec` declares `Py_TPFLAGS_HAVE_VECTORCALL`; the `cpyext_bound_slot` helpers check the type whose slot they call; `cpyext_datetime.sizes` guards `PyDateTimeAPI` the way `table` does; `fetch_returned` checks what the fetch handed over before reading the block; the four `cpyext_type_dict` entry points check their type argument.

### Declined

- **`buffer.rs:1161` — a contiguous copy for `PyBUF_READ`.** `pypy/module/cpyext/memoryobject.py memory_from_contiguous_copy` raises exactly this, with exactly this message: `"creating contiguous readonly buffer from non-contiguous not implemented yet"`. pyre already is the implementation.
- **`pystate.rs:198` — release the thread-state dict when a thread exits.** `PyThreadState_Clear` decrefs `c_dict` and nulls it, which is what `pystate.py PyThreadState_Clear` does. PyPy's extra safety net is `ThreadState_dealloc` running from the `ThreadStateCapsule` when the `ExecutionContext` dies; pyre's thread state is a `thread_local!` with no such death, and a TLS destructor that decrefs would run interpreter code at thread teardown. Left as the gap it is: a thread that never calls `PyThreadState_Clear` leaks one dict.
- **`datetime.h:17`, `datetime.h:87`.** `pypy/module/cpyext/include/datetime.h` uses the same `DATETIME_H` guard and declares the same process-wide `PyAPI_DATA(PyDateTime_CAPI *) PyDateTimeAPI`.

### Also in this push

`include/pyre3.14t/` carried ~1350 lines of *implementation*, not declarations, because the variadic entry points cannot have Rust bodies. Those bodies now live in `pyre-interpreter/src/cpyext/src/{getargs,modsupport,abstract,unicodeobject,pyerrors,mysnprintf}.c` — the peers of `pypy/module/cpyext/src/` — compiled into the interpreter and exported from it, so an extension resolves them at load time instead of compiling its own copy. `modsupport.h` 917 → 33 lines, `pyre_format.h` 389 → 29, `abstract.h` 172 → 61. `cpyext-abi.py check` reads the C definitions the way it read the header inlines, so the same 24 non-Rust entry points are still checked against the recorded CPython declarations.

Gate: `cargo test --all --no-default-features --features dynasm,cpyext` — 8268 passed, 0 failed, before and after the move.

— *commented by Claude*

youknowone:
### Codex P2 `PyMemoryView_FromBuffer` — recorded as a TODO, not fixed here

The finding is right about upstream: `PyMemoryView_FromBuffer` in `pypy/module/cpyext/memoryobject.py` copies `c_ndim`, `c_format`, `c_shape` and `c_strides` across verbatim and imposes no dimensionality limit, while this refuses anything but a contiguous one-dimensional view.

Three facts place it outside this PR:

- **It is not a regression here.** The rejection is byte-identical at the merge base (`251ebba8b9c`) and arrived in `0afaed4baae` (#1264). This PR only shifted its line numbers.
- **The cited test does not run.** `test_ndarray_multidim` lives in `TestBufferProtocol`, which carries `@unittest.skipUnless(ndarray, 'ndarray object required for this test')`, and `ndarray` is `None` because `from _testbuffer import *` raises `ImportError` on pyre. That is why `test.test_buffer` is baselined `PASS` on both backends — the class is skipped, not passing.
- **The prerequisite is a module, not this function.** `lib_pypy/_testbuffer.c` is already in the tree; nothing builds it. Until it imports, lifting the restriction has nothing to grade it against — and `BufferView::ViewND` already exists on the interpreter side, so the wiring is the smaller half of the work.

Recorded as a TODO on the function in `85a8e28b2d4`, naming `lib_pypy/_testbuffer.c` as what comes first.

— *commented by Claude*

PR #1441, opened 2026-08-23
check: re-measure the startup subtrahend on a timer, and fit the wasm ratio allowances to what two runs measure; jit-trace build prepass split

youknowone:
Two independent workstreams share this branch (it is the shared work branch on a
tree several sessions write to). They touch disjoint files; the commit list
separates them cleanly.

## A. `pyre/check.py` — the startup subtrahend, and the wasm ratio allowances

`3663a4c3591`, `cbc5e40a2b7`, `3ef94baf33b`, `44be0f13e7a`, `748a0f70bd7`
(plus one line of `pyre/bench/synth/str_getitem_len_hot.py`)

### 1. Startup was measured once and subtracted from benches that ran for the next ten minutes

`measure_startups` ran one sampling pass upfront; `_exec_time` subtracted its
median from every bench for the rest of the job. The comment above
`STARTUP_SAMPLES` rests the median-over-minimum choice on *"the estimate has to
come from the same load conditions as the run it is subtracted from"* — and a
single upfront pass cannot meet that. One macos job read every interpreter's
startup at about twice its neighbour's while raw bench times moved by a fifth of
that; the difference lands whole in `exec` and collapsed a 0.074s exec to 0.009s.

- `_sample_startups` / `_adopt_startups` split out, and `maybe_refresh_startups`
  re-samples once the estimate in hand is `STARTUP_REFRESH_S` old.
- Called at the top of `run_bench` / `run_synthetic_bench` — a **fixture**
  boundary, never between one fixture's backends, because the pairing argument
  requires a fixture's numerator and denominator to be reduced by one estimate.
- A target whose empty program exits non-zero on a refresh **keeps its previous
  estimate** and renders `[unsampled]`; the single pass had written `0.0`, which
  is the right no-subtraction default at t=0 and a silent inflation of every
  later `exec` mid-job.
- `measure_startups` now runs **after** `warmup`. It used to sample whatever
  state the cargo build left behind — cold page cache, every core just busy —
  while the benches it is subtracted from run one process at a time.

**`STARTUP_REFRESH_S` is sized from the logs, not guessed.** The phase between
the startup line and the last comparison is 592s on ubuntu, 402s on macos, 661s
on windows — the rest of those 45–70min jobs is cargo. 300s crosses that 1/1/2
times; **120s** crosses it 4/3/5. Each pass now prints its own wall cost so the
next sizing reads a measurement.

### 2. `str_getitem_len_hot` outgrew its allowance, and the ceiling paragraph had two wrong statements

Post-#1425 the fixture reads **2.87x / 2.94x** across two ubuntu runs of one
base, against the `4.8x` its header carried and the `4.0x` global. Its own note
named the condition for removal — the leg crossing out of the trace module once
per iteration — and #1425 is what changed it. Directive deleted.

The paragraph above `WASM_MAX_DYNASM_RATIO` said the constant *"came back to
4.0"* (the sequence is 3 → 4 → 3.5 → 3.8 → 4.0, so that reads as a restoration
of a value it never came back to) and *"No fixture carries an allowance today"*
while three did. Both corrected, and the census that actually bounds the
constant put in their place.

### 3. An allowance the backend has outgrown used to look exactly like one holding a fixture up

`wasm_ratio_allowed` carried `(name, ceiling)` and the summary printed the
ceiling alone. Finding the 4.8x-over-2.9x case above took parsing 434 comparison
rows out of a job log. It now carries the measured ratio — taken from the values
the gate judged, so a median retry is what gets reported — and the summary reads:

```
wasm/dynasm ratio raised above 4x by `# pyre-check: max-wasm-ratio` for:
  builtin_folds_hot 13x (measured 9.2x), math_folds_hot 13x (measured 3.9x)
```

`WASM_RATIO_FIT_HEADROOM = 1.15` — the margin allowances are already fitted with
— gates the "outgrown" call. Without it `math_folds_hot` (3.93x under a 4.0
ceiling) is named as outgrown on the strength of 0.07x, **on a host that is not
the one its 13x was fitted to**: its header records darwin-arm64 at 8.1–11.3x,
which no ubuntu run measures. The line says the reading is this host's and points
at the fixture's own note.

### Measurements this adds to the file

Two ubuntu runs on base `9be6297e7dd` (PRs #1432 and #1382) gate 273 fixtures in
common:

| | |
|---|---|
| worst **without** an allowance | `short_circuit_boxed_int_cross_fn` 3.01x → 3.50x |
| next | `pickle_terminal_raise_resume` 3.43x / 3.42x |
| run-to-run `\|Δ\|/max` over the 273 | median 4.7%, p75 8.8%, **p90 14.9%**, tail 67% |

So under the plus-15% rule the ceiling's floor is **4.03x — 4.0 is the fitted
value, not slack**, and 3.5 is breached rather than merely tight. And the
`wasm_ratio_gate` docstring's claim that 15% was chosen over *"the ~3% ubuntu
shows between runs"* is wrong: 15% is that runner's p90. Both corrected.

**No gate was widened.** The only movement is one dead allowance removed.

### Verification

`ast.parse` clean; unit checks over `_adopt_startups` (adoption, failed-target
hold, timer gating, `--no-startup-subtract` no-op, `_exec_time` tracking the
refresh) and over the outgrown predicate; `check.py --no-build --backend dynasm`
runs of `--no-synthetic` (10/10) and single fixtures, with `STARTUP_REFRESH_S`
forced low to fire a refresh at every boundary.

⚠️ The code inside `if backend == "wasm"` is **not** exercised locally — this box
has no wasm build. CI is the gate for it.

## B. `jit-trace` build prepass and Charon extraction

`9200d5d8d21`, `85ed655effc`, `819e58589e2` — `scripts/llbc_extract.py`,
`pyre/pyre-jit-trace/build{.rs,/prepass.rs}`, `pyre/scripts/extract-llbc.py`,
the `Cargo.toml`s and one CI workflow line.

**These are another session's commits on this shared branch, not mine.** Their
own messages carry the reasoning and measurements (`--hide-marker-traits`:
96.8s → 81.3s user, artefact 781MB → 752MB, with the cross-process determinism
set compared byte-equal). I could not reach the author — a cross-session message
went unanswered — so I am not representing that reasoning here beyond pointing
at it.

Two things I did do, because this PR carries them:

**`ac160492314` — fixed the `prepare Charon/LLBC (macos-latest)` failure they
caused.** `charon toolchain-path` shares its stream with rustup, which announces
an install when the runner does not have the pinned nightly yet; both call sites
read the capture whole, so `charon_host_triple` built a `rustc` path out of the
notice and macos died before `cargo test` and `check.py` could run. Now routed
through `charon_toolchain_path`, which takes the last non-empty line.
`charon_host_triple` is introduced by `9200d5d8d21` and is not on `origin/main`,
so this was a branch-only regression.

**Audited the `prepass` feature split for a consumer that turns defaults off
without naming it** — the failure mode would be a build script silently writing
placeholder JIT tables. Every site is covered:

| turns defaults off | how `prepass` still arrives |
|---|---|
| workspace `pyre-jit-trace` | `pyre-jit`, its only inheritor, forwards `dynasm`/`cranelift`/`prepass` one by one |
| `pyre-wasm` → `pyre-jit` | names `features = ["prepass"]` |
| `pyre-wasm` → `pyre-jit-trace` (does **not** name it) | only reachable via `wasm-host`, which turns on `pyre-jit/prepass`; unification covers it |
| CI `-p pyre-jit-trace --no-default-features` | `,prepass` added by `85ed655effc` |
| `check.py` / `cpython_tests/run.py` backend builds (`-p pyrex --no-default-features`) | `--no-default-features` applies to the selected package only, and `pyrex` keeps `pyre-jit`'s defaults |
| `-p pyre-jit --no-default-features` in the LLBC extraction | the one path that drops `prepass` on purpose — the point of the change |

No other CI step or script builds either crate with `--no-default-features`.

> **Correction.** That table audits the **`prepass`** axis only. The workspace
> entry's `default-features = false` drops *both* of `pyre-jit-trace`'s defaults,
> and the `dynasm` axis had a real gap: a `pyre-jit` naming no backend (wasm32)
> left `pyre-jit-trace` with neither feature, which stopped it compiling. Two CI
> jobs were red on it; fixed in `3017b55f561`, and the diagnosis is in a comment
> below.

I have **not** reviewed the 2321 moved lines of `build.rs` → `build/prepass.rs`.

---

## Review responses (Codex + CodeRabbit)

Six items, each checked against the code rather than accepted on faith:
**4 fixed, 2 declined.** Full per-item reasoning is in the triage comment on
this PR; the short form:

| item | verdict | commit |
|---|---|---|
| `codegen_cache_key` never hashes `build/prepass.rs` | fixed — the whole `build/` dir is hashed | `e648ae3812d` |
| parallel layout pass stalls on its pipe | fixed — output held in a file, not a pipe | `4bd57a259e5` |
| `host_triple` local shadows the module function | fixed (was latent: nothing in `extract` calls it) | `4bd57a259e5` |
| a raise orphans the running Charon passes | fixed — try/finally; `git diff -w` proves the re-indent is pure | `711bc247a54` |
| per-target timestamps for the startup samples | declined — every pass re-samples every target, and a failed one already renders `[unsampled]` | — |
| error when `invalidate_cargo_unit` removes nothing | declined — a cold unit in a warm layout hits that condition on a correct run | — |

The first four are all in the B workstream (another session's commits on this
shared branch). `711bc247a54` is deliberately its own commit so the ~110-line
re-indent is the only thing under review in it.
— *authored by Claude*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebAssembly benchmark checks by recording measured ratios, identifying outdated allowances, and preserving reliable startup estimates.
  - Fixed JIT and WebAssembly builds to generate complete runtime data instead of placeholders.
  - Improved extraction diagnostics, validation, cache handling, and failure reporting.

- **Performance**
  - Added configurable parallel processing for faster LLBC extraction.
  - Reduced unnecessary rebuilds and improved resource-aware execution.


youknowone:
브랜치에 다른 세션의 커밋 3개(`9200d5d8d21`, `85ed655effc`, `819e58589e2` — llbc/jit-trace 빌드 프리패스)가 섞여 있어, 본문에서 A(`pyre/check.py`)와 B(빌드 스크립트)로 분리했습니다. B는 제가 작성·검증하지 않았고 해당 세션에 별도로 알렸습니다.

리뷰 시 파일 기준으로 갈립니다:
- **A (제 것)**: `pyre/check.py`, `pyre/bench/synth/str_getitem_len_hot.py`
- **B (다른 세션)**: `.github/workflows/pyre-ci.yml`, `Cargo.toml` ×4, `pyre/pyre-jit-trace/build.rs`, `pyre/pyre-jit-trace/build/prepass.rs`, `pyre/scripts/extract-llbc.py`, `scripts/llbc_extract.py`

⛔A에서 **게이트를 넓힌 곳은 없습니다.** 유일한 이동은 `str_getitem_len_hot`이 이미 벗어난 4.8x allowance 하나를 **제거**한 것이고, `WASM_MAX_DYNASM_RATIO`는 4.0 그대로입니다(두 런 census상 4.03x가 하한이라 그 값이 fitted임을 주석에 기록).

— *commented by Claude*

youknowone:
### `prepare Charon/LLBC (macos-latest)` 실패 — 원인 확정, `ac160492314`로 수정했습니다

macos 레그 전체(`cargo test`, `check.py`)가 이것 때문에 skip됐습니다.

```
extract-llbc.py:147 main -> llbc_extract.py:3212 run_cli -> :2103 extract
  -> :1886 charon_host_triple -> :198 run_capture
FileNotFoundError: [Errno 2] No such file or directory:
  'The required toolchain is not installed. Installing...\n\n
    nightly-2026-02-22-aarch64-apple-darwin installed - rustc 1.95.0-nightly (...)\n\n
   /Users/runner/.rustup/toolchains/nightly-2026-02-22-aarch64-apple-darwin/bin/rustc'
```

`charon toolchain-path`가 rustup과 같은 스트림을 공유하는데, 러너에 pinned nightly가 아직 없으면 rustup이 설치 공지를 먼저 찍습니다. 두 호출부 모두 캡처 **전체**를 경로로 읽고 있어서 `charon_host_triple`이 공지까지 포함한 `rustc` 경로를 만들었습니다. (`ensure_charon_std`는 `Path(...).name`이 마지막 세그먼트에 떨어져 우연히 살아남았을 뿐, 같은 버그입니다.)

**마지막 비어있지 않은 줄**을 취하는 `charon_toolchain_path` 헬퍼를 넣고 두 호출부를 그리로 보냈습니다. 실패한 캡처 원문 / 정상 1줄 / 공백 패딩 세 경우로 파싱 규칙을 검증했습니다.

**귀속**: `charon_host_triple`과 `charon toolchain-path` 호출은 이 브랜치의 `9200d5d8d21`(B 워크스트림)이 도입한 것으로, `origin/main`에는 없습니다. 즉 브랜치 한정 회귀이고 A(`pyre/check.py`)와는 무관합니다. B 작성자를 찾는 중인데 아직 확인이 안 돼, PR이 막혀 있어 제가 고쳤습니다 — 원 작성자께서 다르게 고치고 싶으시면 되돌리셔도 됩니다.

— *commented by Claude*

youknowone:
### 정정: 앞서 올린 census 숫자가 틀렸습니다 (`771de0eb105`)

본문 A절의 census 표에서 **"worst without an allowance = `short_circuit_boxed_int_cross_fn` 3.50x"** 는 **틀렸습니다**. 파서가 행을 조용히 버리고 있었습니다.

check.py 표 형식은 `  {name:<35} {cpython:>8} {pypy:>8}` + 백엔드마다 ` {cell:>18}` 입니다. 세 가지 함정이 각각 행을 **에러 없이** 떨어뜨립니다:

| 함정 | 깨지는 이유 | 증상 |
|---|---|---|
| 뒤에서 토큰 세기 | 셀이 `FAIL`/`SNAPDIFF`/`timeout`이면 **한 토큰**(정상은 `0.21s   ~20.7x` 두 토큰) | 실패한 런에서 그 fixture가 사라짐 |
| 고정 열 오프셋 | `{name:<35}`은 **최소폭** — 더 긴 이름은 뒤를 밀어냄 | `unspecialized_long_math_zip_paths`가 3.02x인데 11.32x로 읽힘 |
| `\S+ \S+ \S+` head 검사 | cpython 셀이 `skip (opt)` = **두 토큰** | `skip-cpython` fixture 전부 탈락 (= allowance 가진 둘) |

⛔결정적인 건 **두 런의 교집합**을 쓴 것입니다. `foriter_make_function_body`가 한 런에서 SNAPDIFF라 교집합에서 빠졌는데, **allowance 없는 fixture 중 최악(3.74x)** 입니다.

**행 끝에서 18폭·19스트라이드로 잘라내고 합집합**으로 다시 세니 gated가 273 → **299**개가 되고 순위가 바뀝니다:

| fixture | #1432 | #1382 | allowance |
|---|---|---|---|
| builtin_folds_hot | 8.62x | 9.23x | 13x |
| math_folds_hot | 3.89x | 3.93x | 13x |
| **foriter_make_function_body** | (SNAPDIFF) | **3.74x** | — |
| short_circuit_boxed_int_cross_fn | 3.01x | 3.50x | — |
| str_getitem_len_hot | 2.87x | 2.94x | 제거됨 |

**결론이 바뀝니다.** plus-15% 규칙을 3.74x에 적용하면 **4.30x**를 요구합니다 — 트리의 4.0보다 **위**입니다. 즉 4.0은 slack이 아니라, 최악 fixture에 **7%**만 남기고 있습니다(규칙은 15%를 원함).

⛔그렇다고 4.3으로 올리는 건 아닙니다 — 이 천장은 이미 두 번 fixture 때문에 넓혀졌고 두 번 다 붉었습니다. `foriter_make_function_body`를 고치거나 exempt하는 게 맞습니다. 주석에 그 판단까지 적었습니다.

`str_getitem_len_hot` allowance 제거는 그대로 유효합니다(2.94x × 1.15 = 3.38 < 4.0). 세 fixture는 로그 원본 행과 대조해 검산했습니다.

— *commented by Claude*

youknowone:
## Review triage — 6 items, verified against the code rather than accepted on faith

**3 fixed, 2 declined with a counterexample, 1 deferred.**

### Fixed

**Codex P1 — `codegen_cache_key` does not hash `build/prepass.rs`** (`e648ae3812d`)

Confirmed by reading the function: it hashes `build.rs`, two named `src/*.rs`
files, `source_paths`, `majit-translate/src`, every `{majit,pyre}/*/src`,
`Cargo.lock`, the rustc version and the LLBC inputs — and never `build/`. Since
`85ed655effc` moved the prepass to `build/prepass.rs` (`#[path]`-included by the
wrapper), editing it recompiles the build script and makes Cargo rerun it, and
then the unchanged key restores the previous generated tables over the ones the
rerun just wrote. Fixed by hashing the whole `build/` directory, which also
covers a second module added beside it. No `CODEGEN_CACHE_VERSION` bump is
needed — adding a hashed input already changes every key.

**CodeRabbit — the parallel layout pass can stall on its pipe** (`4bd57a259e5`)

Real, and worth being precise about the failure mode: it is not a deadlock. The
host pass is spawned *without* `stdout=PIPE` (it inherits the parent's stdout),
so `finish_charon_pass`'s bare `proc.wait()` is safe, and the parent does
eventually reach `communicate()`. What is lost is the parallelism the code was
added for: the layout child writes to a 64 KB pipe that nothing reads until the
host pass has finished, and a Charon cargo pass writes far more than that, so
the child blocks at the first full buffer — for exactly the span
`layout_passes_in_parallel` exists to overlap. Output now goes to a
`tempfile.TemporaryFile`, read back only on failure, so the hold-back behaviour
is unchanged.

**CodeRabbit — `host_triple` shadowing** (`4bd57a259e5`)

Real but latent: the local at the top of `extract` does shadow the module-level
`host_triple`, and nothing inside `extract`'s body calls that function, so no
current path is broken. Renamed to `charon_host` anyway — it costs two lines and
removes the footgun for the next edit.

### Declined

**CodeRabbit — per-target timestamps for the startup samples**

`startup_sampled_at` records when a *pass* ran, and every pass re-samples every
target unconditionally, so a target that failed to sample is retried at the next
refresh, at most `STARTUP_REFRESH_S` later. Per-target timestamps would only
change behaviour if a failed target could postpone its own next attempt, which
it cannot. Its staleness is also already legible in the log: `_adopt_startups`
keeps the previous estimate for such a target (deliberately — writing 0.0 would
silently inflate every later `exec`) and renders it as `[unsampled]` rather than
as a fresh `median[min..max]`.

**CodeRabbit — error when `invalidate_cargo_unit` removes nothing**

The proposed condition is "`.fingerprint` exists but no entry matched", and that
has a legitimate cold case: the directory exists as soon as *anything* was built
for that triple in that layout, which does not imply this package was. The first
extraction into `target/llbc-layouts`, or into a layout where only `corpus` had
been built, hits it on a completely correct run — so as a hard error it would
break real runs and as a warning it would cry wolf. The failure it aims at is
already caught downstream: a unit that cargo skips makes Charon emit nothing, and
`finish_charon_pass` refuses that with "Charon emitted no artefact at …". The
documented return value stays informational.

### Deferred

**CodeRabbit — `try/finally` so a mid-pass exception does not orphan the children**

The concern is real (a raise between the host spawn and the drain loop leaves
multi-GB cargo/Charon processes running), but the `finally` has to span the whole
spawn-to-drain region — about 110 lines of `extract` re-indented — and it guards
a path that neither CI nor a local run exercises, so it would land unvalidated
inside an unrelated diff. Better as its own commit where the re-indent is the
only thing under review.

— *commented by Claude*

youknowone:
Correction to the triage above: the `try/finally` item is **fixed, not deferred** (`711bc247a54`).

My reason for deferring was that a ~110-line re-indent of `extract` would land unvalidated inside an unrelated diff. Both halves of that dissolve when it is its own commit: `git diff -w` on `711bc247a54` shows *only* the `try:`/`finally:` lines and the `layout_passes` declaration moved above the spawn — which is a proof that the re-indent changed nothing else, not an assertion about it. The declaration has to move so the `finally` clause cannot read an unbound name, and the host `Popen` stays outside the `try` so a spawn that itself raises has no child to reap.

Final tally: **4 fixed, 2 declined.**

— *commented by Claude*

youknowone:
## Two red jobs, one root cause — fixed in `3017b55f561`

`sandbox e2e + wasm web build (ubuntu-24.04)` and `pyre/check.py (ubuntu-24.04)`
both died at the same compile error, before either ran anything it was there to
run. The check.py job never reached a bench, so nothing on this branch has yet
been graded against a perf gate on ubuntu.

```
error[E0425]: cannot find function `pyre_production_cpu` in module `majit_metainterp::blackhole`
    --> pyre/pyre-jit-trace/src/trace.rs:3283
note: found an item that was configured out
    --> majit/majit-metainterp/src/blackhole.rs:8216
8215 | #[cfg(any(feature = "dynasm", feature = "cranelift"))]
```

**Why this branch and not `main`.** The workspace entry for `pyre-jit-trace` now
carries `default-features = false`, and `pyre-jit` declares
`pyre-jit-trace = { workspace = true }`. A `pyre-jit` that names no backend —
wasm32, where the backend is chosen by `target_arch` — therefore forwards none,
and `pyre-jit-trace` arrives with neither feature where it used to inherit
`dynasm` from that manifest's own defaults. `cargo tree -e features` for the
failing command resolves `pyre-jit-trace` to **`prepass` alone**. `main` at
`84131dc4da8` is green on the same job, and the one commit it has that this
branch lacks touches no manifest and no `trace.rs`.

**The fix** is the gate `majit-metainterp` already puts on its *own*
`builder.cpu` assignment in `build_inline_call_only_bh_builder`. `trace.rs` held
the only ungated copy in the tree; it compiled until now because the workspace
default silently supplied a backend. Turning that off is the more correct state,
so the call site is what moves.

Verified locally with the exact failing command —
`cargo check -p pyre-wasm --target wasm32-unknown-unknown --no-default-features
--features web` — rc=0. (`Checking pyre-jit` succeeding is the proof that
`pyre-jit-trace`'s lib type-checked: cargo folds a build-script package's lib
under its one `Compiling` line, and `pyre-jit` cannot check without that
metadata.)

**My audit above was incomplete, and this is the gap it left.** The table in the
PR description checked every `--no-default-features` consumer for **`prepass`**
and concluded "every site is covered". `default-features = false` on the
workspace entry drops **both** of that manifest's defaults, and I never ran the
same table for `dynasm`. One axis of a two-axis switch is not an audit.

I also want to retract a claim I made while diagnosing this: I read
`Compiling majit-backend-dynasm` in `main`'s wasm-web step as proof that the old
graph pulled the native backend into the wasm32 target. It does not prove that —
`pyre-jit-trace`'s build script build-depends on `pyre-interpreter` with
`features = ["dynasm"]`, so that line is consistent with a host build-dependency,
and my own local wasm check prints it too. The diagnosis does not rest on it.

— *commented by Claude*

youknowone:
## Correction: my previous fix was the wrong shape (`e5e3f4f4e3e`)

`3017b55f561` made the wasm build compile and **that is all it did**. The run
then failed differently: **304 of 449 wasm fixtures panicked** on
`expect("cpu not set")` / `expect("blackhole cpu reference unset")`. Gating the
assignment away is not equivalent to not needing it — the wasm build genuinely
needs `bh.cpu`, and it had been getting one from the `dynasm` that
`pyre-jit-trace`'s defaults were supplying by accident.

So the E0425 was a symptom, and I treated it as the defect.

**The actual condition.** `pyjitpl` binds `BackendImpl` in three arms:

```rust
#[cfg(all(feature = "cranelift", not(target_arch = "wasm32")))]  … CraneliftBackend
#[cfg(all(feature = "dynasm", not(feature = "cranelift"), not(target_arch = "wasm32")))]  … DynasmBackend
#[cfg(target_arch = "wasm32")]  … WasmBackend
```

which is `any(target_arch = "wasm32", feature = "dynasm", feature = "cranelift")`.
`pyre_production_cpu` and `build_inline_call_only_bh_builder`'s own `builder.cpu`
assignment were both gated on the *narrower*
`any(feature = "dynasm", feature = "cranelift")`, so on wasm the backend existed
and the accessor for it did not. Those were the only two occurrences of the
narrow predicate in the tree; both now carry `BackendImpl`'s own condition, and
so does the call site in `trace.rs`.

This also explains why nobody hit it before: on wasm the narrow gate was
satisfied by the accidental `dynasm`, so widening it restores exactly the
behaviour `main` has — now for a reason rather than by inheritance.

`cargo check -p pyre-wasm --target wasm32-unknown-unknown --no-default-features
--features web` passes. Whether the 304 fixtures come back is for CI to say;
a compile check cannot answer it.

**The startup timer did get its first real ubuntu reading in that run**, since
the job reached the benches this time:

```
startup (empty-program user-CPU, median 5 in 1.6s, …): cpython=0.010s[0.005..0.010] pypy=0.013s[0.011..0.015] dynasm=0.098s[0.094..0.099] cranelift=0.094s[0.093..0.097] wasm=0.039s[0.034..0.042]
startup re-measured in 1.6s (…): … dynasm=0.096s[0.092..0.099] …
startup re-measured in 1.6s (…): … dynasm=0.098s[0.093..0.100] …
startup re-measured in 1.6s (…): … dynasm=0.098s[0.095..0.100] …
startup re-measured in 1.6s (…): … dynasm=0.098s[0.091..0.105] …
```

One initial pass plus **four refreshes** — the count `STARTUP_REFRESH_S = 120`
was sized for on ubuntu — at 1.6s per pass, which is the cost the line reports so
the interval can be re-sized from a measurement. dynasm reads
0.098/0.096/0.098/0.098/0.098 across the job, i.e. this run had no drift to
correct; the point is that the estimate is now never more than one interval older
than the bench it is subtracted from. `dynasm 457/457` and `cranelift 457/457`
passed.

— *commented by Claude*

youknowone:
## `e5e3f4f4e3e` verified locally — the full wasm population, not a compile check

I said a compile check could not answer whether the 304 panics come back. It
cannot, but `check.py` can, and it runs the wasm backend on a developer machine:

```
python3 pyre/check.py --backend wasm --synthetic-only          # 439 fixtures
python3 pyre/check.py --backend wasm --no-synthetic --no-build #  10 fixtures
```

```
ALL PASSED: wasm 439/439
ALL PASSED: wasm  10/10
```

449 total, **zero** `expect("cpu not set")` / `expect("blackhole cpu reference
unset")` panics — against 304 failures on the previous head, all of which were
panics inside `majit-metainterp` (298 print the cpu message outright; the other 6
have it truncated by the result column and share `blackhole.rs:10177`). 294 of
the 304 are `synth/`, which is exactly what the first command covers; the other
10 are what the second one covers. The populations match, so this is the whole
failing set and not a sample of it.

Caveats, so the number is read for what it is: this ran with
`PYRE_LLBC_SKIP_FINGERPRINT_CHECK=1` (the shared tree's `build/llbc` is stale),
so field offsets read out of the artefacts are not guaranteed current. That does
not weaken the verdict for *this* defect — `cpu not set` is an `Option` being
`None`, not a mis-read offset — but it does mean the run is not a substitute for
CI's gate overall. A wasm-only invocation also does not evaluate the wasm/dynasm
ratio gate, which needs a dynasm baseline in the same invocation.

— *commented by Claude*

PR #1443, opened 2026-08-23
gc: app-level gc module defects, and the comments that credit a parking gc_op

youknowone:
Nine commits, all in the GC area. Three are behaviour fixes with a gate snippet each, one is a rooting fix, the rest correct comments that state an invariant the tree no longer has.

## Behaviour

**`gc.is_finalized` was always `False`.** `flags::FINALIZER_RUN` is recorded by `begin_finalizer` on every `tp_finalize` branch — `__del__`, generator/coroutine, and the async-gen awaitables — not only `__del__`. The buffer-export branch does not record: it is `tp_dealloc`, not `tp_finalize`.

**`gc.set_threshold` dropped its third value.** `GC_THRESHOLD` is three slots and `set_threshold(1, 2, 3)` round-trips. Only the round trip: the collector still does not read these.

**`gc.collect` and `gc.get_objects` did not bound their generation.** `gc.get_threshold` answers a three-tuple, so this module reports three generations, but `gc.collect(3)`, `gc.collect(-1)` and `gc.collect(99)` all ran a full collection and answered 0 where `gc_collect_impl` raises `ValueError("invalid generation")`. `gc.get_objects` refused every non-`None` generation with a `NotImplementedError`, although `MiniMarkGC::do_get_objects` already filters by one — -1 is every object, 0 is the nursery, 2 is what is not in it, and 1 is the generation this collector keeps empty. The argument is passed through and bounded the way `gc_get_objects_impl` bounds it. The generations partition the heap exactly: `get_objects(0)` 641 + `get_objects(1)` 0 + `get_objects(2)` 2193 == `get_objects()` 2834 on a fresh interpreter.

**`gc.get_referents` omitted an object's heap type.** `subtype_traverse` visits it — "for a heaptype, the instances count as references to the type" — and a class names its metaclass on the same rule. `gc.get_referents(Plain())` was `['A']` against CPython's `['A', Plain]`, and an attribute-less instance answered `[]` against `[C]`. The collector correctly has no edge for `PyObject.ob_type` (it addresses a `try_gc_alloc_stable_raw` header nothing traces), so the edge is materialised at the API boundary next to `pin_unboxed_container_referents`, which exists for the same reason. A tuple subclass already arrives with its type among the collector's own edges, so the pinned referents are scanned first and the type is reported exactly once. A statically allocated type is still not reported, which is why an ordinary class does not name `type`.

**`gc.callbacks` was an empty list nothing calls.** Removed. `moduledef.py` defines none, and the collector-side hook contract keeps its calls allocation-free, so nothing here could run an app-level callback around a collection. `gc.hooks` is the notification surface that does fire.

**`register_module` held the `_xoptions` dict across a decode.** `fsdecode_os_str_wtf8` reaches `call_registered_decode_error_handler`, which calls app-level Python, so a decode is a collection point. The dict is nursery-allocated and moves; it now lives in a `push_roots` slot and each store reads it back. The value half is decoded after the name half, so the only movable object live across a decode is the rooted dict.

## Comments

**Twenty-one comments credited a parking `gc_op`.** It sits under the section headed "GC operations — unsynchronised under the GIL": a GIL assertion, a reentry guard and a bare borrow of the singleton, with no exit safepoint. Every ordering and every root is kept; only the stated reason changes — the transform's own `COND_CALL_GC_WB`/`SETFIELD_GC` order, the slot being what a walker reads and a promotion rewrites, old-gen buying immobility rather than survival, and `gc_op_with_root`'s forwarding coming from a collection driven inside the operation. The comments about real blocking waits, where the GIL is released and a mutator does park, are untouched: the import lock's `before_external_block`, `_queue`'s `Condvar::wait`, faulthandler's futex, and `gc_sync`'s own STW machinery.

**`DEFAULT_NURSERY_SIZE`'s doc** said majit does not implement the cache probe. `estimate_best_nursery_size` does; the constant is the fallback arm of `best_nursery_size_for_l2cache`.

**The two `gc` snippets asserted nothing** — both only printed, and `extra_tests/run.py` passes a snippet on its exit status, so they were vacuously green. Both now assert, and both were red-checked against a deliberately wrong expectation.

## Verification

- `extra_tests/run.py --gated-only`: 99/99 on cpython, dynasm and cranelift.
- Each new snippet passes unmodified on CPython 3.14.6, so it is a parity test rather than a pyre-shaped one, and fails on the pre-fix binary.
- `cargo fmt --all -- --check` and `scripts/check-new-line-citations.py` clean.

## Not in this PR

`young_rawmalloced_objects` / `GCFLAG_VISITED_RMY` is still unported, and it is measurable: an object over `LARGE_OBJECT_THRESHOLD` (135168 B) is born old-gen, and 10 of 10 such blocks are still alive after 400,000 small-object allocations have forced many minors over a 4 MB nursery. It is not a leak — `finish_alloc_in_oldgen` bumps `bytes_made_old_since_cycle`, so large allocations push the major trigger and a major does reclaim them — so the cost is peak RSS and major frequency. The port is invasive enough to want its own change: `oldgen.rs` owns the `RawMallocedObject` records that hold the `alloc_start`/`layout` a free needs, and the minor's trace paths, its tail, the major's precondition and the write barrier's youth predicate all move with it.

## Summary by CodeRabbit

* **New Features**
  * Garbage-collection APIs now report three supported generations and validate generation arguments consistently.
  * `gc.get_objects()` supports valid generation values.
  * `gc.get_referents()` includes heap-allocated type references.
  * Removed the obsolete `gc.callbacks` binding.

* **Bug Fixes**
  * Improved finalizer handling, including generator and async-generator finalization.
  * Strengthened garbage-collection rooting and object relocation safety across runtime operations.

* **Tests**
  * Added coverage for generation validation, audit events, heap-type referents, resurrection, and threshold behavior.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -1874,9 +1874,10 @@ fn get_objects_via_active_runtime(generation: i8, visitor: majit_gc::GetObjectsV
 }
 
 /// Split the same way [`gc_write_barrier_via_active_runtime`] is: the TLS path
-/// cannot park, so the raw address is still current there, while the `gc_op`
-/// fallback can — and a minor collection during that wait would forward `obj`,
-/// leaving the collector to answer about a forwarding stub.
+/// answers out of the box it already holds, so the raw address is still current
+/// there, while the `gc_op` fallback enters the collector — where a minor
+/// driven from inside the operation would forward `obj`, leaving the collector
+/// to answer about a forwarding stub.
 fn get_referents_via_active_runtime(obj: GcRef, visitor: majit_gc::GetObjectsVisitorFn) {
     let mut visit = visitor;
     if gc_box::present() {
--- a/majit/majit-backend-dynasm/src/runner.rs
+++ b/majit/majit-backend-dynasm/src/runner.rs
@@ -723,8 +723,8 @@ fn dynasm_get_referents(obj: majit_ir::GcRef, visitor: majit_gc::GetObjectsVisit
     if gc_box::with_mut(|g| g.get_referents(obj, &mut visit)).is_some() {
         return;
     }
-    // See `MajitGc::get_referents`: the fallback can park, so the argument is
-    // published and reloaded rather than passed as the raw local.
+    // See `MajitGc::get_referents`: the fallback enters the collector, so the
+    // argument is published and reloaded rather than passed as the raw local.
     majit_gc::gc_sync::gc_op_with_root(obj, |g, obj| g.get_referents(obj, &mut visit));
 }
 
--- a/majit/majit-gc/src/lib.rs
+++ b/majit/majit-gc/src/lib.rs
@@ -1321,11 +1321,11 @@ impl GcAllocator for GcHandle {
         gc_sync::gc_op(|gc| gc.get_objects(generation, visitor))
     }
     fn get_referents(&mut self, obj: GcRef, visitor: &mut dyn FnMut(GcRef)) {
-        // Rooted like `write_barrier`: a contended `gc_op` parks this mutator,
-        // so another thread's minor collection can forward `obj` while the call
-        // waits. Reading the raw argument afterwards would hand the collector a
-        // forwarding stub, which `do_get_referents` rejects — reporting a live
-        // object as having no referents at all.
+        // Rooted like `write_barrier`, for the reason `gc_op_with_root`
+        // states: a collection driven from inside the operation forwards `obj`,
+        // and the reload is what keeps the raw argument from reaching
+        // `do_get_referents` as a forwarding stub, which it rejects — reporting
+        // a live object as having no referents at all.
         gc_sync::gc_op_with_root(obj, |gc, obj| gc.get_referents(obj, visitor))
     }
     fn is_tracked(&mut self, obj: GcRef) -> bool {
--- a/majit/majit-gc/src/nursery.rs
+++ b/majit/majit-gc/src/nursery.rs
@@ -47,10 +47,11 @@ pub const TRANSLATION_NURSERY_SIZE: usize = 896 * 1024;
 
 /// env.py `NURSERY_SIZE_UNKNOWN_CACHE`.
 ///
-/// PyPy's translated incminimark normally estimates the nursery from cache
-/// size, but its documented fallback is 4MB.  majit does not yet implement
-/// the platform-specific cache probe, so `GcConfig::default` uses this after
-/// checking `PYPY_GC_NURSERY`.
+/// PyPy's translated incminimark estimates the nursery from cache size and
+/// falls back to 4MB.  `estimate_best_nursery_size` implements that probe, so
+/// this is the fallback arm of `best_nursery_size_for_l2cache`, taken when the
+/// reported L2 is 8MB or smaller — not the default `GcConfig::default` reaches
+/// on its own.
 pub const DEFAULT_NURSERY_SIZE: usize = 4 * 1024 * 1024;
 
 /// Nursery memory region with bump-pointer allocation.
--- a/pyre/bench/synth/gc_pypy_frontend.py
+++ b/pyre/bench/synth/gc_pypy_frontend.py
@@ -3,12 +3,15 @@ import gc
 import sys
 
 
-# The generation argument is integer-unwrapped but otherwise ignored: every
-# value is accepted and none of them selects anything.  What `collect` answers
-# is not asserted here -- it is a spec axis, so pyre answers an int where this
-# fixture's oracle answers None, and `extra_tests/snippets/stdlib_gc.py` is
-# where the return value is pinned.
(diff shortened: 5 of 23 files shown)


youknowone:
Local gate lanes on the pushed head:

- `extra_tests/run.py --gated-only` — 99/99 on cpython, dynasm and cranelift
- `extra_tests/parity_tests/run.py` — all parity tests pass on all three
- `cargo test -p majit-gc -p pyre-object` — 273 + 284 passed, 0 failed
- `cargo fmt --all -- --check` and `scripts/check-new-line-citations.py` clean

Each of the three new snippets also passes unmodified on CPython 3.14.6 and fails on the pre-fix binary, so they are parity tests rather than pyre-shaped ones.

— *commented by Claude*

youknowone:
Thanks — checking each finding against both upstreams. Two are the deliberate deviations this PR is about, one is refuted, and one pointed at a real gap that I have now fixed (though not for the stated reason).

CI on this PR was also red on `synth/gc_pypy_frontend` for exactly the reason §1.1/§1.2 name — the pypy-oracle fixture still asserted the frontend these commits replaced. Reconciled in 083bcebf8b3.

### §1.1 `gc.collect` bounds, §1.2 `gc.get_objects` generations — deliberate, filed

Both are `[3.14-spec]` deviations, adjudicated with AGENTS.md's six tests rather than taken by preference. Measured, all three:

| probe | CPython 3.14.6 | pypy3 7.3.22 | pyre |
|---|---|---|---|
| `collect(-999)` / `(-1)` / `(3)` / `(999)` | `ValueError: invalid generation` | accepted, `None` | `ValueError` |
| `get_objects(0)` / `(-1)` | works | `NotImplementedError` | works |
| `get_objects(1000)` / `(-1000)` | `ValueError` | `NotImplementedError` | `ValueError` |

The artefact for §1.2 (test 2) is in-tree: `lib-python/3/test/test_gc.py test_get_objects_arguments` asserts `ValueError` for `1000` and `-1000` and `TypeError` for `"1"` / `1.234`. It is **not** `@requires_gil_enabled`, unlike its `test_get_objects_generations` neighbour, so it holds on 3.14t too — test 6 is satisfied.

For §1.1 the artefact is the measured 3.14.6 run above. It is also the internally consistent choice: the module reports three generations (`get_threshold` answers a three-tuple, `set_threshold` writes three slots), so accepting a generation that `get_objects` rejects would leave the two entry points disagreeing about how many exist.

The audit payload deliberately does **not** depart: `referents.py get_objects` reports a constant `-1` and `gc_get_objects_impl` reports the argument, and nothing pins it — `lib-python/3/test/test_audit.py test_gc` compares event *names* only (`[event[0] for event in events]`). So PyPy's shape stands there, and per-axis is the intended outcome of test 5, not an oversight.

### §1.3 `gc.get_referents` heap type — refuted, the artefact is in this diff

`pyre/extra_tests/snippets/gc_get_referents_reports_a_heap_type.py` is in the reviewed file list and **passes unmodified on CPython 3.14.6**, which is an admissible artefact under test 2 ("a measured run at the pinned version"). The edge is `subtype_traverse`'s — *"for a heaptype, the instances count as references to the type"* — and the snippet also pins the negative (`type not in gc.get_referents(Plain)`, since a statically allocated type is not visited) and that it is reported exactly once.

Test 4 is a clean negative and worth recording: `pypy/module/gc/referents.py` contains **no** `@jit.*`, `_immutable_*` or `rgc` hint anywhere in the file, and `_list_w_obj_referents` carries no decorator. Nothing about PyPy's shape is load-bearing here.

### §2 `is_finalized` — citation accurate, conclusion does not follow, but it found a real gap

`@jit.dont_look_inside` does decorate `UserDelAction._run_finalizers`. It does not support refusing the change:

- The hint governs **tracing of the function**, not the finalizer-run bit. Test 4 asks whether a hint governs *the value being changed*; PyPy has no such value — as §3 itself notes, PyPy's `gc` exports no `is_finalized` at all.
- `dont_look_inside` makes the region opaque, and the write is strictly inside it. A hint that hides a body cannot be the reason to refuse a write in that body.
- The artefact is in-tree: `lib-python/3/test/test_gc.py test_is_finalized` asserts `gc.is_finalized(lazarus)` is True after `__del__` resurrects it.

**But the finding was worth raising**, because pyre's `_run_finalizers` did not carry the hint at all — the port dropped it, while its own doc comment quoted the upstream line range that contains it. Now added.

It is inert for codegen, which is why it went unnoticed: `look_inside_graph` already rejects the graph because it loops (`res = res && !contains_loop`), and `unsafe_loopy_graphs` has only diagnostic consumers. So the opacity was holding by a property of the loop rather than by a decision about the body — the hint states the decision.

### §3 `gc.is_finalized` public surface — agreed, and it is CPython parity

Agreed this is pre-existing and not introduced here. Worth noting it is not an unanchored extension: `gc.is_finalized` is a CPython API and `lib-python/3/test/test_gc.py test_is_finalized` is the in-tree assertion it answers.

— *commented by Claude*

PR #1445, opened 2026-08-23
Fix BlockingIOError 3rd argument handling via PyNumber_Check

### LINKED ISSUE #1150: BlockingIOError with an oversized third argument matches neither CPython nor PyPy
## Summary

`BlockingIOError(errno, strerror, n)` where `n` is a number that does not fit an
index-sized integer produces a state neither reference runtime produces: pyre
stores the value as `filename` (PyPy's answer) while truncating `args` to two
elements (CPython's answer for a *filename* third argument).

## Measurement

Same script on all three, 3.14 / 7.3.20 / pyre-dynasm at `ea731d7`:

| expression | CPython 3.14 | PyPy 7.3.20 | pyre |
|---|---|---|---|
| `BlockingIOError(1,'x',3)` | `written=3`, `args=(1,'x',3)` | same | same |
| `BlockingIOError(1,'x',2**70)` | **`ValueError: cannot fit 'int' into an index-sized integer`** | `filename=2**70`, `args=(1,'x',2**70)` | `filename=2**70`, **`args=(1,'x')`** |
| `BlockingIOError(1,'x',Idx())` where `Idx.__index__` returns `2**70` | **`ValueError: cannot fit 'Idx' into an index-sized integer`** | `filename=<Idx>`, `args=(1,'x',<Idx>)` | `filename=<Idx>`, **`args=(1,'x')`** |
| `BlockingIOError(1,'x','f')` | `filename='f'`, `args=(1,'x')` | `filename='f'`, `args=(1,'x','f')` | `filename='f'`, `args=(1,'x')` |
| `BlockingIOError(1,'x','f',5)` | `filename='f'`, `args=(1,'x')` | `args=(1,'x','f',5)` | `filename='f'`, `args=(1,'x')` |

pyre already follows CPython rather than PyPy on the two string rows, so the
oversized-number rows are the only ones where it agrees with neither.

## Cause

`exceptions.c:oserror_init` routes the third argument by `PyNumber_Check`, not
by whether the conversion succeeds:

```c
if (Py_IS_TYPE(self, (PyTypeObject *) PyExc_BlockingIOError) &&
    PyNumber_Check(filename)) {
    self->written = PyNumber_AsSsize_t(filename, PyExc_ValueError);
    if (self->written == -1 && PyErr_Occurred())
        return -1;          /* the ValueError propagates */
}
else {
    self->filename = Py_NewRef(filename);
}
```

Once the argument answers `PyNumber_Check`, it is a `written` count and an
overflow is an error — there is no fallback to the filename branch. pyre takes
the fallback, so the value lands in `filename` while the `args` slice that the
`written` branch performs has already happened.

`Idx()` is in the table because `PyNumber_Check` is true for anything with
`nb_index`, so a plain object with an oversized `__index__` reaches the same
path.

## Notes

Not fixed in-place because `pyre/pyre-interpreter/src/builtins.rs` has
uncommitted work in this worktree; filing so the finding is not lost.

— *commented by Claude*

sigmaith:
## Summary

In the `OSError` family the third constructor argument is the filename. For `BlockingIOError` it is `characters_written` instead — the byte count a partial write got through before it blocked. One slot, two meanings, so the constructor has to decide which one a given value is.

CPython decides in two steps (`Objects/exceptions.c oserror_init`):

1. `PyNumber_Check` picks the branch.
2. Only then is the value converted, and a conversion that fails is an error. It propagates; there is no falling back to the filename branch.

pyre collapsed the two steps into one — it ran the conversion and used its success or failure as the branch test. So a failing conversion quietly turned into "then it must be a filename".

PyPy does fall back that way (`interp_exceptions.py:637-643`), which is fine for PyPy: it keeps the full `args` tuple on both branches. pyre trims `args` on the filename branch, like CPython. Falling back after the trim has already happened produces a state **neither** upstream produces — the value stored as `filename`, but `args` already cut short.

The issue reports the two overflow rows; measuring the whole third-argument space turned up four more.

### Measured

CPython 3.14.6 (`lib-python/stdlib-version.txt` pins `v3.14.6`) vs `pyre-dynasm`. Only the divergent rows are listed; `3` / `True` / `__index__`→5 / `'f'` / `b'f'` / `None` already matched and still do.

| third argument | CPython 3.14.6 | pyre before |
|---|---|---|
| `2**70` | `ValueError: cannot fit 'int' into an index-sized integer` | `filename=2**70`, `args=(1,'x')` |
| `__index__`→`2**70` | same, naming `'Idx'` | `filename=<Idx>` |
| `1.5` | `TypeError: 'float' object cannot be interpreted as an integer` | `filename=1.5` |
| `1j` | `TypeError: 'complex' object …` | `filename=1j` |
| `__int__`-only | `TypeError: 'N' object …` | `written=7` — accepted where CPython rejects |
| `__float__`-only | `TypeError: 'OnlyFloat' object …` | filename |

All six match after the patch. The overflow message names the **original** operand's type, not the `__index__` result's.

## What changed

- **`builtins.rs` `os_error_fill_slots`** — the branch is chosen by `number_check`; the conversion runs through `getindex_w_written`, whose error propagates. The filename arm is untouched.
- **`baseobjspace.rs`** — two helpers beside the precedent they follow: `number_check` (`abstract.c PyNumber_Check`) and `getindex_w_written` (`PyNumber_AsSsize_t` with `ValueError`, same shape as the adjacent `getindex_w_index`). `number_check` is deliberately wider than what the conversion accepts — that width is what sends an `__int__`-only operand into the numeric branch to fail there.
- **`os_error_fill_slots` returns `Result`** so the failure can leave the constructor (`oserror_init`'s `return -1`). Both callers already returned `Result`, so each needed one `?`.
- **`cpyext/number.rs`** — it carried its own copy of the predicate, already drifted: the complex clause was missing, so `PyNumber_Check(1j)` answered 0 where CPython answers 1. It now calls the shared `number_check`.

## Testing

- The 14 third-argument shapes above, plus OSError-family controls (`OSError(2,'No such file','/tmp/x')`, the zero- and one-argument forms, `FileNotFoundError`), diffed against CPython 3.14.6 — no differences.
- `cpython_tests/run.py --filter test_exceptions` — PASS, no regressions.
- `cargo test -p pyre-interpreter --features dynasm` — 376 passed, 0 failed.

No JIT change, so the benchmark suite was not re-measured.

Closes #1150

## Self-review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recognition of numeric objects, including objects supporting standard numeric conversion methods.
  - Improved handling of oversized index values by reporting a clear `ValueError` that identifies the original operand type.
  - Correctly propagates errors when recording the number of characters written during operating system error handling.
  - Ensured consistent numeric checks across supported interfaces.

### DIFF
--- a/pyre/pyre-interpreter/src/baseobjspace.rs
+++ b/pyre/pyre-interpreter/src/baseobjspace.rs
@@ -4211,6 +4211,41 @@ pub(crate) unsafe fn getindex_w_index(index: PyObjectRef) -> Result<i64, PyError
     }
 }
 
+/// `abstract.c PyNumber_Check` — `nb_index` / `nb_int` / `nb_float`, or a
+/// complex.  Looked up on the type, so an instance attribute does not answer.
+/// Deliberately wider than what [`getindex_w_written`] accepts: an
+/// `__int__`-only operand has to enter the numeric branch and fail there.
+pub(crate) unsafe fn number_check(w_obj: PyObjectRef) -> bool {
+    if w_obj.is_null() {
+        return false;
+    }
+    unsafe {
+        pyre_object::pyobject::is_int_or_long(w_obj)
+            || pyre_object::is_float(w_obj)
+            || pyre_object::pyobject::is_complex(w_obj)
+            || lookup(w_obj, "__index__").is_some()
+            || lookup(w_obj, "__int__").is_some()
+            || lookup(w_obj, "__float__").is_some()
+    }
+}
+
+/// `PyNumber_AsSsize_t(w_obj, PyExc_ValueError)` — [`getindex_w_index`] with
+/// the exception `exceptions.c oserror_init` passes.  The overflow names the
+/// ORIGINAL operand's type, not the `__index__` result's.
+pub(crate) unsafe fn getindex_w_written(w_obj: PyObjectRef) -> Result<i64, PyError> {
+    match int_w(space_index(w_obj)?) {
+        Ok(i) => Ok(i),
+        Err(e) if e.kind == PyErrorKind::OverflowError => Err(PyError::new(
+            PyErrorKind::ValueError,
+            format!(
+                "cannot fit '{}' into an index-sized integer",
+                object_functionstr_type_name(w_obj)
+            ),
+        )),
+        Err(e) => Err(e),
+    }
+}
+
 #[inline(never)]
 unsafe fn setitem_bytearray(obj: PyObjectRef, index: PyObjectRef, value: PyObjectRef) -> PyResult {
     let mut value = value;
--- a/pyre/pyre-interpreter/src/builtins.rs
+++ b/pyre/pyre-interpreter/src/builtins.rs
@@ -6794,11 +6794,14 @@ fn os_error_parsed_errno(args: &[PyObjectRef]) -> Option<PyObjectRef> {
 /// filename from `args_w` per `_init_error` (line 652).  Must run after the
 /// `w_class` retag so the `BlockingIOError` numeric-filename special-case can
 /// see the resolved class.
-fn os_error_fill_slots(exc: PyObjectRef, args: &[PyObjectRef]) {
+/// `Err` when the `characters_written` conversion fails — `oserror_init`
+/// propagates it out of the constructor (`return -1`).
+fn os_error_fill_slots(exc: PyObjectRef, args: &[PyObjectRef]) -> Result<(), crate::PyError> {
     use pyre_object::interp_exceptions;
-    // Both the errno parse and the `BlockingIOError` test below run `int_w`,
-    // which is a user `__index__`, so the arguments are read off the root
-    // stack rather than out of the caller's slice once either has run.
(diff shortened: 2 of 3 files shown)


pyre/pyre-interpreter/src/baseobjspace.rs
```diff
@@ -4211,6 +4211,41 @@ pub(crate) unsafe fn getindex_w_index(index: PyObjectRef) -> Result<i64, PyError
     }
 }
 
+/// `abstract.c PyNumber_Check` — `nb_index` / `nb_int` / `nb_float`, or a
+/// complex.  Looked up on the type, so an instance attribute does not answer.
+/// Deliberately wider than what [`getindex_w_written`] accepts: an
+/// `__int__`-only operand has to enter the numeric branch and fail there.
+pub(crate) unsafe fn number_check(w_obj: PyObjectRef) -> bool {
```
youknowone, line 4218: 👍

youknowone (approved):
Thank you so much!

PR #1452, opened 2026-08-23
interp, _io, collections, setobject, listobject, jit-trace: method-cache lookups, O(1) set delete, and a class-attribute fold

youknowone:
Independent slices, each audited against the vendored `pypy/` / `rpython/` tree before landing.

### `interp`: route the `lookup_where` callers through the method cache
typeobject.py `lookup_where` **is** `lookup_where_with_method_cache`, and `lookup` is its value half; the uncached `_lookup_where` is reached only when `version_tag is None`. The `__abs__` / `__float__` / `__round__` resolutions and the itertools `__iter__` / `__next__` override probes went through `lookup_where` / `lookup_where_pair` instead — both walk the MRO uncached, the latter rebuilding the pair from two single-register `dont_look_inside` projections of that walk.

The comment on `builtin_abs_obj` justified the uncached walk by the method cache "projecting only the value half". It does not: the cache holds `(w_class, w_value)` pairs, which `walk_method_cache_gc` forwards.

### `setobject`: delete a set member in O(1)
`shift_remove_index` pulls the whole tail down on every delete, so discarding the first of *n* keys *n* times is quadratic: 200000 `s.discard` took **162.9s** against pypy's 0.008s. `rordereddict.py _ll_dict_del_entry` marks the slot dummy and is O(1); `IndexMap` has no dummy, and `swap_remove_index` is the O(1) primitive it does have. It reorders the remaining members, which set iteration does not specify. Porting the dummy is the rordereddict convergence.

### `jit-trace`: fold a class attribute that `object.__getattribute__` returns unchanged
The instance-slot fold needs a mapdict storage index, so a name living only on the type had none and residualized `space.getattr` whole. This arm is the `space.get(w_descr, w_obj)` tail of `Object.descr__getattribute__`, where an absent `__get__` returns the value unchanged.

`class_attr_fast_path` accepts the receiver only when the type does not override `__getattribute__`, the value is neither a data descriptor nor carries `__get__`, its own type is not a heaptype, and `find_map_attr` proves this instance does not shadow the name. A devolved map declines. The value is baked under the pins `walker_guard_mapdict_instance_shape` already emits: the layout `GuardClass`, the exact `w_class`, the type `version_tag`, and a `guard_value` on the map. A rebind goes through `mutated`, which recurses into subclasses, so a base-class rebind revokes a derived instance's fold too.

The `can_move` gate is load-bearing, not conservative. A **recorded** `ConstPtr` is forwarded (`remove_constptrs_in` → `LoadFromGcTable` at emit, `gcreftracer` at run), but `write_residual_call_result_to_dst` first parks the `OpRef` in the walker's `registers_r`, and none of the mutator extra areas registered in `pyre-jit/src/eval.rs` walks that bank. The ungated folds are not counter-examples: `try_walker_specialize_load_type_attr` bakes a type object and `emit_module_dict_cell_fold` a `malloc_typed` cell, neither of which moves, while a class attribute is an arbitrary object.

### `_io`
- **`BytesIO` gets its own `flush`.** `W_BytesIO.typedef` has no `flush`, so it inherits `W_IOBase.flush_w`, which raises only when `_CLOSED()`. But `W_BytesIO.close_w` overrides `W_IOBase.close_w` **without** setting `__IOBase_closed`, so a closed BytesIO flushed without raising. `_io.BytesIO` on the pinned 3.14 carries `flush` in its own type dict and raises `ValueError` once closed; `stdlib_io_bytesio.py` asserts both the dict key and the raise.
- **Abstract base types are qualified with their module** (`_IOBase` → `_io._IOBase`, and the Raw/Buffered/Text siblings).
- **An omitted `readline` limit is distinguished from an explicit `None`.** `W_TextIOWrapper.readline_w` keeps the interp2app default as the untranslated null sentinel and rejects only an explicit `None`; the default becomes `PY_NULL` and `next_w` preserves the missing sentinel rather than passing `None`.

### `collections`: initialize the `defaultdict` factory through its slot
`app_defaultdict.defaultdict.__init__` writes through the slot descriptor (`defaultdict.default_factory.__set__`) so a subclass `__setattr__` cannot intercept initialization.

### `listobject`: cite `is_plain_int1`, cover the bigint that does not fit
`is_int` and `is_long` are `py_type_check` against distinct payload types, so the two arms cannot both hold. The new test pins the `_fits_int()` conjunct: 2\*\*63 and 2\*\*70 are rejected, and a list holding one does not take integer storage.

### `extra_tests`: the int identity unaries answer a plain int
intobject.py builds `descr_trunc` / `descr_floor` / `descr_ceil` from `_self_unaryop`, whose body is `self.int(space)`. `W_IntObject.int` returns `self` only when `type(self) is W_IntObject` and otherwise `space.newint`; `W_BoolObject.int` is always `space.newint`.

### `docs`: make PyPy module presence the porting boundary
`AGENTS.md` states up front that module presence follows PyPy: verify the `pypy3` oracle imports a module and identify its owner before implementing it, and a skipped CPython test does not override that boundary.

---

### Withdrawn during review

A `s.add(x)` / `s.discard(x)` bound-method fold was on this branch and **has been removed**. The arm applied `set_add_value` during the walk, because the walker records IR without executing it and the traced iteration would otherwise drop its element. The `list.append` arm it copied that from applies a plain memory write; a set store hashes and compares, so the apply is safe only when neither can reach the interpreter and the receiver is not virtual. Neither holds in general — `synth/dict_set_key_eq_operand_order` adds keys whose `__hash__` / `__eq__` are Python methods, and `synth/set_hash_protocol` adds to a set allocated inside the loop. Both abort five times, after which the abort ceiling bans the green key; cranelift went 23 red on macos and windows while dynasm stayed green, because `we_are_jitted()` has its only arming site in the cranelift compiler. Restricting the element to an exact `int` cleared 16 of 23 and left the virtual-receiver seven — one symptom, not the premise.

The `discard` arm went with it: the `discard` cost was the quadratic `shift_remove_index`, fixed in `setobject` above, and with that gone the arm bought nothing.

### Known gap

The class-attribute fold has **no corpus fixture**. No `.jitstats` file moved across 463 fixtures on either backend, and the three nearby fixtures cover adjacent receiver shapes only — `attr_cache_invalidation` reads an instance slot, `attr_instance_shadows_class` exercises the decline path, `class_attrs_methods` uses a type receiver. A fixture for the instance-receiver / no-shadow / hot-read shape, gated with `spec-folds` rather than a ratio ceiling, is the follow-up.

https://claude.ai/code/session_012H7KYToch6UHjdJHWdhE9i


youknowone:
Three more commits, and the measurements the earlier description deliberately did not claim.

### What landed

- `b6c67bb0719` **setobject: O(1) delete.** `shift_remove_index` pulls the whole tail down per delete, so discarding the first of n keys n times is quadratic. Reverting the walker fold left the number unchanged, which is what ruled the fold out as the cause. `swap_remove_index` is the O(1) primitive `IndexMap` has; it reorders remaining members, which set iteration does not specify. Porting `_ll_dict_del_entry`'s dummy is the rordereddict convergence.
- `e718166b335` **jit-trace: class-attribute LOAD_ATTR fold.** A name that lives only on the type has no mapdict storage index, so it residualized `space.getattr` whole.
- `1182e3ad0ee` **jit-trace: drop the `set.discard` arm.** Its premise was that `discard` carried the same two-`CallMayForceR` tax as `add`; the cost was the quadratic delete underneath. With that fixed the arm buys nothing.

### Measured

Load-independent witnesses, both engines, identical results — this is the part worth trusting:

| witness | pyre | pypy3 7.3.20 |
|---|---|---|
| 200000 unique `s.add` → `len` | **200000** | 200000 |
| instance shadows the class attr mid-loop | 100000 | 100000 |
| class attr rebound mid-loop | 100000 | 100000 |
| **base**-class attr rebound mid-loop | 100000 | 100000 |

The `len` row is the point of the `set.add` fix: it was 199999, one element recorded but never applied. The base-class row is the guard case worth calling out — `mutated` recurses into subclasses, so a base rebind revokes a derived instance's fold.

Timings, interleaved best-of-7 so contention hits both engines:

| loop | before | after | pypy3 | ratio |
|---|---|---|---|---|
| `s.discard` × 200000 | 162.9s | **0.032s** | 0.0023s | 14x |
| `a.x` × 2000000 | 0.49s | **0.0056s** | 0.0018s | 3.1x |
| `s.add` × 200000 unique | — | **0.032s** | 0.0047s | 6.9x |

**Caveat on those timings.** The box was at load ~60 from concurrent builds and pypy3's own baseline inflated ~2.2x against its quiet-machine value, so the absolute seconds are soft and the ratios are approximate. The two order-of-magnitude collapses (162.9s → 0.032s, 0.49s → 0.0056s) are far outside that noise; the residual 3–14x gaps to pypy are directionally real but should be re-measured on a quiet box before anyone plans against them. These loops are **not** at parity with pypy — they are no longer pathological.

### Verification

- `check.py --backend dynasm`: **ALL PASSED 462/462**, on a freshly re-extracted LLBC (`pyre-jit` was stale and the build gate caught it) and a fresh release build.
- `cargo check --workspace --all-targets --features dynasm`: clean.
- New unit tests `discard_from_the_front_keeps_later_members` and `is_plain_int1_rejects_a_bigint_that_does_not_fit_i64` pass.

— *commented by Claude*

youknowone:
### CI caught a regression in my own `set.add` arm — reverted in `5893fb7105b`

**What CI showed:** `ALL PASSED: dynasm 463/463` but `FAILED: cranelift 23 failed, 440 passed`, identically on macos-latest and windows-latest.

**Attribution took care.** PR #1454 — the main tip this branch sits on — also shows a failing `check.py` job, which made this look inherited. It is not: #1454's log prints `ALL PASSED: cranelift 463/463`, and its job-level failure comes from the later pypy cross-check lane. Main's cranelift is green, so the 23 were ours.

**Isolated by A/B, not by argument.** `specialize.rs` and `residual_call.rs` are outside the LLBC closure, so one build plus two env knobs gave all four configurations instead of a re-extract per bisect step:

| config | result |
|---|---|
| both arms ON | FAILED |
| class-attr fold OFF | FAILED |
| **set.add arm OFF** | **ALL PASSED** |
| both OFF | ALL PASSED |

The class-attribute fold is exonerated and stays.

**Root cause.** The walker records IR without executing it, so the arm applied `set_add_value` during the walk — otherwise the traced iteration drops its element (measured: 199999 of 200000). `list.append` does the same thing safely because its apply is a plain memory write. A set store is not:

- `synth/dict_set_key_eq_operand_order` hot-loops `s.add(Recorded(i, log))` where `__hash__`/`__eq__` are **Python methods**, so the apply ran user code from inside the walk;
- `synth/set_hash_protocol` warms with `s = set(); s.add(i % 8)`, so the receiver is **fresh per iteration** and the `may_force` residual forces a set the trace would have kept virtual.

Both abort five times, after which the ceiling bans the green key — `abort_ceiling_banned=1`, `abort_ceiling_refused` in the tens of thousands, and every `abrt_*` counter at 0. That is the `loops_aborted 0 -> 5, loops_compiled N -> N-1` signature on nearly every failing fixture.

**Why only cranelift:** `we_are_jitted()` has exactly one arming site repo-wide, `JittedGuard::enter` in the cranelift compiler. On dynasm it is a constant `false`. My local verification ran dynasm only, which is how this shipped.

**Why revert rather than narrow.** Restricting the element to an exact `int` cleared 16 of 23 and left the seven virtual-receiver cases — it addresses one of two broken premises. Re-landing needs both: builtin hash/eq **and** an already-escaped receiver, validated on both backends over the full corpus. `set.add` returns to the generic residual; the `setobject` O(1) delete and the class-attribute fold are unaffected and keep their wins.

**Verified locally after the revert:** `check.py --backend cranelift` **463/463** and `--backend dynasm` **463/463**, on freshly re-extracted LLBC (`pyre-jit`'s closure carries `pyre-interpreter` sources, so it needed extracting too).

— *commented by Claude*

PR #1456, opened 2026-08-23
cpyext: load an extension as its header describes it; and the stdlib gaps cffi's and trio's own suites found

youknowone:
Thirty-eight commits. The pip work this branch started with has landed; what is
left is the layer under it — loading real C extensions, and the stdlib gaps two
third-party test suites found when they were run against pyre as-is.

The two suites are the measure here, because a hand-written driver only probes
what its author already suspects:

| suite | before | after |
|---|---|---|
| cffi 2.1.1, its own pytest suite (21 files, ~1190 cases) | `test_new_ffi_1` all-error, `test_recompiler` SIGABRT + 14 failed, `test_function` 13 failed | **2 failed**, both `test_char16_t`, neither a cffi defect |
| trio, its own suite (34 files) | 6 files with pyre-owned failures | **0 pyre-owned failures** |

Every behaviour row below was measured against a CPython 3.14.6 control on the
same host with a scripted probe run under both interpreters and the two outputs
diffed; each is byte-identical after the change unless the row says otherwise.

# cpyext — loading an extension the way the header describes it

- **PEP 793 export slots.** A module whose init returns a `PyModuleDef_Slot`
  array was not built at all; `Py_mod_exec` / `Py_mod_create` are now walked.
- **Entry points.** An extension declares its own `PyInit_*` / `_cffi_pypyinit_*`
  names; pyre now exports what the extension declares instead of guessing one
  spelling. api-mode cffi modules export `_cffi_pypyinit_` because pyre defines
  `PYPY_VERSION` — that is the reason the whole `test_new_ffi_1` file errored.
- **Module definitions read at the wrong offsets are now refused** instead of
  producing a module built from garbage.
- **The names an extension imports when it never reads pyre's headers.**
  PyO3 and Cython declare the C ABI themselves, so their objects import
  `Py_TYPE`, `Py_IS_TYPE`, `_Py_IncRef` / `_Py_DecRef`, the `PyList_GET_*`
  accessors and the rest as *symbols* where pyre had published only macros.
  A wheel pip had just built failed to `dlopen`, naming no symbol.
- **`METH_CLASS` rows get the class, `METH_STATIC` rows get no receiver.**
- A failed `dlopen` reports **what** it failed with, not that it failed.

# compiler / `_ast`

- rustpython pin moved to `48a3a1f`; its columns are read as characters.
- A tree compiled from `ast` objects carries a source, so positions index into
  something.

# Exception construction

`make_exc_type` always installs the `__new__` it is passed, so a module-defined
`OSError` subclass built through the base `Exception.__new__` never parses
`(errno, strerror)`. `_socket`, `_ssl` and `_io` now build theirs through the
family constructor, and `termios.error` derives from `Exception` as
`Modules/termios.c` has it.

# `_socket`

| | before | after | CPython 3.14.6 |
|---|---|---|---|
| any op on a closed socket | `OSError('Bad file descriptor')`, `errno=None` | `OSError(9, 'Bad file descriptor')` | same |
| `("localhost", 80)` on `AF_INET6` | `OSError('invalid IPv6 address: localhost')` | resolves | resolves |
| a bogus name, either family | `OSError('name or service not known: …')`, `errno=None` | `gaierror(8, 'nodename nor servname provided, …')` | same |
| `EAI_*` module constants | absent | the set `netdb.h` names | same |
| `gethostbyname("")` / `("<broadcast>")` | `gaierror` | `0.0.0.0` / `255.255.255.255` | same |
| `connect(("<broadcast>", 9))` | `gaierror` | connects | connects |
| `gethostbyaddr` of an unresolvable name | `herror`, no errno | `gaierror(8, …)` | same |
| `gethostbyaddr` of an IP with no PTR | `herror`, no errno | `herror(1, 'Unknown host')` | same |
| `htons(70000)` | `28689` | `OverflowError` | `OverflowError` |
| `ntohs(-1)` | `65535` | `ValueError` | `ValueError` |
| `getservbyport(70000)` | looked up port 4464 | `OverflowError` | `OverflowError` |
| `inet_pton(1234, …)` | message-only `OSError` | `OSError(47, …)` | same |
| `bind(("127.0.0.1", 70000))` | `port must be 0-65535` | `bind(): port must be 0-65535.` | same |
| `getnameinfo(("127.0.0.1",), 0)` | `ValueError` | `TypeError` | same |

`errno=None` is invisible to every caller that classifies by `e.errno` — trio
does `if exc.errno in _closed_stream_errnos` in two places, and three
`test_highlevel_socket.py` tests failed on the wrong exception type because of
it.

Also: `recvmsg_into` with two buffers panicked. A `Vec` of scope-owning root
guards drops front-to-back, so element 0's truncate killed element 1's roots;
the vector now holds one scope.

# `signal`, `select`, `termios`, `fcntl`

These four are one chain — `test_ki_in_repl` alone exposed all of it, each fix
uncovering the next.

- **`signal`** hands the app-level handler the frame it is called for. It was
  passing `None`.
- **`select`** publishes the 18 `KQ_NOTE_*` / internal `EV_` names it omitted,
  and `kevent()` binds its six keywords. The surface is now byte-identical to
  the control's 33 names.
- **`termios`** published **77 of the 196** integer names darwin defines. Now
  196/196, 0 extra, 0 mismatched — 94 read from `libc`, 25 spelled out with the
  value `<sys/termios.h>` / `<sys/ttydefaults.h>` / `<sys/ttycom.h>` gives.
  Graded by a three-way diff: names against CPython, values against a compiled
  `#ifdef` probe over the headers.
- **`termios`** built every message with `format!("<syscall>: {e}")` over a
  `std::io::Error`, so a failure read `(19, 'tcgetattr: Operation not supported
  by device (os error 19)')`. `wrap_oserror` and `PyErr_SetFromErrno` both spell
  it with `strerror` alone.
- **`select.kevent`** had no `__repr__` at all, so the six fields a caller set
  were invisible in any log line that printed one.
- **`fcntl`** took only an integer third argument — `fcntl.ioctl(fd, req, buf)`
  raised `TypeError: ioctl() arguments must be integers`. Both functions now
  take the buffer argument they are documented to take, including the
  `mutate_flag` dispatch and the 8-byte guard canary `fcntlmodule.c` writes
  after the staged argument, so a 1-byte overrun raises
  `SystemError("buffer overflow")` rather than corrupting the stack. Five
  probes, ~40 cases, all identical to the control.

# Module exceptions

`repr` read the whole registered name, so seventeen classes spelled their
module twice — `termios.error(19, 'boom')` where `_PyType_Name` gives
`error(19, 'boom')`. It now reads the accessor the `__name__` getter reads, so
the two channels cannot disagree. Three were also registered under the wrong
name: `_ssl`'s seven are `ssl.SSLError` and friends, `_locale.Error` is
`locale.Error`, and pyexpat's is `xml.parsers.expat.ExpatError`.

`__name__`, `__qualname__`, `__module__`, `str(cls)` and `repr(instance)` are
identical to the control for all nineteen classes probed.

# Assorted

- A std stream write that got a non-`str` raises `TypeError` instead of
  silently dropping the bytes.
- `tuple` / `frozenset` unbound `__hash__` runs the base implementation from the
  unbound descriptor, not the subclass override.

# Constant surfaces

A three-way diff of the integer surface of 17 modules — names against CPython
3.14.6, values against the SDK headers — found **110 names absent and one wrong
value**. Added to `_socket` (55), `mmap` (14), `fcntl` (12), `posix` (10),
`syslog` (8), `time` (4) and `termios` (1); `socket.IP_MAX_MEMBERSHIPS` carried
linux's 20 where `<netinet/in.h>` says 4095 on darwin. `mmap.PROT_NONE` is
published by neither `mmapmodule.c` nor pypy's `mmap`, so it is dropped rather
than added.

`_locale` published `CODESET` alone, so 54 of the 55 keys
`_localemodule.c langinfo_constants` names could not be spelled even though
`nl_langinfo` accepted them. Those come from `libc`, which carries each
platform's own numbering.

**`signal.NSIG` was 64 on every unix.** `<signal.h>` says 32 on darwin and
glibc's `_NSIG` is 65, so it is now read per platform, keeping 64 where the
header is silent as `signals.c` does. 65 exceeds a word, so `pypysig_flags_bits`
is ported as the `N_LONGSIG`-element array it is upstream. `valid_signals`
passed a literal 64 to its `sigismember` sweep and reported a signal with no
name. `SIGEMT`, `SIGINFO` and `SIGIOT` are added.

**`_sre.MAXGROUPS` published `2**31 - 1` where 3.14 answers `2**30 - 1`.**
`sre.h` halves `INT32_MAX`; the vendored engine crate halved its own unsigned
`MAXREPEAT` and landed one bit wide. `re/_parser.py` bounds a pattern's group
count and two group references on the number.

**`ast.PyCF_TYPE_COMMENTS` was `0x40000000`** — the number `consts.py` uses
because it keeps `PyCF_ASYNC_HACKS` on `0x1000`. `Include/cpython/compile.h`
numbers it `0x1000`, and 3.14 has no `PyCF_ASYNC_HACKS`. **`_symtable.DEF_BOUND`**
carried `DEF_TYPE_PARAM` beside the three bits `pycore_symtable.h` names.

# `_ast` — type comments are collected, and refused where the grammar has no rule

`ast.parse(src, type_comments=True)` was accepted and returned a tree with every
`type_comment` `None` and `type_ignores` `[]`. No parser change was needed: the
parse result already carries its token list and `convert.rs` was discarding it
with `.into_syntax()`, while all five node types already had an unfilled
`runtime_type_comment` field.

`lexer.c`'s three rules are ported — the `"# type: "` prefix where each space
matches a run of spaces **and tabs**, the TYPE_IGNORE split on `ignore` followed
by end-of-comment or an ASCII non-alphanumeric byte, and the newline an ignore
alone on its line takes into its tag — and attachment walks the five spans the
grammar accepts a `TYPE_COMMENT` in.

Collecting is only half the flag. A `TYPE_COMMENT` is a **token**, so a rule
that does not accept one leaves the parser looking at a token it cannot shift:
`x += 1  # type: int`, `x: int = 1  # type: int` and `f()  # type: int` are all
`SyntaxError` in CPython, and pyre parsed all three. What attachment leaves over
is now that refusal — `invalid syntax`, spanning the type text rather than the
`#`, at the first such comment in source order, with the offset counting
characters as `_PyPegen_byte_offset_to_character_offset` does.

Writing the refusal exposed a **live attachment defect**: an assignment took its
comment from anywhere on the value's line, but the rule reads it as the very
next token. With a `;` between them the comment belongs to the assignment it
follows.

| source | CPython | pyre, before |
|---|---|---|
| `y = 1; z = 2  # type: int` | `Assign` at col 7 | `Assign` at col **0** |
| `y = 1; z = 2; w = 3  # type: int` | `Assign` at col 14 | `Assign` at col **0** |
| `y = 1; z += 1  # type: int` | `SyntaxError` | silently attached to col 0 |

Each comment now records where the code before it ends, and an assignment takes
one only when nothing but spaces and tabs stands between. Without the refusal
that misattachment had no symptom — it just built a wrong tree.

Graded by a **76-case differential corpus against CPython 3.14.6: 0
mismatches.** It found four defects a hand-check missed — the two above, plus
`fn parameter` hard-coding `("type_comment", self.none())` so every `arg` case
failed regardless of collection, and the EOF-ignore newline being guarded on a
trailing `\n` already existing.

# cpyext — a heap corruption glibc saw and darwin did not

`PyBytes_FromStringAndSize(NULL, n)` allocated a header-sized block, and
`realize_pending` links that same block to the `bytes` it builds -- so the block
C is still filling becomes that object's mirror. Every later `make_ref` of it
then reached `stamp_ob_size`, which writes `ob_size` **at offset 24 of a 24-byte
allocation**. Under glibc those eight bytes overlap the next chunk's size field,
so the neighbour's `free()` reported `free(): invalid pointer` and
`cpyext_locks` died there with no other output. The darwin allocator leaves that
slack unused, which is why `cargo test` was green on macOS and windows.

The block is now allocated at `CPyVarObject` width, and its `ob_size` is set as
it is allocated -- `pyobject.py:108` does the same with
`pyvarobj.c_ob_size = itemcount`, so `Py_SIZE` answers the requested length for
the whole window in which C owns the buffer.

# Gates

Re-run in full on the rebased tree (base `bf260229c01`):

- `pyre/cpython_tests/run.py --backend dynasm` — **223 PASS / 0 FAIL / 0 CRASH /
  0 TIMEOUT**, no regressions. Graded against the binary CI builds for it,
  `--no-default-features --features dynasm`: adding `cpyext` fills
  `EXTENSION_SUFFIXES`, which un-skips `test_importlib.extension.*` and then
  fails it on a `_testsinglephase` no pyre tree carries.
- `cargo test --all --no-default-features --features dynasm,cpyext` — **181
  result lines, all ok, 8387 passed, 0 failed.**
- `pyre/extra_tests/parity_tests/run.py` — **all parity tests pass.**

The branch adds twelve fixtures: `ast_type_comments`,
`compile_flag_and_symtable_constants`, `darwin_module_constant_surface`,
`exception_repr_names_the_bare_class`, `locale_langinfo_item_constants`,
`select_kevent_repr`, `signal_range_is_the_platform_nsig`,
`socket_address_argument_conversions`, `sre_constant_surface`,
`std_stream_write_rejects_non_str`,
`termios_error_carries_errno_and_strerror`, `unbound_hash_is_base_implementation`.
Each is `cpython=OK dynasm=OK`, with the pypy divergence declared where the
fixture pins one. The langinfo and signal fixtures are scoped to darwin and
linux: windows has no `<langinfo.h>` and neither `SIGUSR1` nor the POSIX
delivery model.

# Known and not fixed here

`pyexpat`'s parser reports an error one column further right than expat does
(`<a><` is column 4, expat says 3). `bump_char` advances the column **after**
consuming, so every diagnostic names the position past the offending character
rather than at it; closing it means capturing the failing token's start at each
of the module's 42 error sites and grading them against real expat, which is a
change of its own rather than an adjustment.

**A parser diagnostic that is not a type comment still carries no location.**
`ast.parse("x = = 1")` raises `SyntaxError` whose `lineno`, `offset` and `text`
are all `None`, and whose message is the underlying parser's
(`Expected an expression at byte range 4..5`) rather than `invalid syntax`. The
type-comment refusal above locates itself because this layer computes that
position; the general path hands the parser's error straight through.

The constant-surface gaps an earlier
draft listed here are closed above; what is left is not addressable by adding a
number — honest capability booleans (`ssl.HAS_*`), build and environment
identity, opaque handles rendered as ints, and modules that are genuinely
absent or partial (`_datetime`, `curses`, `readline`, `sqlite3`).

**A `gc:` commit was withdrawn from this branch.** It made a closed `_io` stream
withdraw its finalizer registration, whose whole diff was setting an existing
`GCFLAG_IGNORE_FINALIZER` bit — but that flag's three readers in `collector.rs`
had never executed, because nothing had ever set it. Setting it armed them, and
`test.test_datetime` went from 6/6 ok to 5/6 SIGSEGV; a fresh `origin/main`
worktree and an ablation of the single call both confirmed it. #1464 has since
landed the collector work properly. Post-rebase, `test_datetime` is 4/4 ok plus
a fifth clean run inside the suite.

`raise_signal` from a non-main thread was listed here in an earlier draft; it
was re-measured and pyre already matches — the handler runs on the main thread,
and does not run before `raise_signal` returns in the worker.

## Summary by CodeRabbit

* **New Features**
  * Expanded CPython extension compatibility, including stable ABI support, type slots, reference counting, exceptions, and runtime-state APIs.
  * Added broader macOS platform constants across networking, signals, terminal control, time, logging, and system modules.
  * Improved socket, `fcntl`, `ioctl`, and buffer support for bytes-like and memory-mapped data.
  * AST conversions now preserve source locations, including multibyte text ranges.

* **Bug Fixes**
  * Corrected exception representations, error details, hashing behavior, stream finalization, signal handling, and standard-stream type validation.

* **Tests**
  * Added extensive cross-platform parity and CPython ABI coverage.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3241,7 +3241,7 @@ dependencies = [
 [[package]]
 name = "rustpython-codegen"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "bitflags 2.13.0",
  "indexmap",
@@ -3264,7 +3264,7 @@ dependencies = [
 [[package]]
 name = "rustpython-common"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "ascii",
  "bitflags 2.13.0",
@@ -3287,7 +3287,7 @@ dependencies = [
 [[package]]
 name = "rustpython-compiler"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "rustpython-codegen",
  "rustpython-compiler-core",
@@ -3301,7 +3301,7 @@ dependencies = [
 [[package]]
 name = "rustpython-compiler-core"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "bitflags 2.13.0",
  "bitflagset",
@@ -3317,7 +3317,7 @@ dependencies = [
 [[package]]
 name = "rustpython-host_env"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "bitflags 2.13.0",
  "cc",
@@ -3351,7 +3351,7 @@ dependencies = [
 [[package]]
 name = "rustpython-literal"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "hexf-parse",
  "is-macro",
@@ -3430,7 +3430,7 @@ dependencies = [
 [[package]]
 name = "rustpython-sre_engine"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "bitflags 2.13.0",
  "num_enum",
@@ -3442,7 +3442,7 @@ dependencies = [
 [[package]]
 name = "rustpython-unicode"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "icu_casemap",
  "icu_locale",
@@ -3456,7 +3456,7 @@ dependencies = [
 [[package]]
 name = "rustpython-wtf8"
 version = "0.5.0"
-source = "git+https://github.com/RustPython/RustPython.git?rev=dd2cc4d77a625661e5ca189f1cf9be199ecaa434#dd2cc4d77a625661e5ca189f1cf9be199ecaa434"
+source = "git+https://github.com/RustPython/RustPython.git?rev=48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201#48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201"
 dependencies = [
  "ascii",
  "bstr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,16 +72,16 @@ pyrex = { version = "0.0.2", path = "pyre/pyrex" }
 # parity fixes from #8550 were merged.  `fix_code_filenames` needs `IndexMut`
 # to edit nested code constants in place, and `loads` relies on the #8552 hook
 # to preserve invalid code bytes, so this pin must not move below either.
-rustpython-common = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-rustpython-literal = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-rustpython-compiler = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-rustpython-compiler-core = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
+rustpython-common = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+rustpython-literal = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+rustpython-compiler = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+rustpython-compiler-core = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
 ruff-text-size = { package = "rustpython-ruff_text_size", git = "https://github.com/RustPython/ruff.git", tag = "0.15.19-rustpython" }
 num-complex = "0.4"
-rustpython-host_env = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-rustpython-wtf8 = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-rustpython-unicode = { git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
-sre-engine = { package = "rustpython-sre_engine", git = "https://github.com/RustPython/RustPython.git", rev = "dd2cc4d77a625661e5ca189f1cf9be199ecaa434" }
+rustpython-host_env = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+rustpython-wtf8 = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+rustpython-unicode = { git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
+sre-engine = { package = "rustpython-sre_engine", git = "https://github.com/RustPython/RustPython.git", rev = "48a3a1f8e04eb4f1ea16c6324e936e6ece3f0201" }
 
 # majit JIT framework
 majit = { version = "0.0.2", path = "majit/majit" }
--- a/include/pyre3.14t/listobject.h
+++ b/include/pyre3.14t/listobject.h
@@ -12,8 +12,15 @@ extern "C" {
 #endif
 /* list. */
 
+/* Each is exported as well as spelled here, for the reason `Py_TYPE` is: a
+   caller that declares the prototype itself rather than including this header
+   still has to find a symbol.  The declaration comes before the macro that
+   replaces it. */
+PyAPI_FUNC(Py_ssize_t) PyList_GET_SIZE(PyObject *);
 #define PyList_GET_SIZE(ob) PyList_Size((PyObject *)(ob))
+PyAPI_FUNC(PyObject *) PyList_GET_ITEM(PyObject *, Py_ssize_t);
 #define PyList_GET_ITEM(ob, i) PyList_GetItem((PyObject *)(ob), (i))
+PyAPI_FUNC(void) PyList_SET_ITEM(PyObject *, Py_ssize_t, PyObject *);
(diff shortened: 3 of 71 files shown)


youknowone:
CI red 3건을 조사했습니다.

| job | 소유 | 내용 |
|---|---|---|
| `sandbox e2e + wasm web build (ubuntu)` | **이 브랜치** | `call.rs`가 `module::select::kevent_type()`을 무조건 이름 지어 `--features sandbox` 빌드가 E0433. `select`는 sandbox에서 cfg-out입니다. |
| `cargo test (ubuntu)` | **이 브랜치** | rustpython 핀 상향 이후 `fstring_missing_comma_uses_recursive_expression_range` 실패. `character_offset`은 스칼라 중간에 떨어질 수 없어서, "mid-scalar이면 거부"라는 예전 불변식이 성립하지 않습니다. |
| `pyre/check.py (ubuntu)` | **main** | pypy lane 4건(`foriter_stopiteration_exception_event`, `traceback_lineno_sentinel`, `utf8_check_untrusted_bytes`, `utf8_surrogatepass_error_span`)이 main 자체 run [32673897453](https://github.com/youknowone/pyre/actions/runs/32673897453)에서도 동일하게 red입니다. |

앞의 두 건은 고쳤습니다. f-string 쪽은 span을 거부하면 `f'{α β}'`가 `f-string: expecting '}'` / `6, 7`로 떨어지고, 투영하면 `invalid syntax. Perhaps you forgot a comma?` / `4, 6`이 됩니다 (3.14는 `4, 7`). 후자가 엄밀히 더 가깝기에 그 값을 pin하고 남은 1글자 갭의 원인을 주석에 적었습니다.

이어서 상수 표면 배치를 넣었습니다: 17개 모듈 정수 표면을 CPython 3.14.6 / pypy3와 3-way diff해서 이름 110개 누락 + 값 1개 오류를 확인하고, 값은 전부 SDK 헤더에서 되읽어 채웠습니다. `signal.NSIG`는 단순 상수가 아니라 동작 버그였습니다 — 64 고정이라 `getsignal(40)`이 CPython에서 `ValueError`인데 성공했습니다. darwin 32 / glibc 65로 나누고, 65는 한 워드를 넘으므로 `pypysig_flags_bits`를 upstream대로 `N_LONGSIG` 배열로 포팅했습니다.

로컬 게이트: CPython suite 223 PASS / 0 FAIL / no regressions, `cargo test --all --no-default-features --features dynasm,cpyext` exit 0, parity의 새 fixture 3개 모두 `cpython=OK dynasm=OK`. sandbox 빌드는 컨테이너의 linux/arm64에서 `cargo check --features sandbox,dynasm` 통과를 확인했습니다.

— *commented by Claude*

youknowone:
`cargo test (ubuntu)`의 두 번째 red를 근본 원인까지 추적했습니다. 컨테이너의 linux/arm64에서 결정적으로 재현됩니다 (~1분).

```
test the_locks_an_extension_holds ... FAILED
stderr: free(): invalid pointer
```

`PyBytes_FromStringAndSize(NULL, n)`이 미러 블록을 헤더 크기(24바이트)로만 잡는데, `realize_pending`이 **같은 블록**을 완성된 `bytes`에 링크합니다. 그 시점부터 그 객체에 대한 모든 `make_ref`가 `var_length` 분기를 타서 24바이트 할당의 offset 24에 `ob_size`를 씁니다.

glibc에서는 24바이트 요청이 32바이트 청크를 받고 그 마지막 8바이트가 **다음 청크의 size 필드**와 겹칩니다. 즉 이웃을 깨뜨리고, 다음 `free()`가 `free(): invalid pointer`를 냅니다 — stdout도 backtrace도 bytes 언급도 없이. darwin 할당자는 그 여유 공간을 쓰지 않으므로 같은 out-of-bounds 쓰기가 조용히 지나갑니다. 그래서 `cargo test (macos)`와 `(windows)`는 내내 green이었습니다.

`publish_length`의 doc은 "`ensure_mirror`가 var-length 종류를 링크하는 유일한 경로이고 필드 크기만큼 블록을 잡는다"고 적고 있었는데, 두 번째 경로가 있었습니다. 나머지 헤더 크기 경로인 `PyUnicode_New`의 pending 블록은 영향이 없습니다 — `var_length`가 `str`에는 `None`을 답합니다.

수정: pending 블록을 `CPyVarObject` 크기로 잡고, `PyBytes_FromStringAndSize`가 그러듯 길이를 즉시 publish하며(C가 버퍼를 채우는 동안 `Py_SIZE`가 답해야 하므로), `_PyBytes_Resize` 이후에도 갱신합니다.

검증: linux/arm64에서 `cargo test -p pyrex --no-default-features --features dynasm,cpyext` 전부 통과(수정 전 `cpyext_locks` 실패 → 수정 후 ok). darwin에서 CPython suite 223 PASS / no regressions, `cargo test --all --no-default-features --features dynasm,cpyext` exit 0(169 ok).

parity fixture 두 개도 플랫폼 범위를 좁혔습니다. Windows는 `<langinfo.h>`가 없어 55개 item key가 모든 인터프리터에서 부재하고, `SIGUSR1`도 POSIX 전달 모델도 없습니다. 그리고 `valid_signals()` 멤버십은 enum 멤버십이 아닙니다 — glibc의 realtime signal은 `Signals` 이름이 없어 CPython에서도 평범한 int로 나옵니다.

— *commented by Claude*

youknowone:
## Codex parity review — the three flags are answerable from the C source

All three deviations from PyPy are CPython **observables**, and the governing
artefact for each lives in `Modules/`, which the review's search (PyPy sources
plus `lib-python/3`) does not cover. Citations, quoting the line:

### 1. `_sre.MAXGROUPS` — `2**31 - 1` → `2**30 - 1`

`Modules/_sre/sre.h:19-21`, the 64-bit branch:

```c
#if SIZEOF_SIZE_T > 4
# define SRE_MAXREPEAT (~(SRE_CODE)0)
# define SRE_MAXGROUPS ((SRE_CODE)INT32_MAX / 2)
```

published at `Modules/_sre/sre.c:3419`:

```c
ADD_ULONG_CONSTANT(m, "MAXGROUPS", SRE_MAXGROUPS);
```

`INT32_MAX / 2` is `1073741823`. `python3.14 -c "import _sre;
print(_sre.MAXGROUPS)"` answers `1073741823`, and pyre answers `1073741823`
after the change.

The review is correct that `lib-python/3` pins no *value*, and that is the
reason the fixture exists: `test_re.py:703-707` reads `MAXGROUPS` out of the
module under test, so both sides of every comparison move together and a wrong
value is invisible, and `test_re.py:2924` only asserts `>= 0`. The fixture
header states exactly this.

Worth noting for the diff itself: the two macros have **different bases** —
`SRE_MAXREPEAT` is unsigned all-ones, `SRE_MAXGROUPS` halves `INT32_MAX`. The
vendored engine crate derived `MAXGROUPS` from its own `MAXREPEAT`, which is
precisely how it landed one bit wide.

### 2. `select.kevent.__repr__`

`Modules/selectmodule.c:1895` defines `kqueue_event_repr`, installed at `:1981`:

```c
{Py_tp_repr, kqueue_event_repr},
```

That PyPy's `W_Kevent` registers no `__repr__` at all is not incidental to the
change — it is stated in the doc comment on the added method, as the divergence
it is.

### 3. `pyexpat` exception name

`Modules/pyexpat.c:2292`:

```c
state->error = PyErr_NewException("xml.parsers.expat.ExpatError",
```

The class is reachable as `pyexpat.error`, `pyexpat.ExpatError` and
`xml.parsers.expat.ExpatError`; the last is the name it is *registered* under,
and it is what `__module__` and `repr()` of an instance show.

— *commented by Claude*

PR #1459, opened 2026-08-23
jit-trace: fold a Python-level str(int) to the jit_int_str render call, and key ALLOW_UNBOXING_INTS on pyre's int width

youknowone:
Two independent changes, plus the fixture gate and the row-count fix that follow from the first.

- `jit: fold a Python-level str(int) to the jit_int_str render call`
- `bench: gate the store_attr_direct fold on for_iter_direct_store_double`
- `jit-trace: correct the SPEC_FOLD_ROWS length after the rebase merged two rows`
- `mapdict: read ALLOW_UNBOXING_INTS off pyre's int width, not its pointer width`

---

## `str(int)` folds to `jit_int_str`

A JIT-compiled loop calling `str(i)` was **slower than the interpreter**: 5,000,000 iterations of `acc ^= len(str(i))` inside a function, min user-CPU of 3 runs, startup-subtracted, read 272 ns/iter on dynasm against 40 on cpython and 8 on pypy. `loops_compiled` confirmed the loop was compiled, so this was compiled-code cost, not a compile decline.

**Why there was no fold.** `BUILTIN_FOLDS` in `jit_builtin_folds.rs` is keyed on `BuiltinCodeFn` via `builtin_code_fn_of`, but `builtins.rs` binds the name `str` to `gettypeobject(&STR_TYPE)` — a **type object**, not a builtin function. No row in that table could ever have matched `str`, however many were added. The working precedent is `try_walker_specialize_float_call` / `try_walker_specialize_int_call`, which fold a call whose callable is a pinned type object; this adds `try_walker_specialize_str_call` beside them.

**No new helper and no new ABI.** `pyre_object::unicodeobject::jit_int_str` already existed, was already `#[majit_macros::elidable]`, was already registered in `jit_fnaddr.rs`, and `jtransform` already lowered the graph-level `UnaryOp { op: "str" }` over an Int operand to it. The Python-level call site was the only one not taking that channel. Upstream records the same shape — `call_r(ConstClass(ll_str__IntegerR_SignedConst_Signed), i49, descr=<Callr 8 i EF=3>)` followed by `guard_no_exception`.

The fold emits: guard-callable on the `str` type object, `GuardClass(INT_TYPE)`, exact-`w_class` guard, unbox, `CallR(jit_int_str, [Int] -> Ref, ELIDABLE_EFFECT_INFO)`, `GuardNoException`. The two guards are both needed and neither implies the other — `ob_type == &INT_TYPE` excludes `bool` and `W_LongObject`, `w_class == int` excludes int subclasses that override `__str__`.

**The cross-check must not allocate.** Every fold cross-checks its helper against the authentic boxed result before recording, but this helper allocates a `W_UnicodeObject`, and a nursery collection under it could move `arg_obj` / `boxed_result`, which the walker holds as raw pointers with no root scope. So the rendering is split out as `int_str_text(v) -> String` and the fold compares text.

`[Int] -> Ref` needs no wasm vouching: `majit-backend-wasm/src/codegen.rs` maps `Type::Int | Type::Ref => ValType::I64`, so the signature is `(i64) -> i64` — the arity-keyed family. Only the mixed `Float1` shape needs `set_faithful_residual_call_addrs`.

Measured on dynasm, same harness, startup-subtracted:

| loop body | fold off | fold on | cpython | pypy |
|---|---|---|---|---|
| `acc ^= len(str(i))` — the fold's subject | 272 | **70** | 40 | 8 |
| `s = str(i); acc ^= ord(s[0]) ^ len(s)` | 344 | **136** | 54 | 6 |

The second row's remaining 136 is **not** all `str`: 70 + 64 = 134, where 64 is a loop-invariant `ord(s[0]) ^ len(s)` that this fold does not touch and `PYRE_FBW_NO_SPECIALIZE=str_call` does not move. That is tracked separately; do not read 136 as this fold's ceiling.

New fixture `pyre/bench/synth/str_int_render_hot.py` carries both gates:

- `# pyre-check: spec-folds=str_call` — `_check_spec_folds` runs the fold census on a native backend and fails if the declared label never fires. This is what gates the fold itself; a throughput ceiling cannot single out one fold among several.
- `# pyre-check: max-pypy-ratio=15` — placed **between the two arms**: `PYRE_FBW_NO_SPECIALIZE=str_call` reads 38.2x pypy, the default reads 11.5x.

Arm-checked in both directions: a bogus `spec-folds` label fails, the real one passes; the ceiling fails with the fold suppressed.

## `for_iter_direct_store_double` declares its fold

That fixture's declared subject is the `store_attr_direct` fold, and it gated nothing about it at any `trace_limit` — its `max-pypy-ratio=0.5` measures throughput only. One `# pyre-check: spec-folds=store_attr_direct` line puts it on the same channel as the new fixture.

## `SPEC_FOLD_ROWS` length

`SPEC_FOLD_ROWS` is a fixed-size array whose length is written as a literal. `main` appended a `frame_lasti` row while this branch appended `str_call`; the rebase merged both (different lines, no textual conflict) and carried only one side's count. The array declared 70 entries and held 71 — a `cargo check` E0308, i.e. every build job would have failed. Corrected to 71.

---

## `ALLOW_UNBOXING_INTS`

mapdict.py's `ALLOW_UNBOXING_INTS = LONG_BIT == 64` was ported as `usize::BITS == 64`. In RPython those are the same quantity — `LONG_BIT` is the machine word and `W_IntObject.intval` is a C `long`. In pyre they are not: the interpreter's int is `i64` on every target, while pointer width is 32 on `wasm32-unknown-unknown`. So int attributes stayed boxed on the guest alone.

That pyre's `LONG_BIT` is 64 everywhere is stated four times in the tree, none of them under a cfg:

| site | value |
|---|---|
| `pyre_object::intobject::W_IntObject.intval` / `w_int_get_value` | `i64` |
| `sys/vm.rs` `module_ns_store(ns, "maxsize", …)` | `w_int_new(i64::MAX)` — the guest really does print `sys.maxsize == 2**63-1` |
| `majit_metainterp::ruleopt::proof::LONG_BIT`, and `ruleopt::rust_codegen`'s lowering of the `LONG_BIT` token | `64` |
| `objspace/descroperation.rs` comment | "RPython's target constant `LONG_BIT` is 64 for pyre's i64 word" |

Census: no other production site in `pyre/` or `majit/` uses `usize::BITS` or `target_pointer_width` to stand in for int width — the two remaining runtime hits are both `#[cfg(test)]`.

The const is now `true`. Enabling the int arm on wasm32 reuses machinery that is already live there: the unboxed slot is an i64 `TypedItemsBlock` and float unboxing already drives it, and wasm32 has native `i64`, so upstream's emulated-longlong reason for the 32-bit disable does not apply.

**Two wasm baselines move, both onto values the native backends already record:**

| fixture | before (wasm) | after (wasm) | dynasm / cranelift |
|---|---|---|---|
| `mapdict_unboxed_type_change_attr` | 3 loops / 2 bridges / 404 guards | 4 / 0 / 4 | 4 / 0 / 4 |
| `for_iter_select_receiver_swap` | 1 / 4 / 802 | 1 / 5 / 1002 | 1 / 5 / 1002 |

Both now agree with dynasm and cranelift field for field, which is why they are re-recorded rather than investigated. The first fixture's header had already predicted this outcome in writing — it explained the 404 as "no unboxed fold ⇒ the trace pins no `Terminator.allow_unboxing` quasi-immutable ⇒ the guard is bridged instead of the loop retraced". That paragraph is deleted.

Native control: dynasm reproduces its own committed baselines on all four touched fixtures. The host build is unaffected by construction — on a 64-bit host `usize::BITS == 64` and `true` compile identically.

---

### Withdrawn from this PR

An earlier revision added a `jit-stats lane compiled nothing` gate — red when one backend's baseline records zero compiles while a sibling's records some — plus a `# pyre-check: uncompiled-backends=` escape hatch on `for_iter_direct_store_double` and `trace_too_long_inline_multiframe`. **That commit is dropped: follow-up measurement refuted its premise, and the gate had no true positives.**

`pypyjit.set_param("...threshold=1...")` makes code that is *not the fixture body* compile. A control script consisting only of `import pypyjit` + `set_param` compiles 1 loop on dynasm at `trace_limit` 10–50 and 2 at ≥60 (flat through 340), and **0 on the wasm guest at every limit**. `MAJIT_LOG=1`'s `[bcenc-audit]` names identify them: the control compiles `_clear` at tl=40 and `_requires_builtin_wrapper` + `_clear` at tl=70 — and each fixture compiles *exactly those and nothing else*. Both names are `importlib` internals; neither appears in either fixture. Neither fixture's own functions compile on dynasm at its pinned limit.

Subtract that constant and the two arms agree exactly at every rung, including the abort counters (`trace_too_long_inline_multiframe`: dynasm `loops_aborted` 31 − 1 = 30 = wasm's 30; `fbw_blackhole_adopted_single_frame` 21 − 1 = 20 = 20; `multi_frame` 10 = 10). A plain `def f(x): return x + 1` loop first compiles at `trace_limit=12` on **both** backends.

So neither firing was a wasm-lane loss: at the pinned limits no lane had body compiles to lose, which makes both of the gate's hits false positives on their own terms. Across all 1304 committed lanes the predicate had exactly those 2 hits.

The predicate is also structurally wrong, which matters more than the hit count — a tripwire is supposed to sit at zero. It compares body + harness constant across backends, and that constant is backend- and limit-dependent, so it fires with a false diagnosis on an entire by-design class: any fixture that arms `threshold=1` and pins a low limit. Both firings bear that out — each minted an exemption comment asserting the guest needs a larger limit, which the body accounting refutes.

Scope of the evidence: the by-name identification is dynasm-only (wasm is trivially zero); cranelift's committed baselines are consistent with the same constant but were not name-verified. The follow-ups are tracked separately.


youknowone:
Force-pushed: the second commit (`check: fail a backend that compiled nothing its sibling backends compile`) is **dropped**, so any review of it is stale. This PR is now the `ALLOW_UNBOXING_INTS` fix alone.

Reason: follow-up measurement refuted that commit's premise, and its own commit message stated the false claim ("the guest needs about five times the native limit"). `pypyjit.set_param("...threshold=1...")` causes non-fixture code to compile — a control script of just `import pypyjit` + `set_param` compiles 1 loop on dynasm at `trace_limit` 10–50 and 2 at ≥60, and 0 on the wasm guest at every limit. `MAJIT_LOG=1`'s `[bcenc-audit]` names them: `_clear`, and `_requires_builtin_wrapper` at the higher limits — both `importlib` internals, and each of the two exempted fixtures compiles exactly those and nothing else. Neither fixture's own functions compile on dynasm at its pinned limit, so the guest's zero was never a wasm-lane loss, and the gate's only 2 hits in 1304 committed lanes were both this artefact.

Details are in the PR body under "Withdrawn from this PR".

— *commented by Claude*

youknowone:
> This adds a JIT-only `str(int)` implementation in `pyre-jit-trace` … fix the generation/inlining defect that leaves the call opaque rather than teaching the walker this operation separately.

The generation defect is real and is the right long-term fix, but two of the premises here are not accurate for this patch, and the third names a blocker that is measured and tracked.

**1. There is no JIT-only `str(int)` implementation.** The fold's call target is `pyre_object::unicodeobject::jit_int_str` — it lives in the interpreter's object layer, is registered in `pyre-interpreter/src/jit_fnaddr.rs`, and is the *exact* target `majit-translate`'s `jtransform` already lowers a graph-level `UnaryOp { op: "str" }` over an `Int` operand to (`CallTarget::function_path(["jit_int_str"])`, covered by `transform_graph_lowers_int_str_to_jit_int_str_residual_call`). Upstream records the same operation for the same source — `call_r(ConstClass(ll_str__IntegerR_SignedConst_Signed), i49, descr=<Callr 8 i EF=3>)` followed by `guard_no_exception`.

So the recorded trace *is* the one the generated pipeline emits. What this patch supplies is the **entry** into that existing channel from the Python-level call site, which was the only site not taking it. No second rendering of int→str exists to drift: `jit_int_str` is `w_str_new_managed(&int_str_text(v))`, and `int_str_text` is what the fold compares against.

**2. Divergence from the interpreter is structurally blocked at record time.** Before it records anything, the fold runs the interpreter's own path — `force_plain_eval()` + `call_function_impl_result` — and cross-checks the authentic boxed result against `int_str_text`. A mismatch declines the fold. Combined with the guards (callable identity on the pinned `str` type object, `GuardClass(INT_TYPE)`, exact-`w_class`), an interpreter change to what `str(int)` returns cannot silently survive into compiled code; it turns the fold off.

**3. The mechanism is not introduced here.** `origin/main` carries **67** `try_walker_specialize_*` entry points; this patch makes it 68. `try_walker_specialize_int_call` and `try_walker_specialize_float_call` are the same shape line for line — `std::ptr::eq(concrete_callable, <T>_type_obj)` on a pinned type object, because `builtins.rs` binds `int`/`float`/`str` to `gettypeobject(&…)` rather than to a `BuiltinCodeFn`. `SPEC_FOLD_ROWS` goes 70 → 71.

**Disposition: deferred, with the blocker named.** Routing this through the descent instead requires the descent to execute the builtin's body, and today it does not. Measured: `PYRE_FBW_NO_SPECIALIZE=builtin_len` costs 11x wall (0.32s → 3.66s) with `loops_compiled` **unchanged** — the call is residualized, not aborted, i.e. the descent never reaches the body. A census of blocked wrappers put 448 of 487 blocking on a call the descent never runs, so lowering alone does not open that wall. Until that lands, the fold is the only channel from a Python-level `str(i)` to `jit_int_str`, and the alternative is not "the generated pipeline emits it" but "the call stays a `CallMayForce` residual" — which is the 272 → 70 ns/iter this patch measures, on a loop that was slower than the interpreter.

Happy to file the descent item as a blocking follow-up on this PR if you'd rather it not merge ahead of the general fix.

— *commented by Claude*

Issue #1449, opened 2026-08-23
Heap types are born in old-gen, so no minor reclaims them and add_subclass's scan makes class creation quadratic


youknowone:
Creating classes against one shared base is **9.9x pypy3** and **2.7x CPython 3.14**, and the whole gap is one linear scan that upstream also has but never pays. Bounding the scan by hand removes 78% of it and puts pyre *ahead* of CPython.

## Measured

`type('X', (), {})` 100000 times, macOS arm64, release `pyre-dynasm`, pypy3 7.3.22, CPython 3.14.6.

| arm | pyre | pypy3 | CPython 3.14 |
|---|---|---|---|
| **A** shared base (`object`) | **2.160s** | 0.219s | 0.794s |
| **B** a fresh base every 100 classes — the scan can never exceed 100 | **0.465s** | 0.153s | 0.915s |
| **C** = A + `gc.collect()` every 1000 | 1.475s | 0.350s | 0.995s |

`len(object.__subclasses__())` when arm A finishes: pyre **13706**, pypy3 746, CPython 288.

A -> B is 4.6x on pyre and 1.4x on pypy3; on CPython B is *slower* than A. So 78% of pyre's arm-A cost is the length of one list, and the class-creation path itself (arm B) is 3.0x pypy3 and faster than CPython.

## Why the list is long: a heap type is born old, so no minor can reclaim it

```python
gc.collect()
for i in range(20000):
    C = type('X', (), {})          # every one of them is garbage immediately
for _ in range(400000):
    _ = [None]                     # ~ 100 minor collections over a 4MB nursery
```

`len(object.__subclasses__())` at each point:

| | after the 20000 | after the nursery churn | after `gc.collect()` |
|---|---|---|---|
| pyre | 20168 | **20168** | 169 |
| pypy3 | 1340 | 286 | 261 |
| CPython 3.14 | 400 | 400 | 172 |

**Not one dead type is reclaimed by a minor.** Sampling `len(object.__subclasses__())` every 500 iterations of arm A gives `670, 1170, 1670, 2170, ...` — exactly +500 per 500 classes, monotonic to 33670, dropping only when a major fires. pypy3's same sample oscillates between 473 and 3060.

The cause is the allocator `w_type_new` picks:

- `w_type_new` (`pyre-object/src/typeobject.rs`) takes its block from `gc_hook::try_gc_alloc_stable_raw(W_TYPE_GC_TYPE_ID, W_TYPE_OBJECT_SIZE)`.
[...]

PR #1462, opened 2026-08-24
majit-gc: the young non-moving birth #1449 asks for, plus #1382's three real review threads

youknowone:
Two independent pieces, both carried over from #1382: the three review threads
that turned out to be real, and the `young_rawmalloced_objects` machinery #1449
asks for.

## 1. `majit-gc`: a young non-moving birth (#1449, partial)

`GCFLAG_VISITED_RMY` was defined in `flags` and had **zero uses** anywhere in
the tree. It is the flag `external_malloc(..., alloc_young=True)` needs, and
without it every oversized allocation went straight to the old generation,
where only a major collection can reclaim it.

This adds the whole mechanism:

* `OldGen::try_alloc_young` — the `alloc_young` arm of `external_malloc`. It
  never consults `SMALL_REQUEST_THRESHOLD`, because an arena block must be old.
* `OldGen::free_young_rawmalloced_objects` — the end-of-minor step that
  promotes a `GCFLAG_VISITED_RMY` record into `old_rawmalloced_objects` and
  frees the rest.
* `MiniMarkGC::visit_young_rawmalloced_object` — the stamp, plus the lists an
  old object owes (`remembered_set` when `TRACK_YOUNG_PTRS` is clear,
  `old_objects_with_cards_set` when `HAS_CARDS`) and the promotion's share of
  `bytes_made_old_since_cycle`.
* `MiniMarkGC::remove_young_arrays_from_remembered_set` — run first in the
  minor, before the STATE_MARKING re-gray and every root walk.

### The part that is not four functions

Twelve existing sites answered "not in the nursery" with "old", and each is a
place a young raw-malloced object would be read as alive-and-old while this
collection can still free it:

| site | what it did |
|---|---|
| phase-1c jitframe root arm | a young block is in `rawmalloced_payloads`, so `oldgen.contains` was true and it was traced as an **old jitframe** — freed at the end of the same minor with its children kept |
| `classify_young_owner` | `!nursery.contains(owner)` ⇒ "survived", so a dead block kept its address-keyed side-table entry |
| `deal_with_young_objects_with_finalizers` | moved a registrant to the old list with no survival test |
| `invalidate_young_weakrefs` | no young-target branch |
| `rrc_young_object_alive`, `rrc_trace_c_edges_young`, `_rrc_minor_free` | "an object outside the nursery survives every minor" |
| `rawrefcount_create_link_pyre` / `_pyobj` | pyre had collapsed upstream's **two** questions (which list, which identity table) into one; they come apart exactly here |
| `bh_probe_stale_young_slots` | a young holder read as an old holder, so every young reference it carried was reported as a missing write barrier |
| `do_get_objects` generations 0 and 2 | `gc.get_objects(2)` would report a generation-0 object as old |
| the three slot loops of `trace_and_update_object` + the dirty-card loop | no visit hook |

### What takes the new birth

`alloc_with_type` and `alloc_with_type_rooted` — the entry points whose
**small** arm already calls `do_collect_nursery()`. Their callers therefore
already present a complete root set at an allocation, which is exactly the
precondition a young birth adds nothing to. In production that reaches the JIT
through `alloc_nursery`, `alloc_nursery_typed`, `alloc_varsize_typed`
(`dynasm_malloc_array`, `dynasm_malloc_str`) and `malloc_big_fixedsize`, on both
backends.

One refusal, and it is upstream's own contract rather than a pyre adaptation:
**a weakref type keeps the born-old birth.** `malloc_fixedsize`'s large arm opens
with `ll_assert(not contains_weakptr, "'contains_weakptr' specified for a large
object")`, and `invalidate_young_weakrefs` decides survival by reading a
forwarding pointer a non-moving object never has. A weakref payload is a single
`weakptr` slot, so nothing of that type approaches the threshold anyway.

A **destructor** is not a refusal. `malloc_fixedsize` returns early only for
`needs_finalizer and not is_finalizer_light` — the heavy arm — so a light
destructor falls through to the large arm, takes `alloc_young=True`, and is
appended to `young_objects_with_destructors` by the shared tail.
`TypeInfo.destructor` is documented as that light kind. Following it needs one
adaptation, because upstream leaves the resulting combination undefined:
`deal_with_young_objects_with_destructors` reads `is_forwarded`, which opens
with `ll_assert(self.is_in_nursery(obj), "Can't forward an object outside the
nursery.")`. So pyre reads `GCFLAG_VISITED_RMY` for a rawmalloced member — the
same witness `free_young_rawmalloced_objects` reads a few steps later, and final
by the time either runs, since both `collect_roots_in_nursery` and the
`collect_oldrefs_to_nursery` drain precede them.

The heavy arm needs nothing: pyre models it as a post-allocation
`register_finalizer` queue, and `deal_with_young_objects_with_finalizers`
already had its rawmalloced branch.

### `malloc_big_fixedsize`, and the write-barrier stamp it earns

Upstream's helper calls `do_malloc_fixedsize_clear` — the ordinary
`malloc_fixedsize` — so the size picks the arm and an oversized one is born
young. pyre called `alloc_oldgen_typed` instead. Both backends now call
`alloc_nursery_typed` (`MiniMarkGC::alloc_with_type`), which is that entry
point.

That is what lets `gen_malloc_fixedsize` apply rewrite.py's
`remember_write_barrier(v_result)`, which pyre had deliberately withheld. The
load-bearing half of upstream's *"this is always true because it's a fixed-size
object"* is about cards, not size: fixed-size means
`external_malloc(typeid, **0**, ...)`, so no `GCFLAG_HAS_CARDS` and
`extra_flags = 0`, so no `GCFLAG_TRACK_YOUNG_PTRS` — and a young rawmalloced
object without that flag is appended to `old_objects_pointing_to_young` the
moment `_visit_young_rawmalloced_object` promotes it, so its slots are traced
without a barrier ever running. Unreached, it is freed and its slots never
mattered. That is also why `_gen_call_malloc_gc` must *not* stamp in general:
"v_result might be a large young array", and a card-marked array is born
carrying `TRACK_YOUNG_PTRS`.

The stamp is skipped under `MAJIT_GC_YOUNG_RAWMALLOC=0`, which births the same
object old — the gate is process-wide and read once, so asking it at rewrite
time answers for every trace the process compiles. A test pins both directions.

**One pyre-only arm had to be split out.** `handle_new` also routes a
`SizeDescr::non_moving` descr to `gen_malloc_fixedsize`, at any size — and
`PYFRAME_DESCR_GROUP.size_descr` is a live one. The size-dispatching helper would
have put a small frame **in the nursery**, which is exactly what that flag
exists to prevent. That arm now calls a separate
`malloc_big_fixedsize_oldgen_fn`, sharing the calldescr the way
`malloc_array_oldgen_fn` shares `malloc_array_descr`, and takes no stamp.
Upstream has no such twin because it never reaches this generator for a
small object.

### Why #1449's own prescription is wrong, and what stays open

#1449 says the fix is to give `try_gc_alloc_stable_raw` (the allocator
`w_type_new` picks) this young path. **That is unsound today**, and the reason
is in `pyre-object/src/gc_interp.rs`'s module doc: born-old is deliberate.
pyre's interpreter has no shadow-stack pass, so a `PyObjectRef` held only on
the Rust stack of a bytecode handler is not a root. A born-old block survives
that; a young non-moving block does not — the next minor would **free** it. The
hazard is a use-after-free rather than the dangling-move that doc describes,
but the precondition it names is the same one, and it is still missing.

So #1449's heap-type half stays open, re-scoped: it is blocked on the
shadow-stack pass that `gc_interp.rs` already names as the thing that retires
the born-old stepping stone.

`MAJIT_GC_YOUNG_RAWMALLOC=0` restores the born-old behaviour without a rebuild,
so a defect that only appears once these objects can die at a minor is
bisectable against one binary. Registered in `majit/gate-triage.md` with its
retirement condition.

### On reach: measured, and what the measurement does not say

I could not construct a Python workload that takes the new birth. Four probes
under `MAJIT_GC_LIFETIME_LOG=1`, counting the `kind=raw-young` line:

| probe | result |
|---|---|
| `[0] * 40000` in a loop | `rawyoung=0` |
| `b"x" * 300000` in a loop | `rawyoung=0` |
| compiled loop building 300 KB bigint digit arrays (`loops_compiled=1`) | `rawyoung=0` |
| compiled `str`/`list`/`bytes` builders, 30 M log lines | `rawyoung=0` |

The instrument was armed, not silent: the same run logs `alloc=35251
free=28271`. Every oversized object those workloads allocate comes from
`try_gc_alloc_stable_raw` (born old) or the no-collect spill
(`spill_to_oldgen_or_null`), and neither may take this path.

That is a statement about workloads, not about reach. The API is reachable —
`dynasm_malloc_array` and `dynasm_malloc_str` go through `alloc_varsize_typed`,
which defers to `alloc_with_type` — and a test now pins that boundary so a
future short-circuit in the varsize wrapper cannot quietly put compiled
allocations back in the old generation. `malloc_big_fixedsize` is the arm
upstream itself calls dead (*"Never called as far as I can tell"*), so it is
matched for parity, not for a number.

So: correct, mostly latent, and it does not on its own move an RSS figure.
Saying so is the point — the alternative was to adopt it somewhere it would
have shown up and been wrong.

## 2. `check.py`: the three review threads from #1382 that were real

I graded all 19 unresolved threads on #1382; 16 were already fixed or wrong on
the merged tree. These three were not.

* **The build recipe did not include the build environment.** `build_backend`
  runs cargo with no `env=`, so `RUSTFLAGS` and the five lowering gates
  `pyre-jit-trace/build/prepass.rs` folds into `codegen_cache_key` decide what
  lands in the artefact while `build_recipe_digest` could not see them. A run
  under `PYRE_MIR_FRAMESTATE=0` stamped a digest a later unset run reproduced
  exactly, and `--no-build` then measured a differently-lowered JIT. Unset now
  hashes distinguishably from empty, the same distinction `prepass.rs` draws.
* **`Cargo.lock` was excluded from the build window unconditionally**, so an
  external rewrite mid-build (a `git checkout`, a sibling worktree's `cargo
  update`) was indistinguishable from cargo's own resolution.
  `settle_cargo_resolution` runs `cargo metadata` before the window opens —
  the same resolution, no compile — and when it succeeds the lock is judged
  like every other input. Measured: 5.3 s, and it leaves `Cargo.lock`
  byte-identical.
* **`run_selfcheck`'s compiled-loop floor was a constant 1**, which any single
  loop reaches. `wrapper_subclass_load_method_self` compiles eight, so seven of
  its guarded shapes could stop reaching the JIT with the fixture still green —
  the exact failure the floor exists to catch. Each fixture now declares its
  **measured** floor with `# pyre-check: selfcheck-loops=<n>`; the directive is
  required, not defaulted, because defaulting is what let the multi-loop
  fixtures sit under a floor that could not see them.

## Verification

* `cargo test -p majit-gc --lib` — 278 passed, 0 failed.
* `cargo test -p majit-gc --features gc_stress` — 280 passed, 0 failed.
* `cargo clippy -p majit-gc --all-targets` — no new warnings (the one
  `chunks_exact` hit is pre-existing at `HEAD`).
* `python3 pyre/check.py --no-build --synthetic-only` — **dynasm 453/453,
  cranelift 453/453, wasm 445/445**, all three backends.
* The loop floors are measured, not guessed: every fixture was run on all three
  enabled backends, twice, and the two passes are identical. The backends agree
  on every fixture, so no floor is a lowest-common-denominator compromise.
* The floor gate is armed, not merely quiet: raising
  `getframe_inlined_callee_lasti_escape`'s floor from 2 to 99 produces
  `FAIL loops_compiled=2 < 99: a shape this guard covers stopped reaching the
  JIT`.
* LLBC re-extracted (`scripts/extract-llbc.py --force`) and `--check` returns 0,
  so the offsets the backends were built against match this tree.

Declared floors: `wrapper_subclass_load_method_self` 8,
`identity_dict_view_iteration_regression` 7, `frame_lineno_mid_replay_regression`
4, three at 3, one at 2, eight at 1. The two `selfcheck-interpreted` fixtures
stay exempt.

## Summary by CodeRabbit

* **New Features**
  * Large collecting allocations can now start in the young generation, improving memory lifecycle handling and collection behavior.
  * Added optional loop-census reporting to identify which compiled traces were generated, including WebAssembly support.
  * Self-checks can now require specific code paths or functions to compile.

* **Bug Fixes**
  * Improved handling of young large objects across promotion, finalization, weak references, barriers, and major collection.
  * Preserved old-generation allocation for non-moving objects.

* **Documentation**
  * Expanded guidance for GC diagnostics, benchmark self-checks, and reachability analysis.


youknowone:
CI triage for the first run:

* **`cargo fmt --check`** — mine. Fixed in fa9d7c4.
* **`pyre/check.py` on macos / ubuntu / windows** — **not this branch's.** main is
  red on exactly these three jobs at `62d2029` (run 32673897453), with a
  byte-identical failure signature:

  ```
  pypy: exited 1 (last non-empty stdout line '') | AssertionError: consume_generator
  pypy: exited 1 (last non-empty stdout line 'rebuilt: -1') | AssertionError: (-1, 31)
  pypy: exited 1 (last non-empty stdout line '') | ModuleNotFoundError: No module named '_json'
  pypy: exited 1 (last non-empty stdout line '') | AssertionError
  ```

  All four are `pypy` runner failures in `extra_tests/parity_tests`, and the
  `pyre/check.py` stage itself and `extra_tests/run.py --gated-only`
  (99/99 on cpython, dynasm and cranelift) both passed in the same job.

Everything else was already green: `cargo test` on all three hosts, the
Charon/LLBC prepare on all three, sandbox e2e + wasm web build, cpyext ABI,
pre-commit.

— *commented by Claude*

youknowone:
Answering the Codex parity review.

### §2 — `MAJIT_GC_YOUNG_RAWMALLOC=0` births oversized collecting allocations old

Correct, and deliberate. It is a rollback gate, **default on**, with an entry in
`majit/gate-triage.md`; the deviation exists only when it is explicitly set.
`external_malloc(..., alloc_young=True)` is what the default arm does. The
switch is there because this change makes a class of object start dying at a
minor that previously could only die at a major — if that surfaces a defect
elsewhere, the bisect needs to be a run, not a rebuild.

### §3 — upstream gives a *light*-destructor large allocation the young birth

This one is right on the facts and I have amended the code to say so, because
the previous comment implied upstream forbade both halves when it only forbids
one.

`malloc_fixedsize` refuses `needs_finalizer and not is_finalizer_light` — the
**heavy** arm — and sends it to `external_malloc(..., alloc_young=False)`. A
light destructor falls through, reaches the large arm, takes `alloc_young=True`,
and is appended to `young_objects_with_destructors`. pyre's `TypeInfo`
destructor is exactly that light kind, so my `type_alloc_may_be_young` is
stricter than upstream here.

What I did not do is relax it, because upstream's own precondition rules the
combination out. `deal_with_young_objects_with_destructors` decides survival
with `is_forwarded`, and `is_forwarded` opens with

```
ll_assert(self.is_in_nursery(obj), "Can't forward an object outside the nursery.")
```

which a rawmalloced address never satisfies. With `ll_assert` compiled out the
read returns false and the destructor runs on a reachable object. So the pairing
is undefined upstream rather than modelled — consistent with
`malloc_big_fixedsize`'s own "Never called as far as I can tell" — and pyre takes
the arm that is defined. Earning the other one back means giving the young
destructor sweep a second survival witness, not loosening this predicate. The
doc comment now carries that reasoning and the citation.

### §4 — structural adaptations

Both readings match the intent. No change.

Also in `ceeed6a`: a test at `GcAllocator::alloc_varsize_typed`, the entry point
`dynasm_malloc_array` and `dynasm_malloc_str` reach the collector through. It
defers to `alloc_with_type`, so an oversized varsize allocation from compiled
code does take the young birth; the test pins that so a future short-circuit in
the varsize wrapper cannot quietly put those allocations back in the old
generation. Proven armed — it fails under `MAJIT_GC_YOUNG_RAWMALLOC=0`.

— *commented by Claude*

youknowone:
Pushed `f3746fd`, `87e2525`, `ae525d0` — the two upstream deviations named in
the parity review are now implemented rather than documented.

### The light-destructor young birth

`malloc_fixedsize` returns early only for `needs_finalizer and not
is_finalizer_light`, so a light destructor falls through to the large arm, takes
`alloc_young=True`, and is appended to `young_objects_with_destructors` by the
shared tail. `type_alloc_may_be_young` now excludes the weakref flag alone,
`finish_alloc_young_nonmoving` runs `register_young_object_if_needed`, and
`deal_with_young_objects_with_destructors` reads `GCFLAG_VISITED_RMY` for a
rawmalloced member instead of a forwarding pointer — which is the one place
upstream leaves undefined, since `is_forwarded` opens with
`ll_assert(self.is_in_nursery(obj), ...)`. That witness is final by then: both
`collect_roots_in_nursery` and the `collect_oldrefs_to_nursery` drain run before
the call, and `free_young_rawmalloced_objects` reads the same flag after it.

The heavy arm needed nothing — pyre models it as a post-allocation
`register_finalizer` queue, and `deal_with_young_objects_with_finalizers`
already had its rawmalloced branch.

### `malloc_big_fixedsize` → `malloc_fixedsize`, and the stamp

`dynasm_malloc_big_fixedsize` and `gc_malloc_big_fixedsize_helper` now call
`alloc_nursery_typed` (`MiniMarkGC::alloc_with_type`) rather than
`alloc_oldgen_typed`, matching `do_malloc_fixedsize_clear`. That earns back
rewrite.py's `remember_write_barrier(v_result)`, which pyre had withheld.

I was wrong in `fa429f3` about what earning it needs — I said a descr-level
plainness predicate for the destructor half. It needs no such thing: with the
destructor half fixed above, the only type-level exclusion left is the weakref
flag, which upstream rules out by the same `ll_assert` its own stamp relies on.
What it does need is the rollback gate read at rewrite time, and that is done —
`MAJIT_GC_YOUNG_RAWMALLOC=0` births the object old and suppresses the stamp
together. A test pins both directions; with the gate off the barrier reappears.

The load-bearing half of upstream's *"always true because it's a fixed-size
object"* is about cards: fixed-size means `external_malloc(typeid, **0**, ...)`,
so no `HAS_CARDS`, so `extra_flags = 0`, so no `TRACK_YOUNG_PTRS` — and
`_visit_young_rawmalloced_object` then appends the object to
`old_objects_pointing_to_young` at promotion, tracing its slots with no barrier
having run. That is also why `_gen_call_malloc_gc` must not stamp in general.

### One thing this nearly broke

`handle_new` also routes a `SizeDescr::non_moving` descr to
`gen_malloc_fixedsize`, **at any size** — and `PYFRAME_DESCR_GROUP.size_descr`
is a live one. A size-dispatching helper would have put a small frame in the
nursery, which is exactly what that flag exists to prevent. That arm now calls a
separate `malloc_big_fixedsize_oldgen_fn`, sharing the calldescr the way
`malloc_array_oldgen_fn` shares `malloc_array_descr`, and takes no stamp.
`test_non_moving_new_declines_nursery_at_small_size` is what caught it.

The new CALL_R can collect where the old one could not; that is the same gcmap
contract `malloc_array` has always had — both go through `gen_call_malloc_gc` →
`emitting_an_operation_that_can_collect()` and both carry
`EffectInfo::MOST_GENERAL`.

### Gates

* `cargo test --all --no-default-features --features dynasm` — **8242 passed, 0
  failed** over 168 binaries (8237 before this branch, +5 new tests)
* `cargo test -p majit-gc --features gc_stress` — 285 passed, 0 failed
* `cargo clippy` on the four touched crates — no new warnings; the remaining
  ones are `majit-translate`'s, untouched here
* every new test proven armed by inverting its witness or by the gate

— *commented by Claude*

PR #1464, opened 2026-08-24
gc: the ignore_finalizer hint, the copy/move write barriers, collect(gen), a --heapsize that reaches the collector, and a refused allocation told apart from a missing hook

youknowone:
Four commits on top of `62d2029eb21`.

## `gc: release the queued cpyext mirrors before collect returns`

`gc.collect()` drains the cpyext dead queue before returning, so a mirror
released by the collection is gone by the time the call answers.

## `gc: ignore_finalizer, the copy/move write barriers, move_out_of_nursery, collect(gen) and set_max_heap_size`

Ports the `incminimark` surface that had readers but no writers:

- `ignore_finalizer` (`GCFLAG_IGNORE_FINALIZER`) — the flag had three readers
  and no setter. Wired to `rgc.may_ignore_finalizer`'s upstream call sites:
  `generator.frame_is_finished`, the coroutine arm of `_invoke_execute_frame`,
  and `W_IOBase.maybe_unregister_rpython_finalizer_io` for
  `_IOBase`/`BufferedReader`/`TextIOWrapper.close`.
- `writebarrier_before_copy` / `writebarrier_before_move` /
  `manually_copy_card_bits`, together with the
  `updated_old_objects_pointing_to_pinned` flag they read, set at the two sites
  that change the list.
- `move_out_of_nursery` and the `GCFLAG_SHADOW_INITIALIZED` arm of
  `_trace_drag_out` (`copy = False`), which had zero uses.
- `collect(gen)` via `minor_collection_with_major_progress(force_enabled)`,
  and `set_max_heap_size`.
- `builtin_id` hands the audit hook the value read back out of the pinned slot
  rather than the pre-normalization one.

16 new collector tests.

### A CPython-3.14 divergence in the hint's position

Upstream emits `rgc.may_ignore_finalizer(self)` *ahead* of the
`"can't send non-None value to a just-started %s"` raise. Ported verbatim that
loses a warning CPython emits:

| runtime | `c.send(1)` then `del c` |
|---|---|
| CPython 3.14.6 | TypeError **and** `coroutine 'f' was never awaited` |
| pypy3 7.3.22 | TypeError, **no warning** |

A rejected `send` does not start the coroutine, so this is not a 3.11↔3.14
axis — the hint's position is the cause. Under spec=CPython the hint goes
*behind* the raise, pinned by
`extra_tests/snippets/coroutine_never_awaited_survives_a_send_typeerror.py`.

Because the hint has no other app-level effect, a passing snippet alone would
not show the plumbing is live, so the tree was built once with upstream's
ordering as a red arm:

| arm | result |
|---|---|
| hint ahead of the raise (upstream) | `AssertionError: []` |
| hint behind the raise (this PR) | pass |

### Deliberately not done

`ll_arraycopy` is not switched to `writebarrier_before_copy`: the copy sites
already emit the unconditional `try_gc_write_barrier`, which is strictly more
conservative, and no production allocator sets `HAS_CARDS`, so the swap would
buy nothing and add a dependency on the nursery `TRACK_YOUNG_PTRS` discipline.
The list-shift sites cannot take `writebarrier_before_move` without a
`try_gc_owns_object` probe — the `std::alloc` fallback `ItemsBlock` carries no
`GcHeader` — and that probe is the hot query the codebase removed. The
obligation is documented at `alloc_in_oldgen_with_cards` instead.

`young_rawmalloced_objects` + `GCFLAG_VISITED_RMY` (and `malloc_big_fixedsize`
as a young rawmalloc, which depends on it) is not attempted: the youth
predicate is spread across eight `is_nursery_object_start` copy-decision sites,
and a partial port frees live objects.

## `launcher: consume a leading --heapsize and forward the interact option to the child`

`set_max_heap_size` has exactly one upstream production caller, and pyre's
half of it was parsing the value and throwing it away, with a comment saying
why: *"pyre's GC has no runtime heap-limit knob."* It does now.

- `pypy_interact.py main` — the controller resolves the `k`/`m`/`g` suffix and
  prepends `--heapsize <bytes>` to the sandboxed program's arguments
  (`extraoptions + arguments[1:]`), instead of discarding it.
- `targetpypystandalone.py entry_point` — `--heapsize N` in `argv[1]`, and no
  other position, is consumed and cut out (`argv[:1] + argv[3:]`), so it
  reaches neither `sys.argv` nor `sys.orig_argv`; `app_main.py` fills both from
  the list `entry_point` hands on.
- `parse_heapsize` refuses a count above `i64::MAX` (`bytes > sys.maxint`).

Verified by equivalence with the channel that already worked, same program and
same limit:

| arm | result |
|---|---|
| no limit (control) | `allocated 4000000` |
| `PYPY_GC_MAX=41943040` | `using too much memory, aborting` |
| `--heapsize 41943040` | identical |

### An open defect this surfaced

The first heap-limit event should raise `MemoryError` and does not.
`pyre_object::lltype::malloc_typed_managed` reads the collector's OOM `NULL` as
*no GC hook installed* and falls back to `std::alloc`, so the object is
allocated outside the GC heap, the program keeps running, and the next major
collection takes the `max_heap_size_already_raised` arm and aborts. The
collector comments that promise "the interpreter allocation chokepoint raises
`MemoryError`" are measured false. `PYPY_GC_MAX` has behaved this way all
along — the row above is the proof, since it runs untouched code. Aborting on
the *second* event is upstream-faithful; only the first event's `MemoryError`
is missing.

**The remaining `MemoryError` gap is still open**, but the fourth commit
addresses the classification half of it — see below.

## `gc_hook: separate a refused allocation from a missing hook at the four sites that merged them`

`majit_gc::GcAllocOutcome` splits `NoRoute` (no hook installed, so the caller's
own `malloc_raw` path *is* the whole heap) from `Failed` (a GC owns the heap and
could not satisfy the request), and its doc records that `malloc_raw` is the
wrong answer for the second: the object holds managed references the collector
never traces, behind a header it never walks.

`try_gc_alloc_stable_raw` collapsed the two into one null and documented the
collapse as information-free — *"every traced caller already treats those two
cases identically"* — which contradicts `GcAllocOutcome`'s doc in the same
crate. Its **48 callers** all answer that null with `malloc_typed` /
`malloc_raw` / `Box`. It now classifies, so null means `NoRoute` alone and a
`Failed` reaches `gc_alloc_failed`. `malloc_typed_managed`, `w_list_new`'s
rooted collecting allocation, and `w_dict_new`'s non-stable arm made the same
merge on their own `Option<*mut u8>` results and now classify too. Four edits
cover roughly fifty-one sites; `pyframe.rs` `alloc_frame_block` already did
this correctly and is the model.

**This is a latent-correctness fix, not a demonstrated one.** Behaviour is
unchanged in every configuration measured, and I could not construct a repro
that routes an OOM null through any of the four sites: list-heavy, dict-heavy
and `PYRE_JIT=0` programs at four different `--heapsize` limits all still abort
at the second event. On the stable path a `Failed` is reachable only through the
size-overflow guard, since `alloc_in_oldgen_clear` memclears its result
unconditionally — a null there would already segfault. What is fixed is the
contract, and the contradiction between the two doc comments.

Left alone deliberately: the items-block sites in `object_array.rs` fall back to
a `std::alloc` block, which is a gated supported configuration rather than an
off-heap accident; `weakref.rs` and `unicodeobject.rs` answer a `Failed` with an
*immortal* object, which stays in the root set and is traced.

## Gates

| lane | result |
|---|---|
| `check.py --synthetic-only` | 453/453 × dynasm, cranelift — ALL PASSED 2/2 |
| `extra_tests/run.py --gated-only` | 100/100 × cpython, dynasm, cranelift |
| `cargo test -p majit-gc` | 289 passed |
| `cargo test -p pyrex --lib` | 19 passed |
| `cargo test -p pyre-object --lib` | 292 passed |
| `cpyext_types` (dynasm,cpyext) | 2 passed |
| `cargo fmt --all -- --check` | clean |
| citation gate, rpython module parity | clean |

— *authored by Claude*

## Summary by CodeRabbit

* **New Features**
  * Added generation-aware garbage collection controls and runtime heap-size configuration.
  * Added support for moving objects out of the nursery and selectively ignoring unnecessary finalizers.
  * Added launcher support for the `--heapsize` option.
  * `gc.collect()` now completes eligible native deallocations before returning.

* **Bug Fixes**
  * Improved coroutine and generator finalization behavior, including more accurate “never awaited” warnings.
  * Improved cleanup and finalizer handling when closing buffered and text I/O objects.
  * Fixed object pinning, allocation failure handling, and garbage-collection tracking in runtime operations.
  * Improved memory updates during object copying and movement.

### DIFF
--- a/majit/majit-gc/src/collector.rs
+++ b/majit/majit-gc/src/collector.rs
@@ -1059,6 +1059,9 @@ pub struct MiniMarkGC {
     /// minor left at least one pinned object in the nursery.
     any_pinned_object_kept: bool,
     old_objects_pointing_to_pinned: Vec<usize>,
+    /// incminimark.py `updated_old_objects_pointing_to_pinned`: set whenever
+    /// the list above changes, and read by `writebarrier_before_copy` alone.
+    updated_old_objects_pointing_to_pinned: bool,
     /// incminimark.py:311,430 `max_number_of_pinned_objects` and
     /// `pinned_objects_in_nursery`.
     max_number_of_pinned_objects: usize,
@@ -1203,6 +1206,7 @@ impl MiniMarkGC {
             surviving_pinned_objects: Vec::new(),
             any_pinned_object_kept: false,
             old_objects_pointing_to_pinned: Vec::new(),
+            updated_old_objects_pointing_to_pinned: false,
             max_number_of_pinned_objects,
             pinned_objects_in_nursery: 0,
             nursery_objects_shadows: AddressMap::default(),
@@ -1268,6 +1272,14 @@ impl MiniMarkGC {
     ///
     /// `has_gc_ptrs_in_var` should be true for arrays containing GC pointers
     /// (i.e. `has_gcptr_in_varsize(typeid)` in RPython).
+    ///
+    /// Every caller today is a test, so no card array exists in a running
+    /// heap.  A production one arrives owing two things that the per-item
+    /// barrier does not supply, because a card records a *position*: a bulk
+    /// copy into it owes [`Self::writebarrier_before_copy`], and any move of
+    /// items within it — a list insert, remove, splice or reverse — owes
+    /// [`Self::writebarrier_before_move`], which is what generalizes the cards
+    /// back to "any item may be young" before the positions change.
     pub fn alloc_in_oldgen_with_cards(
         &mut self,
         type_id: u32,
@@ -3669,6 +3681,36 @@ impl MiniMarkGC {
         self.allocate_shadow(obj_addr)
     }
 
+    /// incminimark.py `move_out_of_nursery`: "called twice, it should return
+    /// the same shadow object, and not creating another shadow object.  As a
+    /// safety feature, when called on a non-nursery object, do nothing."
+    ///
+    /// `rgc.move_out_of_nursery`'s contract is what makes the eager copy sound:
+    /// "Only use for immutable objects!"  The shadow is filled once and the
+    /// nursery original may still be written to afterwards, so a mutable object
+    /// would have the two copies disagree until the next minor.
+    pub fn move_out_of_nursery(&mut self, obj_addr: usize) -> usize {
+        if !self.is_in_nursery(obj_addr) {
+            return obj_addr;
+        }
+        let shadow = self.find_shadow(obj_addr);
+        let hdr = unsafe { header_of(obj_addr) };
+        if unsafe { (*hdr).has_flag(flags::SHADOW_INITIALIZED) } {
+            return shadow;
+        }
+        unsafe { (*hdr).set_flag(flags::SHADOW_INITIALIZED) };
+        let type_id = unsafe { (*hdr).type_id() };
+        let payload_size = self.size_for_typeid(obj_addr, type_id, "move_out_of_nursery");
+        // Payload only: `allocate_shadow` already wrote the shadow's own header
+        // and the flags the old generation needs, and the copy must not undo
+        // them.  The next minor reads SHADOW_INITIALIZED and skips its own copy
+        // for the same reason.
+        unsafe {
+            std::ptr::copy_nonoverlapping(obj_addr as *const u8, shadow as *mut u8, payload_size)
+        };
+        shadow
+    }
+
     /// minimark.py `id_or_identityhash(gcobj)`.
     /// Return a stable address usable as identity hash.  For nursery
     /// objects, returns the shadow's address (which is where the object
@@ -3841,6 +3883,7 @@ impl MiniMarkGC {
                 let holder_hdr = unsafe { header_of(holder_addr) };
                 if unsafe { !(*holder_hdr).has_flag(flags::PINNED) } {
                     self.old_objects_pointing_to_pinned.push(holder_addr);
+                    self.updated_old_objects_pointing_to_pinned = true;
                     unsafe { (*holder_hdr).set_flag(flags::PINNED) };
                 }
             }
@@ -3946,11 +3989,21 @@ impl MiniMarkGC {
             // Remember VISITED and re-apply it after the copy.
             let shadow_was_visited =
                 unsafe { (*(shadow_hdr as *const GcHeader)).has_flag(flags::VISITED) };
+            // incminimark.py `_trace_drag_out`: `if (tid &
+            // GCFLAG_SHADOW_INITIALIZED) != 0: copy = False`.
+            // `move_out_of_nursery` has already filled this shadow, and its
+            // caller has been reading it ever since.  Copying again is not
+            // merely redundant: HAS_SHADOW and SHADOW_INITIALIZED "have no
+            // meaning in non-nursery objects", so the second copy is what would
+            // carry the nursery flag word onto an object that has the right one.
+            let shadow_initialized = unsafe { (*hdr_ptr).has_flag(flags::SHADOW_INITIALIZED) };
             // minimark.py:1518-1520: clear HAS_SHADOW before copy so the
             // flag does not propagate to the shadow object itself.
             unsafe { (*hdr_ptr).clear_flag(flags::HAS_SHADOW) };
-            unsafe {
-                std::ptr::copy_nonoverlapping(header_ptr as *const u8, shadow_hdr, total_size);
+            if !shadow_initialized {
+                unsafe {
+                    std::ptr::copy_nonoverlapping(header_ptr as *const u8, shadow_hdr, total_size);
+                }
             }
             if shadow_was_visited {
                 unsafe { (*(shadow_hdr as *mut GcHeader)).set_flag(flags::VISITED) };
@@ -5816,6 +5869,7 @@ impl MiniMarkGC {
         // VISITED set at this point; the oldgen sweep clears it later.
         self.old_objects_pointing_to_pinned
             .retain(|&obj_addr| unsafe { (*header_of(obj_addr)).has_flag(flags::VISITED) });
+        self.updated_old_objects_pointing_to_pinned = true;
         // incminimark.py:2528-2529 — VISITED still distinguishes survivors, so
         // this is where each mirror learns whether its object made it.
         if self.rrc.enabled {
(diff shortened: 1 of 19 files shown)


youknowone:
Went through all six findings. Two applied in `58bd978b9f4`, four declined with reasons below.

## Applied

**`take_heapsize_option` accepted counts above `sys.maxsize`** (CodeRabbit). Real inconsistency: `parse_heapsize` on the controller side refuses `> i64::MAX`, but the child side took the whole `u64` range and let `usize::try_from(..).unwrap_or(MAX)` clamp it silently. Both sides refuse it now.

**The cpyext drain assertion could be satisfied by an automatic collection** (CodeRabbit). Correct — `assert during > before` did not establish that the *explicit* collection is what drained. Automatic collection is now disabled across the setup and restored afterwards, and the count is read once more after the Owners are built, so the test asserts `settled == before` and then `during > settled`. That also newly confirms an explicit `gc.collect()` still drains while collection is disabled.

## Declined

**`Parser::from_iter` does not skip `argv[0]`** (Codex, P1). This is not what lexopt does. `from_iter` is `Parser::new(args.next(), args)` and `from_env` is `Parser::new(source.next(), source)` — the same constructor, both consuming the first element as `bin_name`; the doc says *"The first item from the iterator **must** be the binary name"*. `from_args` is the one that does not skip, and it is not what this code calls. Positive control: `pyre --heapsize 104857600 -c 'import sys; print(sys.argv)'` prints `['-c']` and runs. Were this real, every `-c` and script invocation would break, and the 460-fixture and 100-snippet gates would not start.

**`move_out_of_nursery` should re-grey its shadow onto `more_gray_stack`** (CodeRabbit, flagged as the merge blocker). No upstream basis. `incminimark.py` `_allocate_shadow` allocates through `_malloc_out_of_nursery`, copies `header(obj).tid` into the shadow, registers it in `nursery_objects_shadows` and returns — it touches no gray stack and has no `STATE_MARKING` branch, and neither does `move_out_of_nursery`. The shadow does not become the object until the next minor; until then the marking phase traces the nursery object itself. The port here is line-for-line upstream.

**`manually_copy_card_bits` needs a length/bounds assertion** (CodeRabbit). Upstream carries exactly one assertion in that function, `ll_assert(self.card_page_indices > 0, "non-positive card_page_indices")`, which is ported. Adding a second has no upstream basis.

**Move `maybe_unregister_rpython_finalizer_io` before the stored `flush_error` return in buffered close** (CodeRabbit). This would move away from upstream, not toward it. `interp_bufferedio.py` `close_w` has `self.maybe_unregister_rpython_finalizer_io(space)` on the line immediately after `self.buffer = None`, and both sit *after* the `try/finally`. When flush raises, the `raise` in the `except` clause propagates and upstream runs **neither**. Skipping the unregister on that path is what matches upstream on the finalizer axis, which is the axis this PR changes.

That said, the review did point at a real deviation next to it, just not the one it named: pyre clears `self.buffer` before returning the stored flush error, where upstream does not reach that line at all. It predates this PR, and correcting it stales the LLBC closure, so I have left it and am recording it here rather than folding it in silently.

## Gates

Re-run on the current base (`1eeff808c82`), which is why the fixture count reads 460 rather than the 453 in the description.

| lane | result |
|---|---|
| `check.py --synthetic-only` | 460/460 × dynasm, cranelift — ALL PASSED 2/2 |
| `extra_tests/run.py --gated-only` | 100/100 × cpython, dynasm, cranelift |
| `cpyext_types` (dynasm,cpyext) | 2 passed |
| `cargo test -p pyrex --lib` | 19 passed |
| `cargo fmt --all -- --check`, citation gate | clean |

— *commented by Claude*

youknowone:
Follow-up to the triage above — the deviation I said I was leaving is now fixed, and the two questions I left open are settled from code.

## `de139bbb787` — the buffered-close deviation

`interp_bufferedio.py close_w` reaches `self.buffer = None` and the `maybe_unregister_rpython_finalizer_io` call on the line after it only once the `try`/`finally` has completed, so a flush that raised carries its error out past both; the raw close in the `finally` is all that still runs. pyre returned the stored flush error *after* releasing the buffer, so it released a buffer upstream keeps. The flush-error return now comes first and neither line runs on that path. The unregistration was already behind it — only the release was not, which is why the original review comment pointed the other way.

`interp_textio.py close_w` splits the same way through `except … raise` / `else` and already matched, so it is untouched.

Behaviour is unchanged at app level, as expected for an ordering fix between two internal releases. Spot-checked against CPython 3.14 on the exact path (raw whose `write` raises, so flush fails and raw close succeeds): both print `first close raised: flush boom` / `second close quiet` / `closed: True`. I did not add a gate snippet for it — what this changes is buffer release and finalizer registration, neither of which is app-observable, so a passing snippet would prove nothing.

## The items-block question, settled

I said the `object_array.rs` fallback needed the tracing question answered before I would touch it. Answered: `list_object_custom_trace` carries an explicit regime split on `try_gc_owns_object(list.items)` — a GC-owned block is forwarded through the `items` field slot, and a `std::alloc` block has **each live element forwarded in place**. So that fallback is fully traced and is genuinely not the untraced-managed-reference hazard. No change needed, now on evidence rather than caution.

## Why the allocation-outcome fix is inert, settled

`oom_pending` is `mem::take`n in exactly two functions: `alloc_with_type_slow` and `alloc_with_type_rooted_slow`, the two *collecting* nursery allocators. `alloc_with_type_no_collect` never reads it, and the stable/old-gen path cannot fail anyway — `alloc_in_oldgen_clear` memclears its result unconditionally, so a null there would already segfault, and only the size-overflow guard produces one.

The interpreter's dominant allocation paths are exactly the two that do not observe the flag. So under a heap limit `oom_pending` is set and never read: nothing returns null, the program keeps allocating, and the next major takes `max_heap_size_already_raised`. That is the whole mechanism, and it is why the classification fix changes no observable behaviour — the null is never produced on those paths, not that the classification is wrong.

Reviving the first event's `MemoryError` therefore means making the no-collect and stable paths observe `oom_pending` as well. That is separate work and not in this PR.

## Gates

Re-run on `1eeff808c82` after the change.

| lane | result |
|---|---|
| `check.py --synthetic-only` | 460/460 × dynasm, cranelift — ALL PASSED 2/2 |
| `extra_tests/run.py --gated-only` | 100/100 × cpython, dynasm, cranelift |
| `cpyext_types` (dynasm,cpyext) | 2 passed |
| `cargo fmt --all -- --check`, citation gate | clean |

— *commented by Claude*

PR #1473, opened 2026-08-25
jit-trace: descent blocker reporting, zero-length aggregate folding, and five fold retirements

youknowone:
Rebased onto `main`. Fourteen commits from the descent-wall work, plus one conflict resolution that repairs a compile break already on `main`.

## `main` does not currently compile

`df55038caae` (#1435) added a `load_deref` row to `SPEC_FOLD_ROWS` without moving the declared length, so `main` declares `[(&str, &str, &str); 72]` against a 73-row literal. That is the same defect `ad34e40112f` had just fixed for the two rows before it. The conflict in `diag.rs` landed on that constant, and it is resolved to the literal's actual post-retirement length (73 upstream rows − 5 retired here = **68**) rather than to either side.

## What is in the branch

**Descent diagnostics and reach**
- `jit-trace: report every un-lowered helper a declined descent could reach` — the blocker log names the whole set, not just the first symbol in byte order.
- `jit-trace: add a kill switch for the descent's un-lowered-helper scan`, and `bound the canonical-helper descent's host recursion` (the descent recurses on the host stack; corpus max depth is 4).
- `jit_fnaddr: publish through signature-checked helpers`, and `publish abs_uses_builtin and w_bytearray_find`.

**Lowering**
- `jtransform: allocate the enum base for a payload-less sum variant`, `lower the niladic Option aggregate ctor to an allocation`, `fold the r_uint signedness marker like intmask`.
- `rtyper: fold a zero-length shaped aggregate to a prebuilt instance` — `Tuple<>` / `Array<T; 0>` were gc-allocating a zero-field descr carrying no type id; upstream allocates nothing.

**Extraction**
- `llbc: extract with debug-assertions off` — under the dev profile 325 `debug_assert!`s were being translated into real jitcode ops.

**Interpreter**
- `baseobjspace: build get_and_call_function's reloaded args as a Vec` — the fixed `[PY_NULL; 8]` leg ended in `<[T; N]>::index(&array, ..len)`, a `RangeTo` subslice that stays an inherent-impl call with no registry entry, dropping this graph and every dunder dispatch reaching one through it to the legacy walker.

**Fold surface**
- `jit-trace: retire five gate-neutral specialization folds` — `builtin_isinstance`, `builtin_issubclass`, `builtin_type`, `builtin_hasattr`, `load_type_name_attr`. Suppressing each is gate-neutral, so each was a shortcut over a path the tracer already handles. `SPEC_FOLD_ROWS` 73 → 68.
- `bench/synth: declare spec-folds on five fixtures`.

## Verification

`python3 scripts/extract-llbc.py majit-rlib pyre-object pyre-interpreter pyre-jit` → `cargo build --release -p pyrex` → `python3 pyre/check.py --backend dynasm`: **ALL PASSED, dynasm 473/473**.

https://claude.ai/code/session_019gCeUzbGWCK8S6vqnXh416

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect results when optimizing tuple subscripts inside hot loops.
  * Corrected construction and handling of `Option` and `Result` values, including empty variants.
  * Preserved integer identity for unary positive operations on builtin integers.
  * Improved handling of zero-length tuples and arrays.

* **Performance**
  * Added optimized support for tuple indexing and type attribute access.
  * Improved specialization and fallback behavior for compiled code.

* **Documentation**
  * Updated optimization configuration guidance and benchmark coverage.
  * Added controls and diagnostics for bounded optimization analysis.


youknowone:
Rebased onto `origin/main` and pushed four more commits. `pyre/check.py --backend dynasm` is **476/476 ALL PASSED** at the tip.

**`majit: decline the short-preamble inline when a used_box producer is dropped`** — a silent wrong-answer fix, pre-existing on `main` and not introduced by this branch. `t += c[0] + c[-1]` with `c` a tuple parameter printed `14704252783379614800` instead of `15000000`.

`ExtendedShortPreambleBuilder::setup` drops a pure short-preamble op whose args do not resolve, and declines the whole inline for a guard, an ovf producer, or an op `short_preamble.jump_args` carries. A `used_boxes` entry took none of those arms — it is carried by the extended LABEL (`jump_to_existing_trace` builds `current_label_args` as `label_args + used_boxes`), not by the short preamble's own JUMP. Because the box is then in `inputargs_set`, every later short op that reads it stays "resolvable" and survives its producer's drop, and `inline_short_preamble` resolves it through the LABEL-slot seed rather than a replay result. The slot holding the invariant's unboxed `intval` received the accumulator on the back edge — `Label(…, v53)` closed by `Jump(…, v257)`.

Upstream has no inline-time drop at all (`_map_args` raises `KeyError`), so declining and falling back to `jump_to_preamble` is the orthodox answer. Regression fixture `pyre/bench/synth/tuple_mixed_sign_subscript_invariant.py`, verified red on a build with the fix reverted.

**`jit-trace: serve a specialised-pair subscript by descending w_tuple_getitem`** (+ its shared plumbing commit) — replaces the hand-emitted item read with a real sub-jitcode descent. The three arity-2 specialisations come out op-for-op identical to the hand fold. The canonical layout is deliberately left on `try_walker_specialize_subscr_tuple`: descending there looked better on a straight loop (108 → 69 ops) but cost a compiled bridge and ran 2.4x slower in `selfrec_bridge_nontail_promote`.

**`objspace: return the receiver from pos for an exact int`** — the `is_long` arm already spelled this rule; the machine-int arm rewrapped unconditionally. No observable change (pyre's int identity already reports equal-valued ints as the same object); verified against pypy3 across magnitudes, an int subclass, an overridden `__pos__`, both bools, float, complex, and the TypeError path.

— *commented by Claude*

PR #1475, opened 2026-08-25
wasm: lower a vouched void residual and widen the in-guest allow-list; jit: fold `s[i]` and `repr(int)`, and three defects the folds exposed

youknowone:
Fifteen commits on top of `main`, in three groups: two new trace-time folds, the wasm backend's residual-call lowering, and the defects the folds reached once they were exercised.

---

# 1. Two folds

## `jit: fold s[i] on an exact str to a jit_str_getitem call`

`try_walker_specialize_subscr` recognized dict, tuple, specialised-pair and list receivers; a `str` fell through to the generic `CallMayForce`, which forces virtualizables and clears the heap cache across itself. Measured on dynasm with the identical loop body over each receiver type (5M-iteration function-level loops, min of 5 user-CPU runs, startup-subtracted, ns/iter):

| receiver | ns/iter |
|---|---|
| `list[int]` | 1.8 |
| `list[str]` via `ord(seq[0])` | 2.2 |
| `tuple[int]` | 4.2 |
| `bytes[int]` | 24.3 |
| `str[int]` | **51.1** |

`jit_str_getitem` is the scalar arm of `descr_getitem` restricted to an exact `str` receiver and a machine-int index; it declines everything else with `PY_NULL`, so `IndexError`, `__index__` coercion and negative indices stay in the interpreter and the trailing non-null guard carries the decline back. Both operands are guarded exactly: a `str` subclass may override `__getitem__`, and a `bool` index shares `int`'s `intval` while carrying its own type.

The fold removes the may-force residual and nothing else: **51.1 → 45.5** ns/iter, against pypy's 0.7. The remaining 45 is the boxed code point — `len(s[0])` reads 44.3 and `ord(s[0])` 45.0, so the cost does not depend on the consumer, and pypy never allocates it at all. Upstream `_getitem_result` allocates a fresh `W_UnicodeObject` the same way; virtualizing it is a separate lever, so `str_subscr_hot` gates the fold with `spec-folds=` and carries no throughput ceiling.

`bytes[int]` keeps the residual.

## `jit: answer repr(int) through the same int-render fold as str(int)`

`repr(i)` on an exact `int` renders the same decimal text `str(i)` does, and took the opaque `bh_call_fn` residual where `str(i)` already folds to `jit_int_str` (cpython / pypy / dynasm, ns/iter):

| | cpython | pypy | dynasm |
|---|---|---|---|
| `acc ^= len(str(i))` | 43.8 | 7.7 | 74.1 (already folded) |
| `acc ^= len(repr(i))` | 40.0 | 7.2 | **100.0** → 72.5 |

`repr` is bound as a `BuiltinCodeFn` rather than a type object, so the callable check takes a second identity rather than a second entry point; `is_repr_builtin` names it by the wrapped code, so a rebound global keeps the residual. The existing cross-check against `int_str_text` holds the two callables to the same answer.

Not a `BUILTIN_FOLDS` row: every row's helper answers without allocating, and rendering an int builds a string.

## `bench: declare the folds three sole-exerciser fixtures already fire`

A corpus census (all 461 synth fixtures under `PYRE_FBW_SPEC_CENSUS=1`) shows 70 fold labels firing somewhere and 22 firing in exactly one fixture. Three of those sole exercisers carried no `# pyre-check: spec-folds=` line, so a label that stopped firing would have gone unnoticed everywhere: `math_folds_hot` (8 labels), `pickle_terminal_raise_resume` (2), `getframe_method_call_residual_body_once` (1). Each declaration is arm-checked in both directions.

---

# 2. wasm residual calls

## `wasm: vouch jit_mapdict_unboxed_read_f's call descr as its real ABI`

The helper is `(i64, i64, i64) -> f64` and its recorded descr maps to the same signature, but a mixed float/word shape is outside the backend's arity-keyed `call_indirect` family, so every call went out through the `jit_call` host trampoline. 400000 iterations of `acc += p.x` over a float attribute: **398958 crossings before, 0 after**. The byte-identical loop over an int attribute crosses 0 both times, and dynasm shows no float-vs-int gap at either point.

## `wasm: lower a vouched void residual over its descr's parameter types`

`(i64, i64, i64, f64) -> ()` had no arm at all — the typed family returned `None` on a void result and the uniform word family cannot carry a float parameter. `TypedResidualSig` now spells the result `Option<ValType>` and the predicate admits the void-recorded CALL opcodes.

The predicate also now declines an op the i64, void-word or true-void family can express, rather than leaving that to the emit site's arm order: the emit site tried those three first while the signature-collection loop did not, so an op both could claim collected a type index nothing branches to. That set was empty before and stays empty after.

4000000 iterations of `c.v = 1.5`: **3998959 crossings and 0.69s before, 1 crossing and 0.03s after**, against the int store's 1 crossing and 0.02–0.03s at both points.

## `wasm: name the callee behind each PYRE_WASM_CALL_HIST slot` / `key the call histogram by importer`

The probe printed a bare function-table slot, which named nothing without an external disassembler, and it keyed on that slot alone — so the two populations that funnel into `jit_call_trampoline` summed into one row. Slots now resolve through the module's element and name sections to a demangled symbol, and the histogram keys on the importer (`src=trace` for `env::jit_call` / `env::jit_call_compact`, `src=host` for `pyre_jit::jit_call_host`). A run with the probe off reads and parses nothing.

## `wasm: extend the in-guest uniform-i64 allow-list from 5 callees to 17`

With the probe fixed, the census over the 468-fixture synth corpus is decisive: every residual host crossing arrives on the in-guest path (`src=host`, 537113) and **none from a compiled trace** (`src=trace`, 0) — so the codegen-side work above was complete and the whole remaining cost lived on a path codegen never sees. The twelve callees added here are the heaviest of that population.

| | before | after |
|---|---|---|
| corpus host crossings | 537113 | **100560** (−81%) |
| the twelve added callees | top of the histogram | absent |

Attribution is clean because the post-extraction, pre-patch census came back byte-identical at 537113.

Each addition is declared `pub extern "C" fn(i64, ...) -> i64`, which is what makes comparing `fn as usize` prove the ABI and not just the identity; the list stays an exact-function allow-list and a mismatched arity still falls through to the reflective path. The two void-returning callees the census also names are left out — the helper's result is `Option<i64>` and a void callee's result handling is not settled here.

---

# 3. Three defects the folds reached

Two came from the Codex review on this PR and one from CI; all three were judged by measurement, on `pyre-dynasm` with the JIT on against the same binary with `PYRE_NO_JIT=1`, plus CPython and pypy3 as references.

## A str index wider than `usize` — two sites

`jit_str_getitem` cast its `i64` index with `as usize`. The interpreter's machine int is 64-bit on every target (`sys.maxsize` is `i64::MAX` on the wasm guest too), so `2**32` reaches the helper as a valid exact-int operand and truncates to `0` on wasm32, answering `s[0]` where an `IndexError` is owed.

The fixture leg added alongside the fix then failed on wasm anyway, and reached a second site: `getitem_str` in `baseobjspace.rs` carried the identical cast, so `"abcde"[2**32]` answered `'a'` on the guest even with a cold trace. Swept the other sequence types on the guest — `bytes`, `bytearray`, `list`, `tuple`, `range` all already raise — so `str` was the only one.

## A shared render is visible where `is_w` compares storage

`descr_repr` renders with the `@jit.elidable` `ll_int2dec` and wraps the characters in a **separate** `space.newutf8` allocation; `jit_int_str` performs both in one call, and all three channels that named its effect called it elidable. The pure pass then shared one call between two renders of the same operand, and `is_w` gives a `str` of `_len() > 1` storage identity — so `str(int)`, already on `main`, had the same defect as the `repr` arm added here.

100000 iterations at `i = 123456`:

| | `repr(i) is repr(i)` | `str(i) is str(i)` |
|---|---|---|
| CPython | 0 | 0 |
| pypy3 | 0 | 0 |
| pyre, plain eval loop | 0 | 0 |
| pyre, compiled trace | **98558** | **98558** |

All three channels — the helper attribute, `jtransform`'s `UnaryOp { op: "str" }` lowering, and the walker fold — now record `EF_CAN_RAISE`, and the compiled trace reads 0. Recovering the elidable half needs the render and the wrapper split into two ops, the shape `emit_box_long_inline` already gives the bigint arms; that is follow-up work, because `W_UnicodeObject` carries derived `byte_len` / `len` fields that upstream reads off the returned `rstr` and pyre's `Wtf8Buf` value box has no descr for.

The same review flagged `s[i]` for the same reason, and that one is a **false positive**: the result is one code point, `is_w` unique-ifies a `str` of `_len() <= 1` by its WTF-8 bytes, and all four of CPython, pypy3, pyre's eval loop and pyre's compiled trace answer `is` True 100000/100000. That call stays pure — it only moves to `EF_ELIDABLE_OR_MEMORYERROR`, which is what `getcalldescr` picks when the only failure is the allocation.

## `bench: print str_subscr_hot's wide code points as integers`

`check.py` removes `LANG` / `LC_ALL` / `LC_CTYPE` / `PYTHONIOENCODING` from a measured child, so the Windows runner resolved the fixture's piped stdout to the ANSI codepage and the U+4E2D in the last line raised `UnicodeEncodeError`, failing the dynasm and cranelift lanes with "cpython crash". It was the only fixture in the synth corpus printing a non-ASCII character.

---

# Verification

`pyre/check.py`: **dynasm 486/486, cranelift 486/486, wasm 478/478**, RC=0. The only banded note is the pre-existing `inline_freevar_after_mayforce` drift, unchanged.

`str_int_render_hot`'s three baselines move by `guard_failures 404 → 405, loops_compiled 4 → 5`. Control: running the same binary with the new identity leg removed reproduces the recorded 404 / 4 exactly and reads 8.6x pypy against the fixture's ceiling of 15 — so the `+1 / +1` is the added loop and not the effect reclassification.


youknowone:
## Codex review dispositions

All three findings were judged by measurement rather than by reading, on `pyre-dynasm` with the JIT on against the same binary with `PYRE_NO_JIT=1`, plus CPython and pypy3 as references.

### 1. "Reject indices that do not fit in wasm32 usize" — **confirmed, fixed**, and it was two sites, not one

`jit_str_getitem` cast its `i64` index with `as usize`. The interpreter's machine int is 64-bit on every target — `ALLOW_UNBOXING_INTS`' own comment in `mapdict.rs` spells this out, `sys.maxsize` is `i64::MAX` on the wasm guest too — so `2**32` reaches the helper as a valid exact-int operand and truncates to `0` on wasm32. `usize::try_from` now does the conversion, which subsumes the negative-index check it replaces.

The fixture leg added alongside it then failed on wasm *anyway*, and reached a second site: `getitem_str` in `baseobjspace.rs` had the identical cast, so `"abcde"[2**32]` answered `'a'` on the wasm guest even with a cold trace. Swept the other sequence types on the guest for the same shape — `bytes`, `bytearray`, `list`, `tuple`, `range` all already raise — so `str` was the only one.

### 2. "Preserve fresh identities for repeated string subscripts" — **false positive**

`s[i]` returns a single code point, and pyre ports pypy's `is_w`: a `str` with `_len() <= 1` compares by its WTF-8 bytes, not by storage. Two separate allocations therefore already answer `is` True, so sharing one is not observable.

Measured over 100000 iterations of `a = s[2]; b = s[2]; a is b`:

| | shared |
|---|---|
| CPython | 100000 |
| pypy3 | 100000 |
| pyre, plain eval loop | 100000 |
| pyre, compiled trace | 100000 |

The call stays pure. It does move to `EF_ELIDABLE_OR_MEMORYERROR`, which is what `call.py:292-299` picks when the only failure is the allocation, and the fold's note now records the invariant above instead of asserting that identity is not readable.

### 3. "Keep repr(int) out of the identity-changing render fold" — **confirmed, and wider than `repr`**

Same probe on `repr(i)` / `str(i)` at `i = 123456`, 100000 iterations:

| | `repr(i) is repr(i)` | `str(i) is str(i)` |
|---|---|---|
| CPython | 0 | 0 |
| pypy3 | 0 | 0 |
| pyre, plain eval loop | 0 | 0 |
| pyre, compiled trace | **98558** | **98558** |

`_len() > 1` gives storage identity under `is_w`, so the shared box is visible. The root cause is structural rather than the `repr` admission: `descr_repr` (`intobject.py`) renders with the `@jit.elidable` `ll_int2dec` and wraps the characters in a **separate** `space.newutf8` allocation, while `jit_int_str` performs both in one call. Marking that pair elidable is what let the pure pass share it — so `str(int)`, already on `main`, carried the same defect.

All three channels that named the effect now record `EF_CAN_RAISE`: the helper attribute, `jtransform`'s `UnaryOp { op: "str" }` lowering, and the walker fold. The probe reads 0 on the compiled trace afterwards, and `str_int_render_hot` gains the leg so it stays that way.

Recovering the elidable half needs the render and the wrapper split into two ops — the shape `emit_box_long_inline` already gives the bigint arms. That is filed as follow-up work rather than done here, because `W_UnicodeObject` carries derived `byte_len` / `len` fields that upstream reads off the returned `rstr` and pyre's `Wtf8Buf` value box has no descr for.

### Control for the re-recorded baselines

`str_int_render_hot`'s three baselines move by `guard_failures 404 -> 405, loops_compiled 4 -> 5`. Running the same binary with the new identity leg removed reproduces the recorded `404 / 4` exactly and reads 8.6x pypy against the fixture's ceiling of 15 — so the `+1 / +1` is the added loop and not the reclassification.

### Verification

`pyre/check.py`: dynasm 486/486, cranelift 486/486, wasm 478/478, RC=0. The only banded note is the pre-existing `inline_freevar_after_mayforce` drift, unchanged from before these commits.

— *commented by Claude*

PR #1479, opened 2026-08-25
Portal trace bracket nesting, and the review findings #1470 left open

youknowone:
Follow-up to #1470, which merged while its review was still being worked
through. Seven commits: one revert, one behaviour fix with its own gate, and
the review findings that were still open.

## The revert

`2699b190308` "jit: root a ref constant from the moment the assembler bakes it"
is reverted. Three defects were raised against it and none is answered:

* it writes through a `&[ConstSlot]` obtained from a shared reference —
  `walk_jitcode_constants_refs_in` uses `ConstSlotR::set` precisely because
  `Vec::as_ptr() as *mut` would be a write through shared-reference provenance;
* it patches by nursery address, so an address reused after a collection
  (ABA) mis-patches an unrelated slot;
* a drain that bakes a reference but publishes no jitcode never clears the
  roots it took, so they are retained without bound.

It is reverted for those three, **not** as a crash fix — measured, the tree at
`238091009c1` carries the same commit and its CI run `32832823865` was green
including the CPython-suite job.

## The behaviour fix

The portal carried `execute_frame`'s bracket as three `?`-terminated calls, so a
trace hook that raised took the hooks after it down with it. `execute_frame`
nests them instead: `call_trace` in the outer `try`, `return_trace` in the inner
`finally`, `leave` in the outer one.

Measured with a profiler and a trace function both installed on a compiled loop,
profile events for the loop frame:

| trace hook raises at | before | cpython 3.14.6 | pypy3 |
|---|---|---|---|
| `call`   | `[]`        | `['return']`          | `['return']` |
| `return` | `['call']`  | `['call', 'return']`  | `['call', 'return']` |

`a_raising_trace_hook_still_owes_the_leave_event.py` is red before the fix on
both arms and green after.

## Review findings

* **Refuted by measurement, then fixtured.** The claim that a trace callback
  installing `sys.setprofile()` from the `call` event loses the loop's
  `c_call` / `c_return` to folding: measured 2500/2500 and 10000/10000 at two
  tails, matching cpython 3.14.6, whole event set matching pypy3. Kept as
  `a_profiler_installed_from_a_call_event_keeps_c_events.py` because nothing
  covered it.
* **Refuted by measurement.** The claim that the portal bypasses
  `ExecutionContext::enter` so callbacks see `f_back is None` and escaped exits
  skip cleanup: caller chain 400/400 correct, exception traceback
  `['<module>', 'raiser']`, and a frame captured out of a compiled callee still
  walks `['leaf', 'hot', '<module>']` after further compiled runs — cpython,
  pypy3 and pyre agree.
* **Applied.** `with_typed_decision_key`'s recovered `is_being_profiled = true`
  path now has a test asserting the installed cell's `comparekey` carries
  `values[1] == 1`. `pypy_hidden_applevel_takes_only_code` no longer swallows
  `TypeError` for arguments that are accepted, and exercises `object.__init__`
  after marking it. `global_tracer_armed_before_a_hot_loop`'s global arm asserts
  the callee reports no `return` or `line` events. Two upstream citations name
  the symbol rather than a line number.

`pyre/check.py`: dynasm 485/485, cranelift 485/485, wasm 477/477, 3/3 backends.

https://claude.ai/code/session_01QCct5Z2Yn3Be6Pz22GgNZw

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved profiling and tracing behavior for JIT-compiled code, including reliable call, return, and builtin events.
  * Preserved correct exception and cleanup behavior when trace hooks raise errors.
  * Fixed recovery of profiled loop data during JIT compilation.
  * Clarified which callable types are accepted by hidden application-level handling.

* **Refactor**
  * Simplified constant management and JIT frame execution while preserving runtime behavior.

* **Tests**
  * Added regression checks covering profiling, tracing, exception handling, and callable classification.


youknowone:
Review triage. One finding was real and is fixed; one is unmeasurable with the instrument that would settle it; one was already refuted.

## §2 — fixed in `e4d3d608f0a`

Confirmed by measurement. The bracket read `w_exitvalue` back off its own result word, so a `return_trace` that raised turned the result into `Err` and the leave hook was handed `None`. `pyframe.py execute_frame` keeps `w_exitvalue` as a local across the inner `finally`.

Profile events for the loop frame, trace hook raising at `return`:

| | `call` arm | `return` arm |
|---|---|---|
| before | `[('return', None)]` | `[('call', None), ('return', **None**)]` |
| after | `[('return', None)]` | `[('call', None), ('return', **100**)]` |
| pypy3 | `[('return', None)]` | `[('call', None), ('return', 100)]` |
| cpython 3.14.6 | `[('return', None)]` | `[('call', None), ('return', None)]` |

cpython and pypy3 disagree here, so the fixture is graded against pypy3. `a_raising_trace_hook_still_owes_the_leave_event.py` now checks the argument as well as the event, and is red on the pre-fix binary.

## §1 — the structural claim is right, but the fix cannot be restored and the hazard is currently unmeasurable

The gap is real: `rpython/jit/codewriter/assembler.py` appends the ref to `self.constants_r`, a list hanging off a GC-traced `Assembler`, so upstream's build window **is** traced; pyre's `Vec<ConstSlotR>` is not until `install_jitcodes`.

Un-reverting is not the answer. `2699b190308` carries three defects, none of them addressed:

* `patch_build_const_refs` writes the pool through `jitcode.constants_r.as_ptr() as *mut i64` — a write through shared-reference provenance. `walk_jitcode_constants_refs_in` names that exact construction as the thing not to do, which is why it goes through `ConstSlotR::set`;
* `note_build_const_ref` deduplicates by raw address, so a nursery address reused after a minor collection leaves the new constant neither rooted nor patched, and the patch then rewrites its slot to the **old** object's forwarded address;
* a drain that publishes nothing never reaches `patch_build_const_refs`, so its entries stay rooted until some later drain does.

Reachability of the window is unproven, and I could not settle it:

* `PYRE_PROBE14=1` on the snippet that commit shipped as its own demonstration reports `RELOCATION APPLIED #1: constants_r slot 0x6fc05f070 -> 0xbd00306b8 (walk 1, pool_slot 5/7)`. The slot still held its **nursery birth address** at walk 1, so no collection crossed bake→publish on that body; the published-pool walker then forwarded it correctly.
* GC stress cannot grade it. The stress sites are `#[cfg(feature = "gc_stress")]` and `pyre/pyrex/Cargo.toml` ships the feature commented out, so `MAJIT_GC_STRESS=1` on a normal release binary is silently inert — control: `walk=34` with and without the flag. Built with the feature on, the flag is live (`walk=27` → `walk=650`) but the binary segfaults on `print("x")` (rc=139) before any JIT, and on the JIT snippet exits 138 with `slots_cumulative=0`, i.e. nothing was ever compiled.

So the revert stands on the three defects, and the hazard is left open rather than papered over with more unverified rooting machinery. A correct fix roots each baked ref on the **shadow stack**, remembers the slot index, **reads the forwarded address back** at publication and writes it via `ConstSlotR::set`, and clears the region from a `Drop` guard so an empty drain releases it too — no raw-address dedupe, no provenance violation, no retention on the empty path.

## §3 — pre-existing, and its observable predictions do not hold

Marked pre-existing by the review itself. Measured across cpython 3.14.6, pypy3 and pyre: the caller chain a hook sees is correct 400/400 (now asserted by `global_tracer_armed_before_a_hot_loop.py`), the exception traceback reads `['<module>', 'raiser']`, and a frame captured out of a compiled callee still walks `['leaf', 'hot', '<module>']` after further compiled runs. All three agree.

`pyre/check.py`: dynasm 485/485, cranelift 485/485, wasm 477/477.

— *commented by Claude*

PR #1481, opened 2026-08-25
jit-trace: adopt a residual's fastlocals writes into the walk, on both the forcing and the non-forcing arm

youknowone:
An `f_locals` write made by a function the tracer is *running* — `sys._getframe(1).f_locals["x"] = v` inside a callee — was lost on the iterations the caller's loop is recorded. Both losses are now closed.

## What was wrong

The walk holds each local as a box and reads it out of the virtualizable shadow, the way `pyjitpl.py _opimpl_getarrayitem_vable`'s standard leg does. A residual that writes the frame's fastlocals writes the frame the interpreter holds — which is not the object the walk reads. The `[fbw-post-residual]` probe added here prints exactly that:

```
[fbw-post-residual] origin=flush heap=0x77e70bb28 frame=0x77e63a898 same=false
                    moved=[2] shadow_stale=[2]
```

`same=false` is the finding: `refresh_virtualizable_shadow_from_heap` reads a different object than the residual wrote, so the shadow never learns, and every later read — and the values the loop closes on — carry what the local held before the call.

## Two recordings, two paths to the same loss

| | first recording (iteration 1040) | second recording (iteration 2105) |
|---|---|---|
| the callee | a residual that enters a frame | inlined; the store itself is the residual |
| force | fires, escape, walk aborts | **none at all** |
| how the write dies | the escape resume image is built from the shadow | nothing forces, so `CloseLoop` hands the compiled loop the stale box |

The second arm forces nothing because `framelocalsproxy_setitem`'s `force_locals` returns early on `(*frame).vable_token == 0`, and the walk arms that token on its snapshot rather than on the live frame. The store still lands on the array; only the escape is missing.

## The fix

`adopt_residual_locals_writes` reads the moved slots back out of the frame's `locals_cells_stack_w` after such a residual, as a recorded `GETARRAYITEM_GC_R` rather than a folded constant, so compiled code performs the same load after its own force. It now runs on both arms — the not-forcing one placed before `restore_escape_flush_undo`, where the write is still in the live frame and the restore holds that slot back anyway.

What makes the re-read sound is not the force but `moved`: the diff is taken against an image of the region captured before the call, so only slots the call itself changed are adopted, and a slot the walk is ahead on is never in the set.

**The baseline must be the flush image whenever one is armed for the frame.** Diffing against the pre-call image instead adds the *loop variable* to `moved` — the flush writes it, the residual does not — and the compiled loop then re-reads its own counter out of the array and runs one iteration ahead. Measured: dynasm went 1 → 3490 failures while cranelift was fixed by the same change, which is what a two-backend run is for.

## GC

The pre-call image is held in shadow-stack slots for the duration of the call (`gctransform/framework.py` push_roots/pop_roots), not in a plain `Vec`. The residual is a collection point, and a minor collection forwards the frame's own array in place while copies outside the root stack keep their pre-collection words — every surviving slot would read as a write.

## Verification

- `flocals_writethrough_forcing_residual.py`, bounded at 4000 rounds so it clears both recordings: `OK` on both backends, and with `PYRE_FBW_NO_ADOPT_RESIDUAL_LOCALS=1` it reports `2 of 4000, first: [(1040, 1039), (2105, 2104)]` on each. CPython and `PYRE_JIT=0` both `OK`. Passes to 20000 rounds.
- `check.py --backend dynasm,cranelift`: **484/484 on each, FAIL 0.**
- `cargo test --all --features dynasm`: 181 suites, **8346 passed, 0 failed.**

## Also here

- `PYRE_FBW_NO_ADOPT_RESIDUAL_LOCALS` gets its `gate-triage.md` row; `every_live_gate_has_a_triage_entry` was failing without it.
- `SPEC_FOLD_ROWS` drops the row count from its type. `origin/gc-decouple` carries the same change with a better doc comment — drop this commit if that lands first.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved preservation and write-through behavior for local variables during traced and residual execution.
  - Enhanced rollback handling so local values remain consistent across exceptional and resumed execution paths.
  - Improved moving garbage-collection support for pending state.

- **Diagnostics**
  - Added optional debugging information to identify local values involved in aborts and execution discrepancies.
  - Added a diagnostic accessor for inspecting tracked heap state.

- **Tests**
  - Added coverage for local-variable write-through behavior during long-running traced loops.

- **Documentation**
  - Documented local-variable materialization behavior and updated feature-gate guidance.


youknowone:

**§1 (parity regressions): 1 reclassified, 0 fixed.**
The flagged item is the write-through `f_locals` proxy answering the same way under the tracer as it does in the interpreter. PyPy's `getdictscope()` has no write-through because PyPy predates PEP 667; pyre's `FrameLocalsProxy` is pre-existing (the review files it under §3 itself), and the new fixture already declares the divergence in its header (`# pyre-check: pypy-diverges`). What the patch changes is that the JIT stops disagreeing with pyre's own interpreter. Not a regression.

**§2a (unrooted pre-call image): fixed — 5aebbe6c9b5.**
The finding is real, though not for the stated reason: `PyObjectRef` is `*mut PyObject`, so the comparison is pointer equality and never dereferences. The actual hazard is that the residual is a collection point — a minor collection forwards the frame's own `locals_cells_stack_w` in place while a plain `Vec` of copies keeps the pre-collection words, so every surviving slot reads as a write. In the arm where the residual does not force, the walk's shadow can be ahead of the frame, and the adoption would then put the array's value over a box the frame was never told about.

The copies now go through shadow-stack slots held across the call — the `push_roots`/`pop_roots` idiom the review cited, already used in `state.rs` around `FrameBox::new`. Verified: the regression fixture is `OK` on both backends with the adoption on and still fails at iteration 1040 with it off.

**§2b (`GetfieldRawI` on the locals array): reclassified to §3, deferred.**
`frame_locals_cells_stack_array` carries an in-code TODO naming `GetfieldGcR` as the orthodox emission and the still-real blocker: a direct swap panics in `gc_alloc_nursery_shim` on cranelift because the post-getfield write-barrier path reaches a nursery allocation with a missing slow-path stub. The helper has six callers; this patch is the sixth. The review lists the same helper under §3.

**§3/§4:** the two §3 entries are the same two mismatches above, dispositioned there. §4 is empty.

— *commented by Claude*

youknowone:
### Codex parity review, re-run at `a5a9537`

**§2 is now `None`** — the unrooted pre-call image was the only live item and `5aebbe6c9b5` closed it. The `GetfieldRawI` item is no longer filed as introduced here.

**§1 stands as reclassified.** The review's argument for why it cannot be a structural adaptation is that `pypy/interpreter/pyframe.py`'s `fast2locals` path carries `@jit.unroll_safe`. That hint says how PyPy's JIT may trace the loop; it says nothing about whether the proxy writes through. PyPy 7.3.x is a 3.11-level interpreter and predates PEP 667 entirely, so there is no PyPy behavior here to be in parity *with* — which is why the fixture carries `# pyre-check: pypy-diverges`.

The review also names the governing requirement itself (`lib-python/3/test/test_frame.py:327-330`) and files the interpreter-side proxy under §3 as already present on `upstream/main`. So what this patch changes is that the traced path stops disagreeing with pyre's own interpreter on a behavior 3.14 requires.

CI on this tree: 14/14 green.

— *commented by Claude*

PR #1474, opened 2026-08-25
gc: the young raw-malloced generation, and the max-heap MemoryError the dispatch loop owes

youknowone:
Two independent GC slices on the same branch.

## 1. The max-heap `MemoryError` (2 commits)

Under a heap limit (`PYPY_GC_MAX` / `--heapsize`) the **first** breach was silent on every path, so a program's only sign of the limit was the **second** breach's `out_of_memory("using too much memory, aborting")` — the rung that names a Python exception was skipped entirely.

The breaching collection is usually driven by the dispatch-loop safepoint, which returns `()` and cannot raise, so upstream's `raise MemoryError` in `major_collection_step` had no carrier. `EB_MEMORY_ERROR` (bit 5 of the eval-breaker word) carries it: armed where upstream raises, taken by both dispatch loops, raised there through `handle_exception`.

The two channels are mutually exclusive per breach, selected by `reserving_size`. Arming both latches the undelivered one onto the *next* breach — an `oom_pending` no allocation was waiting for is taken by the next unrelated one, failing an allocation that breached nothing.

`pyre/pyrex/tests/heap_limit_memory_error.rs` pins it end to end, on both dispatch loops and on both allocation paths. What it deliberately does **not** assert is that the program survives its handler: after the first breach the live set is at the limit by definition, so whether any given handler fits in the remaining headroom moves non-monotonically with the limit (80/96/112/128 MiB flips a surviving handler to an aborting one and back). The stable observable is that the limit becomes a Python exception.

## 2. The young raw-malloced generation (3 commits)

`OldGen` gains `young_rawmalloced_objects` and its membership set, allocated by `try_alloc_young_rawmalloc` — individual `raw_malloc` at every size, since an arena block must be old — and emptied by `sweep_young_rawmalloced`, which promotes onto `old_rawmalloced_objects` or frees. The collector side is `alloc_young_rawmalloc_clear`, `visit_young_rawmalloced_object` and `free_young_rawmalloced_objects`, wired into `drag_out_root`, the three slot paths in `trace_and_update_object`, and `do_collect_nursery`.

[...]

PR #1494, opened 2026-08-26
jit-trace: seed keyword-only defaults from a namespace mapping, and ask the is-None scan about the operand the branch tests

youknowone:
Three commits: the keyword-only default seeding, a narrowing of the
`is`-against-None inline gate, and a correction to the `w_kw_defs` parity note
the first one carried.

## `interp,jit`: rebuild a definition's keyword-only defaults as a namespace mapping

Ports `Function.init_kwdefaults_dict` so an inlined callee's keyword-only
defaults are seeded from a namespace mapping's cells instead of three residual
calls per default.

| shape | before | after |
|---|---:|---:|
| 1 keyword-only default, module-level callee | 18.33 ns | **0.43 ns** |
| 5 keyword-only defaults | 105.13 ns | **1.40 ns** |
| callee defined in the loop, 1 / 5 | 25.06 / 282.03 ns | **7.97 / 12.69 ns** |

0.43 ns is what the same callee costs with a *positional* default, so the
keyword-only fill is gone rather than merely cheaper.

The seeding stands on three separate mechanisms, each of which had to be got
right:

* The slot that holds the mapping has to be pinned, or a callee defined inside
  a loop reads the first iteration's cells forever. **Correction, per the Codex
  parity review:** an earlier version of this text claimed `w_kw_defs` is not
  among `Function._immutable_fields_` upstream and that there is no
  invalidation to record against. Both are false — `'w_kw_defs?'` is declared,
  and `function_set_kwdefaults` already notifies `QuasiImmutSlot::WKwDefs`. The
  orthodox pin is therefore the quasi-immutable marker `defs_w` uses, not the
  `GuardValue` this emits; the swap waits on the marker install's
  `guards_the_callee_function` gate, which the baked module-level callee this
  fold targets does not satisfy. Behaviour here is correct either way — a
  rebind deopts through the guard instead of invalidating — but the JIT hint is
  stricter than upstream's. Corrected in `a7a9cad42c8`.
* A baked cell belongs to **one** mapping, so without that guard a callee
  defined inside a loop reads the first iteration's cells forever.
* The axis is `heap_cache().is_unescaped(callable_guard_op)`, not
  `guards_the_callee_function`: a trace-allocated callee is a fresh function
  each iteration at a recycled address, and the peeled body's load then
  disagrees with the heap cache.

This change moves no gated `.jitstats` key on its own, so it would have shipped
unguarded. It carries `spec-folds=kwonly_defaults_inline` (suppressing the
resolve declines the whole inline, `loops_compiled` 1 -> 2, so the census and
the counter cross-check each other) plus `kwdefaults_invalidation`, a selfcheck
fixture that moves each of the three mechanisms one at a time — including the
two defects above.  It declares `selfcheck-compiles=hot`, the shape
`synth_selfcheck_compiles` asks of a `selfcheck` marker.

## `jit-trace`: ask the is-None scan about the operand the branch tests

The gate's own comment names the branch's operand — "when the multiframe inline
int-specializes **the tested local**" — but the predicate asks whether **any**
incoming binding lands unboxed. So one int parameter decided the fate of every
other, and `def clamp(v, lo=None)` with `if lo is None:` was declined because
`v` is an int. The binding is also computed outside the per-pc closure, so
whenever it is true the sibling kept-stack arm's `stack_depth_at` test is
masked entirely.

`liveness::branch_operand_local` now names the slot, and answers only when the
producing instruction is a LOAD_FAST of any spelling **and** nothing in the body
writes that slot. Both halves are required, because the answer comes from the
binding the call site supplied.

That second half is load-bearing on three existing fixtures, each of which
would otherwise have begun inlining on an unproven path —
`getframe_residual_callee_own_frame` tests a local assigned from
`sys._getframe()`, `polymorphic_slot_retype` tests one unpacked from a tuple
and then reassigned, `nested_loop_gate_switch` tests a parameter the body
rebinds. All three still decline and their baselines are unmoved.

Measured on dynasm against pypy 7.3.24, interleaved, min of 7 per reading,
three rounds each at load 8:

| shape | dynasm | pypy | ratio |
|---|---:|---:|---|
| `clamp(v, lo=None)`, `if lo is None:` | 5.07 ns | 3.45 | 1.47x |
| same callee, keyword-only `*, lo=None` | 5.36 ns | 5.36 | **1.00x** |
| `read_n(buf, size=-1)` — still declined | 644 ns | 4.59 | |
| same callee with no identity test | 0.54 ns | 0.00 | |

The two new fixtures are the discriminator, on one binary:
`optional_none_arg_inline` (`lo=None`) reads `loops_compiled` 2 -> 1 with
`caro_funcentry` 1 -> 0; `is_none_unboxed_operand_decline` (`size=-1`) stays at
2. Before the change both read 2. dynasm, cranelift and wasm agree field for
field — the scan's third arm, a blanket wasm decline, was removed by
`b99bf57bde4` (#1092).

`test.test_pickletools` passes. It is the **only** surviving oracle for the
miscompile this arm exists for: `test.test_descr`, the other one `a036267b06f`
bisected to, is now baselined FAIL for an unrelated reason (`test_slots` counts
`len(gc.get_objects())` across a loop). That is why the fixture pair is added
rather than relied on from the suite.

The scan's header comment is corrected while here: it claimed the opcode lowers
to an `is`/`is_not` identity residual call, but the codewriter emits a bare
`ptr_eq`/`ptr_ne` whose `Kind::Int` result feeds the exitswitch. What the shape
actually needs is the tested value popped as a Ref, which is what the arm is
about.

## Verification

`check.py` **dynasm 492/492, cranelift 492/492, wasm 484/484**, plus
`test.test_pickletools`, measured at base `53304bdd23e` and re-run in full at
`1579a6a2411` after a concurrent session rebased the branch there.

Now rebased onto `ef42e062da9` and **re-gated there in full**: dynasm 492/492,
cranelift 492/492, wasm 484/484, off a fresh LLBC extract that
`extract-llbc.py --check` confirmed before any leg ran, with `HEAD` unmoved
across all three. Every hunk the branch still carries is byte-identical to what
the earlier runs judged, so the two agree about this diff and the new run also
covers #1474's GC change under it.

That rebase's only casualties were two changes main has since landed itself,
both dropped here: `6ad2595c699` gave `PYRE_GC_SIZE_AUDIT` its `gate-triage.md`
row — in §7, where this branch had put it in §6c — and #1492 gave the two hook
selfchecks the same `selfcheck-compiles=hot` line.

## Not included

Routing `d[k]` through the oopspec-tagged dict arm was investigated and
deliberately **not** done. PyPy emits one CSE-able `ll_call_lookup_function`
plus a `getinteriorfield_gc_r`; pyre's arm that actually runs
(`walker_probe_exact_dict_hit`) emits one call tagged `OopSpecIndex::None`, so
the faithfully ported `_optimize_call_dict_lookup` can never fire on a
subscript — while the only `DictLookup`-tagged arm, which that probe shadows,
costs **three** calls because the index must be re-validated under the dict lock
without a GIL. Trading one opaque call for three does not pay on the fixtures:
`dict_update_hot` builds a fresh dict every iteration and reads two different
keys, so `(dict, key)` never repeats. Separately, neither dict arm has a
`SPEC_FOLD_ROWS` row, which is why the shadowing went unnoticed — that is the
cheapest next step.

https://claude.ai/code/session_01EFbvNce7TD6xB2Xe2Zh3Vs

## Summary by CodeRabbit

* **New Features**
  * Improved handling of keyword-only default values, including updates, deletions, reassignment, non-string keys, and dictionary-like operations.
  * Added support for optimized calls using stable keyword-only defaults.
* **Bug Fixes**
  * Improved consistency between runtime calls and exposed default-value mappings.
  * Added safeguards for changed or unsupported default-value mappings.
* **Tests**
  * Added parity coverage and benchmark scenarios for optional values, default invalidation, and optimized execution paths.


youknowone:
On §2 — **the review is right and the patch's comment is wrong.** Checked
directly:

* `pypy/interpreter/function.py` `Function._immutable_fields_` does list
  `'w_kw_defs?'`, as its last entry after `'w_text_signature?'`. My comment
  claims the opposite.
* The second half of that sentence is wrong too. `fset_func_kwdefaults` is a
  plain store *upstream*, but pyre's `function_set_kwdefaults` already calls
  `function_notify_quasi_immut(obj, QuasiImmutSlot::WKwDefs)`, and the whole
  chain behind it exists: `descr.rs` declares `quasi("w_kw_defs",
  FUNCTION_W_KW_DEFS_OFFSET)`, `function_w_kw_defs_descr()` is mapped to
  `QuasiImmutSlot::WKwDefs`, and all of it landed in `14654ae6e8c` (#1336) —
  well before this branch. So there *is* an invalidation to record against, and
  I asserted there was none.

The sibling field makes the mismatch concrete: `defs_w?`, guarded a few lines
up in the same function, uses `record_quasiimmut_field` — the orthodox route —
while this one emits a per-call `GetfieldGcR` + `GuardValue`. Stricter than
upstream's hint, and a guard upstream does not pay.

Two things I am *not* doing blind, and why:

1. `walker_guard_function_field`'s own doc — "pyre's setters do not yet force
   the quasi-immutable invalidation, so a `GuardValue` stands in for it" — is
   stale for the same reason, and it covers `code?` / `w_func_globals?` /
   `closure?[*]` as well. Swapping this one field over without settling the
   other three would leave the file saying two different things.
2. `defs_w` gates its marker on `guards_the_callee_function`, because the
   install dereferences `callable_guard_op` and a baked `ConstPtr` must not be
   loaded through. That predicate is false for a **module-level** callee — which
   is exactly the shape this fold targets and measures 0.43 ns on. So the swap
   is not a substitution of one call for another; it needs the constant-callee
   case answered first, then a build and the three-leg gate, with
   `kwdefaults_invalidation` (added in this PR) as its oracle — that fixture
   exists precisely because an earlier attempt at the marker route handed back
   stale values on alternate iterations.

Action taken now: the false comment is corrected rather than left standing, and
the quasi-immutable swap is filed as the follow-up with the blocker named. The
behaviour in this PR is unchanged and correct — a rebind deopts through the
guard instead of invalidating — it is the JIT hint that is stricter than
upstream, as the review says.

— *commented by Claude*

PR #1507, opened 2026-08-26
jit-trace, majit: descend `~x`, name the load_deref declines, and move two descent walls

youknowone:
Eight commits on top of `main` (#1495). Two of them land measured descent
progress; the rest are the review fixes and the doc re-derivation from the
previous round.

## Descent

- **`~x` descends** (`b392348dfcf`, `af6583e0695`). `invert` is split into
  `invert`/`invert_inner`, and `try_walker_specialize_unary_invert` sub-walks
  `invert_inner`'s jitcode. `unary_invert_int` is retired: it is the second
  orthodox retirement, after `subscr_specialised_pair`.
- **The unary RBigInt wrappers are retargeted** (`760f2ba7104`).
  `descroperation::bigint_neg` / `bigint_invert` reach a descent as unlowered
  helpers, so the `<RBigInt as Neg>::neg` call inside them is never seen and
  the existing operator retarget never applies. `front::mir` now swaps the
  wrapper itself, guarded on a single `RBigInt` argument and an `RBigInt`
  destination.

  Measured with `PYRE_FBW_DEBUG_ABORT=1`: `~<long>` used to decline with
  `OrthodoxSubWalkTraceUnsupported { pc: 74 }` naming
  `descroperation::bigint_invert`; it now declines at `pc: 18` naming
  `target:<synthetic-transparent-ctor pyre_object::pyobject::PyObject>`. The
  fold still reports `fired=0` for a `long` operand — the wall moved off
  pyre's own code and onto the transparent-ctor class. `~<int>` still
  descends (`consulted=1 fired=1`).

- **The `len` descent wall moved twice** (`8a4694f13cd`, `d711d313bff`). A
  pair of exact builtins cannot reach an override, so `compare` decides that
  before running the override probe; and the swap block's type identity test
  is now spelled `std::ptr::eq` rather than `NonNull`'s `!=`, which goes
  through `NonNull::ne`. With `PYRE_FBW_NO_SPECIALIZE=builtin_len
  PYRE_FBW_DESCENT_SCAN_OFF=1 PYRE_FBW_INLINE_DIAG=1`, the `[subwalk-abort]`
  went `try_compare_override` → `NonNull::ne` (`abort_pc=245`) →
  `compare_slot` (`abort_pc=119`). No fold is retired by this.

## Diagnostics and docs

- `c8fb37a3d27` gives every `try_walker_specialize_load_deref` decline a
  name under `PYRE_FBW_DEBUG_ABORT`. Measured: 38 of 38 declines are
  `cell-not-constant`, so every later guard in that function is unreached and
  its `fired=0` is a missing capability, not dead weight. The doc comment
  says so instead of asserting the fold is dead.
- `8b385730891` re-derives the fold-layer counts in `design.md` §3.8 (69
  definitions, 73 rows, 9,886 body lines, 16,880 file lines; 23 fixtures
  naming 44 folds, 29 uncoupled rows) and records the commands that produce
  them, so the next reader re-derives rather than trusts the sentence.
  `gate-triage.md`'s stale "55 rows" is corrected the same way.
- `18c73fac505` corrects three math-fold doc comments, including recording
  that `MathFloatDomain::Total` spares `fabs` its operand guards but not the
  shared driver's `fold_finite_float_result` screen, so `fabs(inf)` and
  `fabs(nan)` decline.

## Gate

⚠️ `pyre/check.py --backend dynasm` has **not** been run against the final
three commits. The shared worktree was running four other builds and
check.py invocations at load average 15/41, which makes a timing gate
unreadable. `cargo fmt --all -- --check` and
`scripts/check-majit-boundary.py` are both clean; the probes above were run
on a binary verified fresh (see below). CI is the gate for this PR.

⛔ One measurement note worth carrying: a `pyre-dynasm` build in a busy
shared worktree can consume a codegen-cache generation that predates the
edit, and the symptom is a wall that refuses to move. Rebuilding the
byte-identical tree moved both walls above. Confirm the symbol your edit
removes is absent from the newest generation's `symbolic_fnaddr_paths`
before reading a probe.

https://claude.ai/code/session_019gCeUzbGWCK8S6vqnXh416

## Summary by CodeRabbit

- **Improvements**
  - Improved unary inversion handling for built-in integer values, including safer support for exact integer types.
  - Preserved expected behavior for subclasses and boolean values.
  - Added safer fallback handling when optimized execution is unavailable.
  - Improved handling of supported opaque big-integer operations.
  - Enhanced diagnostics for specialization decisions and declined optimizations.

- **Documentation**
  - Updated optimization and specialization documentation with corrected metrics, coverage details, reproducible counting guidance, and validation criteria.

### DIFF
--- a/majit/majit-translate/src/front/mir.rs
+++ b/majit/majit-translate/src/front/mir.rs
@@ -9213,6 +9213,34 @@ impl<'a> Lowering<'a> {
             op_kind
         };
 
+        // Retarget the interpreter's own unary RBigInt wrappers to the same
+        // residuals.  A descent reaches `descroperation::bigint_neg` /
+        // `bigint_invert` as an unlowered helper and never walks their
+        // bodies, so the `<RBigInt as Neg>::neg` call inside is not a call the
+        // retarget above ever sees.  Guarded like the divmod projections:
+        // sole operand and destination both the opaque `BigInt` ADT.  Same
+        // pure-target-swap, fail-safe contract.
+        let op_kind = if let OpKind::Call {
+            target: CallTarget::FunctionPath { segments },
+            args,
+            ..
+        } = &op_kind
+            && args.len() == 1
+            && tyref_is_rbigint(&call.dest.ty, self.llbc)
+            && first_arg_ty
+                .as_ref()
+                .is_some_and(|ty| tyref_is_rbigint(ty, self.llbc))
+            && let Some(residual) = crate::front::rbigint_call::unop_wrapper_residual_path(segments)
+        {
+            OpKind::Call {
+                target: CallTarget::FunctionPath { segments: residual },
+                args: args.clone(),
+                result_ty: ValueType::Ref(None),
+            }
+        } else {
+            op_kind
+        };
+
         // `RBigInt::digits{,_mut}` is only the Rust view adapter for
         // RPython's direct `self._digits` GcArray field.  Project the field
         // itself so the annotator receives the `[Signed]` SomeList registered
--- a/majit/majit-translate/src/front/rbigint_call.rs
+++ b/majit/majit-translate/src/front/rbigint_call.rs
@@ -275,6 +275,38 @@ pub(crate) fn compiler_bigint_residual_path(segments: &[String]) -> Option<Vec<S
     )
 }
 
+/// The interpreter's own unary RBigInt wrappers.
+///
+/// A descent reaches `descroperation::bigint_neg` / `bigint_invert` as an
+/// unlowered helper and stops there, so the `<RBigInt as Neg>::neg` call in
+/// their bodies is never seen and
+/// [`bigint_unop_residual_path`](crate::front::bigint_binop::bigint_unop_residual_path)
+/// never applies.  Swapping the wrapper itself reaches the same two residuals
+/// the operator spelling reaches, the way [`divmod_projection_residual_path`]
+/// does for the binary wrappers.
+pub(crate) fn unop_wrapper_residual_path(segments: &[String]) -> Option<Vec<String>> {
+    let residual = match segments.last().map(String::as_str) {
+        Some("bigint_neg") => "jit_bigint_neg",
+        Some("bigint_invert") => "jit_bigint_invert",
+        _ => return None,
+    };
+    if !segments
+        .iter()
+        .rev()
+        .skip(1)
+        .take(2)
+        .eq(["descroperation", "objspace"])
+    {
+        return None;
+    }
+    Some(
+        ["pyre_interpreter", "objspace", "descroperation", residual]
+            .into_iter()
+            .map(str::to_string)
+            .collect(),
+    )
+}
+
 /// Zero-checked interpreter seams for the two projections of
 /// `rbigint.divmod`. Each returns a direct RBigInt handle at source level, so
 /// the target swap is ABI-identical to the other pointer residual mappings.
@@ -312,6 +344,44 @@ mod tests {
 
     fn segs(parts: &[&str]) -> Vec<String> {
         parts.iter().map(|part| part.to_string()).collect()
+    }
+
+    #[test]
+    fn maps_the_unary_wrappers_and_nothing_else() {
+        for (wrapper, residual) in [
+            ("bigint_neg", "jit_bigint_neg"),
+            ("bigint_invert", "jit_bigint_invert"),
+        ] {
+            assert_eq!(
+                unop_wrapper_residual_path(&segs(&[
+                    "pyre_interpreter",
+                    "objspace",
+                    "descroperation",
+                    wrapper,
+                ])),
+                Some(segs(&[
+                    "pyre_interpreter",
+                    "objspace",
+                    "descroperation",
+                    residual,
+                ])),
+            );
+        }
+        // The owner is load-bearing: a same-named function anywhere else is
+        // not this wrapper and keeps its own body.
+        assert_eq!(
+            unop_wrapper_residual_path(&segs(&["pyre_object", "longobject", "bigint_neg"])),
+            None,
+        );
+        assert_eq!(
+            unop_wrapper_residual_path(&segs(&[
+                "pyre_interpreter",
+                "objspace",
+                "descroperation",
+                "bigint_clone",
+            ])),
+            None,
+        );
     }
 
     #[test]
--- a/pyre/design.md
+++ b/pyre/design.md
@@ -531,13 +531,22 @@ recorder, or the green accounts for.
 
 ### 3.8 The fold layer: hand-written compensation for an opaque objspace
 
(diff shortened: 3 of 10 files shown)


youknowone:
Went through every finding. Two were real and are fixed in `04d9afed419`; one description is accurate but its proposed fix is not; the two P1s rest on premises I can check, and one of them does not hold.

## CI: `cargo fmt --check` was red on `main`, not on this PR

Fixed in `a4b438a1ff1` so the eleven gated jobs can run, but the hunk is not mine: `git diff origin/main HEAD -- pyre/pyre-jit-trace/src/jitcode_dispatch/inline_call.rs` is **empty**, and `main`'s own fmt job was still `queued` when this PR's went red. It is the `kwonly_defaults_inline` `spec_gate` call — `#1494`'s row meeting `#1499`'s row macro, a combination neither PR's CI saw.

## CodeRabbit — `design.md:547`, reproducibility commands

Two of the three hold, and I ran each one to check:

| claim | verdict |
|---|---|
| `rg -c '=> ("'` has an unmatched `(` | **confirmed** — `regex parse error: unclosed group`, exit 2. The command never ran. |
| `rg -c` over `pyre/` and `majit/` gives per-file counts, not a total | **confirmed** — it answers 67/1/1, not 69. |
| `ls pyre/bench/synth/*.py` is non-recursive so it cannot scan nested paths | **refuted** — non-recursive is the point. It answers 484, the published number; a recursive walk answers 530, and 530 − 484 = 46, exactly the over-count the same sentence already attributes to `_pending`/`foriter57`/`iter57`. |

Fixed with `-F` and a summing form, and the corpus command is now marked deliberate so nobody "fixes" it to a `find`.

Chasing this turned up a defect neither review caught: **the count itself was stale.** The rows are **74**, not 73 — `kwonly_defaults_inline` arrived with `#1494` — and `spec_folds!` closes at `diag.rs:418`, not `:417`. Body/file lines moved 9,886/16,880 → 9,865/16,933. Both miscounts survived for the same reason: the recipe printed beside the number did not run. Every command in that section is now quoted as it must be typed and was executed against this commit, producing 74, 69, 484.

## CodeRabbit — `specialize.rs:7284`, null `w_class`

The description is right: `walker_exact_builtin_class` uses `None` for both "not an exact builtin" and "exact builtin whose `w_class` is null", so the second form declines. That is deliberate and documented at the helper — `is_exact_builtin_instance` accepts a null `w_class`, and such an operand cannot be pinned.

**The proposed fix is unsound.** Dropping the guard when `operand_class` is absent leaves the trace pinned on `ob_type` alone, and a builtin subclass shares the payload `ob_type` while carrying its own `w_class`. That subclass would take the fast path and be answered as a plain int — the silent rewrap `walker_guard_exact_w_class` exists to prevent, per its own doc comment.

The sound version is to *pin the null*: emit `getfield(w_class); ptr_eq(NULL); guard_true` for that form. That admits the shape without weakening anything. It is a reach change with **unmeasured reach**, though — I have no fixture that produces a null-`w_class` exact builtin, and this tree has already landed one arm that turned out to birth zero objects in production. Filing it rather than shipping it blind.

## Codex P1 — "Generate unary invert tracing from the public interpreter path"

Agreed on the destination, and that is what the change is walking toward rather than away from. Two facts about where it stands:

`invert` itself **cannot** be entered, and not for want of trying. Its `__invert__` override probe is `dont_look_inside` and is the *second* operation the body executes, and its bool slot raises a deprecation warning that reaches `lookup_exc_class`. Publishing an fnaddr for `invert` is separately blocked by a `&str` fat-pointer argument, which the residual ABI cannot carry one word per slot. `invert_inner` is that same body past both obstacles — it is not a parallel implementation.

What the walker sub-walks is `invert_inner`'s **real jitcode**, so a change to the interpreter body is picked up with no walker edit. What stays hand-written is the *entry guard* — `guard_class` plus `guard_exact_w_class` — which is exactly what lets the caller skip the two arms above and side-exit a subclass to the residual, which still runs the whole of `invert`. The net is one hand-written fold retired (`unary_invert_int`), not one traded for another.

## Codex P1 — "Preserve PyPy's comparison dispatch control flow"

I do not think this one holds, and the file says so a dozen lines below the hunk.

`try_compare_override` has **no upstream counterpart**. PyPy's `_make_comparison_impl` is one `comparison_impl` that swaps operands when the right-hand type is a proper subtype and then invokes the inherited implementation. pyre splits that into an override probe plus a slot-only path — the comment right under the change states it: *"`try_compare_override` already performs that ordering for generic instances and genuine builtin overrides; this is the corresponding path for a builtin subclass which inherits (rather than overrides) its comparison slot."*

So the step this skips is a pyre-only decomposition, not an upstream control-flow step, and AGENTS.md's "control-flow order follows PyPy unconditionally" has nothing here to be out of order with. What upstream does for two exact builtins is the subtype swap and the slot — which is precisely the path taken, since the change skips only the probe and the comment says the ordering below still runs (`bool` is a proper subclass of `int`, and both are exact builtin instances).

The probe is also provably vacuous for that pair rather than merely empirically so: `subclass_special_override` opens with `if is_exact_builtin_instance(obj) { return None; }`, and `is_instance` is false for both. It cannot answer anything but `None`.

Happy to be shown an upstream line where `comparison_impl` does something for two exact builtins that this skips — that would settle it the other way.

— *commented by Claude*

youknowone:
Correction to my previous comment on the null-`w_class` finding: I said I would file it as a reach item with unmeasured reach. It is measurable from the source alone, and the answer is **zero reach**. Closing it rather than filing it.

`pyobject.rs` documents the population directly: a fresh builtin is born carrying `w_class == get_instantiate(ob_type)`, and *"the read-only singletons (`True` / `False` / `None` / `Ellipsis` / `NotImplemented`) leave `w_class` null and are always exact."* Those five are the whole of the null-`w_class` exact-builtin form.

The admission gate in `try_walker_orthodox_unary_invert` runs **before** the `walker_exact_builtin_class` read:

```rust
let admitted = unsafe {
    !pyre_object::is_bool(operand_obj)
        && (is_long || pyre_object::is_int(operand_obj))
        && pyre_object::is_exact_builtin_instance(operand_obj)
};
if !admitted { return Ok(None); }
```

Intersect the two sets and it is empty: `True` and `False` are `bool` and fail the first clause; `None`, `Ellipsis` and `NotImplemented` fail the second. So `walker_exact_builtin_class` cannot answer `None` for an operand that reaches it here, and "the new `~int` and `~long` descent therefore always falls back to the generic residual for that form" describes a form this entry point cannot receive.

`bool` is excluded on purpose, incidentally, and not as an oversight: `invert`'s bool slot raises a deprecation warning, which is one of the two arms `invert_inner` exists to sit past.

`324e96bbfc9` records the argument beside the `let ... else` so the next reader does not have to re-derive it. The conditional-guard fix stays declined — it would leave the trace pinned on `ob_type` alone, and a builtin subclass shares `ob_type` while carrying its own `w_class`.

— *commented by Claude*

PR #1512, opened 2026-08-26
jit: fold LOAD_SUPER_ATTR, so super(C, self).m(x) costs what x + 1 costs

youknowone:
`LOAD_SUPER_ATTR` lowers to one opaque `bh_load_super_attr_fn` residual plus two
`bh_super_attr_unwrap_fn` residuals, and nothing folded any of them, so every
`super().m(x)` rebuilt the proxy, re-walked the MRO suffix and re-bound the
descriptor on each iteration. All three were declared `MayForce`, which also
cleared the optimizer's heap-field cache per call, and all three carried
`RuntimeHelperKind::None`, which `fbw_callee_body_replay_scan` convicts by
absence of a tag — poisoning every `super()`-bearing body out of its inline.

Measured on a FOR_ITER-driven loop calling `def m(self, x): return
super(C, self).m(x)`, 500k iterations, best of 3, one case per process:

| body | before | after | pypy3 | CPython 3.14 |
|---|---|---|---|---|
| `return x + 1` (control) | 1.0 | 0.7 | 0.8 | 26.6 |
| `return A.m(self, x)` (control) | 146 | 112 | 0.9 | 39.6 |
| `return super(C, self).m(x)` | **1309** | **0.8** | 0.8 | 56.3 |

`super(C, self).m(x)` now costs what `x + 1` costs. The `return A.m(self, x)`
row is an unrelated control that this branch does not address; it is here to
show the machine, not a result.

The loop trace loses its 2 `CallMayForceR`, 4 `CallR` and 5 `NewWithVtable`, and
its `SetfieldGc` count falls from 35 to 4; `loops_compiled` falls from 2 to 1.

## What changed

**The tags.** `RuntimeHelperKind::{LoadSuperAttr, SuperAttrUnwrap}`, grouped
with `LoadMethodSelf`. `super_attr_unwrap` moves to `PlainCannotRaise`: it is an
`is_method` class test and one of two field reads, so the may-force shape bought
a `ForceToken`, a `Setfield last_instr`, a `GuardNotForced` and a
`GuardNoException` per call, twice per method-form site.

**The scan.** `SuperAttrUnwrap` gets `replay_safe_read` standing (re-running one
reads the same two words again); `LoadSuperAttr` gets `LoadAttr`'s
`defer_helper` standing (same question about the same descriptor bind).
Together they are what lets a `super()`-bearing body inline at all — worth
1309 → 398 ns on their own.

**The two-entry window.** `fbw_callee_body_has_load_method_self_residual` is
renamed `fbw_callee_body_has_two_entry_method_push` and now also matches
`SuperAttrUnwrap`. LOAD_SUPER_ATTR's method form pushes the same two live
operand-stack entries `LOAD_ATTR`'s does; it could not reach that window while
the residuals poisoned the body, and with them admitted it can.

**The fold.** `super_attr_fast_path` answers when `W_Super.getattribute` reduces
purely to `w_method_new`; `try_walker_specialize_load_super_attr` emits that
allocation, and `try_walker_fold_super_attr_unwrap` reads the two fields back so
the `Method` virtualizes away.

Two details are load-bearing:

* The unwrap fold uses the **cached** getfield the producing fold primed, so no
  op is recorded and the result resolves to a constant. The uncached read leaves
  the following CALL a callable slot with no concrete ref — `[inline-decline]
  why=callable slot carries no concrete ref`, and 657 ns/iteration, *slower than
  leaving the residual alone*.
* `(*w_descr).ob_type` is pinned to exactly `FUNCTION_TYPE`. `get()` dispatches
  `is_method_descriptor` first and answers a `method_descriptor` with
  `builtin_bound_method_new`, whose result has a different `type()` and a
  `module` field the `Method` template does not write.

The fold emits one runtime guard **fewer** than the ordinary method load: there
is no instance-map shadow guard, because `W_Super.getattribute` never consults
the receiver's dict, so `o.name = x` cannot shadow `super().name`. Invalidation
rests on `objtype`'s version tag alone: every class the suffix walk reads is an
ancestor of `objtype`, and `mutated()` recurses into subclasses.

## Verification

* `lib-python/3/test/test_super.py` — 40 tests, OK (3 skipped).
* A 14-case probe (property / classmethod / staticmethod / slot wrapper /
  `method_descriptor` decline paths, zero-arg `super()`, `super(Mid, self)`
  level-skip, class-mode `super(D, D)`, instance-attribute shadowing, and
  redefining `A.m` and an intermediate `F1.v` mid-loop) agrees byte-for-byte
  across pyre-JIT, `PYRE_NO_JIT=1` and CPython 3.14.
* `check.py --synthetic-pattern '*super*'` — 3/3 on dynasm, cranelift and wasm.

## Not fixed here

Zero-argument `super()` in a FOR_ITER-driven inlined callee still declines: its
`__class__` freevar read is a `load_deref` residual, and that helper's
`replay_safe_read` admission is deliberately withheld — `fbw_state.rs` records
the measurement, it un-retires a loop that then answers `f_lasti` with the
resume coordinate. The two-argument form and every `while`-driven shape fold.

`global_is_const_foldable` bakes a type-valued global's *value* into the trace
with no invalidation channel, so rebinding `super` (or any type-valued global)
after a loop is hot is answered stale. That predates this branch and is
unchanged by it: the residual received the same baked constant.

## Summary by CodeRabbit

* **New Features**
  * Improved JIT optimization for `super(...).attribute` access.
  * Reduced repeated proxy creation and method handling overhead in eligible code paths.
  * Added a benchmark covering `super` attribute access within inlined callees.

* **Performance**
  * Added runtime fast paths and specialization for common `super` method and attribute patterns.
  * Added JIT statistics for the new benchmark across supported compilation backends.


youknowone:
Gates run locally on this branch, at the rebased head:

- `check.py --synthetic-only --synthetic-pattern '*super*'` — 3/3 on dynasm, cranelift and wasm.
- `cargo fmt --all -- --check` clean; `scripts/check-majit-boundary.py` clean.
- `lib-python/3/test/test_super.py` — 40 tests OK.

**Not run locally**: the full `check.py` corpus (481 fixtures) and `cargo test --all --no-default-features --features dynasm,cpyext`. Two things in this diff widen a decline into an admission and deserve the corpus specifically:

1. `LoadSuperAttr` joining `defer_helper` makes every `super()`-bearing body `DeferredCall` rather than `Dirty`, which is the same admission class whose `load_deref` sibling is documented as costing aborts. The `Dirty` → `DeferredCall` move is what the `guard_failures` / `loops_aborted` gates would catch.
2. Both fixture headers drop from `max-pypy-ratio=220` to `20`. Locally pypy is clamped to the timing floor at the committed N, so no ratio gate was evaluated — the 20 comes from an N=8000000 run where both fixtures measured ~2.2x pypy3. On a slower CI host that gate is now live where it previously was not.

— *commented by Claude*

Issue #1227, opened 2026-08-14
FOR_ITER: a user-defined `__next__` never enters the trace — no inline route exists, and the seqiter route is dead code


youknowone:
`for x in it`, where `it` is an instance of a user-defined class with a `__next__` method, compiles to a single opaque residual. The user `__next__` frame never enters the trace.

`bench/synth/iteration_protocol.py` is the worst row in the exception/iteration suite: **~30x PyPy on dynasm, ~42x on cranelift**.

## Measured

`MAJIT_LOG=1`, the compiled outer loop (`loops_compiled=1`, `loops_aborted=0` — the loop does compile):

```
v15 = CallMayForceR(pyre_interpreter::runtime_ops::jit_next, v10)   x4
v66 = CallMayForceR(pyre_jit::call_jit::bh_get_iter_fn, v51)        x2
```

38 of 105 ops.

At full bench size (`N = 20000`) with `PYRE_FBW_INLINE_DIAG=1 PYRE_FBW_DEBUG_ABORT=1 PYRE_FOR_ITER_GATE_DIAG=1`, the run prints **zero** `[inline-foriter-gate]` lines. The only declines are importlib startup frames and the `CountDown(5)` constructor. **The inline gate is not declining — it is never consulted.** This is not a gate to widen.

## Why

There is no `__next__` lookup anywhere in `pyre-jit-trace`. Two FOR_ITER callee routes have ever existed:

| route | location | covers |
|---|---|---|
| `try_walker_specialize_for_iter_next` | `pyre-jit-trace/src/jitcode_dispatch/specialize.rs` | `range` / `zip` only, a C-level advance with no Python callee |
| `try_walker_specialize_seqiter_getitem_next` | `pyre-jit-trace/src/jitcode_dispatch/inline_call.rs` | a user `__getitem__` behind a generic `W_SeqIterObject` |

A user class defining `__next__` directly matches neither.

**The second route is dead code.** `5c2d37523d5` (#1126) deleted its sole call site in `residual_call.rs` and left the body in place. It raises no warning because `pyre-jit-trace/src/lib.rs` carries `dead_code` in its crate-level `#![allow(...)]`; `rg` across the tree returns exactly one hit, the definition.

## The obstacle #1126 named

That commit replaced the call site with a comment stating its reason:

[...]

PR #1535, opened 2026-08-27
jit-trace: key a FOR_ITER off the JitCode the walk is executing

youknowone:
`walker_foriter_green_key` derived the FOR_ITER green key from
`ctx.fbw_mode.snapshot_sym` — the walk's **outer portal** — and returned `None`
for every inline sub-walk rather than run a callee offset through the caller's
pc tables, which would answer with a Python pc belonging to neither loop.

Both FOR_ITER specializations key their demotion descr off that value, so both
declined inside an inlined callee:

* `try_walker_specialize_instance_next` — inlines a user class's Python
  `__next__` at the residual;
* `try_walker_specialize_for_iter_next` — range/zip; its own comment named the
  green key as the reason.

The effect is that **a `for` loop written inside a function the walker inlines
kept the opaque `ForIterNext` residual**, running the iterator's `__next__` as a
real frame in the plain interpreter, while the identical loop written in the
loop-owning function was folded.

`diag.rs`'s `fbw_foriter_body_from_op_pc` already resolves the same coordinate
callee-first out of `ctx.inline_callee_consts`, and carries a doc explaining
exactly this hazard. Only the green key never got the same treatment; this PR
gives it that, and drops the two `inline_subwalk` early-returns that existed for
it.

## Measured

Against a binary built from the same base with no other delta, both binaries
read from the same directory (a binary read out of `target/release` against one
read out of a scratchpad shows a fixed +0.03s startup gap that reads as a 30%
"regression" on a 0.1s program):

| program | base | this PR |
|---|---|---|
| `pass` (startup) | 0.02s | 0.01s |
| attribute loop, no FOR_ITER anywhere | 0.10s | 0.11s |
| `for` over a user iterator in the loop-owning function | 1.07s | 1.06s |
| **`for` over a user iterator inside an inlined callee** | **5.79s** | **1.08s** |
| `range` `for` inside a callee | 1.41s | 1.36s |

The in-callee form now matches the top-level form. `loops_compiled`,
`bridges_compiled` and `guard_failures` do not move on any of them.

## Guard

`bench/synth/foriter_instance_next_in_inlined_callee.py`, gated by the fold
census: `instance_next` reads `fired=1` with the sub-walk route open and
`fired=0` with it closed.

No `max-pypy-ratio`: at a size that keeps the fixture's own runtime sane, pypy's
execution-only time clamps to `EXEC_TIME_FLOOR_S` and check.py skips the ratio
(`~`) on exactly the runs a ceiling would have to gate — this run reads
`~34.2x` / `~39.0x`, both unenforced. `.jitstats` cannot gate it either, since
no recorded counter moves. `foriter_call_body` covers the same fold from the
loop-owning function; only the callee position exercises this green key.

## Gate

```
ALL PASSED: dynasm 502/502
ALL PASSED: cranelift 502/502
```

(run at base `8624de80174`; the branch has since been rebased onto
`7a2cb98b1e7` and is being re-gated locally.)

Related: #1227 (the route this extends; its top-level half landed in #1270).


youknowone:
Checked the parity review's finding 1 (and the pre-existing finding 3 it rests on). **The premise is factually wrong: `InlineCalleeConsts.w_code` holds the `W_Code` wrapper, not a raw `CodeObject*`.**

`callee_code_key` is built at `inline_call.rs:4491`:

```rust
let callee_code_key = w_code as pyre_object::PyObjectRef as usize;
```

and that `w_code` is the third value `resolve_inlinable_callee` returns (`mod.rs:7813`), which is the **wrapper**:

```rust
let w_code = pyre_interpreter::function_get_code(callable);
...
if (*(w_code as *const pyre_object::pyobject::PyObject)).ob_type as *const () as usize
    != code_type_addr { return None; }                    // w_code IS a PyCode object
let raw = pyre_interpreter::w_code_get_ptr(w_code as pyre_object::PyObjectRef)
    as *const pyre_interpreter::CodeObject;               // raw is DERIVED from it
...
Some((w_code, (*raw).arg_count as usize, !closure.is_null()))   // returns the wrapper
```

`raw` is a local used only for the arg count and the generator-flag test; it is never stored. So both green-key producers pass the same kind of pointer:

| path | first argument to `make_green_key` |
|---|---|
| snapshot root (`mod.rs`) | `live_code_wrapper(jitcode.raw_code())` — wrapper |
| inline sub-walk (this PR) | `consts.w_code` = `function_get_code(callable)` — wrapper |

The keys are in one namespace, and the demotion key does match the warm-state key for the callee.

For completeness, the `jitcode_index` half is also insensitive to the distinction: `ensure_jitcode_index` canonicalises through `canonical_code_key_opt` (`state.rs:300`), which accepts a raw `CodeObject*` **or** a `PyCode` wrapper and reduces both to the raw key.

Finding 3 ("`w_code` is initialized with `callee_code_key`, a raw compiled-code pointer") is the same misreading and is not a pre-existing representation error either.

— *commented by Claude*

PR #1539, opened 2026-08-27
dicts, sets, gc, cpyext: an rordereddict port behind the dict and set storages, the roots a collection re-walked, and both halves of PEP 590

youknowone:
21 commits on top of `origin/main` (`05cc0827eb8`), which now carries the
`_ssl` transport rework of #1544 — this branch is rebased under it, and the two
files both sides rewrote were merged region by region against the RPython
sources rather than by taking a side.

## dicts, sets — the storage is an rordereddict port

`pyre/pyre-object/src/rordereddict.rs` is new: a `u32` index table over an entry
vector, with the `FREE`/`DELETED`/`VALID_OFFSET` sentinels, the
`(i << 2) + i + perturb + 1` probe, `ll_dict_grow`'s compaction and
`ll_dict_remove_deleted_items`' shrink. The object, unicode, bytes, int, module
and identity dict storages and the set element table are that type now.

A delete writes `DELETED` and clears the entry rather than shifting the tail, so
it is O(1) and every surviving slot number stays put — which is what forces the
rest: **a slot is not a position**. Positional cursors became slot cursors
across marshal, gc, `type_methods`, `typedef`, `baseobjspace` and the dict/set
view iterators.

Three defects the unit tests did not catch, each with a fixture: the two
strategy switches still ran `for i in 0..len { get_slot(i).unwrap() }` and
panicked on a dict a key had been deleted from; `marshal`'s set arm read slots
`0..len` and raised `ValueError: unmarshallable object`; and `RDict::generation`
counted a reindex, a compaction and a clear but not a growth, so
`scan_dict_key_reentrant` trusted a slot number it was owed a restart on. The
witness for that last one has to read `Vec::capacity`, not the data pointer — an
allocator that extends in place returns the address it already had, which passed
on dynasm and cranelift and left wasm wrong.

`ll_dict_lookup`'s `d.paranoia` restart is **not** ported and cannot be: `&self`
tells the compiler nothing writes through it, so the re-read meant to notice a
mutating `__eq__` folds away — measured correct at `opt-level=0` and losing a
present key at `opt-level=1`. The guarantee stays with `callback_free_dict_op!`
and `scan_dict_key_reentrant`, which re-derive their pointers at every step.

## gc — work a collection was repeating

`walk_pyframe_roots_area` ran the module-dict cell walk once per frame over the
globals dict and once per frame over the builtins dict; a frame chain names the
same few dicts repeatedly. `gc.collect()` CPU at a given Python stack depth:

| depth | before | after |
|---|---|---|
| 0 | 3.3–4.0 ms | 1.9–3.8 ms |
| 200 | 10.2–11.2 ms | 4.1–4.2 ms |
| 800 | 31.5–33.7 ms | 6.4–6.6 ms |

Four per-cycle clones removed; `do_collect_full` goes through
`major_collection_step` so the `STATE_SCANNING` resets actually run;
`rawmalloced_contains` gets a hashed page bitmap after its `[lo, hi]` envelope
was measured to reject nothing. The write barrier drops its shadow-stack root
bracket, `root_stack_depth` moves onto the `ShadowStack` it caps per
`ShadowStackPool`, and `ModuleDictEntries` stops siphashing an owned `String` on
every global read that misses the per-code cache.

## cpyext — both halves of PEP 590

Type level: `PyType_FromSpec` answered `SystemError: ... does not support slot
82` and took the importing module with it. Instance level:
`Py_TPFLAGS_HAVE_VECTORCALL` was declared in `object.h` and read by nothing, so
`slot_call` built a tuple and a keyword dict for `PyVectorcall_Call` to take
apart again. `inherit_slots` copies `tp_vectorcall_offset`, which only
`inherit_mirror_slots` did. Measured **3.9x / 2.1x / 1.4x** at 0 / 1 / 2
arguments against the same object without the flag.

## The two CI failures this branch had

**`cargo test` aborted on all three platforms.** `method_cache_mut`'s
`debug_assert!` reads the process-global `gil_is_initialized()` to excuse a
per-thread condition, so the first test to reach `compile_source` armed it for
every other libtest thread — including those that never entered the runtime and
held no GIL. The unwind then crossed `dict_eq_hook::hash_w_hooked`, which is
`extern "C"` and so `nounwind`, turning one assertion into SIGABRT with no
`failures:` summary. `install_hash_hook` now enters the runtime; its own doc
already names the contract ("call at the top of any `#[test]` that constructs an
object/str-keyed dict"), and `RUNTIME_THREAD`'s destructor gives the
registration back on thread exit. The assert is unchanged.

**`check.py dynasm` reported two regressions.** `test.test_ssl` went
PASS -> TIMEOUT: `_SSLSocket.shutdown` drove `while !connection_peer_closed(...)`
with no bound, where `_stdssl._SSLSocket.shutdown` stops after its second zero
return ("Don't loop endlessly; instead preserve legacy behaviour of trying
SSL_shutdown() only twice") and returns the socket even without the peer alert.
`test.test_asyncio` had been promoted out of the cell recording it as a flake
(`TIMEOUT`, "stalls in about one arm in four, at a test that wanders") and
stalled again in the same run; the cell is restored.

## Merging with #1544

Both sides rewrote `_ssl/mod.rs` and `_socket/interp_socket.rs`. Resolutions:

- `read()` — took #1544's `pump` structure and dropped this branch's
  ZERO_RETURN arm. `_read` returns an empty result *successfully* when
  ZERO_RETURN coincides with `SSL_RECEIVED_SHUTDOWN`; the arm tested
  `shutdown_started` (locally initiated — the opposite state) and raised.
  #1544's `break data` on an empty vec already is the upstream behaviour.
- `send`/`recv` — took #1544's `socket_send_bytes` / `socket_recv_bytes` and
  restored this branch's readiness wait at the socket-method layer, because
  `send_raw` calls `wait_for_data(True)` and `recvinto` calls
  `wait_for_data(False)` before their syscalls. The helpers stay raw: they are
  the `_c.send` layer, and the TLS pump calls them directly.
- `receive_transport` — took #1544's descriptor-based version; the
  `receive_transport_limited` rename could not survive, since #1544 deleted the
  `socket_recv` field it called.

## Verification

`cargo test --all --no-default-features --features dynasm,cpyext` is the gate
that was red; it is green on the rebased tree. Each conflict resolution was
`cargo check`ed before the rebase continued.

Not run locally: `pyre/check.py`. The `test_ssl` fix is argued from the upstream
bound rather than from a local suite run, and CI's dynasm/ubuntu leg is what
tests it.

https://claude.ai/code/session_01PuYePQknDcMCUsy8omQ1fh

## Summary by CodeRabbit

* **New Features**
  * Added PEP 590 vectorcall support for compatible extension types, improving call efficiency and compatibility.
  * Added TLS key logging configuration through the SSL context.
  * Improved socket timeout, readiness, address handling, and platform-specific behavior.
* **Bug Fixes**
  * Corrected dictionary and set iteration after deletions.
  * Improved exception context handling and traceback resumption.
  * Enhanced garbage collection, memory-leak handling, and SSL non-blocking operations.
* **Tests**
  * Expanded coverage for vectorcall, socket behavior, garbage collection, and updated runtime test results.


youknowone:
Rebased onto `origin/main` (`903151ee679`) and pushed. The five regressions the last `pyre/check.py dynasm (ubuntu-24.04)` run reported are addressed.

**`test_socket` — five failures, two causes.** `RSocket.__init__`'s tail was never ported: the two Linux creation flags stayed in the type the socket keeps, and a `SOCK_NONBLOCK` socket recorded no timeout. `test_socket_consistent_sock_type` read `526337`, `test_SOCK_CLOEXEC` `524289`, `test_SOCK_NONBLOCK` `2049`, and `testInitNonBlocking` `None` where `0` was expected. The fifth, `testAutobind`, was separate: `pack_sockaddr` decided the abstract namespace on the first byte alone, so an empty path took the regular branch and counted a terminator into `addrlen`; the kernel read that as an abstract address one null byte long rather than as the auto-bind request. `getsockaddrarg` names both members of the extension in one test and gives both `path.len + offsetof(struct sockaddr_un, sun_path)`.

**`test_ssl`, `test_imaplib`, `test_smtpnet` — one cause.** `ce8d356d786` moved the positive-timeout model to RPython's shape, where `settimeout` ends in `_setblocking(self.timeout < 0.0)` and the duration belongs to `_select`, but `_ssl` still read the transport through the raw helpers. A would-block on a socket carrying a positive timeout therefore surfaced as `SSLWantReadError` or an EOF instead of waiting. `_ssl` now waits where `_ssl_select` does, and an exchange that outlasts its socket's timeout is cut off with the `TimeoutError` `wait_for_data` raises.

**`test_zipfile`** — `test_execute_zip2`/`test_execute_zip64` failed with `OSError: [Errno 12] Cannot allocate memory`. Both spawn a subprocess, and neither reproduces on linux-aarch64 or darwin-arm64; the module is PASS on both. Left as-is for this run to re-check.

Verified on linux-aarch64 under Apple `container` (`test_socket`, `test_ssl`, `test_imaplib`, `test_smtpnet`, `test_zipfile` — all PASS, no regressions) and on darwin-arm64, where `pyre/check.py` reads dynasm 512/512, cranelift 512/512, wasm 505/505.

One commit here is not this branch's work: `exception_escape_hot_callee_tb_node_once` and `exception_traceback_frame_lineno` are re-recorded. #1553 and #1388 both touched those files and were rebased onto each other, and the post-merge CI run for each merge was cancelled, so neither recording was measured against the merged tree. Main's own run at `67505f239e9` reports the identical numbers this records — `guard_failures 808 -> 1015`, `1008 -> 1015`, `bridges_compiled 6 -> 4`, `guard_failures 825 -> 624`.

— *commented by Claude*

Issue #1311, opened 2026-08-17
jit(blackhole): the multi-frame chain admits registers_r[frame_reg] on a bare `!= 0`, and a non-frame word reaches record_application_traceback


youknowone:
`pyre-cranelift` SIGSEGVs on `test.test_unittest`. Reported here because the crash is main-owned and **not root-caused** — please do not close it by hardening the dereference.

## Reproduction

```
python3 pyre/cpython_tests/run.py --backend cranelift --filter test_unittest
```

Reproduces on the release `pyre-cranelift` of five worktrees (pyre-0, pyre-1, pyre-3, pyre-6, pyre-7), 5/5, byte-identical fault. pyre-3's binary is dated 2026-08-14 and was run entirely inside its own worktree with its own stdlib, so this predates the current GC-rooting work. `pyre-dynasm` is clean on the same module.

## The damage path (established)

- Faulting deref: `pyre/pyre-interpreter/src/pytraceback.rs:415` — `(*frame).pycode`, reached via `record_application_traceback+176`, `EXC_BAD_ACCESS` at `0x10d`.
- The value is bound at `pyre/pyre-jit/src/call_jit.rs:2583-2589`: when the resumed chain is multi-frame, `virtualizable_ptr` is taken from `frame.registers_r[frame_reg]` where `frame_reg` comes from `portal_red_regs_at(jitcode_index)`.
- The only admission test is `if frame_ptr != 0` (`call_jit.rs:2587`). Every consumer downstream — `call_jit.rs:2639`, `:2655`, `:728`, `pytraceback.rs:408` — checks `is_null()` and nothing else.
- One measured dump had the word `0x86c86dc48`: a real, aligned, non-null heap address that is not a `PyFrame`. Reading its `+0x18` as `pycode` gave `0x102957c80` (static data inside the binary image); that object's `+0x10` gave `0x5`, and `[0x5+0x108]` = `0x10d` faulted.
- A separate session reported `frame_ptr == 2` on the same slot. If both dumps are real, that is **two distinct bad values through one unguarded slot** — `2` would be a tagged small int, a different provenance from a mis-typed heap pointer.

## What is NOT established

Why that red register holds a non-frame in a multi-frame chain. The candidates are (a) `portal_red_regs_at` returning a `frame_reg` that is valid-looking but wrong for that jitcode, (b) a section that never filled its red, (c) a stale word. There is no evidence separating them.

[...]

Issue #1151, opened 2026-08-11
blackhole: find_catch_after_resume_live skips an unbounded prefix before the trailing -live-


youknowone:
## Summary

`BlackholeInterpreter::find_catch_after_resume_live`
(`majit/majit-metainterp/src/blackhole.rs:1377`) skips an **unbounded** number
of startpoints before it looks for the trailing `-live-`. Its caller comment
asserts a bound that the loop does not impose, so the `catch_exception` it
returns is not necessarily the one belonging to the operation that raised.

## The asserted invariant vs. the code

The caller (`handle_exception_in_frame`, `blackhole.rs:1216-1224`) says:

> Ops before the trailing live belong to that translated operation. Once the
> live is crossed, only its immediately-following catch is valid.

The second sentence *is* enforced — after `crossed_trailing_live` is set, the
very next startpoint must be a `catch_exception` or the scan gives up. The first
sentence is not enforced by anything in the loop:

```rust
let mut crossed_trailing_live = false;
for q in points {                       // every startpoint after resume_live_pos
    let op = code[q];
    if op == self.op_catch_exception {
        return crossed_trailing_live.then_some(q);
    }
    if crossed_trailing_live {
        return None;
    }
    if op == self.op_live {
        crossed_trailing_live = true;   // first `-live-` seen, wherever it is
    }
}
```

Before the first `-live-` the loop consumes any number of ops of any kind, with
no block boundary and no distance bound. If the raising operation's own block
has no trailing `-live-`/`catch_exception` pair, the scan walks into whatever
follows and can return a `catch_exception` that belongs to a *different*
translated operation. `route_to_catch` then stores the exception and jumps to
that unrelated handler.

## Relationship to #174

#174 (closed) covers the producer side: the flattener maintains "one can-raise
op per block, `catch_exception` adjacent to its trailing `-live-`" by a hoist
plus a `debug_assert!` that is compiled out in release. This is the consumer
side of the same invariant — even granting the per-block guarantee, the scan
here is not bounded to the block, so it does not fail closed if the producer
side ever stops holding.

## Status

Raised by a Codex parity review of #1126. The behaviour above is what the code
does; I could **not** construct a reaching case, so whether any jitcode pyre
generates today drives the scan past its intended operation is unproven.
Recording it rather than changing the scan, because narrowing it without a
reproducer risks turning a currently-caught exception into an uncaught one.

A cheap first step, if someone wants to close the gap without a reproducer, is
[...]

PR #1543, opened 2026-08-27
codecs, interp: the assignment-target SyntaxError diagnostics, the incomplete-input classification past OtherError, a decoder's buffer acquisition, and an overflowing handler position

youknowone:
Four commits, each measured against CPython 3.14.5 **and** 3.14.2 (the version the CI leg runs) before it was written, with the previous binary kept as the control.

## 1. Assignment-target `SyntaxError` diagnostics

`python.gram` names an assignment target with `_PyPegen_get_expr_name` and picks between three messages:

| grammar rule | message |
|---|---|
| `invalid_named_expression` alt 3 | ``cannot assign to X here. Maybe you meant '==' instead of '='?`` |
| `invalid_assignment` → `RAISE_SYNTAX_ERROR_INVALID_TARGET` | `cannot assign to X` |
| `invalid_assignment` augassign alt | `'X' is an illegal expression for augmented assignment` |

Which one is reached turns on the target's precedence, on whether the target or the value was parenthesized, on the value's leading `bitwise_or`, and on how many targets the statement has — none of which a scan of the source text answers. The old `dict_comprehension_assignment_error` scanned for a top-level `=` and re-parsed the text to its left, so it could only restore the diagnostic when that text happened to parse as a dict comprehension.

`assignment_target_error` parses the whole source with ruff's error recovery, takes the `Assign` / `AugAssign` that the *reported* location falls inside, and reads the answer off the tree.

Measured with the previous release binary as the control, against CPython 3.14.2 (the version the CI leg installs):

| corpus | rows | matched before | matched after | fixed | regressed |
|---|---|---|---|---|---|
| hand-written target matrix | 138 | 83 | 137 | 54 | **0** |
| generated sweep (targets × values × statements) | 507 | 310 | 502 | 192 | **0** |

The six rows still wrong are all unclosed-bracket spans that answer identically before and after (`x = f(` reports `end_offset` 6 where CPython reports 0), plus `def\nf() = 1`, where the parser's own message differs. None of them is an assignment diagnostic.

`syntax_error_python314_offsets.py` gains 114 error rows and 18 that compile; 54 of the error rows reported differently before this change, 0 after. The whole file passes under CPython 3.14.2 and 3.14.5, and still fails under pypy3 7.3.20 at the row its `pypy-diverges:` header names, so that declaration stays honest.

`assignment_span` also moves ahead of `fstring_span`, `scope_span` and `generator_span` in the span chain. `msg` was already assigned in that order, so the two disagreed: `def f(): (yield a) = 1` reported the assignment message on the f-string shape test's span.

## 2. `PyCF_ALLOW_INCOMPLETE_INPUT` past `OtherError`

`_PyPegen_run_parser` tests `_is_end_of_source(p)` — `tok->done ∈ {E_EOF, E_EOFS, E_EOLS}` — **before** the second, error-reporting parse, so a diagnosable error does not beat "incomplete": the whole rule is whether the tokenizer reached the end of the input.

pyre's shape test answered `false` for every `ParseErrorType` outside the spelled-out lexical/bracket shapes and `OtherError`, so `1 if 2` (`ExpectedToken`) and `1 1` (`UnexpectedExpressionToken`) raised `SyntaxError` where CPython raises `_IncompleteInputError`. The position test that arm already applies now covers the remaining shapes for `eval` and `single`.

`exec` is left alone. `_PyTokenizer_FromString(str, exec_input=1, …)` supplies the missing trailing newline, so a NEWLINE ends the statement and the failure is final; only the shapes that read *past* a newline stay incomplete there, and those are exactly the arms above plus the `OtherError` spellings (`@dec`, `x = \`).

Measured over 116 sources × 3 modes, previous binary as the control:

| corpus | rows | matched before | matched after | fixed | regressed |
|---|---|---|---|---|---|
| expression / statement heads | 67 | 32 | 51 | 19 | **0** |
| strings, f-strings, blocks | 48 | 29 | 29 | 0 | **0** |

The 16 + 19 still-wrong cells are f-string shapes and `''` / comment-only sources in `single`, both untouched here.

## 3. Codecs and the earlier follow-ups

Carried from the first commit on this branch: `decode_input_bytes` acquires one `PyBUF_SIMPLE` export through `simple_buffer_bytes`; `call_decode_error_handler` / `call_registered_encode_error_handler` read the handler position with `int_w(...)?` so an out-of-range position leaves as that conversion's `OverflowError`.

**Correction to that commit's message.** It claims the end-of-source test "also settles `1 1` and `1 if 2`". It does not — those two carry `ExpectedToken` and `UnexpectedExpressionToken`, which fell into the `_ => false` arm. Section 2 is the actual fix; `compile_incomplete_input_mode.py` was red until it landed.

## 4. Four exception doc blocks

`EXC_UNICODE_TRANSLATE_ERROR_TYPE`, `ExcKind::UnicodeTranslateError`, `ExcKind::ModuleNotFoundError` and `ExcKind::SyntaxError` described their ports as identity-only with the per-class fields, the `__init__` signatures and `__str__` still to come. Those slots are flattened onto `W_BaseException` and the constructors write them.

## Gate

Run end to end at `45509ea5b62`, HEAD unchanged from start to finish:

```
LLBC 0 · LLBC --check 0 · dynasm 0 · cranelift 0 · wasm guest 0 · wasm runner 0
cargo test --all 0
pyre/check.py 0 — ALL PASSED: dynasm 503/503, cranelift 503/503, wasm 496/496
extra_tests/run.py --gated-only — dynasm 119/119, cranelift 119/119
extra_tests/parity_tests/run.py — syntax_error_python314_offsets.py
    cpython=OK dynasm=OK cranelift=OK pypy=xfail
```

The whole snippet corpus goes 366/373 → 367/373 on both backends: `compile_incomplete_input_mode.py` leaves the failure list and the other six are unchanged and ungated. A spot-check of the vendored CPython suite over the eleven syntax- and exception-adjacent modules reports no regressions, and `test.test_codeop` fails the same single assertion before and after.

## What this does not fix

Nothing in the JIT half of the report this branch came from. Recorded so the next pass starts from measurement rather than from the summary:

- **The exception a compiled run exits with reaches the frame's own table from two places, and only one of them is gated.** `deliver_exit_frame_exception` (the `handle_jit_outcome` path) calls `exit_frame_handler_needs_unwritten_stack` and propagates when the exit pc sits in a `lasti` entry whose operand slots the compiled run never wrote back; the hot `eval_loop_jit` arm after `maybe_compile_and_run` calls `handle_exception` with no such test. Either the hazard is real and the hot site shares it, or it is not and the gate is skipping a frame's own `finally`. Four Python-level shapes — function-entry trace and loop trace, escape past a `finally`, escape caught by the frame's own sibling clause, and a frame owning both a `with` and a `finally` — agree across CPython 3.14.2, pyre-dynasm, pyre-cranelift and pypy3 7.3.20, so it is not observable from Python today. It stays a code-level hazard: one delivery, two sites, one test between them, no fixture pinning either answer. The next step is a counter on `exit_frame_handler_needs_unwritten_stack` to find out whether it fires at all.
- **Why that gate exists at all.** PyPy's `jit_merge_point` sits in `PyFrame.dispatch`'s loop and `handle_bytecode` wraps `dispatch_bytecode` in `except OperationError: … handle_operation_error(…)`, so the Python-level handler search is *inside* the traced region and `ExitFrameWithExceptionRef` genuinely means "no handler in this frame" — `warmspot.py:998-1005` just re-raises. pyre's handler search is interpreter machinery outside the trace, so its exit-with-exception does not carry that meaning, and both delivery sites are pyre-only patches for the difference.
- **#1151** (`find_catch_after_resume_live` skips an unbounded prefix before the trailing `-live-`) is untouched. **#1311**'s stated repro no longer reproduces at this base: `pyre/cpython_tests/run.py --backend cranelift --filter test_unittest --jobs 1` runs `test.test_unittest` and reports PASS 1 / CRASH 0 four times out of four (the runner scores CRASH as its own bucket and exits non-zero on it, so this is an exit-status verdict). The bare `!= 0` admission and its `bh_debug_enabled()` dump are still there, so the question is now which source still drives a non-frame word into it — not whether that module crashes.
- The `RERAISE` non-exception-TOS and `asyncio` `TIMEOUT` reports did not reproduce: 1200 timeout iterations, 200k-iteration hot loops, `PYRE_RERAISE_DIAG=1`.
- `exc_kind_matches` (`interp_exceptions.rs`) has no callers outside its own unit test, and its hardcoded hierarchy omits `NotImplementedError` is-a `RuntimeError`. Latent, not observable today; left as it is rather than widened on a dead path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved syntax error messages and source locations for invalid assignment targets.
  - Corrected incomplete-input detection across compilation modes, including unfinished strings and expressions.
  - Error handlers now correctly report `OverflowError` for unrepresentable positions.
  - Improved codec decoding support for additional buffer-providing objects.
- **Tests**
  - Expanded coverage for syntax diagnostics, compiler behavior, codecs, buffer inputs, and valid assignment forms.


youknowone:
The review findings on this PR are answered in #1556, since this one had already merged.

- **`_codecs` `pin_root` return discarded** (CodeRabbit, Major) — real. `pin_root` returns the normalized reference and `let _ =` threw it away, so `simple_buffer_bytes` was handed the pre-normalization `w_obj`. Fixed by rebinding. The nine sibling `let _ = roots.pin_root(…)` sites in `cpyext/pyerrors.rs` are not the same defect: each re-reads its object through `shadow_stack_get(slot)`.
- **`prefix_span` vs `assignment_span`** (CodeRabbit, outside-diff) — real, and this PR introduced it. `bu'x' = 1` reports the prefix message on the assignment target's span here; CPython 3.14.2 and the pre-merge binary both give @1,1-1,3.
- **`eval` gets statement-only assignment diagnostics** (Codex) — real. `eval_input` is `expressions NEWLINE* ENDMARKER` and has no assignment rule, so CPython answers `invalid syntax` at the operator.
- **The duplicated unterminated-literal predicate** (CodeRabbit) — extracted.
- **The charmap handler-position `OverflowError`** (Codex, "regression to PyPy parity") — measured, and not a regression: CPython 3.14.2 raises `OverflowError` where PyPy 7.3.20 raises `IndexError`. The exception type is observable, so AGENTS.md L152-154/L187-191 take the 3.14t behaviour. The message spelling is still off, but it belongs to `int_w` project-wide, not to this codec.
- **The `__buffer__` return annotation** (CodeRabbit) — skipped. ANN204 is the reviewer's own ruff configuration; this repo has no `[tool.ruff]` section, no ruff hook, and 359 of 374 snippets carry no annotations.

The follow-up also fixes something none of the reviews caught and this PR shipped: the assignment diagnostic overrides a literal's *own* failure. `b'ä' = 1`, `'\N{' = 1`, `b'\x' = 1`, `f'{x!z}' = 1` and `f(x for x in y, 1) = 2` all reported an assignment message here where CPython (and this branch's base) report the literal's. `invalid_assignment` is a grammar alternative, reached only by a statement whose tokens lexed — a literal whose content is wrong never gets there. The 645 rows this PR measured contain none of those shapes; a 22-row corpus written for message-vs-span collisions found all seven at once.

— *commented by Claude*

PR #1556, opened 2026-08-28
interp, codecs, jit: the assignment diagnostic yields to a literal's own failure and answers the eval grammar, the decoder reads its pinned reference, and two delivery paths get a name

youknowone:
Follow-up to #1543, which is merged. Three of these are review findings on that PR; one is a defect #1543 itself introduced, found by a corpus written for this PR. Everything is measured against CPython 3.14.2 — the version the CI leg installs — with the pre-#1543 release binary as the control.

## 1. The assignment diagnostic overrode a literal's own failure

`invalid_assignment` is a grammar alternative: it is reached only by a statement whose tokens lexed and parsed. #1543's `assignment_target_error` anchors on the *reported* location and takes any `Assign`/`AugAssign` that contains it, which is one step too permissive — a literal whose own content is wrong never reaches an alternative that names a target.

Seven rows, all regressed by #1543 and all right before it:

| source | CPython 3.14.2 | before #1543 | on `main` today | here |
|---|---|---|---|---|
| `b'ä' = 1` | `bytes can only contain ASCII literal characters` | ✅ | `cannot assign to literal here…` | ✅ |
| `b'ä' += 1` | same | ✅ | `'literal' is an illegal expression for augmented assignment` | ✅ |
| `'\N{' = 1` | `(unicode error) … malformed \N character escape` | ✅ | `cannot assign to literal here…` | ✅ |
| `b'\x' = 1` | `(value error) invalid \x escape at position 0` | ✅ | `cannot assign to literal here…` | ✅ |
| `f() = '\N{'` | the escape, at the **value** | ✅ | `cannot assign to function call here…` | ✅ |
| `f'{x!z}' = 1` | `f-string: invalid conversion character 'z'…` | ✅ | `cannot assign to f-string expression here…` | ✅ |
| `f(x for x in y, 1) = 2` | `Generator expression must be parenthesized` | ✅ | `cannot assign to function call here…` | ✅ |

The scan now yields when an escape error, a mismatched f-string delimiter, the non-ASCII bytes literal, an `f-string:`/`t-string:` message or an unparenthesized generator expression has already claimed `msg`.

The 645-row corpora #1543 shipped with do not contain a single one of these. This PR adds a 22-row corpus built for message-vs-span collisions specifically, which is what found them.

## 2. `prefix_span` before `assignment_span` (CodeRabbit, outside-diff)

The `msg` overrides assign in sequence and the last one wins — fstring → scope → generator → assignment → **prefix** — so the span chain has to be that order reversed. #1543 put `assignment_span` third, ahead of `prefix_span`:

| source | CPython | before #1543 | on `main` today | here |
|---|---|---|---|---|
| `bu'x' = 1` | `'u' and 'b' prefixes are incompatible` @1,1-1,3 | @1,1-1,3 | @1,3-1,6 | @1,1-1,3 |

## 3. `eval` has no assignment rule

`eval_input` is `expressions NEWLINE* ENDMARKER`. There is no assignment rule in it and nothing may follow the expression, so CPython answers plain `invalid syntax` at the **assignment operator** for every one of these — and pyre was handing back statement-only diagnostics. Three shapes fail earlier and name their own first token: a starred element of an *unparenthesized* tuple, at its `*`; an unparenthesized `yield`, at the keyword; and a chain `a = b = 1`, at the **first** `=`. A display keeps its own brackets, so `[*a, b] = 1` and `(a, *b) = 1` still report at the operator.

The parser's own account of the token it stopped on — a keyword standing where an identifier belongs, or a token past the end of an expression — likewise describes a grammar `eval` does not have, and reports the plain failure too.

`assignment_operator_span` also had to stop trimming the gap between the target and what follows it: a parenthesized target's range ends *inside* its own parentheses, so the gap opens with a run of closing ones and `(yield) = 1` reported `) = ` rather than `=`. Reading the operator outward from its `=` settles that and `(a) = 1`, `((a)) = 1`, `(a := 1) = 2` with it.

And the operator span only applies when the assignment *is* the whole source. `def f(): (yield a) = 1` stands behind `def`, which is where the expression grammar stopped; the branch now takes the top-level statement holding the reported location and, when that is not the assignment itself, reports at its leading token — the keyword every compound statement opens with, or the `@` of a decorated one. Ten shapes move onto it: `(yield a)`, `(yield from a)`, `(await a)` and `[x async for x in y]` as an assignment or augmented-assignment target inside `def` and `async def`.

Getting there needed the compile `Mode`, which reached the diagnostic block only as `allow_incomplete: Option<Mode>` and so was `None` for every caller outside `compile()`. It is now a separate parameter on `compile_err_to_syntax_error{,_maybe_incomplete}`, threaded from `_symtable`, `imp`, `pyrex`, `pyre-wasm` and the `eval`/`exec` builtins — four of the five had it in hand already.

## 4. The decoder's buffer acquisition read a stale reference (CodeRabbit, Major)

`pin_root` returns the *normalized* reference and is `#[must_use]`; `decode_input_bytes` discarded it with `let _ =` and passed the argument it was given to `simple_buffer_bytes`. The nine sibling `let _ = roots.pin_root(…)` sites in `cpyext/pyerrors.rs` are **not** the same defect — each re-reads its object through `shadow_stack_get(slot)` and never touches the argument again.

## 5. Two delivery paths get a name

Neither changes behaviour; both are off unless their variable is set.

`PYRE_EXIT_FRAME_DIAG` prints one line per `exit_frame_with_exception` delivery to a frame's own exception table, naming which of the two sites it came from and the verdict `exit_frame_handler_needs_unwritten_stack` reached. Only `deliver_exit_frame_exception` acts on that verdict; the eval loop's site reaches `handle_exception` without consulting it, and now prints what the gate would have said. Refusals alone cannot be read — a run that never delivers and a run whose deliveries all pass look identical — so the passes are counted too.

`PYRE_CATCH_SCAN_DIAG` prints, for every `find_catch_after_resume_live` call, how many startpoints the scan consumed before the trailing `-live-`. #1151 records that skip as unbounded and says it could not construct a reaching case; this is the count that would answer it. The scan is split into `scan_catch_after_resume_live` so the number is a return value rather than a side effect.

## 6. Also here

The predicate and 10-line comment shared by the structured `UnclosedStringError` arm and the `OtherError` message-spelling arm are now one function (CodeRabbit).

The `__buffer__` return annotation CodeRabbit asked for is **not** applied: ANN204 comes from the reviewer's own ruff config, this repo has no `[tool.ruff]` section and no ruff hook in `.pre-commit-config.yaml`, and 359 of 374 snippets carry no annotations at all.

Codex's parity review called the charmap handler-position `OverflowError` a regression against `pypy/module/_codecs/interp_codecs.py`, which substitutes `-1` and raises `IndexError`. Measured: CPython 3.14.2 raises `OverflowError`, PyPy 7.3.20 raises `IndexError`. The exception type is observable, so AGENTS.md L152-154/L187-191 take the 3.14t behaviour; nothing changes here. What *is* still off is the message spelling — `int_w`'s project-wide `"int too large to convert to int"` against `"Python int too large to convert to C ssize_t"` — and that is one shared helper whose exact text `builtins.rs` matches on elsewhere, so it is not this PR's to change.

## Measured

721 sources × exec/eval/single, CPython 3.14.2 as the oracle, the pre-#1543 release binary as the control:

| corpus | mode | rows | matched before | matched after | fixed | regressed | still wrong |
|---|---|---|---|---|---|---|---|
| `hand` | `exec` | 138 | 80 | 129 | **49** | 0 | 9 |
| `hand` | `eval` | 138 | 4 | 131 | **127** | 0 | 7 |
| `hand` | `single` | 138 | 79 | 128 | **49** | 0 | 10 |
| `sweep` | `exec` | 507 | 301 | 464 | **163** | 0 | 43 |
| `sweep` | `eval` | 507 | 2 | 493 | **491** | 0 | 14 |
| `sweep` | `single` | 507 | 301 | 464 | **163** | 0 | 43 |
| `eval` | `exec` | 54 | 42 | 54 | **12** | 0 | 0 |
| `eval` | `eval` | 54 | 4 | 54 | **50** | 0 | 0 |
| `eval` | `single` | 54 | 41 | 53 | **12** | 0 | 1 |
| `collide` | `exec` | 22 | 15 | 17 | **2** | 0 | 5 |
| `collide` | `eval` | 22 | 11 | 18 | **7** | 0 | 4 |
| `collide` | `single` | 22 | 15 | 17 | **2** | 0 | 5 |
| | | **2163** | | | **1127** | **0** | 141 |

**1127 rows fixed, 0 regressed.** Every one of the 141 still wrong is a row the control gets wrong too — an f-string span off by one, a `\N{}` escape message, a prefix or literal error inside the *value* of an assignment, and 9 rows where a compound statement holding an assignment also carries a stray line-continuation, which lands on the operator here and on the `yield` in the control while CPython names the opening keyword.

`pyre/extra_tests/snippets/syntax_error_python314_offsets.py` gains 37 rows: 9 pinning a literal's own failure against the assignment message, 28 pinning the eval grammar. It passes under CPython 3.14.2 and 3.14.5, and still fails under pypy3 7.3.20 at exactly the row its `# pyre-check: pypy-diverges:` header names.

## Also on this branch

Two test-consolidation commits ride along: `tests: prune redundant parity regressions` and `tests: consolidate parity coverage`. Together they move `parity_tests/` from 188 scripts to 147, relocate the ones an existing owner already covers into `snippets/` (where CI's `gate=1` subset runs them), and rewrite the `extra_tests/README.md` rule to *extend the existing owner instead of adding a parity script whenever one exists*. `syntax_error_python314_offsets.py` is one of the relocated ones, so the 37 rows above now run under the snippet gate rather than the parity runner. Net: 474 insertions, 1493 deletions across 58 files.

`gate-triage: register the two exit-frame and catch-scan diagnostics` adds the §7 rows `every_live_gate_has_a_triage_entry` requires for the two new `PYRE_*` names; without them `cargo test -p pyrex --test gate_triage_complete` fails.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved syntax-error reporting across execution modes, including more accurate messages and source spans.
  * Error banners now preserve the provided source filename in WebAssembly execution.
  * Corrected object lifetime handling for certain decoding and garbage-collection scenarios.

* **Tests**
  * Expanded compatibility coverage for syntax errors, tracing, pickling, method caching, built-in subclasses, and standard-library behavior.
  * Added regression checks for moving garbage collection and JIT execution paths.

* **Documentation**
  * Clarified parity-test placement, ownership, citations, and success criteria.
  * Documented optional runtime diagnostics for troubleshooting interpreter behavior.


youknowone:
CI and review response. Of the four red `pyre/check.py` jobs, one failure class is this branch's and two are not.

## Mine: two reinstated `pypy-diverges` markers

```
builtin_subclass_attr_mapdict.py         cpython=OK dynasm=OK pypy=XPASS
foriter_stopiteration_exception_event.py cpython=OK dynasm=OK pypy=XPASS
```

`tests: prune redundant parity regressions` restored both files from copies that predate the commits which dropped their `pypy-diverges:` lines — `ac8105408a4` for the FOR_ITER one. The lane is two-sided, so a script that passes while its header declares a divergence is as red as one that fails without the declaration. `67505f239e9` reports `pypy=OK` for both. Both files are byte-identical to main again.

Worth recording for anyone checking this locally: my pypy3 is 7.3.20 and both scripts **fail** under it; CI's toolcache pypy3.11 is 7.3.23 and both pass. The declaration has to be read off the CI rows, never a local run.

## Not mine: the two synthetic jit-stats fixtures are red on main

```
FAIL dynasm synth/exception_escape_hot_callee_tb_node_once  guard_failures 808 -> 1015
FAIL wasm   synth/exception_escape_hot_callee_tb_node_once  guard_failures 1008 -> 1015
FAIL dynasm synth/exception_traceback_frame_lineno          bridges_compiled 6 -> 4; guard_failures 825 -> 624
FAIL wasm   synth/exception_traceback_frame_lineno          bridges_compiled 6 -> 4; guard_failures 825 -> 624
```

I first read these as mine — both are exception fixtures and this branch had added an env-gated diagnostic inside `eval_loop_jit`'s `ExitFrameWithException` arm, which the meta-tracer records — and withdrew the diagnostics. That was wrong twice over. A local `pyre/check.py --backend dynasm --synthetic-pattern` over both, on a build whose `eval.rs` and `blackhole.rs` were byte-identical to main, reported the same 808 -> 1015 and 6 -> 4. And main's own push run for `67505f239e9` (run 33151839974) has all four rows with all four numbers, and the same four `pyre/check.py` jobs red, with none of this branch on it.

So the withdrawal is reverted and the diagnostics stand. Nothing on this branch outside `extra_tests/` reaches a compiled trace: `builtins.rs` syntax-error reconstruction, the one-line `_codecs` rebind, four call sites that now pass a mode, and two diagnostics that are off unless their variable is set.

This is worth someone's attention on its own. dynasm moving 808 -> 1015 while wasm moves 1008 -> 1015 reads as the two backends converging on one number rather than as noise, and `exception_traceback_frame_lineno` gives up two bridges while dropping 201 guard failures. Neither is a re-record I would make from this branch.

## Not mine either: `test.test_zipfile` is intermittent

```
test.test_zipfile: PASS -> FAIL
  LzmaTestsWithRandomBinaryFiles.test_open -> AssertionError: 222341 != 222416
```

Four local runs of the module on the ubuntu-dynasm shape: three passed, one failed with the same assertion and different numbers (`234565 != 234592`). The class draws a fresh random payload per run — `datacount = randint(16, 64)*1024 + randint(1, 1024)` — so the read-back comes up short by a different amount each time, 75 bytes in CI and 27 locally, and a single run passes about three times in four. Main's push run drew a lucky payload and reported `FAIL 0`.

The assertion is `len(testdata1) == len(self.data)` after reading a `ZIP_LZMA` member in 256-byte chunks, so `ZipExtFile._eof` is being set while the decompressor still owes output — `zipfile.__init__` line 1189 sets it from `self._decompressor.eof or self._compress_left <= 0`. `pyre/cpython_tests/run.py` has no retry, so one unlucky draw turns the job red. I have not opened an issue or attempted a fix here.

## CodeRabbit

- **`specialised_pair_consumers.py:90`, restore repeated execution for `identity`** — taken, with a correction. The premise ("the previous test called each consumer for 3,000 rounds") does not hold: `identity` came from `tuple_float_identity_python314.py`, which ran once at module level with no loop at all. The suggestion is still right on its own terms — every other consumer in the file loops `N` times, and float-slot identity under a compiled trace is what the file exists to pin — so `identity()` now runs the same `N` rounds. The assertion the merge dropped, that two floats of different value are not identical, comes back with it. Passes on dynasm, CPython 3.14.2 and pypy3 7.3.20.
- **`README.md:41`, use `built-in`** — skipped. The stated reason is consistency with line 37, and line 37 contains neither spelling; the only other occurrence in the file is `--filter builtin_dict`, a command that has to stay as it is.
- **Docstring coverage 26.32% < 80%** — skipped. That threshold is CodeRabbit's, not this repo's.

## Codex parity review

§1 reports no PyPy-parity regressions. §2's two entries are the `PYRE_EXIT_FRAME_DIAG` and `PYRE_CATCH_SCAN_DIAG` stderr writes, which have no RPython counterpart; that is what `gate-triage.md` §7 exists to record, and both carry a row there naming the epic that retires them. §3's note on the forward catch scan is explicitly pre-existing — this branch only counted it, and the count is the answer: across 875 files the scan returns a catch 0 times in 25,974 calls, with a skip distance never above 4.

— *commented by Claude*

PR #1575, opened 2026-08-29
importing: run _handle_fromlist in the native importer; type_methods: revive the dead __format__ fast path

youknowone:
Two independent changes that came out of reviewing the `IMPORT_FROM` follow-ups.

## `importing`: import the fromlist's submodules in the importer, not in `IMPORT_FROM`

`importhook` answered a fromlist with the leaf module and imported none of the
names the list adds; `import_from` covered that by calling `importhook` itself and
storing the result onto the parent module. `_bootstrap.py` does this in
`_handle_fromlist`, which runs only where `importlib._bootstrap` is in
`sys.modules`.

That is not everywhere. `"importlib._bootstrap" in sys.modules` reads **False** on
the wasm guest, so the whole run there goes through the native `importhook`, and
`_handle_fromlist`'s work simply does not happen. Removing `import_from`'s import
without supplying it breaks every `from . import X` in the stdlib on that backend —
`encodings/__init__.py`'s own `from . import aliases` is the first casualty.

So `_handle_fromlist` is ported into the native importer rather than worked around:
each name the package does not carry is imported as `<__name__>.<name>`, `*` expands
`__all__` once, a non-str item raises `TypeError` naming where it came from, and a
`ModuleNotFoundError` is swallowed only when it names that submodule and
`sys.modules` does not block it.

`absolute_import` also bound each loaded module onto its parent unconditionally.
`_find_and_load` returns a `sys.modules` hit *before* `_find_and_load_unlocked` runs,
so a cached name no longer rebinds the attribute. Confirmed against both oracles:

| | CPython 3.14 | pypy3 |
|---|---|---|
| `import json.decoder` after `delattr` → `hasattr` | `False` | `False` |
| `__import__("json", …, ["decoder"])` → `hasattr` | `False` | `False` |
| `Item in ``from list'' must be str, not int` | ✓ | ✓ |
| `Item in hflpkg.__all__ must be str, not int` | ✓ | ✓ |

`import_from` is now the `getattr` plus the `sys.modules` lookup for
`<__name__>.<name>` that `pyopcode.py import_from` performs, with no import and no
store. The JIT's `ImportFrom` admission comments are reworded to the attribute
protocol, which is what can still run Python there.

Adds the `handle_fromlist_contract` and `import_from_submodule_fallback` parity tests
and the `foriter_import_from_submodule` bench.

## `type_methods`: recognise the shared builtin `__format__` by its body

`format_value_dispatch`'s fast path was guarded by a `BUILTIN_FUNCTION_TYPE` check on
the looked-up `__format__`. A type-level lookup returns a `method_descriptor`, so the
check was false for every builtin and the path below it was **unreachable**:
`f"{k:>5}"` on an exact `int` wrapped its spec in a `str`, walked the generic call
path into `builtin_value_format`, and read the same bytes back out. 2964 of 3796
profile samples sat in that detour.

It now compares the code object against `builtin_value_format` — the body `int`,
`bool`, `float` and `str` publish, and the one whose call is equivalent to
`format_with_spec_public`. `object.__format__` carries its own body and is not
recognised, so a class defining no `__format__` of its own still reaches the
`TypeError` it raises for a non-empty spec.

Interleaved, min-of-7, CPU time, with two controls:

| leg | before | after | |
|---|---|---|---|
| empty loop | 0.060s | 0.059s | *control* |
| `format(k, "")` | 0.289s | 0.288s | *control* |
| `format(k, ">5")` | 1.074s | 0.792s | **+26%** |
| `format("abc", ">8")` | 0.884s | 0.573s | **+35%** |
| `format(1.5, ".2f")` | 0.994s | 0.684s | **+31%** |
| `f"{k:>5}"` | 1.056s | 0.737s | **+30%** |

`bench/synth/foriter_format_with_spec` passes at its existing ceiling of 25 —
10.7x/12.1x/16.2x against pypy, from 13.0x/15.2x/19.7x. jitstats are unchanged.

Adds the `format_dispatch_shapes` parity test (122 rows, identical on CPython 3.14,
pypy3 and pyre) and the `foriter_format_with_spec_user_dunder` bench, whose
`__format__` is a Python method and so takes the dispatch path this leaves alone.

## Gate

| gate | result |
|---|---|
| `cargo fmt --check`, majit-boundary, `--check-headers` | pass |
| `check.py --backend dynasm,cranelift,wasm` | 518/518, 518/518, 511/511 |
| `parity_tests/run.py` | all pass, incl. the 3 new |
| `extra_tests/run.py --gated-only` | 140/140 × 3 |
| `extra_tests/upstream/run.py` | 1/1 × 3 |
| `cpython_tests/run.py --backend dynasm` | no regressions |
| `cargo test --all --features dynasm,cpyext` | 199 suites ok, 0 failed |

New-fixture ceilings are set at twice the slower native reading
(`foriter_import_from_submodule` 70 for 29x/35x, `foriter_format_with_spec_user_dunder`
165 for 60x/81x). Both record `loops_compiled=1 guard_failures=1` identically on all
three backends.

https://claude.ai/code/session_01S3x6uzbc3NxAE9K4t6VzUG

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for `from ... import ...`, including submodule fallbacks, cached modules, lazy fromlists, and module attribute hooks.
  * Corrected formatting dispatch for built-in types, subclasses, descriptors, custom `__format__` methods, and invalid format specifications.
  * Ensured formatting and import operations behave consistently across supported execution modes.

* **Tests**
  * Added comprehensive parity coverage for formatting dispatch, import-from-list behavior, and submodule fallback scenarios.
  * Added benchmarks covering formatting and repeated submodule imports.


youknowone:
Addressed all four Codex findings in e31572e3a76. Each was reproduced before fixing; all four were real.

**P1 — surrogate fromlist names panicked the runtime.** `__import__('encodings', {}, {}, ['\ud800'], 0)` is ordinary Python and returns the package on CPython 3.14 and PyPy. On the wasm backend it aborted:

```
[pyre panic] panicked at pyre-object/src/unicodeobject.rs:819:
w_str_get_value: backing Wtf8Buf is not valid UTF-8 (lone surrogate)
```

`hasattr(module, x)` is now keyed by the item object and the child name is assembled in WTF-8. A name with no UTF-8 spelling can be neither found, cached nor blocked by an importer that names modules by `&str` — which is exactly the outcome `_handle_fromlist` already swallows.

**P1 — non-str `__name__` was read through the wrong layout.** With `__path__ = []` and `__name__ = 123` in `sys.modules`:

| | before | after |
|---|---|---|
| CPython 3.14 / PyPy | `ModuleNotFoundError: No module named '123'` | — |
| pyre | returned the object | `ModuleNotFoundError: No module named '123'` |

`from_name` is an f-string upstream, so the value is coerced through `py_str_wtf8`. The resulting failure names the *parent*, which is not a name a fromlist swallows — hence the raise.

**P2 — the fromlist was drained before the first item was processed.** The difference is observable, and both oracles agree: an error a later `__next__` raises must not preempt the `TypeError` an earlier item owes, and an import performed for one item must be visible to the `__next__` after it. Now `iter`/`next`, which also drops the per-item root vector.

**P2 — descriptor-owner check.** Confirmed real, and a regression this PR introduced: reviving the fast path lost the `descr_check` the dead path had been performing by accident.

```python
class S(str): __format__ = int.__format__
format(S("x"), ">5")
```
CPython raises `TypeError: descriptor '__format__' for 'int' objects doesn't apply to a 'S' object`; pyre returned `'    x'`. It now reads `__objclass__` and performs `descr_check`. An exact builtin — the whole of what the fast path exists for — settles it by pointer identity, so the optimization is unaffected: `foriter_format_with_spec` reads 9.6x/10.7x/14.4x, against 10.7x/12.1x/16.2x before the check was added.

## Verification

15 rows added across the two parity tests. The descriptor `TypeError` is worded differently by CPython (`descriptor ... doesn't apply to a ...`) and PyPy (`'X' object expected, got 'Y' instead`), so those rows canonicalise the owner and receiver — what they actually assert — rather than xfailing the whole file on PyPy over a message. Neither pattern matches any other row.

Full gate re-run: `check.py` 518/518, 518/518, 511/511 · parity all pass · gated 140/140 × 3 · upstream 1/1 × 3 · CPython suite 225/225 no regressions · `cargo test --all` green.

Two notes on that run, since both were investigated rather than assumed: `test.test_socket` timed out once under `--jobs 3` and passes on re-run (225/225); and a `gate_triage_complete` failure over `PYRE_BH_PLAIN_PROBE` came from an unrelated uncommitted edit in the shared worktree, not from this branch — it is not in any commit here and the test is green now.

`foriter_import_from_submodule` read 29x/33x/41x across three runs, so its ceiling moved to three times the slowest — the headroom `import_from_hot` carries for the same startup-dominated shape.

— *commented by Claude*

PR #1593, opened 2026-08-29
listobject, jit-trace: fold `list.pop()` on Object storage

fregataa:
## What

The orthodox `list.pop()` fold recognized Integer storage only, so `pop()` on a list of arbitrary objects left the trace as an opaque `call_may_force` residual with the frame forced around it. This extends it to Object storage.

## Why the gate was not the blocker

`w_list_pop_end_inner`'s Object arm was a single fused `object_pop()` helper, and the sub-walk cannot lower one.

`AbstractUnwrappedStrategy.pop_end` (`listobject.py`) is `l.pop()` under an identity `wrap`, which rtypes to `ll_pop_default` (`rlist.py`): read the last item, null the vacated slot because the element is a GC pointer, then shrink. The arm is respelled that way through the `ll_list_obj_*` leaves, as the Integer arm already is — so this is the upstream decomposition rather than a new shape.

That needs one leaf the Object family was missing, `ll_list_obj_getitem_fast`, plus its `list.obj_getitem` lowering to `getfield_gc_r(items)` + `getarrayitem_gc_r`, mirroring `list.int_getitem`.

Recognition then accepts Object storage alongside Integer. The sampled element is a live ref rather than a scalar, so the commit pins it across the sub-walk's allocations and reads it back out of its slot.

## Why Float stays out

Its arm still pops through a fused helper, and it would end on a `w_float_new` with no `walker_box_float` to elide it. That absent boxing call is exactly what the fold's guard-ordering property rests on for Integer. The Object arm returns the element itself, so no boxing follows its stores at all and the hazard cannot arise.

## Measurements

700k iterations of `append` + `pop()`, user CPU, dynasm/macOS arm64:

| receiver | before | after | `call_may_force` in loop |
|---|---|---|---|
| int list | 0.04s | 0.04s | 0 → 0 (already folded) |
| **object list** | **0.12s** | **0.05s** | **2 → 0** |
| str (Ascii) list | 0.21s | 0.21s | 6 → 6 (still declines) |

Attribution is an A/B on one build with only the recognize gate's Object arm toggled.

Note this does not move `bench/synth/list_ops`: its `list_obj_append_pop` workload holds `str`, which is `AsciiListStrategy`, not `ObjectListStrategy`. Ascii and Bytes append/pop folds are separate work.

## The snapshot re-record

`pickle_terminal_raise_resume`'s recorded `guard_failures` moves 292 → 316. pickle's unpickler stack (`self.stack = []` with `stack.pop()`) is a strategy-flipping object list, so the fold's `GuardClass` / `GuardValue` / strategy guard are now visible in the trace instead of hidden inside the residual.

Evidence for re-recording rather than narrowing the fold:

| | without fold | with fold |
|---|---|---|
| user CPU | 0.13s | 0.13s |
| loops / bridges / aborted | 30 / 0 / 1 | 30 / 0 / 1 |
| ops after opt | 4454 | 4449 |
| optimization time | 27.6ms | 24.2ms |
| guard_failures | 292 | 316 |

316 is deterministic over 5 runs. A purpose-built worst case — a hot loop flipping the receiver's strategy every iteration — takes 201 guard failures and still runs in the same 0.04s as the monomorphic version.

I would rather have this reviewed than take it on my own reading, so say the word if you would prefer the fold narrowed (e.g. declining a receiver whose strategy has flipped) over the re-record.

## Tests

- `extra_tests/parity_tests/list_pop_fold.py` gains two real `ObjectListStrategy` cases. The existing `pop_object` case holds `str`, so despite the name it exercises Ascii — the declined path — and no Object-strategy case existed. Expectations verified against both the CPython and PyPy oracles.
- `check.py --synthetic-pattern 'list*'`: 18/18 on dynasm.

## Status

Draft: the full `check.py` on this rebased branch is still running. I will mark it ready once it reports green, and report here if it does not.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `list.pop()` handling for object-based lists, including optimized execution paths.
  - Improved reliability when reading and removing object elements from lists.

- **Tests**
  - Added coverage for live and unused results from `list.pop()` with object-based lists.
  - Updated benchmark statistics to capture dropped iterator items and revised guard-failure counts.


youknowone:
The rationale looks good 👍

youknowone (approved):
Thank you!

PR #1594, opened 2026-08-30
stdlib, cpyext: finish asyncio I/O and add PyPy sqlite CFFI

youknowone:
## Summary

- complete the stream, subprocess, tempfile, shutil, and SSL behavior needed by the asyncio stdlib path
- port the PyPy CFFI module-registration surface and use it for the PyPy sqlite fallback, with cpyext coverage
- align GC, JIT-cell, type, and cpyext flag tables with their PyPy/RPython and header owners
- repin all RustPython crates to upstream 1db2006fb3c70929477190ebec3cad6189129b36 after RustPython #8609 merged, removing the local RustPython2 overrides

## Validation

- cargo metadata --format-version 1
- the full suites were not rerun after the final rebase and dependency repin, per request

## Summary by CodeRabbit

* **New Features**
  * Added SQLite autocommit controls, database configuration APIs, validation, and improved transaction handling.
  * Added support for loading PyPy-mode CFFI extension modules.
  * Improved TLS shutdown behavior and close-notify handling.
  * Added support for valid file descriptor `-1` paths and direct POSIX process abort.
  * Added PyPy 3.11 standard-library staging for release builds.

* **Bug Fixes**
  * Improved `input()` output visibility and generator finalizer behavior.
  * Strengthened subprocess, temporary-file, tuple comparison, and garbage-collection compatibility.

### DIFF
--- a/.github/pyre-dist-build-setup.yml
+++ b/.github/pyre-dist-build-setup.yml
@@ -1,3 +1,10 @@
+- name: Set up PyPy for generated CFFI stdlib modules
+  id: pypy-stdlib
+  uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+  with:
+    python-version: "pypy3.11"
 - name: Stage the Pyre standard library
   shell: bash
+  env:
+    PYPY3: ${{ steps.pypy-stdlib.outputs.python-path }}
   run: python3 scripts/stage-stdlib.py dist-assets
--- a/.github/pyre-dist-build-setup.yml
+++ b/.github/pyre-dist-build-setup.yml
@@ -7,4 +7,4 @@
   shell: bash
   env:
     PYPY3: ${{ steps.pypy-stdlib.outputs.python-path }}
-  run: python3 scripts/stage-stdlib.py dist-assets
+  run: python3 scripts/stage-stdlib.py dist-assets ${{ join(matrix.targets, ' ') }}
--- a/.github/workflows/release-pyre-release.yml
+++ b/.github/workflows/release-pyre-release.yml
@@ -169,8 +169,15 @@ jobs:
       - name: Install dependencies
         run: |
           ${{ matrix.packages_install }}
+      - name: Set up PyPy for generated CFFI stdlib modules
+        id: pypy-stdlib
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "pypy3.11"
       - name: Stage the Pyre standard library
         shell: bash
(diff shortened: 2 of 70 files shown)


youknowone:
Force-pushed `54dcbf509f8...224fd74908b`: rebased onto `38f0a743e61`, plus two commits for compile breaks the branch was carrying.

**The branch did not compile before this push.** Two separate causes, both outside any `cfg`, so no feature combination built:

`85686e57886 metainterp: convert the two remaining jc_flags reads to JcFlags` — `warmstate.rs:1684` and `:2362` read `jc_flags::JC_TEMPORARY`, a module the flag table no longer has. Both sit in `impl WarmEnterState`. They now use `flags.contains(JcFlags::JC_TEMPORARY)`, which is what the other three reads in the file already do.

`224fd74908b build: follow the repinned RustPython's malachite bump and format changes` — the repin to `1db2006` landed without the three source changes that revision requires:

- `malachite-bigint` stayed at 0.10 while the new revision is on 0.11 (RustPython #8604), putting two `BigInt` types in the graph and breaking `marshal/mod.rs:805`, `pycode.rs:3022`, `pyopcode.rs:560`, `pyframe.rs:5667`. The workspace requirement goes to 0.11; `pymath` asks for `version = "0"`, so the lock collapses onto one copy rather than keeping both.
- `CFormatConversion` is now an alias of `format::FormatConversion`, which has no `Bytes` — that variant moved up to `CFormatType::Bytes`. Both readers follow it; in `spec_format_bytes` the nested match flattens, since `%s` and `%b` already shared a body.
- `FormatSpecError` gained `IntTooLargeForCLong`, leaving `format_spec_err` non-exhaustive. It answers with the same message the upstream mapping gives it.

Rebase conflicts were two, identical in shape: `main` added `parking_lot` and this branch added `bitflags` to the first line of `[dependencies]` in `majit-gc/Cargo.toml` and `majit-metainterp/Cargo.toml`. Both are needed, so both were kept.

**Validation, stated precisely.** `cargo check --workspace --all-targets --no-default-features --features dynasm,cpyext` exits 0 — but that run predates the rebase, which afterwards touched only those two dependency lists. It has not been re-run since.

Not run: LLBC re-extraction, the release build, and `test_site`. Not for any result they gave, but because they could not be started. Four background tasks were killed this evening, including one that was doing nothing but `sleep 60` and reading two counters — six trees were building on this 32 GB box at load 88 with 17 of 18 GB of swap in use, and a task dies with no error line whether a sibling killed it or the OOM reaper did. Gating on headroom does not help, because the waiter is reaped too.

So the two fixes above are compile-verified and nothing else on this branch has been re-validated since the repin.

— *commented by Claude*

PR #1609, opened 2026-08-30
jit, jit-trace: the zero-argument super() re-route, the proxy read's stored objtype, a null constant's shadow, and the two-argument LOAD_SUPER_ATTR split

youknowone:
Eight commits: one blackhole latch repair left over from the ForIter work, then
the `super()` route — the zero-argument re-route, the proxy read, a walker-wide
constant shadow, and the two-argument `LOAD_SUPER_ATTR` lowering.

## The zero-argument call had a permanent trace-site ban behind it

`try_walker_specialize_bare_super_call` re-routed only the receivers
`_super_check` settles without running Python. The rest declined to the generic
FRAMELESS residual, which enters `builtin_super`'s zero-argument tail, whose
`gettopframe()` runs `force_frame` inside an opaque `bh_call_fn` — read back as
`VableEscapedDuringResidualCall`. Every walk over such a call aborted, and after
`MAX_TRACE_ABORT_COUNT` of them the merge point was stamped `JC_DONT_TRACE_HERE`
for the rest of the process.

It is now a **substitution** rather than a fold, handed to the shared tail in
`dispatch_residual_call_iRd_kind`, which supplies the concrete execution, the
raise transcription and the vrefs bracket — which is what admits a receiver whose
`__class__` is a property. Executing at record time and *then* declining is the
shape that runs such a property twice; the shared path never does it.

On `zero_arg_super_check_needs_python`, all three backends: `loops_aborted`
5 → 0, `fbw_blackhole_adopted_single_frame` 5 → 0, `abort_ceiling_banned` 1 → 0.
Execution on dynasm **1.91s → 0.32s**.

## The proxy was holding the answer the fold was re-deriving

`W_Super.getattribute` walks the `w_objtype` `_super_check` settled at
construction. `try_walker_specialize_load_attr_on_super` re-derived it through
`super_attr_fast_path`, which answers only for the receivers that check settles
without Python. The fold now reads `super_obj_type_descr()` off the proxy and
guards that field; for a *virtual* proxy the read is answered from the
emission's own `SetfieldGc`, so ordinary shapes pay nothing.

`load_attr_on_super consulted=1 fired=0` → `fired=1`; dynasm **0.32s → 0.17s**,
cranelift `~72.9x` → `~39.3x`.

## A null constant-pool slot read back as untracked

`dispatch_via_miframe` and the bridge sub-walk setup skipped the
`concrete_registers_r` write for a `constants_r` slot holding 0, leaving it
`ConcreteValue::Null` (the *untracked* sentinel) where `concrete_value_from_slot`
spells a known null `Ref(PY_NULL)`. `inline_call.rs`'s callee-bank twin already
wrote it unconditionally — that asymmetry is what named the defect.

This is not super-specific: a plain call's `null_or_self` reaches the arglist as
a `ConstRef(0)` const-window entry, so any fold destructuring that operand's
shadow as `ConcreteValue::Ref` declined silently.

## The two-argument LOAD_SUPER_ATTR is a call and an attribute load

The `oparg & 2` form now emits `simple_call(global_super, NULL, cls, self)` then
`getattr(proxy, name)`. The zero-argument form keeps the fused residual, whose
`frame` operand the blackhole path reads.

| N=2,000,000, user CPU | before | after |
| --- | --- | --- |
| `super(C,self).m()`, builtin `super` | 0.07s | 0.07s |
| same, `builtins.super = wrapper` | 2.13s | **0.67s** |
| `s = super(C,self); s.m()` (both) | 0.07s | 0.07s |

The remaining distance on the rebound fused leg is named, not guessed:
`Unavail::Top/CallStack` → `Collapse::ParentUnavailable`. `caller_operand_slots`
knows CALL, FOR_ITER and the binary opcodes as caller resume shapes, and a call
the codewriter synthesizes *inside* a compound opcode is none of them, so the
inlined wrapper's own `super(cls, obj)` stays residual. That is structural — a
synthesized call has no Python pc to resume after — and is left alone here.

## Fixtures

- `bench/synth/zero_arg_super_check_needs_python.py` is new: the bench half of a
  route `bare_super_frame_escape.py` covers only as a selfcheck, where
  `_apply_snapshot_gate` never runs and no `.jitstats` baseline is read.
- `bare_super_frame_escape.py` sites F and G had docstrings saying the virtual
  fold does not answer for their shape; the census reads `fired=1` for each, so
  they are corrected.
- `load_super_attr.py` and `super_attr_inlined_callee.py` name the three folds
  that now answer the two-argument form.

## Gates

`cargo test --all --no-default-features --features dynasm` rc=0 (205 suites) and
`python3 pyre/check.py` green on all three backends — dynasm 525/525, cranelift
525/525, wasm 518/518 — with no `.jitstats` snapshot moved. That run graded the
tree **before** this branch was rebased onto `ff24df96a2a`; the same gate is
re-running locally on the rebased tree now, and I will report the result on this
PR.

https://claude.ai/code/session_01VZKxAmCTcGFbJssJrbPSfV

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `super()` handling when used with proxy objects and apparent classes.
  * Corrected attribute resolution for `super` proxies, preserving the class captured when the proxy is created.
  * Improved reliability for zero-argument `super()` calls involving closures, virtualized frames, and fallback execution paths.

* **Performance**
  * Optimized two-argument `super()` attribute access, including calls made from inlined code.
  * Added coverage and performance benchmarks for specialized `super()` behavior.


youknowone:
## Re-gate on the rebased tree

The gate promised in the PR body has run locally on the rebased tree:

```
selfchecks 526/526   snippets 526/526   benches 519/519   rc=0
```

Green, no SNAPDIFF, no time or ratio failure.

## The one red CI job is not this branch

`pyre/check.py dynasm (ubuntu-24.04)` is red, and it is red for a reason that
predates this PR — the CPython-suite step, not the gate:

```
── REGRESSIONS (1) ──
  - test.test_datetime: PASS -> CRASH  rust panic:
    pyre/pyre-object/src/tupleobject.rs:655:18:
    internal error: entered unreachable code: bounds guard above
```

`main`'s own CI reproduces it byte for byte on the same day
([run 33339170682](https://github.com/youknowone/pyre/actions/runs/33339170682),
and again in [33321049072](https://github.com/youknowone/pyre/actions/runs/33321049072)),
so it is `main`'s failure showing through, not something this branch introduces.
Every other job is green, including `pyre/check.py` dynasm on Windows,
dynasm+cranelift on macOS, and cranelift on ubuntu.

`tupleobject.rs:655` is the `_ => unreachable!("bounds guard above")` arm of
`w_tuple_getitem_known`'s `SPECIALISED_TUPLE_II_TYPE` branch, so some caller
reached index >= 2 on an arity-2 specialised tuple after checking bounds
somewhere else. That is worth its own issue; it is out of scope here.

— *commented by Claude*

PR #1611, opened 2026-08-30
jit-trace, gc, interp: an explicit sub-walk framestack rooted per frame residency, and the fixes it lands on

youknowone:
## The sub-walk framestack

`run_sub_jitcode_walk` descended into an inlined callee by re-entering `walk()`
recursively from inside the CALL dispatch. It now drives an explicit heap
framestack — `SubWalkDriver` over `SubWalkFrame`, with `SubWalkExchange` holding
the state a nested activation hands back — which is the shape upstream has:
`rpython/jit/metainterp/pyjitpl.py:2861-2866` `MetaInterp._interpret` runs
`while True: self.framestack[-1].run_one_step()`, pushing and popping via
`newframe` / `popframe`, and never re-enters the interpreter from inside a CALL
opcode.

`note_subwalk_driver_step` also checkpoints the heap cache, and `HeapCache`
derives `Clone` for it: the driver replays the CALL opcode to deliver a child's
return, so it must rewind that one step's speculative preamble without
discarding what was known before it.

### Rooting the frame's Ref bank

`InlineRegisterBankGuard` publishes a callee's Ref bank on
`PyreSym::inline_register_banks` so `walk_active_sym_register_area` walks it.
Without it, a callee fold's freshly minted `OpRef::ConstPtr` — which embeds a
heap pointer inline — sits in a bank the collector never visits.

The recursive form got this for free by nesting one guard activation inside the
next, so every live bank was published at once. Scoping the guard to a single
`SubWalkFrame::drive` call does **not** reproduce that: a parent suspends at a
CALL by returning `SubWalkSuspended` and **stays on `frames`** while the child
runs, so its guard is already dropped for exactly the window in which the child
walks and allocates.

The last commit roots per frame residency instead — `push_frame` enters the
guard after pushing, `pop_frame` drops it before popping. Three properties:

- `drive()`'s loop has exactly two arms, one pushing and one popping, and its
  only `return` follows a pop that empties `frames`, so the stacks stay in step.
- `InlineRegisterBankGuard::drop` is a bare `pop()`, so retiring N guards in any
  order removes exactly the N contiguous top entries; only partial retirement
  would need LIFO, and `pop_frame` always retires the top.
- The published address is the `Vec<OpRef>` heap buffer, so a `frames` realloc
  moving the `SubWalkFrame` does not invalidate it, and the walk cannot
  reallocate the bank because `WalkContext` borrows it as `&mut [OpRef]`.

`inline_register_banks` has exactly one reader and it iterates the whole list —
it is a GC root set, not a current-frame pointer.

## Alongside

- `GETFIELD_GC` consults the descr's storage shape before asking the collector
  for a type-id word; a raw or headerless struct has none. `pyobject::PyType`
  is declared raw for the same reason.
- The `FBW_BRIDGE_DECLINED` reset moves to once per attempt
  (`reset_fbw_bridge_declined`, from `call_jit.rs`). One
  `MetaInterp.interpret()` can cross several merge points, and resetting inside
  `trace_bytecode` erased a decline recorded by an earlier segment of the same
  attempt.
- Stores to the callee's own virtualizable are exempt from the freshness check
  while the call is not deferred.
- Ten fixes already on the branch: `_ctypes` `py_object` container slot, the
  `SyntaxError` incompatible-prefix scan, the int/float/complex unary dunder
  slots, the augmented-assignment operator symbol, quasi-immutable forcing,
  generator yield frame exits, and others.

## Verification status — please read

- `cargo check --workspace --all-targets` is clean (0 errors; the only warnings
  are pre-existing and in files this branch does not touch). LLBC artefacts
  re-extracted and stamped.
- **`pyre/bench/synth/pickle_terminal_raise_resume.{dynasm,cranelift,wasm}.jitstats`
  carry an unverified `guard_failures` value** — 317 / 317 / 335. They came out
  of a rebase conflict where this branch's quasi-immutable commit moved the
  counter 292→293 and main's `list.pop()` fold (#1593) moved it 292→316. The
  merged value is not derivable by arithmetic — this branch also moves
  `loops_aborted` 1→3, which can shift guard counts on its own — and the box
  could not host a measuring run (load 100-180, swap 18 of 19.5 GB). These three
  numbers must be replaced by a real run before merge; the gate is
  bidirectional, so a wrong value is a red whichever way it errs.
- `check.py` has not been run on this tree yet.

The two `WIP:` commit messages say the work was not compiled or run. That was
true when they were recorded and is no longer true of the tip.

— *authored by Claude*

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved moving-GC reliability for `ctypes.py_object`, exception groups, traceback locals, and nursery tracking.
  - Corrected augmented-assignment error messages, unary operator behavior, folded reference-field values, and exception routing.
  - Improved support for larger type registries and deeper inline calls.
  - Improved invalid string-prefix syntax handling and speculative field protection.

- **Tests**
  - Added regression and parity coverage for garbage collection, operators, syntax errors, nursery tracking, unary operations, and deep call chains.


youknowone:
Reviewing the inherited `fbw_state.rs` hunk in this PR (the `callee_owned_frame` exemption) — one concrete concern and one to confirm. Neither is a blocker I can settle here: this box is OOM-bound (swap 20.5 of 21.5 GB) so I could not run `check.py` to see whether either is reachable.

**The `!deferred_call` half looks right.** `rpython/jit/metainterp/virtualizable.py:227-241` stamps `virtualizable.vable_token = TOKEN_TRACING_RESCALL` before a residual call and clears it after, so upstream also treats a residual call as the virtualizable's publication boundary. The exemption's shape maps onto that.

**1. `vable_reg` is latched once and never invalidated.**

`fbw_callee_body_replay_scan` recomputes `fresh_ref_regs[dst]` on every ref-producing instruction (`fbw_state.rs:3936`), so the freshness lattice is flow-sensitive. `vable_reg` is not: it is `get_or_insert`-latched from the first `*_vable_r*` operand 0 (`:3483`) and appears nowhere else except the test at `:3895`. Nothing clears it when that register is overwritten.

So if register R is latched as the frame register and later rewritten with a **non-fresh** ref — a `getfield_gc` result, a residual call result — then `setfield_gc` targeting R still takes `callee_owned_frame`, and the `!fresh_ref_regs[R] || !immutable_field` poison it would otherwise get is skipped. That is a store to an arbitrary live-heap object admitted into a replayable body.

The exemption only *matters* when the original condition would have fired, so the reachable shape is exactly: latch R as the vable, overwrite R with something not fresh, then `setfield_gc` R with a mutable field. Whether a register allocator ever reuses the frame register that way is the open question — but this is a soundness scan, so "the allocator probably doesn't" is not the bar.

Cheapest fix: clear `vable_reg` at the same site that already recomputes `fresh_ref_regs[dst]` (`:3936`) when `dst == vable_reg`. That is strictly more conservative — it can only add poison, never remove it. I did not apply it because the jitstats gate is bidirectional and I cannot regenerate the counters on this box, so a blind tightening would land an unverifiable red.

**2. `deferred_call` is narrower than "residual call".**

It is set only for the `defer_helper` allowlist (`LoadAttr`, `BinaryOp`, `CompareOp`, `SetCurrentException`, `LoadSuperAttr`, …). The `replay_safe_read` calls — `load_global`, `load_deref`, `get_current_exception` — are residual calls that do not set it, while upstream would stamp `TOKEN_TRACING_RESCALL` before them all the same. They are pure reads, so they should not force the frame; the case worth confirming is a raising path (`load_global`'s `NameError`) where an unwind captures the frame into a traceback and publishes it before the store is replayed.

Also note the two `WIP:` commit messages say the work was not compiled or run — that was true when they were recorded, not of the tip. `cargo check --workspace --all-targets` is clean here.

— *commented by Claude*

youknowone:
Follow-up on the three unverified `guard_failures` values flagged in the PR body — only **two of the three are gated**.

- `pickle_terminal_raise_resume.dynasm.jitstats` (317) and `.cranelift.jitstats` (317) are covered by `pyre-ci.yml:681` `pyre-check-linux`, whose matrix is exactly `dynasm` + `cranelift`. If either estimate is wrong, this PR goes red and the run prints the real number.
- `.wasm.jitstats` (335) is covered by **nothing**. There is no wasm leg in that matrix, and `rg -ln 'wasm_check' .github/workflows/` returns nothing — no workflow runs `wasm_check.py`. The only CI job that reads a `.wasm.jitstats` value is the wasm codegen regression test (`pyre-ci.yml:945`), and it hardcodes just `compiles` = `loops_compiled` + `bridges_compiled` (`codegen_test.rs:259`, `:303`), which this change does not touch.

So the wasm number will not go red whether it is right or wrong — it would simply become the next baseline. Once the two Linux legs confirm the real delta I will apply the same correction to the wasm file rather than leave the estimate standing.

— *commented by Claude*

youknowone:
### Review findings — status

Rebased onto `ab5995fdcf6` (17 commits, no conflicts; `git range-diff` reports every patch `=`).

**Codex P1, `cdata.rs` — table publication race.** Fixed in `e93b9b98d9f`. The load/store pair let two threads adding their first `py_object` both read zero, and the second store replaced a table the first had already appended to. Now a `compare_exchange`; the loser's empty list is unreachable and collectable.

**Codex P1, `cdata.rs` — slot index race.** Fixed in `3cc44972d46`. Correct as reported: `w_list_append` releases the list lock before `w_list_len` reacquires it, so two adders can both append and then read the same length. `w_list_append_returning_index` (new, `pyre-object/src/listobject.rs`) answers the index the value landed at under the single guard the append already holds. Its body repeats `w_list_append`'s rather than delegating — that wrapper's shape is what the append fold descends, and routing it through a further call resolves `list_append_jitcode()` to `None`, turning every `list.append` into a `Void` residual.

**CodeRabbit, `majit-gc/src/trace.rs` — `T_MEMBER_INDEX` width.** Fixed in `e93b9b98d9f`: `TypeRegistry::register` now asserts the id fits the 16 bits `encode_type_shape` round-trips it through.

**CodeRabbit, `gc_varsize_item_const_shape.py` — unused parameter.** Fixed (`mid(i)` → `mid(_i)`).

**CodeRabbit, `metaclass.rs` — "retain ordinary `py_object` values after these writes".** Not taken. The weak cell is upstream's own semantics, not a gap this PR introduces. `lib_pypy/_ctypes/primitive.py` stores `weakref.ref(obj)` in `GlobalPyobjContainer.add`, falling back to a closure only for operands `weakref.ref` refuses, and its `'O'` `_setvalue` is exactly

```python
def _setvalue(self, val):
    num = pyobj_container.add(val)
    self._buffer[0] = num
```

with no `keepref` and no `_objects_` entry — there is no `'O'`-specific retention anywhere in `lib_pypy/_ctypes`. The `store_reference` path in `structure.py.__set__` is driven by `ensure_objects(cobj)`, i.e. by whether the CData value carries `_objects`, not by the typecode. Adding retention for `tc == "O"` would make pyre keep objects alive where PyPy does not.

Also in this push: `93b4fbd9693` restores `synth/recursion_memo_branch`'s three jit-stats baselines. They had been re-recorded while "jit: trace generator yield frame exits" was still in the tree; that commit is reverted here, its counters were not. check.py observed `bridges_compiled=14, guard_failures=2716, loops_aborted=1, fbw_blackhole_adopted_multi_frame=1` on cranelift (ubuntu) and again on dynasm (windows) — the four values main already records — so the delta was in the files, not the code.

Locally verified: `cargo check -p pyre-object` and `cargo check -p pyre-interpreter --features cranelift` both clean. Under `--features sandbox,dynasm` the only error is the pre-existing apple-gated `libc::setgroups` in `posix/interp_posix.rs`, which is byte-identical on main and unreachable on the ubuntu runner that leg uses.

— *commented by Claude*

youknowone:
**Codex P1, `cdata.rs` — release/acquire ordering.** Fixed in `7f66d770f4e`. Correct as reported: the exchange installed a pointer to a list this thread had just written, and both it and every reader used `Relaxed`, so nothing ordered the list's initialising writes before the losing thread's load. The exchange now releases and acquires, both loads acquire, and the root walk's write-back releases — the spelling the tree already uses for a CAS-published pointer in `pycode.rs` and `interp_exceptions.rs::register_exc_class_for_kind`.

`virtualref.rs`'s `TRACING_RESCALL_DUMMY_PTR` is the one nearby `Relaxed, Relaxed` exchange left alone on purpose: that token is compared by address and never dereferenced, which its comment already states.

Verified locally: `cargo check -p pyre-interpreter --lib --features cranelift` clean.

— *commented by Claude*

youknowone:
### `check.py cranelift (ubuntu)` — `synth/load_name_builtin_cell_fold`, inherited from main

```
FAIL cranelift synth/load_name_builtin_cell_fold  exec 2.24s > pypy 0.96s
  ratio 2.3x > gate 1x; needs 112x pypy startup to be noise
```

Not from this branch, and I have left it alone rather than editing a fixture I do not own.

* This branch does not touch `load_name_builtin_cell_fold.py`; `git diff origin/main HEAD` on it is empty.
* The last commit to touch it is `ab5995fdcf6` (#1615), merged to main about three hours ago.
* #1615's own message records the reading: *"Run 33315418828 read the fixture at 1.7x on ubuntu dynasm, **2.4x on ubuntu cranelift**, 1.6x on windows and 0.46x on macos cranelift"*. The observed 2.3x is that number.
* It set `max-pypy-ratio` from `2.8` to `1`. The size reduction that was to accompany it — *"N and M drop from 900000000/4000000 to 100000/10000"* — is not in the merged tree: `N = 900000000` and `M = 4000000` appear as unchanged context in #1615's diff, and `skip-cpython` also stayed although the same paragraph says it "goes too". The final sub-commit, `bench: keep load-name pypy ceiling without a floor`, carries no body, so there is no recorded rationale for the ceiling of `1` at the original trip counts.
* The census half passed — `spec-folds=builtin_len` is green, so the fold still fires. Only the ratio gate failed.
* main's own runs since #1615 were all cancelled by the next push, and the `check.py cranelift (ubuntu)` job on the current main run (`ad1c353e935`) was still queued at the time of writing, so this has not yet been observed on main itself.

The fix belongs on main — either restore the `2.8` ceiling or land the intended trip-count reduction alongside the `1`.

Meanwhile the failure this branch *did* own is gone: `synth/recursion_memo_branch` no longer appears, and the cranelift leg went from `1 failed, 525 passed` to `1 failed, 528 passed`.

— *commented by Claude*

PR #1620, opened 2026-08-31
rtyper: preserve tuple identity in enum payload rows

youknowone:
Continues #346 after #1600.\n\nThis registers concrete tuple payload rows for generic enum variants and projects them through the same per-shape class identity used by the synthetic tuple constructor. It follows RPython TupleRepr's GC-managed tuple shape and does not add a permissive union rule or a parallel side table.\n\nVerification:\n- actual pyre-interpreter LLBC regression: mini_buffer_params_preregisters_tuple_payload_row passes\n- existing generic enum variant repr regression passes\n- release prepass/build passes\n- before rebasing #1600 onto current main: phaseA 1724 -> 1724, phaseB 1 -> 0, skip 1715 -> 1714\n- after rebasing onto current origin/main and freshly re-extracting the default LLBC set: phaseA 1716, phaseB 0, skip 1706, direct divergence hits 0\n- mini_buffer_params reaches CodeWriter\n\nThe two census rows are intentionally separate because current main changed the corpus; the fixed-corpus pre-rebase comparison is the direct delta for this patch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tuple payloads in enum variants and option-like values.
  * Fixed nullable values and null array storage to preserve their correct list or string representations.
  * Improved scalar method routing for payload-free enums.
  * Corrected UTF-8 buffer builder detection and mutable accumulator handling.
  * Improved vector extension behavior for translated code.
  * Ensured built-in codec error handlers are registered during interpreter startup.

### DIFF
--- a/majit/majit-gc/src/lib.rs
+++ b/majit/majit-gc/src/lib.rs
@@ -3096,7 +3096,12 @@ pub fn gc_shrink_array(addr: usize, smaller_length: usize) -> bool {
 }
 
 /// Return the active collector's existing TYPE_INFO layout for `addr`.
-/// `None` means it is not a registered variable-size GC object.
+///
+/// `None` means either that `addr` is not a registered variable-size GC
+/// object, or that no backend has installed the query — the same answer
+/// [`gc_owns_object`] gives for an address no GC owns.  A caller that needs
+/// a layout regardless supplies its own, the way `jit_ll_arraymove` falls
+/// back to the array token of the block it allocates.
 pub fn gc_varsize_layout(addr: usize) -> Option<GcVarSizeLayout> {
     if addr == 0 {
         return None;
--- a/majit/majit-translate/src/annotator/bookkeeper.rs
+++ b/majit/majit-translate/src/annotator/bookkeeper.rs
@@ -2331,11 +2331,22 @@ impl Bookkeeper {
                 }
             }
         }
+        let is_enum_variant = {
+            let guard = self.struct_fields.borrow();
+            guard.as_ref().is_some_and(|registry| {
+                n.rsplit_once("::")
+                    .is_some_and(|(parent, _)| registry.is_enum_base(parent))
+            })
+        };
         for (field_name, field_ty) in &fields {
             if field_name == "__class__" {
                 continue;
             }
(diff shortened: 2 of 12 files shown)


youknowone:
Second #346 slice pushed as 15a4f7cf0ee. Payload-free enums already lower to an integer discriminant, so their inherent Rust methods now use the statically registered graph path instead of a Python getattr on SomeInteger. PyPy's corresponding buffer request is an integer BUF_* flag; no union or method-table exception was added.\n\nFresh fixed-corpus result: METHOD-RESOLUTION-GAP 22 -> 11 and all require_contiguous-on-Integer failures disappear. Overall phaseA remains 1716, phaseB 0, skip 1706: the unblocked buffer_bytes path now exposes the next walls (core::array::index and SimpleBufferBytes.data Instance/List), so this is an onion-frontier move rather than a count-only win. Actual-LLBC regression and release prepass/build pass.

youknowone:
#346 progress at fb0402366d5 on PR #1620.\n\nThe remaining _codecs phase-B failure was not a missing function-pointer repr. PyPy marks interp_codecs.register_builtin_error_handlers with @not_rpython and invokes it from the host-only MixedModule constructor. Pyre had moved that work into translated CodecState::new, creating a runtime array of Rust function pointers that upstream never annotates.\n\nThis commit restores the upstream boundary: CodecState::new only creates state; install_builtin_modules performs builtin-handler registration during host setup; register_builtin_error_handlers, install_builtin_modules, and their startup-only caller init_sys_path carry the existing not_rpython marker. The provisional Option/Tuple frontend compensation was removed rather than retained.\n\nFresh LLBC/prepass census:\n- phaseA: 1734 -> 1724\n- phaseB: 1 -> 0\n- MAJIT_RTYPER skip: 1724 -> 1713\n- direct divergence: 0\n\nValidation:\n- cargo check --all --no-default-features --features dynasm\n- cargo test --all --no-default-features --features dynasm\n- python3 pyre/check.py: dynasm 531/531, cranelift 531/531, wasm 524/524\n- actual-LLBC marker regression passed\n\nForce-pushed rtyper2.

PR #1633, opened 2026-08-31
Step the set-like dict-view walks one item at a time, and root their operands across user code

youknowone:
Rebased onto `afc274b880c`. The two bench commits this PR opened with are **gone** — main fixed both reds itself while this branch sat, and replaying my recordings would have made things worse. Details below, because one of them applied cleanly and would have shipped a new red.

## What the branch carries now

| commit | what it does |
|---|---|
| `6909dee9acf` | root both `compare_slot` operands across the set conversion |
| `7030b35acfb` | root the dict-view operands the set ops hold across user code |
| `67c11db8288` | **`Port JitCode lowerer parity fixes`** — not mine, it is the JitCode lowerer work that already lived on this branch |
| `22262b7eb1a` | step the set-like dict-view walks one item at a time (the review fix, below) |

## The two dropped commits

Main's own CI at `d5d186a98ef` (the run after #1624) says both problems are already solved:

| fixture | ubuntu dyn | ubuntu cl | windows | macos dyn / cl | on main |
|---|---|---|---|---|---|
| `load_name_builtin_cell_fold` | 1.7x / 1.8x | 2.3x | 0.9x | 0.7x / 0.5x | **passes** at ceiling 2.5, floor 0.4167 |
| `loop_in_try_tail_unbound_check` | — | — | — | 1.6x / 2.0x | passes, floor 0.9667 |
| `short_circuit_boxed_int_cross_fn` | — | — | — | 1.5x / 2.1x | passes, floor 1.0 |
| `short_circuit_value_kept_stack` | — | — | — | 1.5x / **3.8x** | passes, floor 1.0 |
| `swap_except_return_resume` | 1.4x / 1.8x | 2.3x | ?1.8x | 1.5x / 1.9x | passes, floor 0.8667 |

`pyre/check.py` is `170/170 passed` on ubuntu dynasm and cranelift there; macos's only check.py failure is `cranelift raise_catch 5.0x`, which is reserved.

- **`retire load_name_builtin_cell_fold's pypy ratio again`** — my premise was that no single ceiling covers the host span while still deriving a floor beneath it. #1624 refuted it: `max-pypy-ratio=2.5` covers `[0.5, 2.3]` with floor 0.4167 under it. My commit would have deleted a working gate. It conflicted, which is what made me look.
- **`re-record four ceilings the derived floor no longer fits`** — applied **with no conflict at all**, and would have introduced a new failure: its `short_circuit_value_kept_stack` ceiling is 3.7 against main's macos cranelift reading of **3.8x**. The four floor failures it was written for no longer happen (pyre now reads 0.22s on macos where it read 0.05s), so the commit was simply obsolete.

A bench header is a recording of a measurement, not source. Replaying one onto a newer base replays a stale reading, and a clean apply is not evidence.

## The code review

Codex's parity review of the previous head reported **no regressions introduced** and no other new mismatches. Its two section-3 findings were pre-existing deviations, both in the functions these commits already touch. Both reproduce as wrong answers, so `22262b7eb1a` fixes them.

**1. `descr_isdisjoint` materialised its operand.** Upstream opens `space.iter(w_other)` and tests `space.contains_w(self, w_item)` per step, answering `False` at the first common element. `dict_view_isdisjoint` called `collect_iterable` first:

```python
def gen():
    yield 1
    raise ValueError("boom")
{1: 'a'}.keys().isdisjoint(gen())
```

CPython 3.14 and PyPy both answer `False` having consumed one item. pyre raised `ValueError` and consumed all of it. The set-like branch that iterates the shorter operand was missing too, so it is added.

**2. `_all_contained_in` read a snapshot.** Upstream walks `space.iter(w_dictview)`, so a `contains_w` that resizes the dict is caught by the view iterator's own size check. With a `__eq__` that grows the dict, CPython and PyPy both raise `RuntimeError: dictionary changed size during iteration`; pyre answered `True`. pyre's dict-view iterator already implements the guard — only the snapshot bypassed it.

**3. `compare_slot`'s set-vs-view arm** reduced both operands to sets built from that same snapshot, so `set == view` had the same blind spot from the other side. A set answers `NotImplemented` for a view and the reflected call on the view decides, so the arm now hands the pair to `dict_view_compare` with the operator reflected when the view is on the right. Every non-mutating ordering case already agreed before and after, which is the evidence that the reflection mapping is answer-preserving.

Both walks keep their operands, the iterator and the yielded item in shadow-stack slots and read them back at each use; the item slot is rewritten per turn rather than pinned per item.

## Witness

`pyre/bench/synth/dict_view_setlike_lazy_walk_regression.py`, a `selfcheck` guard with `selfcheck-compiles=warm` so the JIT has to compile the walks it covers. Every expectation is the value CPython 3.14 and PyPy both produce. On the pre-fix binary **8 assertions fail**, before and after warmup:

```
FAIL isdisjoint_keys_hit_then_raise: expected False, got !! ValueError
FAIL isdisjoint_items_hit_then_raise: expected False, got !! ValueError
FAIL isdisjoint_stops_at_first_hit: expected (False, 1), got (False, 5)
FAIL eq_mutating_eq: expected !! RuntimeError, got True
FAIL eq_mutating_eq_len: expected 3, got 4
FAIL le_mutating_eq: expected !! RuntimeError, got True
FAIL set_eq_view_mutating_eq: expected !! RuntimeError, got True
FAIL set_eq_view_mutating_eq_len: expected 3, got 4
```

## Verification

- `pyre/check.py --backend dynasm` — **532/532, ALL PASSED** (531 before; the new fixture is the extra).
- `cargo fmt --all -- --check` — clean.
- `scripts/check-new-line-citations.py` — clean.
- `dict_view_set_ops.py` output unchanged against CPython.
- The two rooting commits remain latent: their witnesses pass on the pre-fix binary too, with and without the JIT, so neither has a witness of its own. That is stated in their commit messages.

The 532/532 run was made just before the last rebase onto `afc274b880c`; the rebuild and re-gate at that base is running now and I will report it here if anything moves.

## Not touched

- `cranelift raise_catch` on macos — reserved.
- `baseobjspace::call_method` / `call_method_result` publish the receiver and args *inside* the `method_descriptor_shortcut` arm, leaving the args slice unrooted across the shortcut probe and the `getattr_str` fallback. Upstream `callmethod.py call_method_opt` receives `arg_w` as a tuple the collector traces, which argues the rooting belongs in the callee — but `pyframe.rs call_mapping_method` publishes in the caller, so the contract wants settling before a hot generic path is changed.

— *authored by Claude*

## Summary by CodeRabbit

* **New Features**
  * Method-call syntax is now supported across configuration and value-lowering paths.
  * Integer and floating-point comparisons in branches are compiled more directly, including negated conditions while preserving expected behavior.

* **Bug Fixes**
  * Improved garbage-collection safety and live mutation handling during dictionary-view and set operations.
  * Preserved live-register ordering when processing individual liveness markers.

* **Improvements**
  * JIT register counts now use a compact format with validation for supported limits.
  * Expanded coverage for method calls, dictionary views, lazy iteration, and optimized comparison branches.

### DIFF
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/dispatch.rs
@@ -865,7 +865,7 @@ fn lower_extended_arg_inner_while(
         quote::quote! { let _ = __builder.live_placeholder(); },
     );
     lowerer.emit_op(
-        OpMeta::conditional_guard_int_eq(
+        OpMeta::conditional_guard_compare(
             Register::int(opcode_reg),
             Register::int(extended_arg_const_reg),
             after_loop.clone(),
@@ -1085,7 +1085,7 @@ fn try_lower_have_argument_guard(lowerer: &mut Lowerer, stmt: &Stmt) -> bool {
         quote! { let _ = __builder.live_placeholder(); },
     );
     lowerer.emit_op(
-        OpMeta::conditional_guard_int_eq(
+        OpMeta::conditional_guard_compare(
             Register::int(local_reg),
             Register::int(const_reg),
             ok_label.clone(),
@@ -2727,7 +2727,7 @@ pub(super) fn lower_dispatch_chain(
                 // Fused goto_if_not_int_eq: branch to the next alternative (or the
                 // arm skip label) if opcode != const.
                 lowerer.emit_op(
-                    OpMeta::conditional_guard_int_eq(
+                    OpMeta::conditional_guard_compare(
                         Register::int(opcode_reg),
                         Register::int(const_reg),
                         miss_label.clone(),
--- a/majit/majit-macros/src/jit_interp/jitcode_lower/lower_control.rs
+++ b/majit/majit-macros/src/jit_interp/jitcode_lower/lower_control.rs
@@ -12,20 +12,6 @@ fn literal_nonnegative_i64(expr: &Expr) -> Option<i64> {
 }
 
 impl<'c> Lowerer<'c> {
-    /// True when any op recorded since index `since` is a float
-    /// comparison — an `OpKind::BinopF` whose result register is int
-    /// banked.  Float arithmetic writes a float result; only the value-
-    /// form `float_lt/le/eq/ne/gt/ge` (`ff>i`) writes an int.  A float
-    /// comparison feeding a conditional guard grows a bridge that hangs the
-    /// compiled trace, so the branch lowerers roll back and use interpreter
-    /// fallback when the condition lowered through one.
-    fn ops_since_contain_float_compare(&self, since: usize) -> bool {
-        self.op_metadata[since..].iter().any(|op| {
-            matches!(op.kind, OpKind::BinopF)
-                && op.writes.iter().any(|w| matches!(w.kind, BindingKind::Int))
-        })
-    }
-
     /// Lower an `if` / `while` condition, peeling a leading `!` into the
     /// branch instead of computing it.
     ///
@@ -40,7 +26,7 @@ impl<'c> Lowerer<'c> {
     /// and does not exist. A condition is `bool` by the language's own typing
     /// rule, which is the guarantee `optimize_goto_if_not` spells out as
     /// `v.concretetype != lltype.Bool: return False`.
-    pub(super) fn lower_condition(&mut self, cond: &Expr) -> Option<(Binding, bool)> {
+    pub(super) fn lower_condition(&mut self, cond: &Expr) -> Option<LoweredCondition> {
         let mut negated = false;
         let mut expr = cond;
         loop {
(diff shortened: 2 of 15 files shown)


youknowone:
Both review findings were real. Fixed in `23c4008eed2`.

**Codex P1 — root the view before the comparison helper.** Confirmed, and it is a wrong answer rather than only an unsound window.

First, that the window is reachable at all — a set subclass's Python `__len__` does run on every arm, in pyre:

```python
class S(set):
    def __len__(self):
        gc.collect(0)
        return set.__len__(self)

S({1}) == d.keys()      # the override runs
d.keys() == S({1})      # so does this one
S({1}) <= d.keys()      # and here
d.keys().isdisjoint(S({9}))
```

Then the discriminator. I built two binaries from this tree differing **only** in the rooting hunk — the second by reverse-applying that hunk and rebuilding — and ran them interleaved in the same environment under `PYPY_GC_NURSERY=64K`, over 20000 comparisons of a collecting `__len__` against a dict view nothing else holds:

```
round 1  unrooted: wrong answers: 1   rooted: wrong answers: 0
round 2  unrooted: wrong answers: 1   rooted: wrong answers: 0
round 3  unrooted: wrong answers: 1   rooted: wrong answers: 0
round 4  unrooted: wrong answers: 1   rooted: wrong answers: 0
round 5  unrooted: wrong answers: 1   rooted: wrong answers: 0
```

The default nursery reports 0 on both, so the shape only surfaces with the nursery small enough for the churn to force the relocation.

Both operands are now published before either length call and read back from their slots, and each length is reduced to its native value where it is taken, so the boxed int the first call returns is no longer held across the second. Rooting went into `dict_view_compare` rather than its callers so every caller is covered.

**CodeRabbit — `raising_after([1])` is a generator, so `iter()` on it runs no user code.** Correct, and it means nothing exercised the receiver root `dict_view_isdisjoint` takes before `space.iter(w_other)`. Added `CollectsOnIter`, whose `__iter__` calls `gc.collect(0)` before returning its iterator, invoked on a temporary view — plus the collecting-`__len__` subclass on both sides of all six operators.

`pyre/check.py --backend dynasm` is 532/532.

— *commented by Claude*

PR #1637, opened 2026-09-01
object, jit, backends: the step-one range iterator shapes, an eval-breaker floor compare, and per-op machine-code offsets

youknowone:
Three independent changes, each measured.

### `object`: the two `step == 1` range iterator shapes

`descr_iter` now picks `W_IntRangeStepOneIterator` or `W_IntRangeOneArgIterator`
when `W_Range.promote_step` is set, which `descr_new` sets when it saw no step
argument. Both shapes hold `stop` instead of the general shape's `remaining`
counter. `stop` is the sole `_immutable_fields_` entry on either shape,
matching `functional.py:766` and `:797`; `start` and `W_Range.promote_step`
are plain fields, since `W_Range` upstream declares no `_immutable_fields_`.

They register behind the unconditional code-object walks as type ids 183 and
184. Type ids are positional — `build_gc`'s registration order assigns them and
`#[pyre_class(type_id = N)]` only drift-checks — and the conditionally-compiled
types have to stay one contiguous trailing slice, because that is what
`active_subclass_range_hierarchy()` truncates under `sandbox`. So the two new
shapes go behind the last unconditional registration, and the gated tail
(posix, `_ssl`, `mmap`, both Windows overlapped owners, `_WindowsConsoleIO`,
`_cffi_backend`) moves up by two in `SUBCLASS_RANGE_HIERARCHY` and in both
`all_subclass_range_aliases` lists.

All three shapes carry the same `range_iterator` type object in their
`w_class`, so `type()` identity holds across them and `is_range_iter` covers
all three.

Measured, same program, one binary, optimized trace between LABEL and JUMP:

| loop | body | iterator ops |
| --- | --- | --- |
| `range(1, n, 1)` (general) | 12 ops | `IntGt`, `IntAdd`, **`IntSub`**, **`SetfieldGc` x2** |
| `range(1, n)` / `range(n)` | **10 ops** | `IntLt`, `IntAdd`, `SetfieldGc` x1 |

No `GetfieldGc` survives in either body — `stop` is loop-invariant and hoists. Because the shape turns on `promote_step`
rather than on `step == 1`, `range(a, b, 1)` keeps the general shape.

Parity: a fixture covering all three shapes with `__length_hint__`,
`__reduce__`, `__setstate__`, pickling and `type()` identity matches CPython
byte for byte, and the vendored `test.test_range` passes against its baseline.

### `jit`: an eval-breaker floor compare instead of a mask

The eval-breaker bits are renumbered so the flags a compiled loop must stop for
sort above the ones it may ignore (`EB_GC_INTERP = 1`, `EB_ASYNC = 2`,
`EB_STW = 4`, `EB_FINALIZING = 8`). The back-edge poll then emits
`UintGe(word, JIT_BREAKER_FLOOR)` instead of a mask and a test, one op fewer
per poll.

### `backend`: a per-op machine-code offset from both native backends

dynasm printed the `MAJIT_DUMP` offset line only from the arms that produce a
result, so a guard's or a discard's bytes were read as part of the preceding
op's span and every span after a trace's first guard named the wrong operation.
Both arches now open a span for those two arms too, 3 -> 5 dump sites each.

cranelift carried no source locations at all. Under the same switch it stamps
each op's trace index as a `SourceLoc` before lowering and reads the regions
back off the compiled buffer, printing `[clif] @0xNNNN op[i] OpCode` next to
dynasm's lines. Regions are recorded only when the dump is asked for; the
emitted bytes are unaffected.

### CI

The previous head of this branch was red on five checks. Graded against main's
own CI at the same time, three of them are main's:

| check | main control | this branch |
| --- | --- | --- |
| `cranelift raise_catch` (macos) | FAIL | FAIL |
| `the_answer_is_the_same_at_every_carrier_depth_cap` | FAILED | FAILED |
| `the_depth_decline_is_the_arm_this_program_reaches` | FAILED | FAILED |
| `dynasm synth/load_name_builtin_cell_fold` (ubuntu) | — | FAIL 1.7x |
| `cranelift synth/load_name_builtin_cell_fold` (ubuntu) | — | FAIL 2.3x |

The two branch-owned rows are on the fixture the eval-breaker change targets,
and they were measured before these commits existed. They are ubuntu-only, so
this run is what grades them.

`pyre/check.py` is green locally on all three backends (dynasm 531/531,
cranelift 531/531, wasm 524/524).

## Summary by CodeRabbit

- **New Features**
  - Added optimized handling for common `range()` iterations, including one-argument ranges and ranges with a step of one.
  - Range iterators now provide consistent type information across supported forms.

- **Bug Fixes**
  - Prevented garbage-collection signals from unnecessarily interrupting optimized execution.
  - Improved range construction and iteration when an explicit step is provided.

- **Diagnostics**
  - Enhanced optional JIT dump output with more precise machine-code and operation mappings.

### DIFF
--- a/majit/majit-backend-cranelift/src/compiler.rs
+++ b/majit/majit-backend-cranelift/src/compiler.rs
@@ -110,6 +110,13 @@ fn majit_verify_enabled() -> bool {
     *ENABLED
 }
 
+/// Whether `MAJIT_DUMP` is set, cached at first access.
+fn majit_dump_enabled() -> bool {
+    static ENABLED: std::sync::LazyLock<bool> =
+        std::sync::LazyLock::new(|| std::env::var_os("MAJIT_DUMP").is_some());
+    *ENABLED
+}
+
 use crate::asm_memory::{CraneliftArenaHandle, CraneliftArenaMemoryProvider};
 use crate::guard::{BridgeData, JitFrameDeadFrame, drop_bridge_payload};
 use majit_backend::deadframe::ExitDescr;
@@ -10579,7 +10586,17 @@ impl CraneliftBackend {
         }
         // else: linear trace — already in entry_block with vars defined
 
+        let dump_offsets = majit_dump_enabled();
         for op_idx in 0..ops.len() {
+            // Carry the trace's op index through codegen as a source location.
+            // Cranelift closes a srcloc region when the next one opens, so
+            // every byte the op emits — including the block plumbing a LABEL
+            // lowers to below — lands in that op's span, and the emitted bytes
+            // are unaffected either way. Only the dump reads the mapping back,
+            // so the regions are recorded only when it is asked for.
+            if dump_offsets {
+                builder.set_srcloc(cranelift_codegen::ir::SourceLoc::new(op_idx as u32));
+            }
             if let Some((_, label_block)) = label_blocks
                 .iter()
                 .find(|(label_idx, _)| *label_idx == op_idx)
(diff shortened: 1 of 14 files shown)


youknowone:
Both findings in section 2 were real and are fixed in `95f59d4`.

Checked against the upstream file rather than argued:

- `functional.py:766` / `:797` — `_immutable_fields_ = ['stop']` on both
  step-one shapes. `start` was wrongly marked immutable; it is now a plain
  field, and the doc comment that repeated the claim is corrected.
- `functional.py:444` — `class W_Range(W_Root)` declares no
  `_immutable_fields_` at all, so `promote_step` is now a plain field too.

Note this is the same deviation section 3 records for the four sibling
`W_Range` fields, which the baseline already marks immutable. I matched
upstream for the field this patch adds and left the pre-existing four alone
as out of scope.

Re-measured after the change: the optimized trace between LABEL and JUMP is
still 12 ops for `range(1, n, 1)` and 10 for `range(n)` / `range(1, n)`, with
no `GetfieldGc` in either body — `stop` is still hoisted, so dropping the two
extra immutability claims cost nothing. `range_shapes` parity with CPython and
the vendored `test_range` are unchanged, and `pyre/check.py` is green on all
three backends (dynasm 531/531, cranelift 531/531, wasm 524/524).

The section 4 notes match what I measured independently: the `__setstate__`
cursor semantics do follow CPython's `test_range.py:459` assertion rather than
PyPy's remaining-count reading, and PyPy and CPython genuinely disagree there.

— *commented by Claude*

youknowone:
@codspeedbot fix this regression

PR #1641, opened 2026-09-01
cranelift: guard-exit coldness and an opt-in loop-closing bridge merge; metainterp: share a guard's resume arrays

youknowone:
Eight commits on `main`, in two independent strands plus three regex-row
follow-ups.

## Cranelift: guard-exit layout, and a loop-closing bridge merge

`cranelift: mark every block only a guard exit can reach cold` — `set_cold_block`
was applied to the block a guard exit starts at and to nothing after it, and
cranelift's coldness is not transitive (`Function::is_effectively_cold` reads
the flag on the block itself plus an unconditional trap). The write-barrier arm,
the bridge dispatch, the closing-jump dispatch and the deadframe return were
laid out inside the trace body. Coldness is now propagated to a fixpoint after
`builder.finalize`, which is where `write_pending_failure_recoveries` puts the
recovery stubs upstream. On `pyre/bench/raise_catch_loop.py` the loop function
goes from 33 cold blocks of 249 to 92. Layout hint only; nothing about behaviour
changes.

`cranelift: merge a loop-closing bridge into the loop's own function` — a bridge
whose closing JUMP lands on a LABEL of the loop it attaches to is retained as a
`MergedRegion`, appended to that loop's stream, and the loop re-emitted with the
region as a block the failing guard branches into (`patch_jump_for_descr`'s
shape, where a bridge hit never runs the failure-recovery stub).

**It is fast, and it miscompiles.** Interleaved on one binary, min per round:
`raise_catch_loop` 1.71s merged / 2.80s not / 2.00s dynasm; a raise-free
`if i % 7` loop 1.56s / 2.47s / 1.74s. But `pyre/check.py --backend cranelift`
against that same binary reports 18 failures with the route on and 2 with it
off, so **16 checks separate the arms** — `dict_set.py` and
`nested_break_not_hot.py` SIGSEGV, `loops_comprehension.py` SIGBUS, each only
once a merge has fired. The frame-too-small hypothesis is refuted (the merged
loop allocates a *deeper* frame, 36 words against 32, and faults inside it);
`exception_escape_hot_callee_tb_node_once` reaches the collector with an invalid
type id, so the fault is a store through a wrong pointer.

So the third commit makes it **opt-in behind `MAJIT_CL_BRIDGE_MERGE`** (off by
default), read once into a backend field so
`bridge_closing_onto_its_owner_merges_into_the_loop` keeps driving the route
without an environment the rest of the process shares. The suspected cause is
recorded in the commit: `prepare_and_merge_regions` runs the GC rewrite per
region in its own namespace, and `merge_regions` replays only the window
`0..value_id_end(region)`, so a rewriter-minted constant at or above that width
is never replayed.

## Metainterp: sharing a guard's resume arrays

`ResumeStorage` holds `Arc<[u8]>` for `rd_numb` and a `SharedResumeSlice` for
`rd_virtuals` / `rd_pendingfields`, so `from_fail_descr` borrows the guard
descr's arrays instead of copying all three at every guard failure.
`SharedResumeSlice` keeps the nullable upstream shape — the empty case, which is
the common one, allocates nothing and reaches its readers as the same empty
slice an empty `Vec` gave them.

An earlier draft of that commit also pooled `ConstPtr` boxes per address in
`Trace`, reading `opencoder.py Trace._refs_dict` as one canonical box per
address. **That is not what `_refs_dict` is**: it interns the *encoded* ref
pool, and the decode side mints a box per read (`_untag`:
`ConstPtr(self.trace._refs[v])`), so two occurrences of one address stay
distinct under `is`. pyre models that with `Rc::ptr_eq`, and sharing the box
made previously-distinct constants compare equal, which moved CSE —
`synth/exception_bridge_traceback_head` to `loops_compiled 1 -> 2`, and
`synth/nested_loop_gate_switch` to `bridges_compiled 6 -> 7`, on all three
backends. Only the resume-array sharing, which carries no box identity, is kept.

## Regex row

`ir, regex: pin the Op size gate at the 240 bytes it reached`,
`macros: state the builder-argument order the JitCode register rewriter depends
on`, and `regex: re-measure the and/or row against a freshly translated
RPython`.

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in optimization that can improve execution of selected loop branches.
  * Added optional diagnostic logging for optimization decisions.

* **Performance**
  * Reduced memory allocation and copying during execution, tracing, and recovery.
  * Improved reuse of runtime frames and temporary buffers.

* **Bug Fixes**
  * Improved reliability when resuming execution and handling nested tracing scenarios.
  * Corrected diagnostic output for compact integer values.

* **Documentation**
  * Updated performance measurements, memory usage figures, and configuration documentation.

### DIFF
--- a/majit/examples/regex/rpython_original/README.md
+++ b/majit/examples/regex/rpython_original/README.md
@@ -362,7 +362,9 @@ C** (555,383 against 4,514,372). So "if you don't change the `and` and `or`
 ... it's not particularly fast" is not a majit artifact and not a Rust
 artifact — it is a property of the spelling, and upstream pays it in the same
 direction and the same order of magnitude. majit pays it 4.1x harder than
-upstream does, which is the one gap in this table that is majit's own.
+upstream does, which is the one gap in this table that is majit's own — read
+that figure with the re-measurement below, which puts it at the top of a
+3.0-4.1x range.
 
 `PYPYLOG=jit-summary:-` on the `and`/`or` build at 262,144 characters says
 where upstream's time goes: 1 loop, **1185 bridges**, `Tracing 0.145s` +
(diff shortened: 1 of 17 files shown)


youknowone:
`pyre/check.py` at the branch tip (darwin-arm64, LLBC re-extracted for all four crates):

    dynasm     2 failed, 529 passed
    cranelift  2 failed, 529 passed
    wasm       2 failed, 522 passed

Cranelift matches dynasm now that the merge is opt-in; the same tree with
`MAJIT_CL_BRIDGE_MERGE` set reports 18 failed / 513 passed.

The 2 are `synth/exception_bridge_traceback_head` and
`synth/nested_loop_gate_switch`, on all three backends, and both pass on this
branch with the two metainterp allocation commits removed — so they are this
branch's. They are jit-stats snapshot mismatches, not wrong answers, and are
left un-re-recorded on purpose (see the description).

Unit tests: `majit-ir` 351 passed, `majit-metainterp` 1760 passed,
`majit-backend-cranelift` 141 passed.

— *commented by Claude*

PR #1646, opened 2026-09-01
object, macros, interp, jit, translate: populate the JIT's two link-time registries on wasm32

youknowone:
`linkme::distributed_slice` rejects `wasm32` — a custom `link_section` static
there must be a plain list of bytes, with no relocation for a pointer field —
so `PYRE_TYPE_OBJECT_FNADDRS` and `PYRE_CLASS_DESCRIPTORS` were both empty in
the guest, and two consumers compensated by dropping work for that target:

- `front::llbc_hints` skipped the `dont_look_inside` stamp on every
  `type_object()` accessor when `TARGET` began with `wasm32`, so the
  codewriter emitted a walked accessor rather than a residual call.
- `pyre-jit-trace`'s prepass dropped every `#[pyre_class]`-derived static
  `PyType` address row for wasm, leaving those statics unbound.

Both registries now carry a wasm32 arm that registers from a `ctor`
constructor. The constructor is emitted beside the item, so it inherits that
module's `cfg` gates the way the link-section element does; a central list
could not, because the accessors sit behind nested gates and one that fell
behind would silently drop an address rather than fail to build.
`for_each_type_object_fnaddr` and `for_each_class_descriptor` read whichever
population the target holds, and the two compensating drops are removed.
`build_gc`'s GC-root completeness oracle, gated off wasm32 because the
registry did not exist there, runs on every target.

### Measured on `pyre/bench/fib_recursive.py`

| | wasm before | wasm after |
|---|---|---|
| generated jitcode census, `type_object` entries | 26 (walked) | **0** — the wasm and native censuses hold the same 3366 entries, none unique to either side |
| `[PREPASS statics] pytype rows` | 136 (0 of 136 registry rows kept) | **225 (89 of 136)** — the native reading |

### What this does not fix

Neither change moves the `fib_recursive` wasm trace-too-long abort:
`abort_diag too_long=1` and `guard_failures=1601` are unchanged. A cranelift
binary built with the wasm feature set (`--no-default-features --features
cranelift`) records the same over-limit trace and reports
`toolong_suppressed=1`, so that abort turns on the feature set rather than on
the target. `MAJIT_LOG=1` reads the two native builds directly: the full build
traces 3648 ops, the feature-matched one 6008 — eight over the 6000 limit.
Why dropping `host_env` lengthens that trace is open and not addressed here.

## Addressing the review

Codex found that binding every host registry row bakes host-process pointers
into a cross-target build. It does, and the hole is wider than the rows it
named: the build script is a build-*dependency* with its own feature
resolution, so a row behind a feature the binary turns off (`mmap_type`,
`_ctypes::cdata_bytes_object` under `--no-default-features --features
cranelift`) has been bound-but-unmatched on **native** builds since it landed.

`disarm_unpaired_build_addrs` closes both axes at the place that already owns
the build → runtime pairing: every build address no runtime name re-pairs is
set to zero in `JitCode.fnaddr`, `constants_i` and `constants_r` before the
table reaches a consumer. A call through zero faults on the first instruction,
a dereference faults at the first field, and a pointer comparison never
matches, so a path that reaches one says so instead of running against
build-process memory.

## The other commits

- `stdlib_warnings.py`'s `module_globals` blocks asserted a single recorded
  warning while naming the loader as `__loader__` alone. `warn_explicit` reads
  the loader through `__spec__.loader` and warns before the source lookup when
  the mapping carries no `__spec__`, so the assertion failed under CPython
  3.14.6 as well as under pyre.
- `_codecs.escape_encode(b"'")` answers `(b"\'", 1)` on CPython, pypy3 and
  pyre; the fixture asserted `(b"'", 1)`.
- `open()` re-spelled the PathLike protocol and called `__fspath__` unbound, so
  an object whose `__fspath__` is a property raised `TypeError: 'property'
  object is not callable`. It now goes through `posix::fspath`, as
  `pypy/module/_io/interp_io.py` `_open` does.

Those three snippets join the CI gate.

## Verification

- `pyre/check.py --backend dynasm,cranelift,wasm`: dynasm 533/533,
  cranelift 533/533, wasm 526/526, no jitstats baseline moved.
- `extra_tests/run.py --gated-only`: 174/174 on cpython, dynasm and cranelift.
- `extra_tests/parity_tests/run.py`: all pass.
- `cargo test --all` (207 test binaries, 0 failures), `cargo fmt --check`.

## The two red check.py legs

Both are the derived pypy floor: `synth/load_name_builtin_cell_fold` on macOS
cranelift (`0.4x < gate 0.416667x`, ceiling 2.5x) and
`synth/loop_callee_shared_mutation` on ubuntu dynasm (`0.8x < gate 1x`, ceiling
6.4x). The same fixture on the same backend reads **0.4x on macOS** and **1.7x
on ubuntu** in this very run, so the ceiling is calibrated for the host where
it reads 1.7x and lowering it to clear the macOS floor would fail that host
against the ceiling. main at this branch's merge base
([run 33565610870](https://github.com/youknowone/pyre/actions/runs/33565610870))
fails the same macOS job with three of the same class, and additionally fails
ubuntu dynasm on `test.test_threading: PASS -> TIMEOUT`, which this branch does
not reproduce (`CRASH 0, TIMEOUT 0`). Details in the PR comment.

https://claude.ai/code/session_017wapwfqfqNe7kFcxQRx85P

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -811,6 +811,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ctor"
+version = "1.0.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "914a755b7c2d4af2bdcff7ce1739e2db9a1b81a9b07123d8015786ae03c0980d"
+dependencies = [
+ "link-section",
+ "linktime-proc-macro",
+]
+
+[[package]]
 name = "ctr"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1806,6 +1816,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "link-section"
+version = "0.19.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "39c29a617ce3df32c08497bdc1ab6e2376e0b17948ac166a2fbe5977c5954cd9"
+
+[[package]]
 name = "linkme"
 version = "0.3.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1824,6 +1840,12 @@ dependencies = [
  "quote",
  "syn",
 ]
+
+[[package]]
+name = "linktime-proc-macro"
+version = "0.2.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7e57c38c1e860fd37c604281cdfb1dd2216977fd76a50f85ba2f388ef3219616"
 
 [[package]]
 name = "linux-raw-sys"
@@ -2696,6 +2718,7 @@ version = "0.0.2"
 dependencies = [
  "bitflags 2.13.0",
  "cc",
+ "ctor",
  "getrandom 0.4.3",
  "indexmap",
  "libc",
@@ -2840,6 +2863,7 @@ name = "pyre-object"
 version = "0.0.2"
 dependencies = [
  "bitflags 2.13.0",
+ "ctor",
  "indexmap",
  "linkme",
  "majit-gc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,6 +138,7 @@ smallvec = "1"
 # link-section distributed slice: `#[pyre_class]` appends each type's GC
 # descriptor to a whole-program slice the JIT driver checks for completeness.
 linkme = "0.3"
+ctor = "1"
 
 # cranelift backend
 cranelift-codegen = "0.134"
--- a/majit/majit-translate/src/front/llbc_hints.rs
+++ b/majit/majit-translate/src/front/llbc_hints.rs
@@ -47,13 +47,6 @@ const CONST_PREFIX_HINTS: &[(&str, &[&str])] = &[
 /// the marker consts present in `llbcs`.
 pub fn harvest_hints_from_llbcs(llbcs: &[Llbc]) -> HashMap<String, Vec<String>> {
     let mut out: HashMap<String, Vec<String>> = HashMap::new();
-    // The `type_object` residual resolves through `PYRE_TYPE_OBJECT_FNADDRS`, a
-    // native-only link-time slice (empty on wasm32, which linkme rejects).
-    // Residualizing there would emit a call to an unpublished address, so keep
-    // those accessors on the legacy walker when building for wasm.
-    let residualize_type_object = !std::env::var("TARGET")
-        .unwrap_or_default()
-        .starts_with("wasm32");
     for llbc in llbcs {
         let function_paths: HashSet<String> = llbc
             .iter_local_fns()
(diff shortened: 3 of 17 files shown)


youknowone:
Confirmed, and the finding is wider than the wasm rows it names.

The two pools come from two compilations of one list. The build script is
compiled for the host, so a module behind `target_arch`, `unix` or `windows` is
in the build pool and not in a cross-target runtime pool — counted statically,
that is `_ssl`'s five `#[pyre_class]` types and five `type_object` accessors
plus `mmap`'s one class for a wasm32 target. It is also a build-dependency with
its own feature resolution (`pyre-interpreter = { features = ["dynasm"] }`,
which carries the crate's default `host_env`), so a row behind a feature the
binary turns off is bound at build time and absent at run time on a **native**
build too: `--no-default-features --features cranelift` is exactly that, and
`mmap_type` and `_ctypes::cdata_bytes_object` have carried the same hole since
they landed.

So restoring the wasm-only filter would close one instance of it and leave the
rest. The fix is at the place that already owns the build → runtime pairing:
`runtime_fnaddr_patch` now takes out every build address no runtime name
re-pairs, before any consumer reads the table.

- `JitCode.fnaddr` and `constants_i` are set to `0`. Zero is the one value that
  cannot be mistaken for either use — a call through it faults on the first
  instruction, and a `PyType` pointer comparison against it never matches — so
  a path that does reach one says so instead of running against host memory.
- `constants_r` is refused rather than zeroed: those slots are roots the
  collector reads out of the pool, so an entry it cannot re-pair is already a
  bad root and a zero there would be a different wrong answer rather than a
  fault. The load panics and names the bindings.

An address the correspondences do know is excluded from the set, because the
build binary's identical-code folding can put a re-pairable name and an
unmatched one on one address.

— *commented by Claude*

youknowone:
### The two red `check.py` legs are the derived pypy floor, and re-recording is refuted

Both failures are one class — pyre came in *under* the floor derived from a
`max-pypy-ratio` ceiling, and the gate asks for the ceiling to be re-recorded:

| leg | fixture | reading |
|---|---|---|
| macOS cranelift | `synth/load_name_builtin_cell_fold` | `0.20s vs pypy 0.53s` → `0.4x < gate 0.416667x`, ceiling 2.5x |
| ubuntu dynasm | `synth/loop_callee_shared_mutation` | `0.07s vs pypy 0.09s` → `0.8x < gate 1x`, ceiling 6.4x |

Re-recording either ceiling is the wrong move, and this run measures why. The
**same fixture on the same backend** reads:

- macOS cranelift: pyre 0.20s vs pypy 0.53s — **0.4x**
- ubuntu cranelift: pyre 0.52s vs pypy 0.26s — **1.7x**

The ratio moves 4x between hosts because the pypy side does. 2.5x is a
well-calibrated ceiling for the host where the fixture reads 1.7x; lowering it
to clear the macOS floor would fail that host against the ceiling instead. The
floor is `ceiling / 6` (#1598), and a single scalar cannot carry both ends
across hosts whose pypy differs this much — which is the same problem #1615
solved for the parity fixtures by keeping their ceilings *without* derived
floors. Extending that to the synth set is a check.py policy change that
belongs in its own PR, not here.

### This branch against main's own run

main at this branch's merge base (`707407841ae`,
[run 33565610870](https://github.com/youknowone/pyre/actions/runs/33565610870))
fails the same macOS job with three of exactly this class —
`attr_store_add_transition` 0.7x < 0.833333x,
`exception_reraise_tb_depth_hot` 0.9x < 0.916667x, `fast_local_swap` 0.5x <
0.7x — and additionally fails ubuntu dynasm on a CPython-suite regression
(`test.test_threading: PASS -> TIMEOUT`). This branch's ubuntu run reads
`CRASH 0, TIMEOUT 0`, so that one does not reproduce here.

Locally, `pyre/check.py --backend dynasm,cranelift,wasm` is 533/533, 533/533,
526/526 with no jitstats baseline moved.

— *commented by Claude*

youknowone:
### The stale comment is real; the crash-target premise is not

Two halves, and only one of them holds.

**Conceded.** `prepass.rs` and `jit_fnaddr.rs` both named a `reject_unpaired_build_addrs` that does not exist and described it as refusing the load, which no version of this does. That wording is what made the finding readable as "installs a pointer nothing checks". Both comments are corrected in a follow-up, along with the function's own doc, which had justified zero as a value that faults on use rather than as the value it actually is.

**Refuted.** Zero is not a poison chosen here. It is the address `jitcode.py` `JitCode.__init__` already spells as `fnaddr=None`, and both consumers test for it *before* they branch:

- `is_callable_fnaddr` is `fnaddr != 0 && !is_symbolic_fnaddr(fnaddr)`. Its own comment says so: stricter than `!is_symbolic_fnaddr`, because `0` is upstream's "no address" and must not be called either.
- Every blackhole `residual_call_*` handler guards on it and returns `reject_unresolved_call`; `reject_unresolved_inline_call` sets `aborted` and returns `LeaveFrame`, which hands the continuation back to the interpreter.
- The walker's residual path tests `func_ptr == 0 || is_symbolic_fnaddr(func_ptr)` at each of its three call sites and returns `ResidualDecline::Symbolic`.

So a residual or inline call reaching one of these declines rather than dispatching. A `PyType` constant is only ever compared — a comparison against zero matches no object, so the guard reading it fails and the path leaves the JIT before anything dereferences it.

The alternative the review proposes — reject or defer the affected JitCode — is strictly worse here. The unpaired names are definitions that do not exist on the target at all (`_ssl` on wasm32, the feature-gated `mmap_type` and `_ctypes::cdata_bytes_object`), so every trace that mentions one is a trace whose guard can never pass. Refusing the body would drop the compilation of code that is reachable and correct; zeroing lets the unreachable arm decline through the mechanism that already exists for exactly this value.

— *commented by Claude*

PR #1658, opened 2026-09-02
jit-trace, majit: a struct descr's tid and size from one cache entry, the symbolic finish payload rooted, an overwritten vable register, and a callee-only resume image

youknowone:
Four independent correctness gaps, each found by reading and each carrying its own
unit test. Split out of `perf-memory`, which also holds an unrelated
cranelift-codegen pin that is not ready.

### 1. A struct descr's tid and payload size come from one cache entry

`descr.py get_size_descr` hands allocation and header emission **one** descriptor
object per STRUCT, so upstream cannot get a tid from one struct and a size from
another. Pyre's frozen `BhDescr::Size` wire value carries a structural key and a
size separately, and `resolve_gc_tid` resolved the key against
`GcCache._cache_size` without ever asking whether the entry it landed on
described a block of this size. A truncated or colliding key stamped the foreign
entry's tid on the allocation, and the collector then walked that type's GC
offsets past the end of the block.

`GcCache::resolve_struct_layout` now returns `(tid, size)` from the same
`SizeDescr` under one lock — the registry is process-global, so two separate
lookups would let a concurrent publication turn a checked hit into an unchecked
fallback. On disagreement `resolved_gc_tid_checked` returns `None` (the caller
declines) and the unchecked `resolve_gc_tid` panics rather than emitting a
header that mis-describes the block.

The unchecked variant fails loudly on purpose, and that is the part worth
reviewing: `register_keyed_size` already documents that two producers can
publish under one key and that genuinely different layouts mean "the key itself
is the defect" (the struct-identity spelling collision `identity_collision_census`
counts). If that collision is live at runtime this panics where it previously
mis-stamped a header — a loud failure replacing a silent one, but a new failure
either way. The gate is where that shows up.

### 2. The symbolic FINISH payload is a GC root

`FBW_FINISH_PAYLOAD` holds an `OpRef::ConstPtr` from the walk's terminate arm
(`fbw_terminate_with_finish`, `fbw_terminate_void_with_finish`,
`fbw_terminate_with_raise`) until `full_body_walk_trace` takes it. Upstream's
`history.py ConstPtr` box sits in the translated GC graph for that whole window;
pyre's copy sits in a TLS `Cell` that nothing walked, so a collection inside the
window left a stale ref behind. Its concrete sibling `FBW_FINISH_CONCRETE` is
already registered at `register_thread_root_areas`; the symbolic half now is
too, and the walker writes the forwarded ref back into the cell.

### 3. `vable_reg` is an identity, not ownership of a register number

In `fbw_callee_body_replay_scan`, `callee_owned_frame` exempts a `setfield_gc`
whose target register is the virtualizable from the freshness/immutability
poison. `vable_reg` was latched at the first `*_vable_r*` and never cleared,
while `fresh_ref_regs` beside it is recomputed on every ref-producing
instruction. Latch register R, overwrite R with a non-fresh ref (a
`getfield_gc` result, a residual call result), then `setfield_gc` R on a mutable
field, and the store escaped `SetfieldGcTargetNotFreshOrMutable` — a store to an
arbitrary live-heap object inside a body the walk may replay from the caller's
CALL.

RPython keeps the virtualizable on the live `MIFrame` box, so the identity dies
with the register's old value. Clearing `vable_reg` at the site that already
recomputes `fresh_ref_regs[dst]` is strictly conservative: it can only add
poison, never remove it.

### 4. A callee-only inline image is not a complete resume chain

`resume.py rebuild_from_resumedata` rebuilds one frame per encoded jitcode
header and `pyjitpl.py capture_resumedata` starts from the portal MIFrame, so a
lone callee with no captured parent is not a resume chain. The
`n_parents == 0 && n_callees == 1` arm of `publish_inline_frames` published one
anyway, which makes `frames[0]` disagree between the two resume decoders. That
shape now falls through to the existing `GuardResumeCoordinateUnavailable`
sentinel like every other incomplete chain.

### Tests

- `checked_gc_tid_rejects_a_foreign_sized_struct_cache_entry`
- `finish_payload_root_walker_writes_back_forwarded_const_ptr`
- `replay_scan_does_not_transfer_frame_identity_across_register_overwrite`
- `parentless_populated_callee_does_not_publish_a_lone_resume_frame` — deliberately
  installs a real `CodeObject` and a populated `-live-`-anchored JitCode so it
  crosses the old `code_ptr.is_null()` fixture bail

### What ran locally, and what did not

- `cargo fmt --check` over the four touched crates — clean.
- `cargo test -p majit-translate --lib codewriter::jitcode` — 12 passed, including
  the new descr test.
- `pyre-jit-trace` / `pyre-jit` **were not compiled in this session**: the box hit
  load 104 with six sibling builds and the job was killed. The non-test code in
  both did compile earlier today, as the Charon extraction over `pyre-jit` on this
  exact tree completed; the three new `#[test]` functions have not been built.
- `pyre/check.py` did not run: the branch was rebased onto main after that
  extraction, so all four `.ullbc` artefacts are stale and re-extracting on this
  box was not viable.

CI is the gate. Items 3 and 4 are tightenings — they can only add declines — so a
moved counter or a SNAPDIFF is the expected shape of a red, and the baseline gets
re-recorded rather than the fix relaxed.

https://claude.ai/code/session_01F69vGeNqTgKVSNqY8WiR4X

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved runtime type resolution to reject mismatched cached layouts, preventing incorrect type assignments.
  - Fixed garbage collection handling for inline finish payloads so references continue to be updated safely when objects move.
  - Prevented stale register information from causing later stores to be misclassified.
  - Strengthened snapshot validation to prevent incomplete frame chains from being published.
- **Tests**
  - Added regression coverage for pointer forwarding, snapshot validation, cache mismatches, and replay behavior.

### DIFF
--- a/majit/majit-ir/src/descr.rs
+++ b/majit/majit-ir/src/descr.rs
@@ -2403,10 +2403,23 @@ impl GcCache {
     /// keys a `_cache_size` slot, so a miss reports the value was already a
     /// tid and the caller keeps it verbatim.
     pub fn resolve_struct_tid(&self, cache_key: u64) -> Option<u32> {
+        self.resolve_struct_layout(cache_key).map(|(tid, _)| tid)
+    }
+
+    /// Resolve both halves that make a typed struct allocation sound: the
+    /// collector `tid` and the payload size owned by the same cached
+    /// `SizeDescr`.  RPython's `GcCache._cache_size[STRUCT]` hands allocation
+    /// and header emission one descriptor object, so these values cannot come
+    /// from different structs.  Pyre's frozen `BhDescr` wire value carries a
+    /// structural key and a size separately; its runtime resolver
+    /// uses this pair to reject a key that lands on a differently-sized cache
+    /// entry instead of stamping the foreign entry's tid on the requested
+    /// block.
+    pub fn resolve_struct_layout(&self, cache_key: u64) -> Option<(u32, usize)> {
         self._cache_size
             .get(&LLType::Struct(cache_key))
             .and_then(|d| d.as_size_descr())
-            .map(|sd| sd.type_id())
+            .map(|sd| (sd.type_id(), sd.size()))
     }
 
     /// Array counterpart of [`resolve_struct_tid`] keyed on
--- a/majit/majit-translate/src/codewriter/jitcode.rs
+++ b/majit/majit-translate/src/codewriter/jitcode.rs
@@ -2118,7 +2118,10 @@ impl BhDescr {
     /// `:348-378` structural identity).  A `_cache_size`/`_cache_array` hit
     /// maps a `cache_key` back to the allocated tid (`gc.py:536-542`); a
     /// miss means the value was already the dense tid (a real tid never keys
-    /// a struct/array cache slot).  Backends call this instead of
+    /// a struct/array cache slot). A cache hit whose payload size disagrees
+    /// with this descr fails loudly: using that tid would make the collector
+    /// trace the cached struct's fields past this allocation. Backends call
+    /// this instead of
     /// `get_type_id() as u32` so a materialized object carries a header the
     /// collector can trace.
     ///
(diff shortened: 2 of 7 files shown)


youknowone:
Force-pushed onto current `main` (`e75174702af`) with the CI fix as a second commit.

**The one real red.** `cargo test` failed on all three OSes with a single error —
`E0063: missing fields call_id, debug_merge_point_py_pc and recursion_greenkey`
at `tests.rs:649`. The rebase applied with no textual conflict but `InlineFrame`
had gained three fields underneath it. The literal now supplies what
`InlineFrameGuard::new` stamps on the first Python portal frame pushed onto an
empty framestack: `call_id: 0` (`MetaInterp.newframe` consumes it), no
`DEBUG_MERGE_POINT` yet, and the portal greenkey a forward inline call carries.

Also folded in CodeRabbit's other finding: `walker_inline_guard_resumes_in_callee`
spelled the parent-chain completeness predicate a second time. Both spellings
compute `n_parents` and `framestack.len()` identically, so it now calls
`inline_snapshot_has_complete_parent_chain` — which is why that helper was
extracted.

**The two `check.py` reds are inherited.** main's own job for the same job name
at `55e6fb87965` prints them:

| job | this PR | main's own job |
|---|---|---|
| `check.py dynasm (ubuntu)` | `test.test_threading: PASS -> TIMEOUT` (`test_3_join_in_forked_from_thread`) | identical |
| `check.py cranelift (ubuntu)` | `for_iter_conditional_store_bridge exec 0.27s > pypy 0.06s  ratio 4.5x > gate 4x` | identical numbers, plus four more (main 5 failed, this PR 1) |

`Codex parity review` produced no report on the previous head — it hit its usage
limit (`try again at Sep 7th`), not a finding.

Still not verified locally: the `pyre-jit-trace` test build was killed three
times on a box sitting at load 47-134 with ~67 MB free. CI is the check.

— *commented by Claude*

PR #1681, opened 2026-09-03
majit-translate, jit-trace, _cffi_backend: take the root bracket out of the jitcode, and the cffi call path it unblocks

youknowone:
Fifteen commits on top of `de7e1a70159`. Two threads: taking the GC root bracket
out of the jitcode, and the `_cffi_backend` call path that unblocks. One
unrelated rooting fix rides along because it was what turned the gate's wasm leg
red.

## Root brackets out of the jitcode

Pyre spells its GC root brackets in interpreter source — `push_roots()`,
`pin_root`, `base()`, `get()`, `root_scope_close`. Upstream RPython never puts
one in a jitcode: `warmspot.py make_jitcodes` runs before the shadow-stack
gctransform, so the bodies the codewriter reads have no bracket in them at all.
All three jitcode consumers already carry their own root set — the compiled
trace's jitframe gcmap, the blackhole's `push_bh_regs`, and the recording walk's
own op cells — which is what makes an erasure at the jitcode level sound while
the native rustc build keeps executing the source bracket unchanged.

`majit-translate` now erases the bracket for the shape whose read-backs can be
answered without the shadow stack: one `pin_root`, every `get` indexed by that
scope's own `base()`, and the pin dominating every read. A bracket that pins in
a loop, or reads a slot the pass cannot name, keeps every op it has.

The pass fired zero times when first written. MIR copies a call argument into a
fresh temporary (`_41 = copy _31`), so an index that is semantically `base()` is
spelled as a copy of it; the pass now follows those copies when nothing else
writes the temporary. Over the two pyre artefacts that is 1196 brackets erased
and 24 kept.

`test_root_bracket_close.rs` states the close invariants over the brackets the
erasure keeps, since an erased bracket publishes nothing and so owes no close —
a test counting closes needs to tell that case from a close the lowering
dropped.

## The `_cffi_backend` call path

- `W_CType`'s declared immutable fields widened, and the argument ctype
  `do_call` reads out of `fargs` promoted, so the per-call `W_CType` dispatch
  folds against a constant.
- `PyObject.ob_type` declared immutable, and a typeptr read answered from
  `get_known_class` rather than only from `PtrInfo::Virtual`. `is_typeptr`
  matches on the field's *name*, and `CPyObject.ob_type` carries that name at
  offset 24 while every consumer here reads the header word at offset 0, so the
  fold is gated on the offset.
- The descent scan's effect rule now reads off the descr, the arm table, and the
  body's own allocations rather than treating every residual call and every
  `setfield_gc` as an effect.
- A residual's ABI words are no longer dereferenced as objects in the locals
  write-back — `get_and_call_function(args_w: &[PyObjectRef])` is a fat pointer
  and the write-back was reading half of it as an object.

## The fix that rides along

`f3ef7a9cae0` is not part of either thread. `exec_or_eval` published the globals
object on the shadow stack before `ensure_exec_builtins` / `ensure_eval_builtins`
and the locals argument only after them. Those two dispatch `setdefault` /
`__contains__` / `__setitem__` on the caller's own mapping, so a dict subclass's
override runs user Python there; a nursery move during it relocated the locals
mapping while the argument word the function still held did not follow.

```python
class NS(dict):
    def setdefault(self, key, value=None):
        gc.collect()
        return dict.setdefault(self, key, value)

exec("r = src.get('x')", NS(), {"src": {"x": 3}})
```

answers `TypeError: '' object is not subscriptable` on the first round. With
more allocation inside the override the same round faults: SIGSEGV on dynasm,
SIGBUS on cranelift, and a `minor_custom_trace_target` GC panic naming a nursery
address in a remembered-set holder. It reproduces identically on `c46ae2e2b35`
with none of this branch applied, so it is not new here; it is the eleventh path
of #1669's ten.

The gate's `synth/exec_namespace_code_object_rooting_regression` reports it on
wasm only, because that host's nursery is small enough that the plain
`gc.collect()` in the fixture already moves the mapping. `pyre CI` was cancelled
on the commit that added that fixture and on its four predecessors, so its wasm
leg has never been green.

## Verification

`cargo test --all --no-default-features --features dynasm` and `pyre/check.py`
were green on the pre-rebase tree (7040 passed / 0 failed; dynasm 541/541,
cranelift 541/541, wasm 1 failed — the fixture above). Since the rebase onto
`de7e1a70159` and the `exec_or_eval` fix, the repro ladder and that fixture pass
on dynasm, but the full chain has **not** yet been re-run to completion; it is
running now and the result will be posted here.

https://claude.ai/code/session_01VKvxTiG1M3K7KuKxVVezaX

## Summary by CodeRabbit

- **Performance**
  - Improved JIT optimization of object type information and immutable fields.
  - Reduced overhead from temporary garbage-collection root scopes where safe.
  - Improved inline-call analysis, cleanup, and handling of generators and nested calls.

- **Reliability**
  - Strengthened garbage collection tracing for immortal objects and cross-module root scopes.
  - Preserved root-scope cleanup during exceptional and result-handling paths.
  - Improved safeguards for repeated failed JIT descents.

- **Documentation**
  - Updated diagnostic gate documentation and configuration guidance for root-scope and descent behavior.


youknowone:
## Gate chain, on the rebased tree with all three fixes

```
cargo test --all --no-default-features --features dynasm   9054 passed, 0 failed
python3 pyre/check.py                                      ALL PASSED: 3/3 backend runs
  dynasm    541/541
  cranelift 541/541
  wasm      534/534
```

The wasm leg is green. `synth/exec_namespace_code_object_rooting_regression`
was the one failure before `f3ef7a9cae0`, and it is the case that fixture's
`exec-separate-locals` check exists for.

Two commits were added since the PR was opened:

- `706a6da57a0` — `optimize_setfield_gc` seeded `known_class` from any
  name-matched typeptr store, and `optimize_getfield_gc`'s second arm folded an
  unstored name-matched read to the virtual's `SizeDescr` vtable, both without
  asking the offset. `is_typeptr` matches on the field's *name*, which
  `CPyObject.ob_type` also carries at offset 24, so a store or read through the
  mirror's field answered for the header word. The read arm beside them already
  asked; these two now do too.

- `80d9b5af9ec` — `do_call` published the function type and the arguments in
  one bracket and numbered the argument slots `fargs_slot + 1`. Arithmetic on a
  bracket's own `base()` result retires it from the jitcode's root-bracket
  erasure, so neither half qualified. Split in two, the function type is one pin
  whose three read-backs all name its own bracket's first slot.

## Where the cffi bench stands

Measured interleaved against `pypy3` on a host under load average 46-71, so the
ratio is the reading and the absolute numbers are not:

| | ns/call | vs pypy3 |
|---|---|---|
| `pypy3` | 7.4 | 1.0x |
| pyre, `PYRE_FBW_DESCENT_SCAN_OFF=1` | 21.9 | 2.96x |
| pyre, production default | 227.9 | 30.8x |

Load-independent, over the same steady loop body rendered from
`MAJIT_LOG_OPT`:

| | steady ops | residual calls |
|---|---|---|
| before the bracket work | 210 | 13 |
| after the erasure landed | 149 | 8 |
| after the `do_call` split | **146** | **7** |

The remaining bracket is now the single largest item in that body:
`capture_shadow_stack_area`, `shadow_stack_cell_len`, `pin_root` and
`root_scope_close` are 4 of the 7 residual calls, plus the `RootScope`
allocation and its two stores, plus the `W_IntObject` box that `pin_root`'s
escape stops the optimizer from keeping virtual.

Erasing it needs the pass to answer a `get` indexed `base() + i`. The
`RootScope::set` hazard does not stand in the way — the pass recognises only
`pin_root`, `base` and `get` on a guard and retires the bracket on anything
else, so a body holding a `set` is already excluded. What does stand in the way
is that the pins run in one loop over `args_w` and the reads in another over
`0..args_w.len()`, and answering the read means proving those two agree. Moving
the pin into the reading loop is not an option: `args_w` is a native copy no
collector rewrites, so argument *j* would go stale while argument *i* converts.

The production-default column is a separate thread — the descent scan's
static answer to a dynamic question — and is not addressed here.

— *commented by Claude*

PR #1717, opened 2026-09-06
wasm: batch nursery allocations and fold raw-load addresses

youknowone:
## Summary

- Batch consecutive inline-eligible `New`/`NewWithVtable` into one nursery bump, matching rewrite `gen_malloc_nursery`. Non-collecting stores keep the run; `LABEL` and collecting ops flush it.
- Let constant-length `NewArray`/`NewArrayClear` join that same bump. Runtime-length arrays stay on the varsize path and flush the pending batch.
- Overflow still allocates each follower through the helper so its frame home is written at its own op. The write barrier is not seeded after the slow/fast join.
- Fold nonnegative constant raw load/store offsets into `MemArg` displacement. PeepSink now folds `i32.const` through `i64.extend_i32_u` and cancels `wrap(extend_u)`.

## Self-review

## Verification

- `cargo test -p majit-backend-wasm --test codegen_test`: 112 passed, 6 ignored runtime integration tests.
- Did not rerun `cargo test --all --no-default-features --features dynasm` or `python3 pyre/check.py` for this follow-up.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved memory allocation efficiency by batching eligible consecutive nursery allocations.
  * Preserved correct handling across collecting operations and runtime-sized array allocations.
  * Improved constant-offset memory access generation.

* **Bug Fixes**
  * Corrected integer extension and wrapping optimizations.
  * Ensured raw loads and stores use their specified memory offsets.

### DIFF
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -3249,6 +3249,100 @@ pub struct NurseryAllocParams {
     pub plain_tids: std::collections::HashSet<u32>,
 }
 
+/// `GcRewriterAssembler.round_up_for_allocation`: raise to the nursery
+/// minimum, then 8-align. Shared by the inline `New*` bump and by
+/// `collect_nursery_batches` so a batched total uses the same size the
+/// per-object arm would have asked for.
+fn aligned_nursery_size(payload: i64) -> Option<usize> {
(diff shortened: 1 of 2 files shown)


youknowone:
#1717 review follow-up is in a new PR: CallMallocNursery lowering, region-boundary batch flush, and the large_threshold restart test.

— *commented by Claude*

PR #1725, opened 2026-09-07
Enable cranelift owner-loop bridge merge by default

youknowone:
## Summary

Turn on cranelift's merge of a loop-closing bridge into its owner function, and land the resume/observe pieces that the exception hot path still needed.

- Re-emit the rewritten owner stream and share its `GcTable`, so a later merge does not resurrect nursery `ConstPtr`s. Snapshot the owner constant pool before codegen `take()` empties it. `MAJIT_CL_NO_BRIDGE_MERGE` restores the out-of-line path.
- Carry failarg `IntBound`s in an optional trailing resume section so a bridge can drop `IntAddOvf`.
- Pin `CHECK_EXC_MATCH` stub-class matches on the exception kind field instead of declining the fold.
- Skip the zero `iadd` a raw load's address adds (already on this branch).
- Record `fannkuch` cranelift jitstats for the merged path (`bridges_compiled` 24 → 93, `guard_failures` 5450 → 19549). Output still matches PyPy.

## Tests

- `cargo test --all --no-default-features --features dynasm`
- `python3 pyre/check.py --backend cranelift` (539 synth + 10 main benches)
- `merged_owner_preserves_reference_constant_after_collection` is ignored under the default parallel harness (sibling MiniMarks share `PENDING_MINOR_TABLES`). Serial: `cargo test -p majit-backend-cranelift merged_owner_preserves -- --ignored --test-threads=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exception matching for generic base exceptions, reducing failures during JIT-compiled execution.
  * Preserved reference tracking during concurrent test-driven garbage collection.
  * Improved recovery of guard state by retaining proven integer bounds.
  * Fixed fail-descriptor lookup after guard movement and optimized zero-offset fixed-address loads.

* **Changes**
  * Bridge regions are now compiled and patched independently; bridge merging and loop re-emission support were removed.
  * Removed obsolete bridge-merge environment-gate entries.

### DIFF
--- a/pyre/bench/fannkuch.cranelift.jitstats
+++ b/pyre/bench/fannkuch.cranelift.jitstats
@@ -1,4 +1,4 @@
-bridges_compiled=93
+bridges_compiled=24
 descr_set_absent=0
 descr_set_ambiguous=0
 descr_set_stale_absent=0
@@ -12,7 +12,7 @@ fbw_rolled_back_with_effects=0
 fbw_store_journal_rollback_failed=0
 field_pos_attached_misplaced=0
 field_pos_spec_misplaced=0
-guard_failures=19549
+guard_failures=5450
 internal_compile_panics=0
 loops_aborted=0
 loops_compiled=6
--- a/pyre/bench/fannkuch.cranelift.jitstats
+++ b/pyre/bench/fannkuch.cranelift.jitstats
@@ -1,4 +1,4 @@
-bridges_compiled=24
+bridges_compiled=93
 descr_set_absent=0
 descr_set_ambiguous=0
 descr_set_stale_absent=0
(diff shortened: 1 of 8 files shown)


youknowone:
fannkuch jitstats should not have been refit. Default-on merge raised `bridges_compiled` 24 → 93 and `guard_failures` 5450 → 19549 with the same printed checksum — extra compiled-guard deopts, not a new target. Local `5e5122600f5` restores the old jitstats file and puts merge back behind `MAJIT_CL_BRIDGE_MERGE`. Not force-pushed: this branch and `origin/perf-exc` have diverged after a rebase onto a newer main.

— *commented by Claude*

pyre/bench/fannkuch.cranelift.jitstats
```diff
@@ -12,7 +12,7 @@ fbw_rolled_back_with_effects=0
 fbw_store_journal_rollback_failed=0
 field_pos_attached_misplaced=0
 field_pos_spec_misplaced=0
-guard_failures=5450
+guard_failures=19549
```
youknowone: regresion

PR #1778, opened 2026-09-11
wasm: skip leftover import holes on LABEL and failargs

youknowone:
## Summary

`unbound_pool_const_seeds` treated LABEL args and jit-debug operands as body reads. After uses were rewritten, a leftover `InputArg` on those sites made `compile_loop` decline the module.

Rebased onto `origin/main` (`9717a2505d5`, #1780). #1780 is the root cause: wasm32 stack-residency made `import_state` leave `InputArg 101` on the peeled LABEL. That fix is kept. This PR skips LABEL / jit-debug args as reads (they are phi dests / listing metadata) and adds unit tests. Unresolved **failargs** still decline: a live failarg is spilled from the zero wasm local (`emit_guard_fail_args_spill`).

The previous ubuntu `check.py` wasm `exit 1` storm was on the pre-#1780 base: compiling traces that still carried the hole. #1780 removes the hole.

`fib_recursive` `max-wasm-ratio=5.5` is unchanged.

## Tests

- `cargo test -p majit-backend-wasm --lib -- import_hole`
- `cargo test -p majit-backend-wasm --lib -- later_read_is_still`

## Self-review

## Summary by CodeRabbit

* **Bug Fixes**
  * Modules with folded import holes on labels, debug metadata, or guard failure arguments are now handled correctly.
  * Valid later uses of the same operand continue to receive the expected pool seed.

### DIFF
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -10582,13 +10582,13 @@ fn unbound_pool_const_seeds(
         }
         match constants.get(&raw) {
             Some(&bits) => seeds.push((raw, bits)),
-            // No producer and no pool entry: a body read would load the
-            // zero wasm initializes the local to. Decline that (the
-            // interpreter runs it correctly). A failarg is only a
-            // snapshot slot — import_state can leave the same leftover
-            // hole on a guard after every real use was rewritten, and
-            // dynasm still compiles the loop.
-            None if failarg => {}
+            // No producer and no pool entry: the local would read as the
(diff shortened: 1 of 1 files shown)


youknowone:
Rebased onto `origin/main` (#1780). The ubuntu wasm `exit 1` storm was on the pre-#1780 base: leftover `InputArg 101` still in the stream, then compiled.

Codex P1 on unresolved failargs: dropped that exemption. A live failarg is spilled from the zero local (`emit_guard_fail_args_spill`). LABEL live-ins are already `defined` after #1780, so the import-hole tests still pass.

— *commented by Claude*

PR #1793, opened 2026-09-12
llbc: ignore spans and source dumps in artefact identity

youknowone:
## Summary

- Charon ULLBC JSON embeds `files[].contents` (whole source, comments included), span `beg`/`end` line/col, and `dest_file` (absolute checkout path). None of those is an input to generated JIT tables.
- A comment in a file-table crate, or a different checkout path, therefore changed the raw artefact bytes and missed the prepass cache.
- `source_text` stays in the digest: the translator uses it to tell `static mut` from `static`. Graph bodies stay too.
- extract keeps the previous artefact bytes when only those dumps moved, so `artefacts=` and the on-disk file stay stable.
- The prepass cache key hashes the semantic feed (`codegen-cache-v20`), so a force extract that only shifted spans still hits.

Follow-up to #1776. File-table comments still run Charon once (no rustc oracle); the prepass after that no longer misses.

## PyPy parity

Cache invalidation only. No interpreter or JIT semantics change.

## Verification

- `python3 scripts/llbc_extract_selftest.py`: all passed
- `rustc --test pyre/pyre-jit-trace/src/ullbc_semantic.rs`: 4 passed
- `corpus.ullbc` digest unchanged after mutating spans / `dest_file`


youknowone:
Extract hung after cargo finished (`Finished dev profile in 3m 38s` at 03:56, runner killed at 06:40, exit 143).

rlib/object/interpreter were exact cache hits. pyre-jit missed the exact key (new `pyre-jit-trace` file in the wide closure) so Extract ran. After Charon wrote `pyre-jit.ullbc`, `keep_previous_if_bodies_unchanged` hashed previous+new by calling `bytes.find` once per span. That is quadratic in file size × span count and never finished.

Reverted in f69b78046a2. This approach is not viable: extract must not parse or rescan the artefact.

— *commented by Claude*

PR #1794, opened 2026-09-12
wasm: do not date residual LABEL RefOp as a home def

youknowone:
## Summary

Rebased onto `origin/main` after #1773. That landing already keeps InputArgRef LABEL phis, declines a producerless LABEL `RefOp` when a later operand reads it, and materializes a sibling `GETARRAYITEM` when one exists — the compile-path Codex P1 asked for.

What remains here: `HomeLiveness` still dated every LABEL arg as defined at the header. A residual `RefOp` is not a phi (`consider_label`); dating it hid the unwritten local from the post-collection reload. InputArgRef / int / float phis are unchanged.

## Tests

- `cargo test -p majit-backend-wasm --lib -- import_hole`

## Self-review

### DIFF
--- a/majit/majit-backend-wasm/src/codegen.rs
+++ b/majit/majit-backend-wasm/src/codegen.rs
@@ -2339,7 +2339,11 @@ impl HomeLiveness {
                     continue;
                 }
                 last_use[a.raw() as usize] = i as i32;
-                if op.opcode == OpCode::Label {
+                // A LABEL arg is a phi def (`consider_label`). A
+                // producerless RefOp is a residual virtualizable slot,
+                // not a phi: dating it here would hide the unwritten
+                // local from the post-collection reload.
+                if op.opcode == OpCode::Label && !matches!(a, OpRef::RefOp(_)) {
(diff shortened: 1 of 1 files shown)


youknowone:
Codex P1: declining in the backend is not the IR fix. `patch_new_loop_to_load_virtualizable_fields` only forwards the expanded entry inputargs; a residual `RefOp` remint on a body LABEL is a different identity, so `emit_op` never rewrites it. Compiling that slot as a wasm local stores null on failarg writeback. This PR only stops that compile; the walker/`compile.py` repair is a follow-up, not a wasm GETARRAYITEM clone.

— *commented by Claude*

youknowone:
Rebased onto `origin/main` (#1773). Codex P1 is addressed there: `materialize_unbound_label_args` plus keeping InputArgRef phis defined. This PR no longer declines every producerless LABEL `RefOp` at the LABEL itself (that would undo #1773). The leftover is `HomeLiveness`: do not date a residual `RefOp` as a header def.

— *commented by Claude*

PR #1805, opened 2026-09-12
perf-exc: exception folds and orthodox exception-edge bridges

youknowone:
Rebased onto current `origin/main`. Exception-path JIT folds plus a line-by-line bridge-parity pass against `compile.py` / `pyjitpl.py` `finishframe_exception`.

## Changes

- Fold `BaseException.descr_reduce` (specialised 2-tuple; peek exception `__dict__`).
- Emit int-strategy `BINARY_SUBSCR` as `dict.lookup` on `dstorage` with the unboxed key (`ll_dict_getitem_with_hash`). Hit continues from the same index.
- Route a pending exception whose handler lives in an **inlined callee** as a bridge, matching `finishframe_exception`. Do not set `bridge_declined_terminally` on uncaught (only backends with `bridge_decline_is_terminal` keep that bit).
- `compile_bridge` runs `propagate_original_jitcell_token` on LABEL `TargetToken`s, as `compile_and_attach` does.
- Cranelift attach comment now cites `patch_jump_for_descr` / `patch_trace`, not a nonexistent `compile.py attach_bridge`.
- Bench ceiling refits from earlier on this branch (exception fixtures). `exception_loop_warmup` keeps main's wasm gate and 6.5 pypy ceiling.

## Not changed

- Cranelift still does **not** merge a bridge into the owner loop (PyPy materializes a separate blob and patches the guard).
- Uncaught `Finish(exc)` / `compile_exit_frame_with_exception` is still a follow-up; that failure blackholes and retries via the jitcounter.

## Measured locally (dynasm)

- `exception_reduce` N=8e6: ~1.2x vs pypy.
- `exc_caught_in_callee_return_loop`: stdout matches pypy `(14352000, 72000)`; 1 loop / 2 bridges.

Assisted-by: Claude

### DIFF
--- a/pyre/bench/synth/exception_metadata_jitstress.py
+++ b/pyre/bench/synth/exception_metadata_jitstress.py
@@ -1,4 +1,4 @@
-# pyre-check: max-pypy-ratio=4
+# pyre-check: max-pypy-ratio=2
 # JIT-stress twin of exception_metadata_hot: `pypyjit.set_param` lowers the
 # trace/function thresholds to 1 so recording fires on the earliest iterations
 # of every section rather than only after the ~1600-iteration warmup. That
--- a/pyre/bench/synth/exception_metadata_jitstress.py
+++ b/pyre/bench/synth/exception_metadata_jitstress.py
@@ -1,4 +1,4 @@
-# pyre-check: max-pypy-ratio=2
+# pyre-check: max-pypy-ratio=3
 # JIT-stress twin of exception_metadata_hot: `pypyjit.set_param` lowers the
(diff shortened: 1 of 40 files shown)


pyre/bench/synth/exception_metadata_jitstress.py
```diff
@@ -1,4 +1,4 @@
-# pyre-check: max-pypy-ratio=4
+# pyre-check: max-pypy-ratio=2
```
youknowone:
```suggestion
# pyre-check: max-pypy-ratio=3
```

PR #1813, opened 2026-09-13
llbc: opaque interpreter modules by reachability, not a name list

youknowone:
## Summary

Census of `pyre-interpreter.ullbc`: 71629 Transparent items, **all** `pyre_interpreter`. Other crates are already Opaque/Foreign, so extra crate-root `--opaque` flags do nothing. 44908 of those items (63%) live under `module::*`.

This uses Charon's documented pair instead of a hand-maintained module name list:

```
--opaque crate::module --include crate::module::_
```

Charon does not walk `crate::module` as a crate root; it translates a module item only if something outside it reaches it. Modules whose `src/module/<name>/` tree mentions `majit_macros::` are `--start-from crate::module::<name>` so their hint-marker consts stay in the artefact (`harvest_hints_from_llbcs`). That set is generated, not listed: `_ast`, `_cffi_backend`, `_codecs`, `_collections`, `_ctypes`, `_hashlib`, `_io`, `_queue`, `_warnings`, `_weakref`, `mmap`, `sys`, `thread`.

`posix` / `gc` / `time` need no special case: core calls reach them and `--include crate::module::_` then translates those items. rustc still compiles the files; they stay in this crate. The rustc-time win is the later `pypy/module` → `pyre-module` split.

`charon_flags=` is already in the stamp, so only `pyre-interpreter` re-extracts. `EXTRACTION_ABI` stays `"1"`.

Replaces the earlier denylist commit on this PR (`_ssl`, `_socket`, …).

## PyPy parity

Not a crate split. Opacity follows Charon reachability plus the existing JIT-hint macros.

## Verification

- `python3 scripts/llbc_extract_selftest.py` (includes `run_child_dirs_whose_rs_mention`)
- Local `pyre-interpreter` extract with the new flags is running; Charon accepted the argv (`--opaque crate::module --include crate::module::_` plus the generated `--start-from` list)
- CI extract will refresh the interpreter artefact from `charon_flags=`

## Summary by CodeRabbit

- **Improvements**
  - Improved interpreter extraction to preserve selected extension modules as opaque while retaining reachable definitions and residual type information.
  - Extraction now identifies relevant modules and Rust-based contents more reliably, improving the completeness and consistency of generated interpreter data.
  - Interpreter extraction arguments are now derived from detected macro-bearing modules and their reachable contents.

- **Tests**
  - Added coverage for identifying immediate subdirectories containing matching Rust sources, including handling of nested paths and missing directories.


youknowone:
Rewrote this PR onto current `main` as a single commit: Charon's `--opaque crate::module --include crate::module::_` pair, plus generated `--start-from` for `src/module/<name>/` trees that mention `majit_macros::`. The hand-maintained native-module denylist is gone.

Local interpreter extract is in progress to measure ullbc size / Transparent counts. Charon already accepted the flags (they appear on the running `charon cargo` argv).

— *commented by Claude*

youknowone:
Local `pyre-interpreter` extract with the first reachability argv:

- Host pass succeeded in 1226 s. Artefact `472584199` bytes (~451 MB) vs the previous `841 MB`. `"opacity":"Transparent"` dropped `71629 → 35141`. `eval_loop_jit` is present. Stamp was not written because the layout pass failed.
- wasm layout failed: `--start-from crate::module::_cffi_backend` and `--start-from crate::module::mmap` do not resolve (`#[cfg(not(target_arch = "wasm32"))]`). Charon then aborted (`ERROR Code failed to compile`).
- Follow-up `36d0bec9665`: keep `--start-from crate` (a non-empty `--start-from` list replaces that default) and use `--start-from-if-exists` for the generated hinted-module roots.

Re-extracting locally with the new flags.

— *commented by Claude*

youknowone:
Re-extract with `--start-from crate` + `--start-from-if-exists` for hinted modules finished cleanly (host 123 s, wasm layout in parallel, stamp written).

| | previous artefact (Sep 5) | first try (module `--start-from` only) | this extract |
|---|---|---|---|
| host ullbc | ~841 MB | 451 MB | **827 MB** (`866750769`) |
| `"opacity":"Transparent"` | 71629 | 35141 | **64898** |
| wasm layouts | 11 MB | failed | **11 MB** (`11612823`), `type_decls` only |

The 451 MB cut was not the opaque/include pair: a non-empty `--start-from` list replaced the default `crate` walk, so Charon started from the 13 hinted modules only. With `--start-from crate` restored, unreached `module::*` items drop (~9% Transparent) and `eval_loop_jit` / hint-marker strings remain. wasm no longer dies on cfg'd-out `mmap` / `_cffi_backend`.

— *commented by Claude*

PR #1801, opened 2026-09-12
Pin rustpython #8687 and consume shared codecs

youknowone:
## Summary

Pin rustpython crates to `53f518ee` (#8687) and consume the engines that PR shared.

- utf-16/32/7, unicode-escape, raw-unicode-escape, and bytes escape go through `rustpython-common::encodings`.
- `numeric_formatting` uses `host_env::locale::localeconv_numeric()` (CHAR_MAX kept) on unix/windows.
- Windows `_uuid` uses `host_env::uuid` when `host_env` is on.
- `get/set_blocking` uses `winapi::get_named_pipe_handle_state`.

The conversion loops live in common. pyre keeps the error-handler glue in `codec_engine.rs`. `charmap` and the Windows code-page path stay local.

## Summary by CodeRabbit

- **New Features**
  - Improved text encoding and decoding across UTF, ASCII, Latin-1, Unicode escape, UTF-7, and related codecs.
  - Added broader support for standard codec error-handling modes.
  - Improved UUID generation and locale-aware numeric formatting with host integration.
  - Added a WebAssembly SSL interface for environments where native TLS is unavailable.

- **Bug Fixes**
  - Improved Windows named-pipe handling, code-page conversions, and filesystem operations.
  - Preserved relevant codec deprecation warnings.
  - Improved consistency of platform-specific behavior across supported environments.

- **Compatibility**
  - Some host-dependent modules and operations remain unavailable without host integration.

PR #1818, opened 2026-09-14
Pin rustpython #8702/#8727 and consume rustls-free ssl

youknowone:
## Summary

Follow-up to #1801. Pin rustpython crates to `RustPython/RustPython` `c382a607` (#8727 on main, after #8702 `3191ecaf3`) and consume the rustls-free `_ssl` core.

- MemoryBIO, PROTOCOL_*/CERT_*/SSL3_*/TLS_ERROR_*, OID tables, and ALPN/hostname live in `rustpython_host_env::ssl` on every target.
- host_env `ssl` is the rustls-free surface; `rustls` is the engine and is native-only. wasm-host must not compile rustls (no JS `UnixTime::now`).
- pyre-native drops leftover copies of that surface. Interpreter `_ssl` stays native-only.
- The wasm backend keeps pointer wrappers and TLS stubs on the shared types.
- `rustpython-ruff_text_size` matches #8700 (crates.io 0.16.5).

RustPython/RustPython#8702 and #8727 are both on main.

Assisted-by: Claude


youknowone:
Codex parity note on `_set_alpn_protocols` raising `SSLError("invalid ALPN protocol list")` vs PyPy `MemoryError`: this matches the native `pyre-native` host path. The public `SSLContext.set_alpn_protocols()` still pre-validates. Leaving the private method on the shared host parser.

— *commented by Claude*

PR #1816, opened 2026-09-14
rjitlog: name boxes i0/p0 and bind writes to tid

youknowone:
## Summary

Follow-up to #1815 Codex review on `88a24d0`.

- `LogTrace.var_to_str`: inputs, results, and failargs share the `i0`/`p0`/`f0` memo.
- `write_trace` / `trace_aborted` take the `start_new_trace` id instead of the process-global latest id.
- `compile_trace` emits `START_TRACE` only after `tracing_done()`.
- Interp-origin / FINISH traces record the green key instead of descriptor id 0.
- `DEBUG_MERGE_POINT` is `MARK_MERGE_POINT`, not a RESOP.

## Test plan

- [x] `cargo test -p majit-metainterp --lib --no-default-features --features dynasm -- rjitlog a_finish_compile an_entry_bridge`
- [ ] `python3 pyre/check.py` (host backends)

## Summary by CodeRabbit

- **Improvements**
  - Improved JIT trace logging with unique identifiers for individual traces.
  - Enhanced trace output with stable, readable variable names and clearer constant references.
  - Improved handling of merge points, bridges, callbacks, and assembler updates.
  - Improved trace-abort reporting and trace metadata for more accurate diagnostics.
  - Added JIT-driver names and source-location details to trace information, including PyPy JIT and iterable-unpacking traces.
  - Improved guard patch and redirection tracking in compiled traces.

PR #1829, opened 2026-09-14
Descend handler-free LoadName/StoreName through unwrap_cell and write_cell

youknowone:
## Summary

Handler-free module `LoadName` / `StoreName` / `StoreGlobal` descend the generated `unwrap_cell` and `write_cell` jitcodes (`typeobject.py`), the same look-inside PyPy uses. The compiled loop keeps the live cell field read/write after the helper's type tests fold on a `ConstPtr`. There is no hand-written getfield/setfield.

A handler-bearing module stays residual. Lifting that gate installs `version?` on `except as` (`DELETE_NAME` → `delitem` always `mutated()`) and aborts the module while-loop.

A non-cell slot is `unwrap_cell`'s identity return; that classification is a trace-time fact, so the loop keeps the `ConstPtr`. A cell still goes through the generated helper.

`write_cell` marks the prebuilt-root dirty bit only on an in-place object-cell pointer store. An in-place `intvalue` store writes no `PyObjectRef`; PyPy's `write_cell` has no barrier on that arm. Calling the `dont_look_inside` mark before classify residualized `CallN` on every module-level int store (`i`, `total`, …).

Also: pin the namespace and re-read the slot before baking a cell write; dirty prebuilt roots on obj-cell FBW rollback; leftover rjitlog spelling from #1816.

## Oracle / measure

`pypy3` of the `load_name` twin: Loop 3 is the inner while-loop as `getfield_gc` / `setfield_gc` on cells, not a walker fold.

After this descent, local dynasm/cranelift of `load_name_builtin_cell_fold` and `global_reassign_invalidation` sit ~1.1× pypy (gates 2.5 and 3). Module-scope `for_iter_*` int stores recovered the same way.

## Test plan

- [x] `pyre-object` `celldict` tests
- [x] local dynasm + cranelift retimes of `load_name_builtin_cell_fold`, `global_reassign_invalidation`, `for_iter_method_branch_inline`, `for_iter_nested_method_inline`
- [ ] `python3 pyre/check.py` on CI (ubuntu dynasm+wasm, cranelift, windows dynasm)


youknowone:
Handler-bearing LoadName/StoreName no longer pin `version?` on the whole module dict: that is what made `except as` (`DELETE_NAME` → `delitem` always `mutated()`) abort 116 traces and knock `loops_compiled` off on other fixtures (e.g. `trace_too_long_inline_multiframe` 54 → 10).

The cell IR now comes from a sub-walk of the generated `write_cell` / `unwrap_cell` jitcodes, and the store descent calls `write_cell` so the prebuilt-root dirty bit is set.

Measured dynasm: reraise jitstress 153 loops / 8 abort / 0.11s; hot 4 loops / 0 abort / 0.16s.

— *commented by Claude*

youknowone:
Correction: the handler-bearing gate is back. We do not descend/fold LoadName/StoreName in a module that has a `try`, so `except as` no longer installs a `version?` watcher on every name access. Handler-free modules descend the generated `write_cell` / `unwrap_cell` bodies instead of a hand-written setfield/getfield.

— *commented by Claude*

youknowone:
Rebased onto `origin/main` (`0c33ff7970d`, includes #1827/#1822/#1681).

Review disposition on `251ec3f`:

- Applied Codex P1: pin `version?` and recheck the slot before `write_cell` descent (`try_walker_orthodox_write_cell`).
- Applied Codex P1: `FbwCellStore::Obj` rollback calls `mark_prebuilt_roots_dirty()`.
- Already landed earlier: store descent calls `write_cell` (prebuilt-root dirty bit); function-entry death ban removed.
- Not applying Codex P2 (lift the handler-bearing gate). That gate is what stops `except as` (`DELETE_NAME` → `delitem` always `mutated()`) from installing a `version?` watcher and aborting 116 traces.
- Not applying CodeRabbit leftover rjitlog/`CALL_INTRINSIC_1` notes; those stay with the leftover rjitlog commit unless that scope expands.

CI: Windows/macOS `for_iter_direct_store_double` failed because a local re-record (`5/45/7421/fbw 44`) did not match CI/main (`4/42/7399/fbw 41`). Restored those two jitstats. The fixture is function-local `STORE_ATTR`/`STORE_FAST`, so the store-cell pin does not change it.

— *commented by Claude*

youknowone:
Review disposition on `06f639e4334`:

**Applied**
- Codex P1: route non-cells through generated `unwrap_cell` (identity skip removed).
- Codex P2: AArch64 redirect logs `newlooptoken.number`. `jl.redirect_assembler` now lives in each backend assembler (`x86` logs `_ll_raw_start`, others log the token number), matching `warmstate.py` / `assembler.py`.
- Codex P2: scientific `str(float)` uses rustc `{:e}` (correctly rounded) plus Python exponent width. `5e-324` is no longer `infe-323`.

**Already applied on earlier commits**
- Codex P1: pin `version?` and recheck the store slot before `write_cell`.
- Codex P1: dirty prebuilt roots on object-cell FBW rollback.
- Codex P1: dirty mark only on `InPlaceObject` (interpreter `write_cell`).
- Codex P1: permanent function-entry retrace ban removed.
- CodeRabbit: `python_float_str` / `screaming_snake_opcode` (`CallIntrinsic1` → `CALL_INTRINSIC_1`).

**Rejected**
- Codex P2: lift the handler-bearing LoadName/StoreName gate. Module-level `except as` is `STORE_NAME` + `DELETE_NAME`; `celldict.delitem` always `mutated()`, so an ungated fold aborts ~116 traces. Handler-free modules still descend.

— *commented by Claude*

PR #1830, opened 2026-09-14
Descend compare, format, except-match, and getattr through generated graphs

youknowone:
CHECK_EXC_MATCH, COMPARE/IS/CONTAINS, FORMAT, and jitted LOAD_ATTR emit `inline_call` of the helper bodies (`compare_value_from_tag`, `format_simple_w`/`format_w`, `getattr`) when those graphs are bound. Residual remains the fallback if a body is unbound.

Also: skip empty exception-cell drains on the CloseLoop back-edge; share the cached BaseException pointer; `type()` reads a specialised `w_class` before `is_exception`; `check_exc_match_against` is `exception_match(type(exc), T)` and `is_type_like_w` is `isinstance_w(..., w_type)`.

## Summary by CodeRabbit

* **New Features**
  * Pyre and PyPy can now discover and install compatible CPython ABI3 wheels on supported platforms.
  * Extension modules using the `.abi3.so` suffix are now supported.
* **Bug Fixes**
  * Improved JIT handling for attribute access, item operations, iteration, formatting, comparisons, and exception matching.
  * Improved propagation of attribute-lookup exceptions during JIT execution.
  * Improved list access and mutation behavior in optimized code paths.
* **Tests**
  * Added coverage for ABI3 packaging tags, extension loading, and updated JIT behavior.


youknowone:
Addressed the latest Codex notes on `9001209704`:

- **setattr name guard / dest write (P1):** a non-constant `w_name` now gets `GuardValue` before the mapdict fold, and a successful `inline_call_r_r` of `setattr` writes `w_none()` to the Ref destination.
- **pre-commit rustfmt:** the new helper matchers are rustfmt-clean.
- **`compatible_tags(None)` (P2):** under pyre/pypy, `interpreter=None` is the running process (`packaging.tags.sys_tags` does that for any impl that is not `cp`/`pp`). CPython extra_tests still treat `None` as “not this interpreter”.

On the 5578 surface comments: that work was an explicit scope expansion (PyPy #5578), not a silent new module. `_imp.extension_suffixes` now documents the 3.14 list vs PyPy’s native-only list; we advertise `.abi3.so` and still omit a bare `.so`.

— *commented by Claude*

PR #1847, opened 2026-09-18
Lower remaining math1 builtins through generated leaves

youknowone:
Follow-up to #1832 after the squash onto main.

- Heapcache live loads prove `0 <= index < arraylen` (wasm32 pointers fit in i32).
- `math.atan` / `exp` / `log1p` / `asin` and the rest of the one-arg float builtins descend unboxed leaves (`_float_atan` … `_float_radians`).
- Interpreter `math1` boxes the pymath `Ok` value via `_float_pos`. The walker still walks a generated leaf so the jitcode has a body; crates.io pymath is not in the Charon artefact.
- `MATH_FLOAT1_FOLDS` is empty. Census unchanged: 103 `spec_folds!` rows, 89 `try_walker_specialize_*`.

## Summary by CodeRabbit

- **New Features**
  - Expanded floating-point math support, including trigonometric, exponential, logarithmic, rounding, comparison, and conversion operations.
  - Added support for additional floating-point constants and byte-preserving float conversions.
  - Improved float-to-integer conversion handling and JIT support for numeric operations.

- **Bug Fixes**
  - Improved array index validation to prevent invalid accesses.
  - Increased compatibility with different float bit-conversion call paths.

- **Tests**
  - Added coverage for expanded math operations, numeric conversions, and opcode handling.


youknowone:
Review follow-up (not yet pushed; branch is rebased onto `origin/main` `4523e35`):

**Done**
- Runtime result-finite guard after descending `exp`/`exp2`/`expm1`/`sinh`/`cosh`/`gamma`/`lgamma`/`pow`, so a later overflow/`0**neg` deopts into the builtin instead of boxing `inf`/`NaN`.
- `Math1ResultCheck` replaces `std::ptr::eq` on `const` `HelperDescent` addresses.
- `gamma`/`lgamma` admit only a finite pymath `Ok`.
- Removed unused `pm1!` / `jit_raw1!` macros.

**Disagree: “synthetic leaves are a pre-descent hand fold”**
The success path is `try_walker_orthodox_descent` → `run_orthodox_helper_subwalk` of a generated `_float_*` jitcode. The result ops come from that walk (`residual_call_irf_f` + in-graph `new_with_vtable`), not from a hand-emitted `CallF`/`wrapfloat`. `crate::module` is Charon-opaque and crates.io pymath is not in the artefact, so the walker cannot enter `math1_pymath`; the leaf is the look-inside body the same way `_float_lt` is for compare. Census is unchanged (`spec_folds!` 103, `try_walker_specialize_*` 89); `try_walker_orthodox_*` is not a specialize row.

— *commented by Grok 4.6*

PR #1865, opened 2026-09-19
jit: one ArrayDescr per list block and object-pointer slice; re-enable the list setitem descent

youknowone:
## Summary

`cpu.arraydescrof(ARRAY)` is one descr per ARRAY lltype. The build-time descr pool and the runtime descrs named the same blocks differently, so the heap cache treated one array as several.

- **List backing blocks** — the `list.int_*` / `list.float_*` / `list.obj_*` oopspec arms emitted array ops with `array_type_id: None`, minting a second ArrayDescr (tid 0) next to the walker's `int_gcarray_descr`. A descended `setarrayitem_gc` was invisible to a cached `getarrayitem_gc_i`, which is what made `array_deopt_resume` answer wrong and why the list setitem descent had been switched off. The arms now carry `LIST_INT_ITEMS_ARRAY` / `LIST_FLOAT_ITEMS_ARRAY` / `OBJECT_REF_GCARRAY_TYPE_ID`, the runtime descrs carry the same identity, and the `BhDescr::Array` bridge returns the runtime descr (also on a `cache_key` hit, since `BhDescr::from_array_descr` drops the spelling). The early return in `try_walker_orthodox_list_setitem` is removed.
- **Float / IntOrFloat arms** — `w_list_getitem_inner` / `w_list_setitem_inner` indexed `float_items` / `int_items` directly (an `abort/` marker for the Float arm, an identity-less `setarrayitem_gc_i` and a `[i64]` read for IntOrFloat). They now go through the `ll_list_*` leaves; `list.float_getitem` gains its oopspec leaf and jtransform arm.
- **Object-pointer slices** — the front's model is that a `&[T]` is one GC array pointer whose length `ArrayLen` reads from the header, but a `"[T]"` spelling was classified header-less. A gateway wrapper's `&[PyObjectRef]` argument was read through `"[*mut PyObject]"` (base 0), its length through an identity-less descr, while the walker built it with `pyobject_gcarray_descr`. `ProjectionElem::Index`, `Index::index`, `<[T]>::len` / `is_empty`, `get` / `first` / `last` and `FixedObjectArray::len` on object pointers now all name `OBJECT_REF_GCARRAY_TYPE_ID` length-prefixed. `wrapper_args_item_descr_index` (a bytecode scan that seeded the heap cache under the wrapper's own descr) and its decline are removed; the descent seeds under the descr it builds the array with.
- **Tests** — `pyre/pyre-jit-trace/tests/array_descr_identity.rs` reads the pool and the list-inner / wrapper jitcodes and pins the above; a unit test asserts the bridged list-int descr is the runtime Arc.

No `spec_folds!` row or `try_walker_specialize_*` function is added; no `.jitstats` baseline is re-recorded.

Left identity-less on purpose: `vable_arraydescrof` (documented — two allocators, one tid cannot describe both blocks; vable ops do not key the array heap cache) and host containers (`Vec` legs, byte views, Bytes/Ascii list item blocks) that have no competing named descr.

## Test plan

- `pyre/check.py --synthetic-only`: dynasm 544/544, cranelift 544/544, wasm 534/534 (fresh LLBC extraction per backend)
- `cargo test -p majit-translate --lib` 3669 passed; `-p pyre-jit-trace --lib` 508 passed; `--test array_descr_identity` 4 passed; `-p pyre-object --lib list` 62 passed

## Summary by CodeRabbit

- **Performance**
  - Improved JIT optimization and consistency for integer, floating-point, and object-backed list operations.
  - Reduced overhead when accessing and updating list elements, including floating-point values.
  - Improved built-in wrapper calls involving object sequences.

- **Bug Fixes**
  - Fixed iterator behavior when garbage collection occurs during iteration.
  - Improved floating-point list element access and updates.
  - Prevented potential shutdown hangs involving the import lock.

- **Tests**
  - Expanded coverage for list access, array descriptors, wrapper calls, and iterator behavior.


youknowone:
The `hermetic-guard` failure (`ValueError: not enough values to unpack (expected 2, got 0)` at `re/_compiler.py:58`) is root-caused and fixed in 3c33e4fa7bf.

`opcode_for_iter` wrote its result through the `&mut PyFrame` it held across `iter_next`. For an old-protocol sequence (`SubPattern.__getitem__`) that step runs Python; a minor collection inside it relocates a nursery frame (one materialised by CALL_ASSEMBLER), so the exhaustion jump landed on the abandoned copy. The live frame kept the fallthrough pc with the iterator on TOS, and `UNPACK_SEQUENCE` unpacked the exhausted iterator → "got 0". It reproduces on the PR base without this branch's commits (stdlib-only reproducer, deterministic under `PYTHONHASHSEED=0`; gone with `MAJIT_NO_BRIDGE=1`), so the branch only moved when it surfaced.

Fix: take the `FrameAnchor` before `iter_next` and do both the push and the exhaustion jump through it, as `opcode_get_iter` already does. Verified: reproducer 0/9 failing (was 5/9), `pyre/extra_tests/pip/run.py --dynasm-only` passes, dynasm synthetic 544/544. The branch is also rebased onto current main.

— *commented by Claude Fable 5.1*

PR #1890, opened 2026-09-21
Drop the cranelift dense fail-arg bindings at a LABEL

youknowone:
## Summary

**cranelift: a gcmap named a word the back edge left an Int in** (`987e873d362`).

A GUARD_NOT_FORCED's fail args are published into the dense `jf_frame` words before its call, and `dense_ref_bindings` lets a later call's gcmap name such a word for as long as its Ref box is alive. The bindings were built in trace order and carried across the loop LABEL, but no JUMP writes those words, so on the back edge each one holds the previous iteration's last publish.

In `pyre/bench/synth/attr_delete.py` the preamble published the ExecutionContext into `jf_frame[4]` and the body published the unboxed `acc` Int there. The body's first nursery slow path then traced `acc` as a reference. With `PYPY_GC_NURSERY=1`, `pyre-cranelift` aborted in 17 of 40 runs (`GCFLAG_HAS_SHADOW but no shadow found`, `invalid type_id` at `parent_site=minor_jitframe_root`, one SIGSEGV). Every bad word reported was a triangular number, i.e. a value of `acc`. `pyre-dynasm` passed 24 of 24.

The bindings are now cleared when a LABEL's block is entered, the counterpart of `aarch64/regalloc.py` `prepare_op_label` keeping only the locations its JUMP writes. The added unit test reads the loop-top call's gcmap and `jf_frame[0]` through a FORCE_TOKEN argument and fails on the old code.

This PR earlier also carried a `class_reassign_hot` wasm baseline and a wasm `memcpy` wrapper. Both were dropped on rebasing onto `4a929a733a4`: #1891 records the same baseline, and #1874 calls the C `memcpy` in the module through its reflected wasm type. On #1874's run `str_fstring` measured wasm 0.77s against dynasm 0.45s.

## Verification

The commit is unchanged by the rebase (`git range-diff` reports it identical to `c54cd35f97d`). At `c54cd35f97d`, on base `8b516323e63`:

- `attr_delete.py` at `PYPY_GC_NURSERY=1` on `pyre-cranelift`: 40/40 exit 0, versus 17/40 aborts before the fix.
- `pyre/check.py`: dynasm 557/557, cranelift 557/557; wasm 545/547, with both failures also failing on `main` at the time.
- `cargo test --all --no-default-features --features dynasm`: 10022 passed / 0 failed.

At `987e873d362`: `cargo test -p majit-backend-cranelift a_loop_labels_first_call_does_not_trace` passes.


youknowone:
Windows `cargo test` was the only real CI failure. `pyre/check.py dynasm` was cancelled because that job failed, not because check.py itself failed.

`test_nullary_true_void_call_uses_indirect_call_without_drop` saw `jit_call_compact` imported (`Some(1)` vs `None`). `RESIDUAL_CALL_ABI` was a process-global atomic, while the faithful-address set is thread-local. A parallel test selecting `Vouched` changed a sibling still compiling under the default `Word` ABI. The ABI is now thread-local, same as `FAITHFUL_RESIDUAL_CALL_ADDRS`. Pushed as `9bd827f7662`.

— *commented by Grok*

